PR #49, opened 2022-12-18
Create DbtRunOperationOperator

pohek321:


cosmos/providers/dbt/core/operators.py
```diff
@@ -326,3 +326,40 @@ def __init__(self, **kwargs) -> None:
...
         return result.output
+
+class DbtRunOperationOperator(DbtBaseOperator):
+    """
+    Executes a dbt core run-operation command.
+
+    :param macro_name: name of macro to execute
+    :type macro_name: str
+    :param args: Supply arguments to the macro. This dictionary will be mapped to the keyword arguments defined in the
+        selected macro. This argument should be a YAML string, i.e. '{my_variable: my_value}'
```
jlaneve, line 337: Is this a YAML string or a dictionary?

pohek321: Oh good catch! I want the operator to take it as a dict and then it gets converted to a string. So we just need to update the class comments. I'm on the road right now, but will be able to do it later tonight.

jlaneve: no rush 🙂

jlaneve (approved):

PR #55, opened 2022-12-19
Add on_kill to DbtBaseOperator to recognize execution timeouts

pohek321:


pohek321:
@jlaneve I think this is what our operator is missing. Have you ever developed one of these?

jlaneve:
I haven't. Have you tested/does this work?

pohek321:
> I haven't. Have you tested/does this work?

Spinning up a sandbox to try it out now. From the looks of it, this is [how the BashOperator is handling it](https://github.com/apache/airflow/blob/main/airflow/operators/bash.py#L201). [OSS docs](https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/models/baseoperator.html) are pretty vague on the subject.

pohek321:
@jlaneve that was it! I put a 15-second execution_timeout on the operator to test and a 60-second `time.sleep` in the execution method so that I could test it - and after adding the `on_kill` it worked as expected.

jlaneve (approved):
nice! good find

PR #69, opened 2022-12-23
Include model tags in parsers as well as a list parameter to filter them

pohek321:
Please take a look at the issue I've attached. This will be a doozie - tags are just one selector type, but they feel the most applicable to our parsers.

In a dbt project, tags can be set in 3 places (see more info [here](https://docs.getdbt.com/reference/resource-configs/tags)):

- [ ] a `properties.yml` file
- [ ] a `.sql` file in the `models` directory
- [ ] a `dbt_project.yml` file

(I've checked the ones that this PR would take care of) - the question is: **Do we want to do it this way?** This feels like the most natural, but we should talk about the following:

- the [other selector types](https://docs.getdbt.com/reference/node-selection/syntax#how-does-selection-work) besides tags (these may or may not apply to how the parsers are parsing)
    - `path:`
    - `config:`
    - `test_type:`
    - `test_name:`
- whether this will be too strenuous on the `DbtDag` and `DbtTaskGroup` process to do this in-memory

### DIFF
--- a/cosmos/providers/dbt/dag.py
+++ b/cosmos/providers/dbt/dag.py
@@ -8,6 +8,7 @@ def DbtDag(
     dbt_args: dict = None,
     emit_datasets: bool = True,
     dbt_root_path: str = "/usr/local/airflow/dbt",
+    include_tags: list = None,
     **kwargs,
 ):
     """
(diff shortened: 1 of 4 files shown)


pohek321:
@jlaneve took a stab at the tags piece of selectors - not sure if this is how we wanted to do it, and it'd probably be best to discuss. I figured some of this (or all of it) would get implemented one way or another.

pohek321:
Implemented in another PR

PR #83, opened 2023-01-11
Convert cli commands to use dbt executable path

pohek321:
Currently, we are sourcing the venv [here](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/providers/dbt/core/operators.py#L152). Instead, we should reference the dbt executable so that implementing the `dbt-ol` wrapper is easier.

Before:
```python
dbt_path = [bash_path, "-c", f"source {self.python_venv} && dbt {self.base_cmd} "]
```

After:
```python
{self.python_venv}/bin/dbt {self.base_cmd}
```

### DIFF
--- a/cosmos/providers/dbt/core/operators.py
+++ b/cosmos/providers/dbt/core/operators.py
@@ -1,8 +1,8 @@
-from typing import Sequence
-
 import json
 import os
 import shutil
+import subprocess
+from typing import Sequence
 
 from airflow.compat.functools import cached_property
 from airflow.exceptions import AirflowException, AirflowSkipException
@@ -11,7 +11,10 @@ from airflow.models.baseoperator import BaseOperator
 from airflow.utils.context import Context
 from airflow.utils.operator_helpers import context_to_airflow_vars
 
-from cosmos.providers.dbt.core.utils.profiles_generator import create_default_profiles, map_profile
+from cosmos.providers.dbt.core.utils.profiles_generator import (
+    create_default_profiles,
+    map_profile,
+)
 
 
 class DbtBaseOperator(BaseOperator):
@@ -78,7 +81,7 @@ class DbtBaseOperator(BaseOperator):
         self,
         project_dir: str,
         conn_id: str,
-        base_cmd: str = None,
+        base_cmd: list = None,
         select: str = None,
         exclude: str = None,
         selector: str = None,
@@ -146,25 +149,34 @@ class DbtBaseOperator(BaseOperator):
         return env
 
     def get_dbt_path(self):
-        bash_path = shutil.which("bash") or "bash"
+        which_dbt = shutil.which("dbt") or "dbt"
 
         if self.python_venv:
-            dbt_path = [bash_path, "-c", f"source {self.python_venv} && dbt {self.base_cmd} "]
+            dbt_path = f"{self.python_venv}/bin/dbt"
         else:
-            dbt_path = [bash_path, "-c", f"dbt {self.base_cmd} "]
+            dbt_path = which_dbt
 
         if self.project_dir is not None:
             if not os.path.exists(self.project_dir):
-                raise AirflowException(f"Can not find the project_dir: {self.project_dir}")
+                raise AirflowException(
+                    f"Can not find the project_dir: {self.project_dir}"
+                )
             if not os.path.isdir(self.project_dir):
-                raise AirflowException(f"The project_dir {self.project_dir} must be a directory")
+                raise AirflowException(
+                    f"The project_dir {self.project_dir} must be a directory"
+                )
         return dbt_path
 
-    def exception_handling(self, result):
-        if self.skip_exit_code is not None and result.exit_code == self.skip_exit_code:
-            raise AirflowSkipException(f"dbt command returned exit code {self.skip_exit_code}. Skipping.")
-        elif result.exit_code != 0:
-            raise AirflowException(f"dbt command failed. The command returned a non-zero exit code {result.exit_code}.")
+    # TODO: Fix this - CompletedProcess(args=['/usr/local/airflow/dbt_venv/bin/dbt', '--help'], returncode=0)
+    def exception_handling(self, return_code):
+        if self.skip_exit_code is not None and return_code == self.skip_exit_code:
+            raise AirflowSkipException(
+                f"dbt command returned exit code {self.skip_exit_code}. Skipping."
+            )
+        elif return_code != 0:
+            raise AirflowException(
+                f"dbt command failed. The command returned a non-zero exit code {return_code}."
+            )
 
     def add_global_flags(self):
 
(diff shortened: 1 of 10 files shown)


pohek321:
@jlaneve & @iancmoritz, after merging this, we should probably tag a new minor release - this is going to break some shit

jlaneve:
hmm I was thinking this would look more along the lines of:

```python
DbtTaskGroup(
  ...,
  dbt_executable="/usr/local/airflow/dbt_venv/bin/dbt"
)
```

This would abstract away the venv, lineage, etc management from our operators and let the user handle it, so you could do something like:

```python
DbtTaskGroup(
  ...,
  dbt_executable="/usr/local/airflow/dbt_venv/bin/dbt-ol"
)
```

And I believe we could also default it to just `dbt` (since, after pip installing it, it'll be in the Docker image's PATH).

@pohek321 what do you think?

pohek321:
> hmm I was thinking this would look more along the lines of:
> 
> ```python
> DbtTaskGroup(
>   ...,
>   dbt_executable="/usr/local/airflow/dbt_venv/bin/dbt"
> )
> ```
> 
> This would abstract away the venv, lineage, etc management from our operators and let the user handle it, so you could do something like:
> 
> ```python
> DbtTaskGroup(
>   ...,
>   dbt_executable="/usr/local/airflow/dbt_venv/bin/dbt-ol"
> )
> ```
> 
> And I believe we could also default it to just `dbt` (since, after pip installing it, it'll be in the Docker image's PATH).
> 
> @pohek321 what do you think?

Ahhh, yeah that makes way more sense - that's a pretty easy change anyway. I'll get it done.

pohek321:
Did some successful testing:

## dbt-ol

### Running `dbt-ol` (with venv)
Dockerfile:
```Dockerfile
FROM quay.io/astronomer/astro-runtime:7.1.0

# install python virtualenv to run dbt
WORKDIR /usr/local/airflow
COPY dbt-requirements.txt ./
RUN python -m virtualenv dbt_venv && source dbt_venv/bin/activate && \
    pip install --no-cache-dir -r dbt-requirements.txt && deactivate
```

dbt-requirements:
```plain text
openlineage-dbt
```

Operator:
```python
DbtDepsOperator(
    ...
    python_venv='/usr/local/airflow/dbt_venv/bin/dbt-ol'
)
```

Result:
```plain text
['/usr/local/airflow/dbt_venv/bin/dbt-ol', 'deps', '--project-dir', '/usr/local/airflow/dbt/jaffle_shop', '--profile', '***_profile']
```
___
### Running `dbt-ol` (no venv)
requirements.txt
```plain text
openlineage-dbt
```

Operator:
```python
DbtDepsOperator(
    ...
)
```

Result:
```plain text
['/usr/local/bin/dbt-ol', 'deps', '--project-dir', '/usr/local/airflow/dbt/jaffle_shop', '--profile', '***_profile']
```
___
## dbt

### Running `dbt` (with venv)
Dockerfile:
```Dockerfile
FROM quay.io/astronomer/astro-runtime:7.1.0

# install python virtualenv to run dbt
WORKDIR /usr/local/airflow
COPY dbt-requirements.txt ./
RUN python -m virtualenv dbt_venv && source dbt_venv/bin/activate && \
    pip install --no-cache-dir -r dbt-requirements.txt && deactivate
```

dbt-requirements:
```plain text
dbt-core
```

Operator:
```python
DbtDepsOperator(
    ...
    python_venv='/usr/local/airflow/dbt_venv/bin/dbt'
)
```

Result:
```plain text
['/usr/local/airflow/dbt_venv/bin/dbt', 'deps', '--project-dir', '/usr/local/airflow/dbt/jaffle_shop', '--profile', '***_profile']
```
___
### Running `dbt` (no venv)
requirements.txt
```plain text
dbt-core
```

Operator:
```python
DbtDepsOperator(
    ...
)
```

Result:
```plain text
['/usr/local/bin/dbt', 'deps', '--project-dir', '/usr/local/airflow/dbt/jaffle_shop', '--profile', '***_profile']
```

pohek321:
@jlaneve sorry for the long comment ☝🏻

jlaneve:
nice! that looks great, can we rename `python_venv` to something like `dbt_executable_path`?

pohek321:
> nice! that looks great, can we rename `python_venv` to something like `dbt_executable_path`?

@jlaneve just committed it up -- thoughts?

cosmos/providers/dbt/core/operators.py
```diff
@@ -95,7 +97,7 @@ def __init__(
         append_env: bool = False,
         output_encoding: str = "utf-8",
         skip_exit_code: int = 99,
-        python_venv: str = None,
+        dbt_executable_path: str = None,
```
jlaneve, line 100: do we want a default argument of just `"dbt"` here? what happens if you don't specify a `dbt_executable_path`?

cosmos/providers/dbt/core/operators.py
```diff
@@ -146,25 +148,33 @@ def get_env(self, context):
...
-        bash_path = shutil.which("bash") or "bash"
+        which_dbt = shutil.which("dbt-ol") or shutil.which("dbt") or "dbt"
 
-        if self.python_venv:
-            dbt_path = [bash_path, "-c", f"source {self.python_venv} && dbt {self.base_cmd} "]
+        if self.dbt_executable_path:
+            dbt_path = self.dbt_executable_path
         else:
-            dbt_path = [bash_path, "-c", f"dbt {self.base_cmd} "]
+            dbt_path = which_dbt
```
jlaneve, line 156: do we still need this `which_dbt` var or can we just use the `dbt_executable_path`?

cosmos/providers/dbt/core/operators.py
```diff
@@ -184,9 +194,9 @@ def add_global_flags(self):
             if global_flag_value is not None:
                 if isinstance(global_flag_value, dict):
                     # handle dict
-                    dict_string = json.dumps(global_flag_value)
+                    yaml_string = yaml.dump(global_flag_value)
                     flags.append(dbt_name)
-                    flags.append(f"'{dict_string}'")
+                    flags.append(yaml_string)
```
jlaneve, line 199: why did this change from json to yaml?

cosmos/providers/dbt/core/operators.py
```diff
@@ -24,7 +26,7 @@ class DbtBaseOperator(BaseOperator):
     :param conn_id: The airflow connection to use as the target
     :type conn_id: str
     :param base_cmd: dbt sub-command to run (i.e ls, seed, run, test, etc.)
-    :type base_cmd: str
+    :type base_cmd: str or list
```
jlaneve, line 29: I _think_ the type hint here should be:

```suggestion
    :type base_cmd: str | list[str]
```

Check out https://stackoverflow.com/a/58859298

cosmos/providers/dbt/core/operators.py
```diff
@@ -78,7 +80,7 @@ def __init__(
         self,
         project_dir: str,
         conn_id: str,
-        base_cmd: str = None,
+        base_cmd: str or list = None,
```
jlaneve, line 83: I _think_ the type hint here should be:

```suggestion
        base_cmd: str | list[str] = None,
```

Check out https://stackoverflow.com/a/58859298

examples/dags/attribution-playbook.py
```diff
@@ -15,7 +15,7 @@
 attribution_playbook = DbtDag(
     dbt_project_name="attribution-playbook",
     conn_id="airflow_db",
-    dbt_args={"schema": "public", "python_venv": "/usr/local/airflow/dbt_venv/bin/activate"},
+    dbt_args={"schema": "public", "python_venv": "/usr/local/airflow/dbt_venv"},
```
jlaneve, line 18: can we update this to use the `dbt_executable_path`?

examples/dags/mrr-playbook.py
```diff
@@ -14,7 +14,7 @@
 mrr_playbook = DbtDag(
     dbt_project_name="mrr-playbook",
     conn_id="airflow_db",
-    dbt_args={"schema": "public", "python_venv": "/usr/local/airflow/dbt_venv/bin/activate"},
+    dbt_args={"schema": "public", "python_venv": "/usr/local/airflow/dbt_venv"},
```
jlaneve, line 17: same here, let's update this example to use `dbt_executable_path`

jlaneve (commented):
apologies for the big review here, but tried to be overly verbose to turn this into a checklist 😬 

overall looking really good! just a handful of minor things

cosmos/providers/dbt/core/operators.py
pohek321: In order for it to just work with dbt installed in the Docker image, we should leave it as None.

cosmos/providers/dbt/core/operators.py
pohek321: I'd leave this personally, but an open to having my mind changed. This will pick up any `dbt-ol` or `dbt` installations by default (outside of a venv). Leaving it this way gives our customers more flexibility (i.e. you want the venv, you specify the path -- otherwise, just add dbt-core to your requirements.txt)

cosmos/providers/dbt/core/operators.py
pohek321: good question! This is what the CLI help docs are expecting. When I changed to the executable path -- it also broke as a dict string.

jlaneve (approved):
this looks great!

Issue #94, opened 2023-01-24
Add  support for "tags" select/exclude to DbtDag & DbtTaskGroup parsers


pohek321:
See dbt docs on cli usage examples [here](https://docs.getdbt.com/reference/node-selection/methods#the-tag-method)

```bash
$ dbt run --select tag:nightly    # run all models with the `nightly` tag
```

Ultimately, in our parsers, we should be able to have a new parameter that looks something like this:

```python
# (Either the select or exclude parameter would be specified with the nightly tag - not both)
jaffle_shop = DbtTaskGroup(
    ...
    select={'tags': ['nightly']} # runs all models that have the nightly tag
    exclude={'tags': ['nightly']} # runs all models that don't have the nightly tag
)
```

Note: We already have this as the `dbt_tags` arg so we'll just need to convert that parameter to be part of the `select` parameter

PR #96, opened 2023-01-24
Convert dbt_tags parameter to general select parameter

pohek321:
Closes [94](https://github.com/astronomer/astronomer-cosmos/issues/94)

I also snuck in there a max line length of 120 so that pre-commit stopped failing.

### DIFF
--- a/cosmos/providers/dbt/dag.py
+++ b/cosmos/providers/dbt/dag.py
@@ -6,7 +6,7 @@ try:
 except ImportError:
     from typing_extensions import Literal
 
-from typing import Any, Dict, List
+from typing import Any, Dict
 
 from cosmos.core.airflow import CosmosDag
 
@@ -25,7 +25,7 @@ class DbtDag(CosmosDag):
     :param emit_datasets: If enabled test nodes emit Airflow Datasets for downstream cross-DAG dependencies
     :param test_behavior: The behavior for running tests. Options are "none", "after_each", and "after_all".
         Defaults to "after_each"
-    :param dbt_tags: A list of dbt tags to filter the dbt models by
+    :param select: A dict of dbt selector arguments (i.e., {"tags": ["tag_1", "tag_2"]})
     """
 
     def __init__(
@@ -36,7 +36,7 @@ class DbtDag(CosmosDag):
         emit_datasets: bool = True,
         dbt_root_path: str = "/usr/local/airflow/dbt",
         test_behavior: Literal["none", "after_each", "after_all"] = "after_each",
-        dbt_tags: List[str] = [],
+        select: Dict[str, Any] = {},
         *args: Any,
         **kwargs: Any,
     ) -> None:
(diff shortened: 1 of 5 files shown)


cosmos/providers/dbt/dag.py
```diff
@@ -36,7 +36,7 @@ def __init__(
         emit_datasets: bool = True,
         dbt_root_path: str = "/usr/local/airflow/dbt",
         test_behavior: Literal["none", "after_each", "after_all"] = "after_each",
-        dbt_tags: List[str] = [],
+        select: Dict[str, Any] = {},
```
jlaneve, line 39: could/should this be something like:

```suggestion
        select: Dict[str, List[str]] = {},
```

cosmos/providers/dbt/render.py
```diff
@@ -25,7 +25,7 @@ def render_project(
     test_behavior: Literal["none", "after_each", "after_all"] = "after_each",
     emit_datasets: bool = True,
     conn_id: str = "default_conn_id",
-    dbt_tags: List[str] = [],
+    select: Dict = None,
```
jlaneve, line 28: could/should this be something like:
```suggestion
    select: Dict[str, List[str]] = None,
```

cosmos/providers/dbt/task_group.py
```diff
@@ -36,7 +36,7 @@ def __init__(
         emit_datasets: bool = True,
         dbt_root_path: str = "/usr/local/airflow/dbt",
         test_behavior: Literal["none", "after_each", "after_all"] = "after_each",
-        dbt_tags: List[str] = [],
+        select: Dict[str, Any] = {},
```
jlaneve, line 39: could/should this be something like:
```suggestion
        select: Dict[str, List[str]] = {},
```

pohek321: Added.

cosmos/providers/dbt/render.py
pohek321: Added.

cosmos/providers/dbt/render.py
```diff
@@ -56,8 +59,24 @@ def render_project(
...
-        if dbt_tags and not set(dbt_tags).intersection(model.config.tags):
-            continue
+
+        # ensures the same tag isn't in select & exclude
+        if "tags" in select and "tags" in exclude:
+            if set(select["tags"]).intersection(exclude["tags"]):
+                raise AirflowException(
+                    f"Can't specify the same tag in `select` and `include`: "
+                    f"{set(select['tags']).intersection(exclude['tags'])}"
+                )
```
jlaneve, line 69: we should be able to move this up, outside of the for loop for efficiency (the input tags won't change across each model)

jlaneve (approved):
just one small comment about where the exclude/select check is, but after that's addressed feel free to merge!

cosmos/providers/dbt/render.py
pohek321: Done.

Issue #89, opened 2023-01-23
Add support for "Path" select/exclude to DbtDag & DbtTaskGroup parsers


pohek321:
See dbt docs on cli usage examples [here](https://docs.getdbt.com/reference/node-selection/syntax#examples)

```bash
$ dbt run --select path.to.my.models     # runs all models in a specific directory
$ dbt run --select path/to/models        # run models contained in path/to/models
$ dbt run --select path/to/my_model.sql  # run a specific model by its path
```

Ultimately, in our parsers, we should be able to have a new parameter that looks something like this:

```python
# (only select or exclude parameter would be specified with staging dir not both)
jaffle_shop = DbtTaskGroup(
    ...
    select={"paths": ["../models/staging"]} # runs all models in the staging directory,
    exclude={"paths": ["../models/staging"]} # runs all models in except the staging directory
)
```

pohek321:
@jlaneve wanted to discuss the following:

How applicable is this? Isn't this already being done with the `dbt_project_name` parameter of the parsers? I can't imagine a use case where this is useful?

PR #99, opened 2023-01-26
Add paths to select and exclude parameters for DbtDag & DbtTaskGroup

pohek321:
closes [89](https://github.com/astronomer/astronomer-cosmos/issues/89)

### DIFF
--- a/cosmos/providers/dbt/render.py
+++ b/cosmos/providers/dbt/render.py
@@ -96,8 +96,7 @@ def render_project(
         # filters out any specified paths
         if "paths" in exclude:
             root_directories = [
-                project.project_dir / path.strip("/")
-                for path in exclude.get("paths", [])
+                project.project_dir / path.strip("/") for path in exclude.get("paths")
             ]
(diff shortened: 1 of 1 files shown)


cosmos/providers/dbt/render.py
```diff
@@ -77,6 +84,24 @@ def render_project(
             if set(exclude["tags"]).intersection(model.config.tags):
                 continue
 
+        # filters down to a path within the project_dir
+        if "paths" in select:
+            root_directories = [
+                project.project_dir / path.strip("/")
+                for path in select.get("paths", [])
```
jlaneve, line 91: Since we're already checking if "paths" exists in select, we shouldn't need the special `select.get("paths", [])`

jlaneve (approved):
Nice! Just one minor comment

cosmos/providers/dbt/render.py
pohek321: Removed the empty list fallback. Resolved

PR #100, opened 2023-01-26
Make region optional in Snowflake connection

dwreeves:
Snowflake allows and if anything seems to encourage not adding in the `region` parameter to a Snowflake connection. The reason is because Snowflake recently [moved to a system](https://docs.snowflake.com/en/user-guide/admin-account-identifier.html#existing-accounts) where the "region" is no longer part of the URI.

Under the hood, Airflow just [passes to](https://github.com/apache/airflow/blob/d9cefcd0c50a1cce1c3c8e9ecb99cfacde5eafbf/airflow/providers/snowflake/hooks/snowflake.py#L285) the Snowflake SQLAlchemy `URL()`, and under the hood Snowflake's `URL()` [allows for](https://github.com/snowflakedb/snowflake-sqlalchemy/blob/7c8effdb8c38f37084d5165961df75cbbbc10ddf/src/snowflake/sqlalchemy/util.py#L49) `region` to be optional.

### Fix

`region` should be allowed to be optional, basically. I wanted to do an approach that did not rely on magically trimming a trailing `.`, so something that explicitly checks for when region is not specified.

There were two approaches I could go with that meets the requirement: either fix it in Python, or fix it in `profiles/snowflake.py`. I went with the former approach because doing lots of things in Jinja2 is usually harder to maintain than doing the equivalent in Python, but it would have looked like this:

```python
"account": "{{ env_var('SNOWFLAKE_ACCOUNT') ~ '.' ~ env_var('SNOWFLAKE_REGION')"
           " if env_var('SNOWFLAKE_REGION')"
           " else env_var('SNOWFLAKE_ACCOUNT') }}",
```

### Pyproject.toml change

I also added `pytest` to the project as it wasn't specified as a dependency, but the `tests/` folder uses it.

### DIFF
--- a/cosmos/providers/dbt/core/utils/profiles_generator.py
+++ b/cosmos/providers/dbt/core/utils/profiles_generator.py
@@ -79,7 +79,12 @@ def create_profile_vars(conn: Connection, database, schema):
 
     elif conn.conn_type == "snowflake":
         profile = "snowflake_profile"
-        extras = {"account": "account", "region": "region", "role": "role", "warehouse": "warehouse"}
+        extras = {
+            "account": "account",
+            "region": "region",
+            "role": "role",
+            "warehouse": "warehouse",
+        }
 
         # At some point the extras removed a prefix extra__snowflake__ when the provider got updated... handling that
         # here.
@@ -88,11 +93,13 @@ def create_profile_vars(conn: Connection, database, schema):
                 extras[key] = f"extra__snowflake__{value}"
 
         # Region is optional
-        region = conn.extra_dejson.get(extras['region'])
-        if region:
+        region = conn.extra_dejson.get(extras["region"])
+        account = conn.extra_dejson.get(extras["account"])
+
+        if region and region not in account:
             account = f"{conn.extra_dejson.get(extras['account'])}.{region}"
         else:
-            account = conn.extra_dejson.get(extras['account'])
+            account = conn.extra_dejson.get(extras["account"])
 
         profile_vars = {
             "SNOWFLAKE_USER": conn.login,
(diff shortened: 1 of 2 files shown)


pohek321:
@dwreeves could you give me an example of what the formatted snowflake URI might look like (obviously without an actual password)? The way this is set up right now ensures that if there's a working snowflake connection in Airflow, then it can be used for cosmos. When I take away the region for a snowflake region, the connection is no longer valid *in Airflow*

dwreeves:
@pohek321 Hmm, that's interesting.

To be clear, if you are on the legacy system for Snowflake account IDs, and you want to construct a URI without the region field, you move it over to the account part of the URI.

I'm on provider package 2.7.0 and my secret looks something like this and it works fine:

```json
{"conn_type": "snowflake", "user": "SOME_USER", "password": "SOME_PASSWORD", "extra": "{\"account\": \"12345.us-east-1\", \"warehouse\": \"SOME_WAREHOUSE\", \"database\": \"SOME_DATABASE\", \"role\": \"SOME_ROLE\"}"}
```

Note that it's `{"account": "12345.us-east-1"}`. Removing the region from that part of the URI for legacy Snowflake connections only works for `us-west-1` I believe.

For good measure, my Airflow dbt integration does the following. (1) Places the secrets directly in the .yml, i.e. I'm not using `{{ env(...) }}`, (2) dump the .yml file in `/tmp` (FWIW, I think I like your approach better than mine):

```python
conn = self.snowflake_hook.get_connection(conn_id)
profile = self.dbt_profile_name: {
    "outputs": {
        self.dbt_target_name: {
            "account": conn.extra_dejson.get("account"),
            ...
        }
    }
}
```

For whatever it's worth, I also am the last person who committed to the docs for the Snowflake connection 😅 https://github.com/apache/airflow/blob/main/docs/apache-airflow-providers-snowflake/connections/snowflake.rst

pohek321:
@dwreeves, I appreciate your patience while I wrap my head around this. 

I think this change would make sense, so in Airflow, your connection would look something like this, right:

Screenshot 2023-01-27 at 2 06 30 PM

dwreeves:
Turns out yes, we can remove the `host` from the UI. 😄 https://github.com/apache/airflow/pull/29208

pohek321:
@dwreeves, good to know! I did what you suggested, and it worked! I did not realize that the `host` parameter was optional. 

However, I was able to break it (and by break, I mean the connection test in Airflow succeeds, but Cosmos cannot authenticate). It broke by putting `<account>.<region>` in the account field as well as `<region>` in the region field -- interestingly enough, Airflow was OK with this (Cosmos was not).

So I added a couple of lines to handle that scenario! If you are cool with the commit, then I'll approve and merge, and we'll tag a new release so you can access these features.

Additionally, thank you so much for your contributions! If you need anything else (or have ideas for improvements), don't hesitate to reach out or open issues. 🍻

pohek321 (approved):
Thanks, @dwreeves! Approved.

PR #101, opened 2023-01-26
Add cancel on kill functionality

dwreeves:
In dbt, [a `SIGTERM` does not cancel any running queries](https://medium.com/@brianepohl/terminating-dbt-in-dagster-kubernetes-job-c53c3bc26012), but a `SIGINT` does. The `SubprocessHook()` does not have a `send_sigint()` method, so we can just copy+paste the source code for it and replace the `SIGTERM` with `SIGINT`.

### DIFF
--- a/cosmos/providers/dbt/core/operators.py
+++ b/cosmos/providers/dbt/core/operators.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import signal
 from typing import List, Sequence
 
 import yaml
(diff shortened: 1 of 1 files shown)


jlaneve (approved):
nice, very good catch and thanks for the contribution!

dwreeves:
@jlaneve Thanks for merging! I have experience integrating dbt and Airflow together and I want to use this integration (I hate managing my own), so I'll be keeping an eye on this project and add a few more things as I get time.

jlaneve:
@dwreeves that would be awesome! Feel free to also open issues with things you think are missing - we can certainly help implement

dwreeves:
Will do. I have at least one nontrivial feature I'd like to see and that will involve some discussion. I'll get to opening an issue hopefully in the next week.

PR #103, opened 2023-01-26
Use targeted database for BigQuery profile

### LINKED ISSUE #102: `dbt_args` with `"db_name"` target is not used for BigQuery profiles
The database (project) used for the dbt target for BigQuery is currently always the project of the service account. For example in a `DbtTaskGroup` if the following is passed:

```python
    run_dbt = DbtTaskGroup(
        group_id="run_dbt",
        dbt_project_name="my_dbt_project",
        dbt_root_path="dbt",
        dbt_args={
            "db_name": "gcp-target-project",
            "schema": "my-dataset",
        },
        dag=dag,
    )
```
`"db_name": "gcp-target-project"` is not used and the GCP project targeted is the one of the service account. This is because in the profiles generator, where `"BIGQUERY_PROJECT"` is always set to the service account project id even though the dataset (schema) is set correctly:
https://github.com/astronomer/astronomer-cosmos/blob/c375f79c03dd36865931273cbc5efd7b722d95ba/cosmos/providers/dbt/core/utils/profiles_generator.py#L112-L118

The fix (I think) would be simply updating it to be "database" instead.

jbandoro:
Fixes #102, the diffs in this PR are from the pre-commit black formatter, when it was actually only a single line update. I tested and it works as expected, though not sure if we want to default to the service-account project if `db_name` is None.

### DIFF
--- a/cosmos/providers/dbt/core/utils/profiles_generator.py
+++ b/cosmos/providers/dbt/core/utils/profiles_generator.py
@@ -180,7 +180,10 @@ def map_profile(conn_id, db_override=None, schema_override=None):
         db = db_override
     else:
         if conn.conn_type != "snowflake":
-            db = conn.schema
+            if conn.conn_type == "google_cloud_platform":
+                db = json.loads(conn.extra_dejson.get("keyfile_dict"))["project_id"]
+            else:
+                db = conn.schema
         else:
             # At some point the extras removed a prefix extra__snowflake__ when the provider got updated... handling
             # that here.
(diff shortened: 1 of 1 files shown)


pohek321:
@jbandoro wanted to confirm this value would be the equivalent of a GCP project_id, but your change will make it configurable, correct?

jbandoro:
> @jbandoro wanted to confirm this value would be the equivalent of a GCP project_id, but your change will make it configurable, correct?

yes that's correct. ATM the target GCP project_id is that of the GCP service account, and this PR will allow the project_id (dbt database) to be specified in `dbt_args`, for the case where a service account has access to many GCP projects.

pohek321:
@jbandoro cool, I think it makes sense to defer a project_id (or database) from the service account when the `db_name` isn't specified, which is why I added that last commit. Thank you for this contribution; it makes a lot of sense and is much appreciated! I will have @jlaneve take a look before we approve, just so that my commit is also getting reviewed. Once we get his 👍 then we can merge!

jlaneve (approved):
nice, this change looks good! eventually as the connection logic continues to grow we may want to find a better/cleaner way to define this, but I think this is good for now

Issue #90, opened 2023-01-23
Add support for "Config" select/exclude to DbtDag & DbtTaskGroup parsers


pohek321:
See dbt docs on cli usage examples [here](https://docs.getdbt.com/reference/node-selection/methods#the-config-method)

```bash
$ dbt run --select config.materialized:incremental    # run all models that are materialized incrementally
$ dbt run --select config.schema:audit                # run all models that are created in the `audit` schema
$ dbt run --select config.cluster_by:geo_country      # run all models clustered by `geo_country`
```

Ultimately, in our parsers, we should be able to have a new parameter that looks something like this:

```python
# (Either the select or exclude parameter would be specified with the materialized:table config - not both)
jaffle_shop = DbtTaskGroup(
    ...
    select={'configs': ['materialized:table']} # runs all models that are materialized as a table
    exclude={'configs': ['materialized:table']] # runs all models that don't materialize as a table
)
```

pohek321:
Here are the following ways to specify configs for a model
1. Using a `config()` Jinja macro with a model
2. Using a `config` resource property in a `.yml` file
3. From the `dbt_project.yml` file, under the `models:` key

See more information [here](https://docs.getdbt.com/reference/model-configs#configuring-models)

PR #104, opened 2023-01-27
Add support for "Config" select/exclude to DbtDag & DbtTaskGroup parsers

pohek321:
Closes [90](https://github.com/astronomer/astronomer-cosmos/issues/90)

I added the label `bug` because I found syntax preventing tags from being read out of `properties.yml` files correctly.


pohek321:
This will convert the DbtModelConfig to this:

```text
DbtModel(name='orders', path='dbt/jaffle_shop/models/orders.sql', config=DbtModelConfig(config_selectors={'schema:stg', 'materialized:view', 'tags:tag_4', 'tags:tag_3', 'tags:yet_another_test_tag', 'tags:tag_1'}, upstream_models={'stg_payments', 'stg_orders'}))
```

...as opposed to this:

```text
DbtModel(name='orders', path='dbt/jaffle_shop/models/orders.sql', config=DbtModelConfig(tags={'tag_4', 'tag_3'}, upstream_models={'stg_payments', 'stg_orders'}))
```

which will really simplify the process for a customer when it comes to using `select`/`exclude` because this is similar to how dbt does it (see [here](https://docs.getdbt.com/reference/node-selection/methods#the-config-method)):

```python
    jaffle_shop = DbtTaskGroup(
        conn_id="snowflake_chronek",
        dbt_args={"dbt_executable_path": "/usr/local/airflow/dbt_venv/bin/dbt"},
        test_behavior='after_all',
        select={ #this could also be an exclude parameter with the same value for the opposite effect
          "configs": ['schema:stg']
        },
        dag=dag,
    )
```

...additionally, it will be easier for our parser to do a single `intersection` instead of multiple for various selector types. LMK if you have any questions

cosmos/providers/dbt/parser/project.py
```diff
@@ -21,19 +21,58 @@ class DbtModelConfig:
...
+        )
+
         # get the unique combination of each list
         return DbtModelConfig(
-            tags=self.tags | other_config.tags,
+            config_selectors=result,
             upstream_models=self.upstream_models | other_config.upstream_models,
         )
 
+    def _config_selector_ooo(
```
jlaneve, line 45: what's `ooo` here?

jlaneve: oh order of operations

cosmos/providers/dbt/parser/project.py
```diff
@@ -21,19 +21,58 @@ class DbtModelConfig:
...
     upstream_models: set[str] = field(default_factory=set)
 
     def __add__(self, other_config: DbtModelConfig) -> DbtModelConfig:
         """
         Add one config to another. Necessary because configs can come from different places
         """
+
+        # ensures proper order of operations between sql models and properties.yml
+        result = self._config_selector_ooo(
+            prefixes=["materialized", "schema"],
```
jlaneve, line 34: what do you think about making this a class variable? are we ever going to call `_config_selector_ooo` with a different set of prefixes?

cosmos/providers/dbt/parser/project.py
```diff
@@ -164,22 +215,38 @@ def _handle_config_file(self, path: Path) -> None:
...
 
-            # parse out the config fields we can recognize
-
-            # 'tags' is either a string or list of strings
-            tags = config.get("tags", [])
-            if isinstance(tags, str):
-                tags = [tags]
+            # config_selectors
+            config_selectors = []
+            for selector in ["materialized", "schema", "tags"]:
```
jlaneve, line 227: can we move the definition of the list of supported selectors to the top of the file as a constant? looks like we use it in multiple places

cosmos/providers/dbt/parser/project.py
pohek321: I think we'll only add to that prefixes list - take a look at my most recent commit and let me know what you think

jlaneve (approved):

PR #108, opened 2023-01-31
Databricks/databricks workflow taskgroup

dimberman:
This pull request creates a new Airflow task group called the DatabricksWorkflowTaskGroup. Databricks workflows are a popular new Databricks product that allows users to optimize their tasks for use on Databricks "job" clusters. The DatabricksWorkflowTaskGroup reads all tasks in the task group and submits them to Databricks as one job. Tasks will need to implement the "convert_to_databricks_workflow_task" function to work in a Databricks task group.

### DIFF
--- a/cosmos/providers/databricks/notebook.py
+++ b/cosmos/providers/databricks/notebook.py
@@ -1,0 +1,225 @@
+"""DatabricksNotebookOperator for submitting notebook jobs to databricks."""
+from __future__ import annotations
+
+import time
+from typing import Any
+
+from airflow.exceptions import AirflowException
+from airflow.models.operator import BaseOperator
+from airflow.providers.databricks.hooks.databricks import DatabricksHook
+from airflow.utils.context import Context
+from databricks_cli.runs.api import RunsApi
+from databricks_cli.sdk.api_client import ApiClient
+
+
+class DatabricksNotebookOperator(BaseOperator):
+    """
+    Launches a notebook to databricks using an Airflow operator.
+
+    The DatabricksNotebookOperator allows users to launch and monitor notebook
+     deployments on Databricks as Aiflow tasks.
+    It can be used as a part of a DatabricksWorkflowTaskGroup to take advantage of job clusters,
+    which allows users to run their tasks on cheaper clusters that can be shared between tasks.
+
+    Here is an example of running a notebook as a part of a workflow task group:
+
+    .. code-block: python
+
+        with dag:
+            task_group = DatabricksWorkflowTaskGroup(
+                group_id="test_workflow",
+                databricks_conn_id="databricks_conn",
+                job_clusters=job_cluster_spec,
+                notebook_params=[],
+            )
+            with task_group:
+                notebook_1 = DatabricksNotebookOperator(
+                    task_id="notebook_1",
+                    databricks_conn_id="databricks_conn",
+                    notebook_path="/Users/daniel@astronomer.io/Test workflow",
+                    source="WORKSPACE",
+                    job_cluster_key="Shared_job_cluster",
+                )
+                notebook_2 = DatabricksNotebookOperator(
+                    task_id="notebook_2",
+                    databricks_conn_id="databricks_conn",
+                    notebook_path="/Users/daniel@astronomer.io/Test workflow",
+                    source="WORKSPACE",
+                    job_cluster_key="Shared_job_cluster",
+                    notebook_params={
+                        "foo": "bar",
+                    },
+                )
+                notebook_1 >> notebook_2
+
+
+        :param notebook_path: the path to the notebook in Databricks
+        :param source: Optional location type of the notebook. When set to WORKSPACE, the notebook will be retrieved
+        from the local Databricks workspace. When set to GIT, the notebook will be retrieved from a Git repository
+        defined in git_source. If the value is empty, the task will use GIT if git_source is defined
+        and WORKSPACE otherwise. For more information please visit
+        https://docs.databricks.com/dev-tools/api/latest/jobs.html#operation/JobsCreate
+        :param databricks_conn_id: the connection id to use to connect to Databricks
+        :param notebook_params: the parameters to pass to the notebook
+        :param kwargs:
+    """
+
+    template_fields = ("databricks_run_id",)
+
+    def __init__(
+        self,
+        notebook_path: str,
+        source: str,
+        databricks_conn_id: str,
+        notebook_params: dict | None = None,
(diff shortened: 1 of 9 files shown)


cosmos/providers/databricks/notebook.py
```diff
@@ -0,0 +1,199 @@
...
+        databricks_conn = hook.get_conn()
+        api_client = ApiClient(
+            host=databricks_conn.host,
+            token=databricks_conn.password,
+        )
+        runs_api = RunsApi(api_client)
+        current_task = {
+            x["task_key"]: x for x in runs_api.get_run(self.databricks_run_id)["tasks"]
+        }[self.dag_id + "__" + self.task_id.replace(".", "__")]
+        while (
```
tatiana, line 134: Should we set some timeout here?

cosmos/providers/databricks/notebook.py
```diff
@@ -0,0 +1,199 @@
...
+        api_client = ApiClient(
+            user=databricks_conn.login,
+            password=databricks_conn.password,
+            host=databricks_conn.host,
+        )
+        runs_api = RunsApi(api_client)
+        run = runs_api.submit_run(
+            {
+                "new_cluster": {
+                    "spark_version": "6.6.x-scala2.11",
```
tatiana, line 171: Should these parameters be configurable?

cosmos/providers/databricks/notebook.py
```diff
@@ -0,0 +1,199 @@
...
+
+        while (
+            runs_api.get_run(current_task["run_id"])["state"]["life_cycle_state"]
+            == "RUNNING"
+        ):
+            print(f"task {self.task_id.replace('.', '__')} running")
+            time.sleep(5)
+
+        final_state = runs_api.get_run(current_task["run_id"])["state"]
+        if final_state.get("life_cycle_state", None) != "TERMINATED":
```
tatiana, line 149: Where can we find out more about the existing current life cycle states?

cosmos/providers/databricks/notebook.py
```diff
@@ -0,0 +1,199 @@
...
+            raise AirflowException(
+                f"Databricks job failed with state {final_state}. Message: {final_state['state_message']}"
+            )
+        if final_state["result_state"] != "SUCCESS":
+            raise AirflowException(
+                "Task failed. Reason: %s", final_state["state_message"]
+            )
+
+    def launch_notebook_job(self):
+        """Launch the notebook as a one-time job to Databricks."""
```
tatiana, line 159: Would there be alternatives to this way of launching the notebook?

cosmos/providers/databricks/notebook.py
```diff
@@ -0,0 +1,199 @@
...
+        from the local Databricks workspace. When set to GIT, the notebook will be retrieved from a Git repository
+        defined in git_source. If the value is empty, the task will use GIT if git_source is defined
+        and WORKSPACE otherwise. For more information please visit
+        https://docs.databricks.com/dev-tools/api/latest/jobs.html#operation/JobsCreate
+        :param databricks_conn_id: the connection id to use to connect to Databricks
+        :param notebook_params: the parameters to pass to the notebook
+        :param kwargs:
+    """
+
+    template_fields = ("databricks_run_id",)
```
tatiana, line 66: WDYT of having the `notebook_path` as a template field as well?

cosmos/providers/databricks/notebook.py
```diff
@@ -0,0 +1,199 @@
...
+            password=databricks_conn.password,
+            host=databricks_conn.host,
+        )
+        runs_api = RunsApi(api_client)
+        run = runs_api.submit_run(
+            {
+                "new_cluster": {
+                    "spark_version": "6.6.x-scala2.11",
+                    "node_type_id": "Standard_DS3_v2",
+                },
```
tatiana, line 173: Should these parameters be configurable by the user?
Do we always create a new cluster? Could the user be running on an existing cluster?

tests/databricks/test_workflow.py
```diff
@@ -0,0 +1,144 @@
+from unittest import mock
```
tatiana, line 1: Since we don't have a CI running this yet, could you share how is the test coverage of the databricks provider ATM?

cosmos/providers/databricks/notebook.py
tatiana: I discussed this with @dimberman  and he suggested, for now, we don't introduce more files - and that if users need, we can add more fields to template fields.

cosmos/providers/databricks/notebook.py
tatiana: We agreed to do this as a separate ticket - @dimberman to help log this ticket and adding the reference here

dimberman: https://github.com/astronomer/astronomer-cosmos/issues/145 linked here

cosmos/providers/databricks/notebook.py
dimberman: @tatiana I've added a state for TERMINATING, the rest of the states should be handled by the error message.

cosmos/providers/databricks/notebook.py
dimberman: Yes I've added that option

tests/databricks/test_workflow.py
dimberman: @tatiana here is the current coverage report

```
cosmos/providers/databricks/notebook.py      72     11    85%
cosmos/providers/databricks/workflow.py      92      8    91%
```

It's a bit difficult to get the numbers up further without a tool like codecov to show specific lines but I think that we can start here and add more tests as we add more functionality. I'd also like to set up codecov for this project sooner than later.

tatiana (approved):
@dimberman looks good. Thanks a lot for addressing the comments!
1. Have the integration tests running in CI/CD (please add a ticket for this before merging?)
2. Review in which repo this code lives

tatiana (approved):

cosmos/providers/databricks/notebook.py
```diff
@@ -0,0 +1,225 @@
...
+                    task_id="notebook_2",
+                    databricks_conn_id="databricks_conn",
+                    notebook_path="/Users/daniel@astronomer.io/Test workflow",
+                    source="WORKSPACE",
+                    job_cluster_key="Shared_job_cluster",
+                    notebook_params={
+                        "foo": "bar",
+                    },
+                )
+                notebook_1 >> notebook_2
```
kaxil, line 53: oh man, love this

PR #113, opened 2023-02-02
Issue 107

### LINKED ISSUE #107: Databricks connection, can't use schema
The databricks connection schema gets set in the db variable here
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/providers/dbt/core/utils/profiles_generator.py#L195
and then ultimately is missed over here because it's expecting it in the schema variable
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/providers/dbt/core/utils/profiles_generator.py#L164

The end result is 
`[2023-01-30, 14:12:14 GMT] {profiles_generator.py:223} ERROR - Please specify a target schema using the schema parameter.`

For the moment I've worked around it by using the schema override option.

MikeWallis42:
Fixes #107 

I had to bump the version of isort because I was getting these errors
https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co

I split out the various profiles to make it easier to unit test and I think I found a couple of edge cases that I've marked with pytest xfail.
It might be easier long term to go further in to breaking this out as each connection/provider seems to be quite distinct in what it requires with little overlap.

Disclaimer, I haven't tested this using cosmos-dev as I haven't been able to get it to run locally.


pohek321:
Hey, @MikeWallis42 I'll spin up a sandbox and test this as part of the review today! Thank you for your contributions to the project!

jlaneve (approved):
nice, this looks like a really great change - going to make it much easier to deal with these connections as the list continues to grow!

pohek321:
Closes [132](https://github.com/astronomer/astronomer-cosmos/issues/132)

pohek321:
@MikeWallis42 thank you for your contributions here! This is excellent. I added a commit to fix the databricks connection as it was breaking, but everything else was perfect! Great work.

MikeWallis42:
@pohek321 Thank you for testing and reviewing for me, I just double checked with the new version and I'm seeing the same issue however. I think my unit tests would have caught it.
Would it be possible to add the unit tests as a github action?

I'll take a look and hopefully submit another PR.

Issue #118, opened 2023-02-03
DbtTaskGroup should not need `dag=dag`


denimalpaca:
Like other task groups, if using the `dag` context manager, the `DbtTaskGroup` shouldn't need the `dag=dag` argument passed in.

PR #119, opened 2023-02-04
Make `dag` arg optional for CosmosTaskGroup

josh-fell:
Closes: #118

Typically when authoring with TaskGroups in Airflow, users will not explicitly pass a DAG object but rather set the TaskGroup within a DAG context manager. Making the `dag` arg optional in the CosmosTaskGroup will allow the same authoring practice. In Airflow, if a `dag` arg is not passed to a TaskGroup, behind the scenes `airflow.models.DagContext.get_current_dag()` is called to retrieve the DAG object. Since CosmosTaskGroup pushes the `dag` arg to its super class, and early enough in construction, the same functional behavior is triggered.


jlaneve (approved):
Nice, thank you!

PR #123, opened 2023-02-06
fixes #72 common API for dbt datasets

### LINKED ISSUE #72: Add utility to get `Dataset` from dbt project + model name
Right now, a user has to write out `DBT://{airflow_conn_id}/{project_name}/{model_name}` themselves, which can be prone to errors if there are typos, capitalization issues, or if the format changes. We should create a dataset util that can be used along the lines of:
```python
from cosmos.providers.dbt import get_dataset
get_dataset("my_conn_id", "project_name", "model_name")
```

We should also use it internally, so there's a single source of truth for how datasets are defined.

MikeWallis42:
fixes #72


jlaneve (approved):
nice, this looks awesome! the only piece of feedback I have is if you think this would be better as something like `DbtDataset` vs `get_dbt_dataset` - I know in the original issue I wrote out `get_dbt_dataset`, but would love to get your opinion too

MikeWallis42:
At this point I don't think it makes sense to provide an object oriented solution.
Perhaps at some point in the future if there's some benefit to be had because of the other providers that emerge.
From a project maintenance point of view it's probably simpler to deprecate this function than a class implementation 😄

Issue #134, opened 2023-02-08
Ensure no environment variables are set to `None` before running the subprocess


jlaneve:
If an env var is `None`, you can get this weird/obscure error (e.g. don't set a role with a snowflake connection):

```
[2023-02-08, 14:37:12 UTC] {subprocess.py:75} INFO - Running command: [‘/home/astro/.venv/dbt/bin/dbt’, ‘run’, ‘--project-dir’, ‘/usr/local/airflow/include/dbt/jaffle_shop’, ‘--models’, ‘customers’, ‘--profile’, ‘snowflake_profile’]
[2023-02-08, 14:37:12 UTC] {taskinstance.py:1772} ERROR - Task failed with exception
Traceback (most recent call last):
  File “/usr/local/lib/python3.9/site-packages/cosmos/providers/dbt/core/operators.py”, line 336, in execute
    result = self.build_and_run_cmd(env=self.get_env(context))
  File “/usr/local/lib/python3.9/site-packages/cosmos/providers/dbt/core/operators.py”, line 261, in build_and_run_cmd
    result = self.run_command(cmd=dbt_cmd, env=env)
  File “/usr/local/lib/python3.9/site-packages/cosmos/providers/dbt/core/operators.py”, line 219, in run_command
    result = self.subprocess_hook.run_command(
  File “/usr/local/lib/python3.9/site-packages/airflow/hooks/subprocess.py”, line 77, in run_command
    self.sub_process = Popen(
  File “/usr/local/lib/python3.9/subprocess.py”, line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File “/usr/local/lib/python3.9/subprocess.py”, line 1741, in _execute_child
    env_list.append(k + b’=' + os.fsencode(v))
  File “/usr/local/lib/python3.9/os.py”, line 810, in fsencode
    filename = fspath(filename)  # Does type-checking of `filename`.
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

jlaneve:
I _think_ the solution here could just be removing env vars with value = `None` in the `get_env` function, but need to test it to be sure.

https://github.com/astronomer/astronomer-cosmos/blob/8072203a30333271af6e5087b53c5170ff9e26bf/cosmos/providers/dbt/core/operators.py#L138-L156

MikeWallis42:
[...]

Issue #135, opened 2023-02-08
Need documentation for the available adapters


pohek321:
It'd be nice to have some documentation on how to properly configure connections & order of operations for db/schema params for our currently supported adapters:

- [x] dbt-snowflake
- [x] dbt-redshift
- [x] dbt-databricks
- [x] dbt-bigquery
- [x] dbt-postgres

PR #140, opened 2023-02-13
refactor dbt adapters

### LINKED ISSUE #107: Databricks connection, can't use schema

MikeWallis42:
fixes #107 #135 

Going a bit further than the last time I've fully separated each adapter out but provided a unit test to make sure that the signatures and return annotations are aligned.

There is no longer a validation of whether there is a database or schema provided in all of the adapters.
We should probably provide a validation function for each to check for missing mandatory fields, particularly as some have varying authentication methods... not that they're implemented just yet.

I've also added unit testing for creation of the profiles yaml file as I touched it.

Finally I added internal references and had a go at starting the external facing documentation for the adapters whilst I had to familiarise myself with the details of each.
For the external facing documentation I focussed on the key differences between the adapters.


pohek321:
@MikeWallis42 going through and testing each connection type to ensure everything is still ship-shape (so far so good!). Could you give a CLI command demonstrating the tests these would allow us to run if implemented? I think that'd help understand these changes high level for those who are looking at this for the first time.

cosmos/providers/dbt/core/profiles/bigquery.py
```diff
@@ -21,3 +27,34 @@
     },
     "target": "dev",
 }
+
+
+def create_profile_vars_google_cloud_platform(
+    conn: "Connection",
+    database_override: Optional[str] = None,
+    schema_override: Optional[str] = None,
+) -> Tuple[str, dict]:
```
jlaneve, line 36: nit: could we type this dict?

cosmos/providers/dbt/core/profiles/databricks.py
```diff
@@ -1,13 +1,45 @@
...
+    from airflow.models import Connection
+
 databricks_profile = {
     "outputs": {
         "dev": {
             "type": "databricks",
             "host": "{{ env_var('DATABRICKS_HOST') }}",
             "catalog": "{{ env_var('DATABRICKS_CATALOG') }}",
-            "schema": "{{ env_var('DATABRICKS_DATABASE') }}",
+            "schema": "{{ env_var('DATABRICKS_SCHEMA') }}",
```
jlaneve, line 12: @pohek321 do you remember if there was a reason we called this `DATABRICKS_SCHEMA` instead of `DATABRICKS_DATABASE`?

pohek321: yes, in databricks-land a database is really a schema and a catalog is a database. Super confusing - I'm putting together a commit to fix and will push it up momentarily.

pohek321: Additionally, I was mirroring all other conns in Airflow that use `schema` as a db parameter (with the exception of Snowflake) -- so that's why it's this way. The commit I'm putting together will also reflect that.

cosmos/providers/dbt/core/operators.py
```diff
@@ -227,7 +228,9 @@ def run_command(self, cmd, env):
         return result
 
     def build_and_run_cmd(self, env: dict, cmd_flags: list = None):
-        create_default_profiles()
+        create_default_profiles(
+            Path(os.path.expanduser("~")).joinpath(".dbt/profiles.yml")
+        )
```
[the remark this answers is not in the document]
MikeWallis42: I had a good read of the [docs](https://docs.getdbt.com/docs/get-started/connection-profiles#connecting-to-your-warehouse-using-the-command-line)
If we were to create a constant for where we loaded the file and also set DBT_PROFILES_DIR (or add it as a global flag) then it could get around any complications where users have loaded a profiles.yml in to the root of their dbt project, thereby superseding the Cosmos setup.

cosmos/providers/dbt/core/profiles/databricks.py
MikeWallis42: The complication that I was having was that the "database" is actually a Unity Catalog which is optional and there for the purposes of data lineage and asset management [docs](https://www.databricks.com/product/unity-catalog).
With dbt it looks like it was introduced as a seperate config item in [v1.4.0](https://github.com/databricks/dbt-databricks/blob/main/CHANGELOG.md#dbt-databricks-140-january-25-2023). Prior to that it was acceptable to use dot seperation in the schema field.

We're not currently using the Unity Catalog and our schemas are AWS Glue databases which reference s3 locations.

As it's optional, if it's not populated then there will be a dependency on fix #134

pohek321:
@jlaneve & @MikeWallis42 can you both take a look at my [last commit](https://github.com/astronomer/astronomer-cosmos/pull/140/commits/902fd45fe27de878807dc9d3a6be48f1e8dbed11)? Our connectivity DAGs were breaking before making this change and this resolved it. Here's what I did purely based on Airflow connection patterns and what's standard in Airflow:

- BigQuery: 
  - Project (DB): use the `database_override` first, if none is specified, then use the `project_id` from the json creds
  - Dataset (Schema): require a `dataset` to be specified via the `schema` parameter. If the schema parameter isn't passed - fail the task
- Databricks 
  - Catalog (DB): use the `database_override` first, and the `conn.schema` second. If neither of those is specified default to the value `hive_metastore` (more details [here](https://docs.databricks.com/data-governance/unity-catalog/hive-metastore.html#default-catalog))
  - Database (Schema): use the `schema_override` for the Databricks Database name to follow Airflow connection parameter patterns. If the schema isn't passed - fail the task
- Postgres: 
  - Database (DB): use the `database_override` first, and the `conn.schema` second. If neither of those is specified then fail
  - Schema (Schema): use the `schema_override`, if the schema parameter isn't passed - fail the task. 
- Redshift:
  - Database (DB): use the `database_override` first, and the `conn.schema` second. If neither of those is specified then fail
  - Schema (Schema): use the `schema_override`, if the schema parameter isn't passed - fail the task.

Open to having a discussion to change them, but since these workflows are being executed *in Airflow* I think it makes the most sense to follow Airflow patterns - i.e. Connection Schema parameter is traditionally used for a database.

cosmos/providers/dbt/core/profiles/databricks.py
pohek321: @jlaneve & @MikeWallis42 can you both take a look at my [last commit](https://github.com/astronomer/astronomer-cosmos/pull/140/commits/902fd45fe27de878807dc9d3a6be48f1e8dbed11)? Our connectivity DAGs were breaking before making this change and this resolved it. Here's what I did purely based on Airflow connection patterns and what's standard in Airflow:

- BigQuery: 
  - Project (DB): use the `database_override` first, if none is specified, then use the `project_id` from the json creds
  - Dataset (Schema): require a `dataset` to be specified via the `schema` parameter. If the schema parameter isn't passed - fail the task
- Databricks 
  - Catalog (DB): use the `database_override` first, and the `conn.schema` second. If neither of those is specified default to the value `hive_metastore` (more details [here](https://docs.databricks.com/data-governance/unity-catalog/hive-metastore.html#default-catalog))
  - Database (Schema): use the `schema_override` for the Databricks Database name to follow Airflow connection parameter patterns. If the schema isn't passed - fail the task
- Postgres: 
  - Database (DB): use the `database_override` first, and the `conn.schema` second. If neither of those is specified then fail
  - Schema (Schema): use the `schema_override`, if the schema parameter isn't passed - fail the task. 
- Redshift:
  - Database (DB): use the `database_override` first, and the `conn.schema` second. If neither of those is specified then fail
  - Schema (Schema): use the `schema_override`, if the schema parameter isn't passed - fail the task.

Open to having a discussion to change them, but since these workflows are being executed *in Airflow* I think it makes the most sense to follow Airflow patterns - i.e. Connection Schema parameter is traditionally used for a database.

cosmos/providers/dbt/core/operators.py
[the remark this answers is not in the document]
pohek321: @MikeWallis42 would we just set that `DBT_PROFILES_DIR` to "~/.dbt/profiles.yml" so that users couldn't override the profiles.yml on their own?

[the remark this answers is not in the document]
MikeWallis42: @pohek321 seems like a good idea to me.
If we have it as a constant then and add it to `DbtBaseOperator.get_env` as well as make use of it in `create_default_profiles` that should close the gap.
I'll add a commit with the changes for review.

cosmos/providers/dbt/core/profiles/databricks.py
MikeWallis42: @pohek321 I think my issue with the databricks one is that the Unity catalog is optional and isn't enabled by default, [ref](https://docs.databricks.com/data-governance/unity-catalog/get-started.html#overview-of-unity-catalog-setup).
Also as per this [page](https://docs.databricks.com/data-governance/unity-catalog/index.html#the-unity-catalog-object-model)

> Catalog: The first layer of the object hierarchy, used to organize your data assets.
> Schema: Also known as databases, schemas are the second layer of the object hierarchy and contain tables and views.

So the schema is a database and catalog is a higher level of grouping.

Would it be better to define it as:-

1. Schema will be the level above an object such as table/view.
 - So in databricks the Airflow connection Schema will be the Databricks Schema
 - For BigQuery the Airflow connection Schema would be a Dataset. 
 - RDBMS: Schema = Schema.
2. Database will be the level above Schema.
 - Databricks: Unity Catalog.
 - BigQuery: Project
 - RDBMS: Database = Database

Is it the case that these connections to an RDBMS would normally use schema for database and then you'd pass the schema in the table name like `my_schema.my_table`?

pohek321: @MikeWallis42 

> So in databricks the Airflow connection Schema will be the Databricks Schema

I think you meant: So in databricks the Airflow connection Schema will be the Databricks *Database*

This is counterintuitive to all other Airflow connection objects because there isn't a single Airflow connection object that uses the `Schema` parameter for an actual schema (except Snowflake). So, this would be an anti-pattern if we defined it differently for Databricks. Databricks docs state that if a catalog *isn't* specified, then `hive_metastore` is used by default (see [here](https://docs.databricks.com/data-governance/unity-catalog/hive-metastore.html#default-catalog)). Our users could overwrite that when instantiating the cosmos class like this:

```python
  jaffle_shop = DbtTaskGroup(
      ...,
      conn_id="databricks_default",
      dbt_args={
          "db_name": "databricks_catalog" # overrides the `hive_metastore` default
          "schema": "databricks_db" # sets a default schema (or databricks database) - the value specified here can be overridden in dbt .yml/.sql files
      },
  )
```

Additionally, dbt treats `catalog.database` databricks objects the same way it treats `db.schema` rds objects. So if you specify the following in a dbt properties.yml:

```yaml
version: 2

models:
  - name: customers
    description: This table has basic information about a customer, as well as some derived facts based on a customer's orders
    config:
      schema: test_schema
```

It will create a be and new "Databricks Database" called `test_schema` like so:
Screenshot 2023-02-15 at 10 04 27 AM

So, in summary - we should keep the pattern that Airflow is following, doing so will also align with how dbt treats these objects:

(RDS `db.schema.table` = Databricks `catalog.database.table` = BigQuery `project_id.dataset.table`

MikeWallis42: @pohek321 thanks for getting back to me so early your time!
ok so basically my original issue should have just been a documentation update. It's quite confusing when you see database and schema being used interchangably across implementations.
- Databricks UI: database
- dbt: schema
- Airflow:
  - Connections; N/A but typically schema is a database.
  - Operators; expects you to pass them in as arguments but uses the terminology catalog and schema.
- cosmos: Pass in `{"dbt_args": {"schema": "my_schema"}}`

So basically there are two methods of using dbt and cosmos to write to a specific databricks database. Schema in dbt_args or like [this](https://docs.getdbt.com/reference/resource-configs/schema) as you showed above.
It will not be possible to add something on the Airflow connection to default this as the databricks connection doesn't to this date.
I doubt I'll be the first to question this so I'm happy to go with the desired implementation just as long as it's documented  😄

pohek321: @MikeWallis42 

Awesome, I went ahead and added a commit to bring clarity to all of this in our docs. See commit [0f37905](https://github.com/astronomer/astronomer-cosmos/pull/140/commits/0f37905cc4f0a0476c4389303622417435f0382c)

pohek321: Marking as resolved and approving.

pohek321 (approved):
@MikeWallis42 thanks for another great contribution!

PR #88, opened 2023-01-20
Add support for python3.7

jbandoro:
Fixes #87

Type hinting updates so can support Python 3.7

Ran the pre-commit hook which did isort and black updates, but there were a lot of E501 Line too long errors that I ignored.

PR #142, opened 2023-02-13
fixes #141 developer typing for dbt_args

### LINKED ISSUE #141: Developer typing for dbt_args
It would be good if there was an easy reference for what keys were available in dbt_args.
Because of the levels of abstraction you have to either know where to go or dig a few layers in to work this out.

MikeWallis42:
fixes #141 

Provided a TypedDict with all available dbt_args from the Operators and added it as typing for the argument on DbtDag and DbtTaskGroup.
Discovered that project_dir passed to DbtBaseOperator was already a Path and not a str so did some small refactoring there to remove the use of the os library.


cosmos/core/airflow.py
```diff
@@ -88,9 +88,11 @@ def get_airflow_task(
     task: Task, dag: DAG, task_group: Optional[TaskGroup] = None
 ) -> BaseOperator:
     """
-    Get the Airflow Operator class for a Task.
+    Get the Airflow operator class for a task.
 
-    :param task: The Task to get the Operator for
+    :param task: The Task to get the operator for.
```
jlaneve, line 93: for consistency:

```suggestion
    :param task: The task to get the operator for.
```

cosmos/providers/dbt/core/operators.py
```diff
@@ -18,73 +19,81 @@
...
+    env: dict[str, str]
+    append_env: bool
+    output_encoding: str
+    skip_exit_code: int
+    cancel_query_on_kill: bool
+    dbt_executable_path: str
+    full_refresh: bool  # Available on seed only
+    macro_name: str  # Available on run operation only
+
+
```
jlaneve, line 50: we're trying to maintain python 3.7 support (which admittedly we should have CI/CD set up for, but we haven't gotten around to it yet). this makes typing a bit tougher - the things I know to look out for:
- no unions via pipes, need to use `typing.Union`
- no `list` or `dict`, need to use `typing.List` and `typing.Dict`

Do you mind making those changes? I've suggested some stuff in-line here, but it may not be 100% correct
```suggestion
class DbtArgs(TypedDict, total=False):
    """
    Maintain all available arguments here so that we can help end users with available kwargs.
    """

    project_dir: Path  # Not available for end users, generated by Cosmos
    base_cmd: Union[str, List[str]]
    select: str
    exclude: str
    selector: str
    vars: Dict[str, str]
    models: str  # Not available for end users, generated by Cosmos
    cache_selected_only: bool
    no_version_check: bool
    fail_fast: bool
    quiet: bool
    warn_error: bool
    db_name: str
    schema: str
    env: Dict[str, str]
    append_env: bool
    output_encoding: str
    skip_exit_code: int
    cancel_query_on_kill: bool
    dbt_executable_path: str
    full_refresh: bool  # Available on seed only
    macro_name: str  # Available on run operation only


```

cosmos/providers/dbt/render.py
```diff
@@ -23,18 +26,19 @@ def render_project(
...
     dbt_models_dir: str = "models",
-    task_args: Dict[str, Any] = {},
+    task_args: DbtArgs | None = None,
     test_behavior: Literal["none", "after_each", "after_all"] = "after_each",
     emit_datasets: bool = True,
     conn_id: str = "default_conn_id",
-    select: Dict[str, List[str]] = {},
-    exclude: Dict[str, List[str]] = {},
+    select: dict[str, list[str]] | None = None,
+    exclude: dict[str, list[str]] | None = None,
```
jlaneve, line 34: just to confirm - does this work in python 3.7 because of the `from __future__ import annotations`? we had a PR a while ago that seemed to make these changes despite the import

https://github.com/astronomer/astronomer-cosmos/pull/88

cosmos/providers/dbt/core/operators.py
josh-fell: The `from __future__ import annotations` import for [PEP 563](https://peps.python.org/pep-0563/) use should handle these typing concerns though. For example, Airflow itself also supports 3.7 and uses PEP 563. No issues thus far. However, `TypedDict` will require some sort of Python version check since it wasn't part of `typing` until Python 3.8. Airflow [uses a `typing_compat` module](https://github.com/apache/airflow/blob/47b67f13da667de4eb69233404d4d5a33574dd58/airflow/typing_compat.py#L32) for this.

MikeWallis42: Yep, there are already a handful of places in the main code base making use of this import, that's where I got the inspiration from.
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/providers/dbt/core/operators.py#L89

I didn't keep in mind that TypedDict didn't exist in 3.7 though so I'll make use of how Airflow handles it.

cosmos/providers/dbt/render.py
MikeWallis42: As I understand it from Python 3.7 they made this annotations future available which implements PEP 563.
https://docs.python.org/3.7/library/__future__.html

I dug out the PR on CPython and noticed a comment from Guido here https://github.com/python/cpython/pull/4390#issuecomment-344116700
Although it must have been called something different originally you can see him checking by using the built-in types.

cosmos/providers/dbt/core/operators.py
```diff
@@ -3,13 +3,14 @@
...
 from pathlib import Path
-from typing import List, Sequence, TypedDict
+from typing import List, Sequence
 
 import yaml
 from airflow.compat.functools import cached_property
 from airflow.exceptions import AirflowException, AirflowSkipException
 from airflow.hooks.subprocess import SubprocessHook
 from airflow.models.baseoperator import BaseOperator
+from airflow.typing_compat import TypedDict
```
josh-fell, line 13: I'm not sure how quickly Airflow will deprecate the `TypedDict` import from this module after [Python 3.7. hits EOL in 4 months](https://endoflife.date/python). Basically Airflow [will drop Python support versions in the next minor or major Airflow version after a Python version reaches EOL](https://airflow.apache.org/docs/apache-airflow/stable/installation/supported-versions.html#support-for-python-and-kubernetes-versions). @jlaneve not sure how long you'd like to support Python 3.7, but it might make sense to port over the compat logic to a native module in Cosmos. WDYT?

dimberman:
@MikeWallis42 any updates on this?

MikeWallis42:
@dimberman apologies, I thought there was an open question about Python support, if that's not the case then I'll take the suggestion made by @josh-fell on duplicating the Airflow handling mechanism.

tatiana:
@MikeWallis42 thanks a lot for the PR. Let's get it merged this week!
Could you rebase and fix any conflicts, please?

cosmos/providers/dbt/core/operators.py
josh-fell: FWIW, Python 3.7 support has been removed from Airflow `main` and presumably starting with 2.7.0 only support Python 3.8+.

jlaneve:
closing this in favor of https://github.com/astronomer/astronomer-cosmos/pull/385

PR #144, opened 2023-02-15
fixes #134 invalid data types getting to subprocess

### LINKED ISSUE #134: Ensure no environment variables are set to `None` before running the subprocess

MikeWallis42:
fixes #134
A small amount of refactoring here too but ultimately I implemented the fix in `DbtBaseOperator.get_env`.
Furthermore I pushed in to that function the profile_vars so that the function has complete responsibility.


MikeWallis42:
This might have exposed another issue for optional arguments whereby dbt expects that as there's an env_var that it will be set.
I'm working through the details on the dev work for the Trino adapter and will hopefully come up with a solution.

MikeWallis42:
@jlaneve @pohek321
OK so as another action, each optional environment variable for each profile will need to have a default set against it.
These defaults will need to be in the same data type as the value that would be expected by the parser.
e.g. in the Trino one I've been looking at, session properties is optional but would resolve to a dictionary.
`"session_properties": "{{ env_var('TRINO_SESSION_PROPERTIES', {}) }}",`
Perhaps a separate set of issues should be raised to update all optional environment variables for all existing connections?

This PR will resolve the immediate issue but then this other issue that I'm describing will become visibile for your Github issue example of an optional Snowflake role.

cosmos/providers/dbt/core/operators.py
```diff
@@ -234,45 +223,33 @@ def run_command(self, cmd, env):
         self.exception_handling(result)
         return result
 
-    def build_and_run_cmd(self, env: dict, cmd_flags: list = None):
+    def build_and_run_cmd(
+        self, context: Context, cmd_flags: list[str] | None = None
+    ) -> SubprocessResult:
         create_default_profiles(DBT_PROFILE_PATH)
+        create_default_profiles()
```
jlaneve, line 230: @MikeWallis42 can't tell if this was an accidental addition or not - can it be removed?

MikeWallis42: @jlaneve apologies it was a rogue duplication, that function expects the path to be passed in now so it would have generated an error, I've removed it.

cosmos/providers/dbt/core/operators.py
```diff
@@ -234,45 +223,32 @@ def run_command(self, cmd, env):
...
         if cmd_flags:
-            for item in cmd_flags:
-                dbt_cmd.append(item)
-
-        ## add profile
-        dbt_cmd.append("--profile")
-        dbt_cmd.append(profile)
+            dbt_cmd.extend(cmd_flags)
+        # add profile
+        dbt_cmd.extend(["--profile", profile])
```
dimberman, line 243: @MikeWallis42 I'm a bit confused here. Wouldn't this go against the "DBT_PROFILES_DIR" environment variable override?

pohek321 (approved):
@MikeWallis42 so sorry for the delay on this. I just tested your changes locally and everything looks great! Approved!

MikeWallis42:
> @MikeWallis42 so sorry for the delay on this. I just tested your changes locally and everything looks great! Approved!

@pohek321 no problems, I know you guys are busy!

PR #146, opened 2023-02-16
Remove duplicated docstrings

kaxil:
Follow up of https://github.com/astronomer/astronomer-cosmos/pull/108/ to remove duplicated docstrings


cosmos/providers/databricks/notebook.py
```diff
@@ -77,22 +76,6 @@ def __init__(
         existing_cluster_id: str | None = None,
         **kwargs,
     ):
-        """
-        Create a Databricks notebook operator.
```
kaxil, line 81: They are already defined in the class, so don't need it in ctor

jlaneve (approved):

PR #147, opened 2023-02-16
Move dbt root directory to dags folder

pohek321:
This will make it so that we can do dag only deploys! Previously, putting `/dbt` project directories in dags would only work locally. This is because airflow prod environments prevent write operations within `/dags`. This restricts dbt because dbt writes to `/logs` & `/target` directories when a dbt command is triggered. So we are going to copy those project files to a place where they can be written, `/tmp`!

At a high-level this is what's changing:
- created `cosmos/providers/dbt/core/utils/file_syncing.py` as an import for functions that can be reused:
  - `is_file_locked` will help us to determine if a file is being used by another process
  - `exclude` sets constants for the directories that we don't care to sync between `/dags/dbt` and `/tmp/dbt`
  - `has_differences` uses filecmp library to compare whether there unsynced changes between `/dags/dbt and `/tmp/dbt`
- updated `cosmos/providers/dbt/core/operators` to use the above utilities to achieve the following flow:
Screenshot 2023-02-17 at 2 14 50 PM

- updated `/examples` & `/astro_tests` so that the `/dbt` root directory is now nested within `/dags`


astro_tests/dags/.airflowignore
```diff
@@ -0,0 +1 @@
+dags/
```
dimberman, line 1: Wait how does this work? You're putting the DAGs directory in the .airflowignore? How does airflow find the DAGs then?

astro_tests/dags/dbt_example_dags.py
```diff
@@ -89,7 +89,7 @@
 
             deps = DbtDepsOperator(
                 task_id=f"{project['project']}_install_deps",
-                project_dir=f"/usr/local/airflow/dbt/{project['project']}",
+                project_dir=f"/usr/local/airflow/dags/dbt/{project['project']}",
```
dimberman, line 92: It might make sense to make this a relative path instead of an absolute path. That would make this example work in more environments (especially when considering local development)

cosmos/providers/dbt/core/operators.py
```diff
@@ -179,7 +191,13 @@ def add_global_flags(self):
         flags = []
         for global_flag in global_flags:
             dbt_name = f"--{global_flag.replace('_', '-')}"
-            global_flag_value = self.__getattribute__(global_flag)
+            # intercept project directory and route it to r/w tmp
+            if global_flag == "project_dir":
+                global_flag_value = (
+                    f"/tmp/dbt/{os.path.basename(self.__getattribute__(global_flag))}"
```
dimberman, line 197: I like the idea of using the existing path name in this path. Leaves the option for multiple simultaneous projects. Is there any risk of cross contamination if there are multiple users or multiple airflows? Maybe we make the base path configurable?

cosmos/providers/dbt/core/operators.py
```diff
@@ -215,14 +233,60 @@ def run_command(self, cmd, env):
                 raise AirflowException(
                     f"The project_dir {self.project_dir} must be a directory"
                 )
+
+        # routing the dbt project to a tmp dir for r/w operations
+        target_dir = f"/tmp/dbt/{os.path.basename(self.project_dir)}"
+
+        changes = True  # set changes to true by default
+        if os.path.exists(target_dir):
```
dimberman, line 241: So if I'm reading this correctly this will check for changes and copy over only when the actual task is run? Would it make more sense to do it every time the DAG is parsed? Especially if these models are showing up in the airflow UI I wouldn't want people to change their models and then not have that reflected

dimberman (changes requested):
This looks sick!

One concern I have is how every time you move a DBT project it makes the PR really messy with "moved but not changed" files.

Unless we don't plan to move this example again we might want to figure out a better way to store these DBT projects

cosmos/providers/dbt/core/operators.py
pohek321: Well, the users are going to make changes to their models in `./dags/dbt/` and it will be reflected on the UI - then when they run the task, it will be copied into `/tmp`

astro_tests/dags/dbt_example_dags.py
pohek321: @dimberman what if we just put the env var for the dags directory in the beginning of the path. Would that accomplish the same thing? Example:

```python

dags_folder = os.environ["AIRFLOW__CORE__DAGS_FOLDER"]

DbtDepsOperator(
  ...,
  project_dir=f"{dags_folder}/dbt/{project['project'}"
)
```

astro_tests/dags/.airflowignore
pohek321: Weird, I don't remember adding this to the `.airflowignore`, I'll remove it

pohek321: I just realized I must've fat fingered this. I meant for it to have `dbt/` in there. @dimberman wouldn't that make more sense?

cosmos/providers/dbt/core/operators.py
pohek321: I don't know if I'm following. Let's put it into context of the jaffle_shop project to ensure we are on the same page. So for this example the project_dir value is `jaffle_shop`, so the `jaffle_shop` directory from `./dags/dbt/jaffle_shop` will get written to `/tmp/dbt/jaffle_shop`. There are file lock mechanisms to ensure that multiple workers don't colide when that copy happens from `/dags` to `/tmp`. I've confirmed this by deleting `/tmp` and executing several `DbtRunOperators` in parallel to ensure they don't get hung-up on each other. I think this will remain true regardless of how many users or Airflows there are running this against tmp. Does that answer your question?

pohek321: @dimberman forgot to @ you, sorry!

pohek321:
> Unless we don't plan to move this example again we might want to figure out a better way to store these DBT projects

@dimberman, maybe it'd make sense to develop sub-modules to move them out of this repo? We could make that another PR though?

dimberman:
@pohek321 submodules sound like a great idea. Let's make that another PR so we can merge from main and make this PR more readable.

pohek321:
> @pohek321 submodules sound like a great idea. Let's make that another PR so we can merge from main and make this PR more readable.

@dimberman can I get a review on [#165](https://github.com/astronomer/astronomer-cosmos/pull/165)

pohek321:
@jlaneve, based on @dimberman's request above; I decided to just remove `/examples` (except the Databricks dags) and `/astro_tests` here in #183. If you're good with that, then I think this will be ready

.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
+clear
```
jlaneve, line 1: is there a reason for this introduction?

pohek321: good catch. might be a typo. will remove

jlaneve (approved):

dimberman (approved):

pohek321:
> @pohek321 is this good to merge? Would love to add this feature ASAP :D.

Yup! Once the checks pass, I'll merge it in!

PR #152, opened 2023-02-17
fixes #59 implement Trino adapter

### LINKED ISSUE #59: Add support for dbt-trino adapter
We should add all of the dbt supported platforms (and then move to the community ones). This will require two things:

- [ ]  default profile that maps to an Airflow connection
- [ ]  new extra in setup.cfg

- See Trino setup in [dbt docs](https://docs.getdbt.com/reference/warehouse-setups/trino-setup). 
- See OSS Airflow Docs for [Apache Trino Connection](https://airflow.apache.org/docs/apache-airflow-providers-trino/stable/connections.html)

MikeWallis42:
fixes #59 

@pohek321, I went a bit crazy and implemented all of the auth types but I'm only able to check ldap.
Can you please double check that you're happy with my implementation of catalog and schema in the context of database_override and schema_override etc.
I could do with a hand on the documentation side of things as well as you've taken lots of screenshots in the other connection docs and I don't want them to be out of theme.

Finally this will be dependant on #134 being merged in as supporting multiple authentication types obviously means lots of optional environment variables that may not get populated.


matthewzahar:
_patiently waiting for this_

cosmos/providers/dbt/core/profiles/trino.py
dimberman, line 1: @MikeWallis42  can you please put this in a "dbt/community" directory? Going forward I think we're going to separate the "officially maintained" and "community maintained" profile types.

dimberman (changes requested):

cosmos/providers/dbt/core/profiles/trino.py
MikeWallis42: @dimberman done, please check you're happy

jlaneve (approved):

Issue #124, opened 2023-02-06
Allow users to add arbitrary arguments to a databricks job Json.


dimberman:
As with any external API, there is always the possibility that our library will run behind new job specs and capabilities. Ideally, we'd like to prevent our users from being blocked by our development speed so they can get their jobs running as quickly as possible. This issue would allow users to add or merge arguments into the generated databricks job json.

User should be able to create a nested dictionary into an `extra_job_params` field, which can be added to the json before the job is submitted.

Here is an example of how that would look.

```
with dag:
    # [START howto_databricks_workflow_notebook]
    task_group = DatabricksWorkflowTaskGroup(
        group_id="test_workflow",
        databricks_conn_id=DATABRICKS_CONN_ID,
        job_clusters=job_cluster_spec,
        notebook_params=[],
        extra_job_params={"webhook_notifications": {
            "email": "foo@bar.com"
        },
    )
```

This extra param would then be merged at the same nesting level
Whoever takes on this task should first look into any potential existing solutions for merging python dictionaries: One library I found at a cursory glance is [mergedeep](https://mergedeep.readthedocs.io/en/latest/#license). If we can avoid owning our own merging logic, that would be ideal.

I would highly recommend writing unit tests for this before writing integration tests, as at the end of the day this is really just generating json objects.

tatiana:
PR: https://github.com/astronomer/astronomer-cosmos/pull/153

PR #153, opened 2023-02-20
Add support to Databricks Workflow Job arbitrary arguments

tatiana:
Closes: #124


cosmos/providers/databricks/workflow.py
```diff
@@ -244,6 +251,7 @@ def __init__(
         python_params: list = None,
         spark_submit_params: list = None,
         max_concurrent_runs: int = 1,
+        extra_job_params: dict = None,
```
jlaneve, line 254: nit: can we type this stronger?
```suggestion
        extra_job_params: dict[str, Any] = None,
```

jlaneve (approved):
overall looks great! just one minor comment, and it looks like the pre-commit CI is failing because a few lines are too long

PR #156, opened 2023-02-21
Add ability to retry individual tasks

### LINKED ISSUE #115: Tasks within a Databricks Workflow Taskgroup should support individual retries.
**Is your feature request related to a problem? Please describe.**
With Databricks Workflow support, we want to give users as close to an airflow-native experience as we can while still getting the advantages of running Databricks tasks as a Databricks Job. One critical feature that Airflow users expect when they run a task within Airflow is the ability to retry a task if it fails. Currently, if a task within a DatabricksWorkflowTaskGroup fails, the user would need to start the entire taskgroup over (and essentially run the entire job again), which is a suboptimal experience for our users.

**Describe the solution you'd like**
Looking at the [Databricks Job API](https://docs.databricks.com/dev-tools/api/latest/jobs.html#operation/JobsRunsRepair) we see that there is a "repair" function that reruns a failed task if we have the job's "repair ID" and the task's name. @fhoda has an example [here](https://github.com/astronomer/databricks-jobs/blob/main/dags/databricks-jobs-dynamic.py#L64-L68) of a system he built for a customer of tracking repair IDs for users.

I think we can take our current system for getting task Ids using job names and mix it with Faisal's repair_id example to make a system where Airflow can send repair requests for individual tasks, giving it control over which tasks. 

Alternatively we can also offer a taskgroup level option to "repair all tasks" which would be more efficient for sharing clusters between tasks. This might be a little bit trickier as we'd need to add a "repair_mode" to the ``_CreateDatabricksWorkflowOperator``.

dimberman:
As a first step in full support for retries within Databricks Jobs, this PR links the databricks "Repair" command to the "_on_retry_callback" of the task. This means that when you click "retry" in the Airflow UI, it should automatically send a repair request to databricks.

Notably this feature only allows for retrying a _single_ task. When sending repair requests to databricks, the databricks Jobs API does not allow for multiple requests. You have to decide all tasks you want to retry at once. With this in mind I think we should also include a "retry all failed" button in the DatabricksTaskGroupOperator on the UX so users can clear tasks and rerun when they have fixed the bug. cc: @bbovenzi

Closes #115


cosmos/providers/databricks/notebook.py
```diff
@@ -9,9 +9,43 @@
...
+    api_client = current_task._get_api_client()
+    jobs_service = JobsService(api_client)
+    current_job = jobs_service.get_run(
+        run_id=current_task.databricks_run_id, include_history=True
+    )
+    repair_history = current_job.get("repair_history")
+    repair_history_id = None
+    if repair_history and len(repair_history) > 1:
+        # We use the last item in the array to get the latest repair ID
+        repair_history_id = repair_history[-1]["id"]
```
pankajkoti, line 40: We're not setting the `repair_history_id` when `len(repair_history)==1`. We would need to handle that case too.

docs/databricks/workflows.rst
```diff
@@ -24,6 +24,17 @@ users can define the individual tasks that make up the workflow. Currently the o
 DatabricksNotebookOperator, but other operators can be used as long as they contain the ``convert_to_databricks_workflow_task``
 function. In the future we plan to support SQL and python functions via the ref:`https://github.com/astronomer/astro-sdk<Astro SDK>`.
 
+Retries
+=======
+
+To retry a task, go to the Airflow UI and click on the "retry" button. This will initiate the task retry.
+Once the task is retried, it will run a repair job using the Databricks "repair" endpoint.
+
+The Databricks only allows one repair at a time. This means that once you retry a task,
```
pankajkoti, line 33:
```suggestion
Databricks only allows one repair at a time. This means that once you retry a task,
```

docs/databricks/workflows.rst
```diff
@@ -24,6 +24,17 @@ users can define the individual tasks that make up the workflow. Currently the o
...
+you cannot retry other tasks until that task completes.
+
+In the future we will support "repair all failed tasks" and "repair downstream tasks".
```
pankajkoti, line 36:
```suggestion
In the future, we will support "repair all failed tasks" and "repair downstream tasks".
```

cosmos/providers/databricks/notebook.py
```diff
@@ -111,6 +145,10 @@ def convert_to_databricks_workflow_task(
         }
         return result
 
+    def _get_databricks_task_id(self):
+        """Get the databricks task ID using dag_id and task_id. removes illegal characters."""
```
pankajkoti, line 149:
```suggestion
        """Get the databricks task ID using dag_id and task_id. Removes illegal characters."""
```

pankajkoti (commented):
LGTM. Minor suggestions.

cosmos/providers/databricks/notebook.py
```diff
@@ -9,9 +9,43 @@
...
+    repair_history_id = None
+    if repair_history and len(repair_history) > 1:
+        # We use the last item in the array to get the latest repair ID
+        repair_history_id = repair_history[-1]["id"]
+    jobs_service.repair(
+        run_id=current_task.databricks_run_id,
+        version="2.1",
+        latest_repair_id=repair_history_id,
+        rerun_tasks=[current_task._get_databricks_task_id()],
+    )
```
tatiana, line 46: @dimberman, do we have follow-up tickets for the following scenarios?
* retrying all failed tasks
* retrying a failed task and running downstream tasks

cosmos/providers/databricks/notebook.py
```diff
@@ -9,9 +9,43 @@
...
+    the future if users ask for it, but for the moment we want to keep things simple while the API
+    stabilizes.
+    :return: None
+    """
+    current_task = context["ti"].task
+    api_client = current_task._get_api_client()
+    jobs_service = JobsService(api_client)
+    current_job = jobs_service.get_run(
+        run_id=current_task.databricks_run_id, include_history=True
+    )
```
tatiana, line 35: Would it make sense for us to check in advance if there is an ongoing repair and give a custom error message, or are we happy to return Databricks error stack traces?

PR #159, opened 2023-02-23
[Databricks] Allow users to install python packages at job and task level

pankajkoti:


jlaneve:
do you mind adding a description showing an example of how to use it?

examples/databricks/example_databricks_workflow.py
```diff
@@ -53,19 +53,23 @@
...
+            notebook_path="/Users/pankaj.koti@astronomer.io/Notebook1",
+            notebook_packages=[{"pypi": {"package": "Faker"}}],
             source="WORKSPACE",
             job_cluster_key="Shared_job_cluster",
         )
         notebook_2 = DatabricksNotebookOperator(
             task_id="notebook_2",
             databricks_conn_id=DATABRICKS_CONN_ID,
-            notebook_path="/Users/daniel@astronomer.io/Test workflow",
+            notebook_path="/Users/pankaj.koti@astronomer.io/Notebook2",
```
jlaneve, line 72: is there a better way to do these local `notebook_paths`? maybe we can create a shared folder in dbr so we don't need to keep changing this?

jlaneve (commented):
can we add docs in this PR too? explaining (a) how to specify them and (b) what the options are (`pypi`, `nb_index`, etc) - we can link back to the databricks docs for this

cosmos/providers/databricks/notebook.py
```diff
@@ -31,6 +31,13 @@ class DatabricksNotebookOperator(BaseOperator):
                 databricks_conn_id="databricks_conn",
                 job_clusters=job_cluster_spec,
                 notebook_params=[],
+                notebook_packages=[
+                    {
+                        "pypi": {
+                            "package": "simplejson"
+                        }
+                    },
+                ]
```
dimberman, line 40: I think we need to be careful in the documentation that we make it clear that anything added at the taskgroup level will be added to all notebooks. This is especially important since this is a functionality that doesn't exist within databricks. It might make sense to not have this example at the notebook documentation and instead put it in the taskgroup, but I'm open on this.

examples/databricks/example_databricks_workflow.py
```diff
@@ -53,19 +53,23 @@
...
+        notebook_packages=[
+            {"pypi": {"package": "simplejson"}},
+        ],
     )
     with task_group:
         notebook_1 = DatabricksNotebookOperator(
             task_id="notebook_1",
             databricks_conn_id=DATABRICKS_CONN_ID,
-            notebook_path="/Users/daniel@astronomer.io/Test workflow",
+            notebook_path="/Users/pankaj.koti@astronomer.io/Notebook1",
```
dimberman, line 64: Maybe we should find a shared folder for our notebooks so we don't keep changing this between PRs :). cc: @tatiana

examples/databricks/example_databricks_workflow.py
```diff
@@ -53,19 +53,23 @@
...
+            notebook_packages=[{"pypi": {"package": "Faker"}}],
```
dimberman, line 65: Is this a real package? Want to be careful we don't break the example DAG and I haven't seen a pypi package with a capital before.

dimberman (changes requested):
Looking great so far!

examples/databricks/example_databricks_workflow.py
pankajkoti: Yes, it is a real package :). I have created the notebook mentioned in the path to use this package and it ran fine with the example DAG for WorkflowTaskGroup.

Here is the PyPI listing for the package: https://pypi.org/project/Faker/

examples/databricks/example_databricks_workflow.py
pankajkoti: yes, using shared notebooks now. https://github.com/astronomer/astronomer-cosmos/pull/159/commits/e7c47f8773d194039ec432eb6226801e0b2793d3

examples/databricks/example_databricks_workflow.py
pankajkoti: yes, using shared notebooks now https://github.com/astronomer/astronomer-cosmos/pull/159/commits/e7c47f8773d194039ec432eb6226801e0b2793d3

cosmos/providers/databricks/notebook.py
pankajkoti: Removed the docstring note from notebook and it already exists in the taskgroup workflow.py docstring.

pankajkoti: Also, updated the docs to mention the suggestion.

cosmos/providers/databricks/notebook.py
```diff
@@ -91,8 +93,10 @@ def convert_to_databricks_workflow_task(
         self, relevant_upstreams: list[BaseOperator]
     ):
         """
-        Convert the operator to a Databricks workflow task that can be task in a workflow
+        Convert the operator to a Databricks workflow task that can be a task in a workflow
         """
+        if hasattr(self.task_group, "notebook_packages"):
+            self.notebook_packages.extend(self.task_group.notebook_packages)
```
tatiana, line 99: By using `extend` here, don't we have the risk of setting multiple times the same package with different versions at the workflow/job level?

From:
https://docs.databricks.com/dev-tools/api/latest/jobs.html#operation/JobsCreate
```
The name of the PyPI package to install. An optional exact version specification is also supported. Examples: `simplejson and simplejson==3.8.0`. This field is required.
```
So we could have something like:
- Workflow level: `simplejson==3.8.0`
- Notebook level: `simplejson==2.0.0`

Would Databricks raise an exception? Or does it silently installs the last dependency definition (at the job level and not workflow)? It would be good to at least document this behaviour.

In future, depending on the usage, it may be worth raising a warning.

examples/databricks/example_databricks_workflow.py
```diff
@@ -53,19 +53,23 @@
...
+            {"pypi": {"package": "simplejson"}},
+        ],
     )
     with task_group:
         notebook_1 = DatabricksNotebookOperator(
             task_id="notebook_1",
             databricks_conn_id=DATABRICKS_CONN_ID,
-            notebook_path="/Users/daniel@astronomer.io/Test workflow",
+            notebook_path="/Shared/Notebook_1",
+            notebook_packages=[{"pypi": {"package": "Faker"}}],
```
tatiana, line 65: What do you think if we have one of the examples with the PyPI package version and perhaps the URL?

cosmos/providers/databricks/notebook.py
pankajkoti: yes, great point. Databricks does not allow duplicate installation of a library and fails the task as in the screenshot attached. Added a warning for the same in the docs.

Screenshot 2023-02-25 at 12 20 37 PM

examples/databricks/example_databricks_workflow.py
pankajkoti: added the example for simplejson at the taskgroup level

tatiana (approved):

dimberman (changes requested):
For the most part this all looks good, but can you please add more description into the PR description. I want to make sure that if we have to look back at this in the future we know exactly what we're looking at.

pankajkoti:
> For the most part this all looks good, but can you please add more description into the PR description. I want to make sure that if we have to look back at this in the future we know exactly what we're looking at.

@dimberman I have updated the description. Please check.

Issue #160, opened 2023-02-23
[Databricks] Running a Notebook task without WorkflowTaskGroup fails


pankajkoti:
The following DAG for DatabricksNotebookOperator without DatabricksWorkflowTaskGroup fails.
```
dag = DAG(
    dag_id="example_databricks_notebook",
    start_date=datetime(2022, 1, 1),
    schedule_interval=None,
    catchup=False,
    default_args=default_args,
)
with dag:
    # [START howto_databricks_workflow_notebook]
    notebook_1 = DatabricksNotebookOperator(
        task_id="notebook_1",
        databricks_conn_id=DATABRICKS_CONN_ID,
        notebook_path="notebook-path",
        notebook_packages=[
            {
                "pypi": {
                    "package": "Faker"
                }
            }
        ],
        source="WORKSPACE",
        job_cluster_key="random_cluster_key",
        existing_cluster_id="existing-cluster-id"
    )
```

The error is as below:
```
[2023-02-23, 16:24:19 UTC] {base.py:73} INFO - Using connection ID 'databricks_conn' for task execution.
[2023-02-23, 16:24:21 UTC] {base.py:73} INFO - Using connection ID 'databricks_conn' for task execution.
[2023-02-23, 16:24:22 UTC] {taskinstance.py:1772} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/cosmos/providers/databricks/notebook.py", line 229, in execute
    self.monitor_databricks_job()
  File "/usr/local/lib/python3.9/site-packages/cosmos/providers/databricks/notebook.py", line 139, in monitor_databricks_job
    current_task = self._get_current_databricks_task(runs_api)
  File "/usr/local/lib/python3.9/site-packages/cosmos/providers/databricks/notebook.py", line 148, in _get_current_databricks_task
    x["task_key"]: x for x in runs_api.get_run(self.databricks_run_id)["tasks"]
KeyError: 'tasks'
[2023-02-23, 16:24:22 UTC] {taskinstance.py:1322} INFO - Marking task as FAILED. dag_id=example_databricks_notebook, task_id=notebook_1, execution_date=20230223T162205, start_date=20230223T162418, end_date=20230223T162422
[...]

PR #162, opened 2023-02-25
[Databricks] Fix issue for using single notebook task launch without taskgroup

pankajkoti:
The latest Databricks Jobs API 2.1 offers a consistent response for
both single task and multi-task job run submit API which fixes the current 
issue mentioned in the ticket.

The PR also adds an example DAG for running the DatabricksNotebookOperator
alone without the DatabricksWorkflowTaskGroup operator.

closes: #160

### DIFF
--- a/cosmos/providers/databricks/notebook.py
+++ b/cosmos/providers/databricks/notebook.py
@@ -128,9 +128,13 @@ class DatabricksNotebookOperator(BaseOperator):
         self._handle_final_state(final_state)
 
     def _get_current_databricks_task(self, runs_api):
-        print(self.databricks_run_id, runs_api.get_run(self.databricks_run_id))
+        print(
+            self.databricks_run_id,
+            runs_api.get_run(self.databricks_run_id, version="2.1"),
+        )
         return {
-            x["task_key"]: x for x in runs_api.get_run(self.databricks_run_id)["tasks"]
+            x["task_key"]: x
+            for x in runs_api.get_run(self.databricks_run_id, version="2.1")["tasks"]
         }[self.dag_id + "__" + self.task_id.replace(".", "__")]
 
     def _handle_final_state(self, final_state):
--- a/cosmos/providers/databricks/notebook.py
+++ b/cosmos/providers/databricks/notebook.py
@@ -193,6 +193,7 @@ class DatabricksNotebookOperator(BaseOperator):
         runs_api = RunsApi(api_client)
         run = runs_api.submit_run(run_json)
         self.databricks_run_id = run["run_id"]
+        print(run)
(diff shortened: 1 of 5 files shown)


cosmos/providers/databricks/notebook.py
```diff
@@ -124,7 +126,10 @@ def monitor_databricks_job(self):
 
     def _get_current_databricks_task(self, runs_api):
         return {
-            x["task_key"]: x for x in runs_api.get_run(self.databricks_run_id)["tasks"]
+            x["task_key"]: x
+            for x in runs_api.get_run(self.databricks_run_id, version=JOBS_API_VERSION)[
```
tatiana, line 130: Is this version necessary? Do we need the major and minor digits?

cosmos/providers/databricks/notebook.py
```diff
@@ -124,7 +126,10 @@ def monitor_databricks_job(self):
...
+                "tasks"
+            ]
```
tatiana, line 132: Do we have any unit tests to cover this change?

cosmos/providers/databricks/notebook.py
pankajkoti: Yes, that version `2.1` is necessary as it gives consistent responses. Both major and minor digits are necessary. I checked that version `2.0` sends the json key as `task` whereas `2.1` sends it as `tasks` across single and multi-task runs.

dimberman (dismissed):
Like @tatiana said, once there are unit tests I'm glad to approve.

cosmos/providers/databricks/notebook.py
pankajkoti: Those were missing, I have added them now.

pankajkoti:
> Like @tatiana said, once there are unit tests I'm glad to approve.

@dimberman I have added them now, please check.

cosmos/providers/databricks/notebook.py
```diff
@@ -94,7 +96,7 @@ def convert_to_databricks_workflow_task(
         Convert the operator to a Databricks workflow task that can be task in a workflow
         """
         result = {
-            "task_key": self.dag_id + "__" + self.task_id.replace(".", "__"),
+            "task_key": self._get_databricks_task_id(),
             "depends_on": [
                 {"task_key": self.dag_id + "__" + t.replace(".", "__")}
```
tatiana, line 101: We should have a unified function to process `_get_databricks_task_id` and this line - otherwise, if the logic changes, we can have issues

tatiana (approved):
Nice catch and quick fix, @pankajkoti , thank you!

Issue #161, opened 2023-02-24
Enable CI/CD for Databricks


tatiana:
We should automate in the CI/CD the following:
* Running example Databricks DAGs
* Running Databricks unit tests

An option would be to follow the patterns from:
https://github.com/astronomer/astro-sdk-template

PR #163, opened 2023-02-27
Create retry buttons for Single task and full databricks jobs

dimberman:
# Description
This pull request adds functionality to the Airflow databricks workflow taskgroup to create "See Databricks Job Run", "Repair All Failed Tasks", and "Repair a single failed task" buttons. These buttons will allow users to easily navigate to the Databricks job run page and repair any failed tasks.

In the "job.launch" task, we add a "Repair All Failed Tasks" button because there is no way to create extra_links in a taskgroup. This button will launch a repair request to Databricks via the JobService API endpoint to repair all failed tasks in the job.

Screenshot 2023-02-26 at 10 22 48 PM

When the repair buttons are clicked, a repair request is sent to Databricks via the JobService API endpoint. After that we flash a message in the UI to let the user know that the repair has started along with the repair id.

Screenshot 2023-02-26 at 10 30 12 PM

In the notebook tasks, we add a "Repair a single failed task" button that allows users to rerun just the associated task. This button will also launch a repair request to Databricks via the JobService API endpoint.

# Reasoning

We originally attempted to use Airflow's built-in retry logic to repair failed tasks, but on trying this we ran into a core difference between how airflow and databricks handle retries. Databricks only allows one repair at a time, meaning that for each task airflow attempts to retry, databricks would start a new job with a new jobs cluster. To efficiently use databricks we need to send all repair requests in a single request, so we created the "Repair All Failed Tasks" and "Repair a single failed task" buttons.


cosmos/providers/databricks/plugin.py
```diff
@@ -0,0 +1,279 @@
...
+) -> list[str]:
+    """
+    Returns a list of all Databricks task IDs for a dictionary of Airflow tasks.
+
+    :param group_id: The task group ID.
+    :param task_map: A dictionary mapping task IDs to BaseOperator instances.
+    :return: A list of Databricks task IDs for the given task group.
+    """
+    task_ids = []
+    print("Getting task ids for db")
```
tatiana, line 48: Should we be using standard libraries `logging` features?

docs/databricks/workflows.rst
```diff
@@ -24,6 +24,17 @@ users can define the individual tasks that make up the workflow. Currently the o
...
+Retries
+=======
+
+To retry a task, go to the Airflow UI and click on the "retry" button. This will initiate the task retry.
+Once the task is retried, it will run a repair job using the Databricks "repair" endpoint.
+
+The Databricks only allows one repair at a time. This means that once you retry a task,
+you cannot retry other tasks until that task completes.
+
+In the future we will support "repair all failed tasks" and "repair downstream tasks".
```
tatiana, line 36: We need to update now that we have the repair of all failed tasks.
Do we have a GitHub ticket for the "repair failed & downstream tasks"?

tatiana:
> (...)  because there is no way to create extra_links in a taskgroup

@dimberman, would it make sense for us to create an issue in Apache Airflow Github to add support to having `extra_links` in task groups as well? Or is there any blockers to be able to accomplish this? I understand this would only be available in future versions of Airflow, but it probably makes sense.

cosmos/providers/databricks/plugin.py
dimberman: This is somethign I wanted to discuss. These classes don't inherit from LoggingMixin so wanted to figure out what best practices are for logging here.

bbovenzi:
> 

I think that could be useful. Although I'm not too familiar with how that would work. We don't define much on a Task Groups right now, since its supposed to be mainly a UI concept. Aren't extra links connected to an operator?

dimberman:
@bbovenzi for the moment they're associated with operators, but I don't see why we can't add them to taskgroups, especially extra_links since they're just UI buttons

cosmos/providers/databricks/plugin.py
```diff
@@ -0,0 +1,279 @@
...
+        # Should we catch the exception here if there is no return value?
+        metadata = XCom.get_one(
+            task_id=ti_key.task_id,
+            dag_id=ti_key.dag_id,
+            run_id=ti_key.run_id,
+            key="return_value",
+        )
+        hook = DatabricksHook(metadata.databricks_conn_id)
+        return f"https://{hook.host}/#job/{metadata.databricks_job_id}/run/{metadata.databricks_run_id}"
+        return f"/foo/bar?dag_id={operator.dag_id}&task_id={operator.task_id}"
```
pankajkoti, line 125:
```suggestion
```
2 returns :)

cosmos/providers/databricks/plugin.py
```diff
@@ -0,0 +1,279 @@
...
+    ) -> str:
+        dag = get_airflow_app().dag_bag.get_dag(ti_key.dag_id)
+        dag.get_task(ti_key.task_id)
+        # Should we catch the exception here if there is no return value?
+        metadata = XCom.get_one(
+            task_id=ti_key.task_id,
+            dag_id=ti_key.dag_id,
+            run_id=ti_key.run_id,
+            key="return_value",
+        )
```
pankajkoti, line 122: Can we not use the below? It looks more convenient method as per Airflow's docstring for `get_one` implementation.
```suggestion
        metadata = XCom.get_value(
            ti_key=ti_key,
            key="return_value",
        )
```

Ref: [databricks/support-for-retries ](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/xcom/index.html#airflow.models.xcom.BaseXCom.get_value)

cosmos/providers/databricks/plugin.py
```diff
@@ -0,0 +1,279 @@
...
+        *,
+        ti_key: TaskInstanceKey | None = None,
+    ) -> str:
+        # Should we catch the exception here if there is no return value?
+        metadata = XCom.get_one(
+            task_id=ti_key.task_id,
+            dag_id=ti_key.dag_id,
+            run_id=ti_key.run_id,
+            key="return_value",
+        )
```
pankajkoti, line 146:
```suggestion
        metadata = XCom.get_one(
            ti_key=ti_key,
            key="return_value",
        )
```
same suggestion as above

cosmos/providers/databricks/plugin.py
```diff
@@ -0,0 +1,279 @@
...
+        Returns a list of task IDs for tasks that have failed or have been skipped in the given DagRun.
+
+        :param dr: The DagRun object for which to retrieve failed and skipped tasks.
+
+        :return: A list of task IDs for tasks that have failed or have been skipped.
+        """
+        return [
+            t.task_id
+            for t in dr.get_task_instances(
+                state=["failed", "skipped", "up_for_retry", "upstream_failed"],
```
pankajkoti, line 182: Do we need `up_for_retry` here? Do we see any possible race conditions here? e.g. A task that was up for retry gets retried by the Airflow scheduler just before a repair is triggered from here.

cosmos/providers/databricks/plugin.py
```diff
@@ -0,0 +1,279 @@
...
+    Returns a list of all Databricks task IDs for a dictionary of Airflow tasks.
+
+    :param group_id: The task group ID.
+    :param task_map: A dictionary mapping task IDs to BaseOperator instances.
+    :return: A list of Databricks task IDs for the given task group.
+    """
+    task_ids = []
+    print("Getting task ids for db")
+    for task_id, task in task_map.items():
+        if task_id != f"{group_id}.launch":
```
pankajkoti, line 50: Will all the needed tasks have this same task ID `f"{group_id}.launch"`?

cosmos/providers/databricks/plugin.py
```diff
@@ -0,0 +1,279 @@
...
+        )
+        flash(f"Databricks repair job is starting!: {res}")
+        return redirect(f"/dags/{dag_id}/grid")
+
+
+repair_databricks_view = RepairDatabricksTasks()
+
+repair_databricks_package = {
+    "name": "Test View",
+    "category": "Test Plugin",
```
pankajkoti, line 267: We will need suitable name and category values here.

cosmos/providers/databricks/plugin.py
```diff
@@ -0,0 +1,279 @@
...
+    "view": repair_databricks_view,
+}
+
+
+class CosmosDatabricksPlugin(AirflowPlugin):
+    name = "my_namespace"
```
pankajkoti, line 273: Is this the default and needed value? Or do we need to change it?

cosmos/providers/databricks/notebook.py
```diff
@@ -9,8 +9,48 @@
...
 from databricks_cli.sdk.api_client import ApiClient
 
+from cosmos.providers.databricks.plugin import (
+    DatabricksJobRepairSingleFailedLink,
+    DatabricksJobRunLink,
+)
+from cosmos.providers.databricks.workflow import DatabricksMetaData
+
+
+def _repair_task_retry_callback(context: Context) -> None:
```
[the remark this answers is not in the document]
dimberman: Yes I can remove this thanks for catching that.

cosmos/providers/databricks/plugin.py
dimberman: Oh neat didn't know this was a thing. Thanks @pankajkoti !

cosmos/providers/databricks/plugin.py
dimberman: Any databricks job will have a launch task, we don't want to relaunch so we don't add it to the repair.

cosmos/providers/databricks/plugin.py
dimberman: We can change it I believe.

cosmos/providers/databricks/plugin.py
dimberman: Yes we need those too because we need to clear those tasks for them to actually be retried. If anything we might want to advise people to turn off retries for databricks taskgroup jobs or default the tasks to not retry. @tatiana @kaxil WDYT?

cosmos/providers/databricks/plugin.py
dimberman: The idea is _not_ to add the group_id.launch task. That's not a databricks task so we shouldn't submit it.

cosmos/providers/databricks/notebook.py
```diff
@@ -9,8 +9,48 @@
...
+    """
+    current_task = context["ti"].task
+    api_client = current_task._get_api_client()
+    jobs_service = JobsService(api_client)
+    current_job = jobs_service.get_run(
+        run_id=current_task.databricks_run_id, include_history=True
+    )
+    repair_history = current_job.get("repair_history")
+    repair_history_id = None
+    if repair_history and len(repair_history) > 1:
```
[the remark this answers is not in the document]
pankajkoti: I still feel that it needs to be >=1. 

Case 1: For the first record, len(repair_history) will be zero and this case is satisfied with both len(repair_history) >=1 and len(repair_history)>1
Case 2: For the second record, len(repair_history) will be 1 and this case won't be satisfied with len(repair_history) > 1
Case 3: For the third record, len(repair_history) will be 2 and this case will be satisfied with len(repair_history) > 1

cosmos/providers/databricks/plugin.py
```diff
@@ -0,0 +1,288 @@
...
+    :param group_id: The task group ID.
+    :param task_map: A dictionary mapping task IDs to BaseOperator instances.
+    :return: A list of Databricks task IDs for the given task group.
+    """
+    task_ids = []
+    print("Getting task ids for db")
+    for task_id, task in task_map.items():
+        if task_id != f"{group_id}.launch":
+            print(f"databricks task id {_get_databricks_task_id(task)}")
+            task_ids.append(_get_databricks_task_id(task))
```
pankajkoti, line 48:
```suggestion
        if task_id == f"{group_id}.launch":
            continue
        print(f"databricks task id {_get_databricks_task_id(task)}")
        task_ids.append(_get_databricks_task_id(task))
```

cosmos/providers/databricks/notebook.py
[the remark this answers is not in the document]
pankajkoti: I checked the API. len(repair_history) is never zero, and the repair history starts with the original record. So I believe what you're saying is right.

cosmos/providers/databricks/plugin.py
dimberman: @pankajkoti we don't need to worry about that as we are only looking at tasks within that task group.

cosmos/providers/databricks/plugin.py
dimberman: @tatiana I was able to set up logging using the LoggingMixin class for Airflow

dimberman: fixed this by using LoggingMixin

pankajkoti (approved):
Mostly looks good to me. One question I have is, once a repair request is sent to Databricks for a failed task(s), do we monitor the status of the task(s)? I mean if a failed task is repaired successfully will it's status change to success in the Airflow UI? 
If we're not handling that at the moment, can we log a ticket for that to check the feasibility for that?

tatiana (approved):
I really liked the plugin strategy, @dimberman , and UI change. One last minor comment on a open thread.

We currently support "apache-airflow>=2.4" and Python 3.7-3.10. Should we log a ticket to ensure we're running the tests/example DAGs against these combinations? Perhaps we can add this to #161, WDYT? cc @pankajkoti 

I'm slightly worried we are introducing 747 lines and that we are not running the tests in the CI yet, but relieved that @pankajkoti is working on it!

cosmos/providers/databricks/plugin.py
tatiana: Do we really need `up_for_retry` here, @dimberman ?

dimberman:
@pankajkoti by clearing the tasks, the airflow scheduler will pick them up as if they've never run and monitor them. 

@tatiana i agree we need CI/CD ASAP. Once @pankajkoti merges his PR, we can address any issues that come up. I don't think anything I've added here is specific to new versions of airflow or python.

dimberman:
Ok as a final test I just ran this against a local airflow using `astro dev start` and can confirm that running the databricks DAG as well as the repairs and job links all work correctly.

cosmos/providers/databricks/plugin.py
dimberman: @tatiana yes, because we can't use Airflow's retries so any task that is "up_for_retry" will just continue to fail unless it is added to the repair job.

PR #165, opened 2023-02-27
Replace astro_tests and examples with submodules

pohek321:
This change will break `astro_tests` and `examples` astro projects away from `astronomer-cosmos` so that changes are more lightweight and modular moving forward.

- astro_tests &rarr; [cosmos-dev (astro-tests)](https://github.com/astronomer/cosmos-dev/tree/astro-tests)
- examples &rarr; [cosmos-dev (base-example)](https://github.com/astronomer/cosmos-dev/tree/base-example)


dimberman (changes requested):
@pohek321 why are we moving the example DAGs into a submodule? As a part of our CI/CD we should be able to run all example DAGs to ensure that they work. With this change we would need to make a change to two repos any time we need to modify an example. Would it be possible to move the dbt projects into submodules (since they won't change) while keeping the DAGs in the main project?

pohek321:
I'm going to go ahead and close this and take a different approach

Issue #145, opened 2023-02-15
Link DatabricksWorkflowTaskgroup timeout to databricks jobs timeout


dimberman:
Description:
To prevent infinitely hanging jobs, it would be useful to have the ability to program timeouts for Airflow tasks that are submitted to Databricks. Currently, Databricks jobs do not have timeouts by default which means that a databricks job could run forever even if the Airflow task times out. By linking Airflow task timeouts to Databricks timeouts, we can ensure that long-running tasks do not block other jobs and resources on the platform.

Proposed solution:

Add a configuration option to DatabricksWorkflowTaskGroup to specify a timeout value. When the Databricks Job is submitted, Airflow should include the timeout value as part of the request to Databricks. This will allow Databricks to enforce the timeout and kill the task if it runs for too long.

Potential challenges:

* Ensuring that the timeout value is propagated correctly from Airflow to Databricks.
* Handling cases where the task takes longer to run than the specified timeout (e.g. due to slow network or data access).
Determining an appropriate default timeout value that works for most workflows.

dimberman:
@pankajkoti and I discussed this ticket and Pankaj is going to do a spike on 

1) tying the airflow timeout to the databricks job timeout when submitting
2) sending a "kill" message to databricks if the task takes longer than the timeout

and report findings here cc: @tatiana

pankajkoti:
**Approach 1: Tying the airflow timeout to the Databricks job timeout when submitting**
The approach would use the execution_timeout param available with the BaseOperator and pass it to the Databricks task which accepts `timeout_seconds` as a param in the payload.

_Pros:_
1. Ensures that the Databricks job will eventually and definitely terminate after the set timeout has elapsed and that the job won't run forever as the timeout is set with the Databricks task specification when triggered.

_Cons:_
1. There might be a slight delay between Airflow task and Databricks task timeout. The Airflow task will be shown as failed first as it launches first and within which we trigger the Databricks task. Ideally, this would be a relatively small number of time units.


**Approach 2: Sending a "kill" message to Databricks if the task takes longer than the timeout**
There is no callback on timeout, but Airflow raises an [AirflowTaskTimeout exception](https://airflow.apache.org/docs/apache-airflow/2.2.0/concepts/tasks.html#timeouts) which means for each task we will need to catch the exception and then trigger a client API call to cancel the job run.

_Pros:_
Can't think of any compared to the first approach

_Cons:_
[...]

PR #166, opened 2023-02-28
[Databricks] Link Airflow task timeout with Databricks Notebook task

pankajkoti:
Based on the analysis and discussion comments on the issue https://github.com/astronomer/astronomer-cosmos/issues/145,
we pass the Airflow task's `execution_timeout` parameter value available with the `DatabricksNotebookOperator` as the `timeout_seconds` parameter for the Databricks Jobs Run API.

closes: #145

### DIFF
--- a/cosmos/providers/databricks/notebook.py
+++ b/cosmos/providers/databricks/notebook.py
@@ -107,7 +107,7 @@ class DatabricksNotebookOperator(BaseOperator):
                 if t in relevant_upstreams
             ],
             "job_cluster_key": self.job_cluster_key,
-            "timeout_seconds": 0,
+            "timeout_seconds": int(self.execution_timeout.total_seconds()),
             "email_notifications": {},
             "notebook_task": {
                 "notebook_path": self.notebook_path,
--- a/cosmos/providers/databricks/notebook.py
+++ b/cosmos/providers/databricks/notebook.py
@@ -107,7 +107,9 @@ class DatabricksNotebookOperator(BaseOperator):
                 if t in relevant_upstreams
             ],
             "job_cluster_key": self.job_cluster_key,
-            "timeout_seconds": int(self.execution_timeout.total_seconds()),
+            "timeout_seconds": int(self.execution_timeout.total_seconds())
+            if self.execution_timeout
+            else 0,
             "email_notifications": {},
             "notebook_task": {
                 "notebook_path": self.notebook_path,
@@ -193,7 +195,9 @@ class DatabricksNotebookOperator(BaseOperator):
                 "base_parameters": {"source": self.source},
             },
             "libraries": self.notebook_packages,
-            "timeout_seconds": int(self.execution_timeout.total_seconds()),
+            "timeout_seconds": int(self.execution_timeout.total_seconds())
+            if self.execution_timeout
+            else 0,
         }
         if self.new_cluster and self.existing_cluster_id:
             raise ValueError(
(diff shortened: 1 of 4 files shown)


pankajkoti:
Example DAGs task showing the status as Timed Out.

Screenshot 2023-02-28 at 10 43 25 PM

dimberman (approved):
LGTM. Seems like this solution is simple and easy to maintain 👍

pankajkoti:
> LGTM. Seems like this solution is simple and easy to maintain 👍 . Have you manually run the tests to make sure this doesn't break any unit tests? We'll need to do that until the CI gets merged.

Thanks. Yes did run the unit tests and made sure that they don't break.

pankajkoti:
> LGTM. Seems like this solution is simple and easy to maintain 👍 . Have you manually run the tests to make sure this doesn't break any unit tests? We'll need to do that until the CI gets merged.

Thanks Daniel. I felt I had run unit tests on this branch, but I had missed running them here, had run them on the other two branches. We did need to alter the tests as they were breaking. Time for the CI to run theses tests sooner :)

cosmos/providers/databricks/notebook.py
```diff
@@ -107,7 +107,9 @@ def convert_to_databricks_workflow_task(
                 if t in relevant_upstreams
             ],
             "job_cluster_key": self.job_cluster_key,
-            "timeout_seconds": 0,
+            "timeout_seconds": int(self.execution_timeout.total_seconds())
+            if self.execution_timeout
+            else 0,
```
tatiana, line 112: Ah, this is interesting, so Databricks considers timeout=0 to be timeout disabled, not for it to instantly timeout :)

tatiana: It would be great to add a unit test for this.

cosmos/providers/databricks/notebook.py
```diff
@@ -193,6 +195,9 @@ def launch_notebook_job(self):
                 "base_parameters": {"source": self.source},
             },
             "libraries": self.notebook_packages,
+            "timeout_seconds": int(self.execution_timeout.total_seconds())
+            if self.execution_timeout
+            else 0,
```
tatiana, line 200: Would it make any sense to subtract the time the AIrflow task took to execute until this point - so we avoid the scenario we discussed:
- Airflow task: times out after 100s, but it spends 10s for setup
- Databricks job: takes 100s, receives timeout of 100s

This would lead the Airflow task would take 110s and fail, but the Databricks job would have been run successfully.

cosmos/providers/databricks/notebook.py
```diff
@@ -193,6 +195,9 @@ def launch_notebook_job(self):
                 "base_parameters": {"source": self.source},
             },
             "libraries": self.notebook_packages,
+            "timeout_seconds": int(self.execution_timeout.total_seconds())
```
tatiana, line 198: Why do we need this logic in two places?

cosmos/providers/databricks/notebook.py
pankajkoti: Yes, Databricks considers 0 to mean forever. That was also the previous value as you can see before my change.
The latest documentation is not that great and also there is no link to previous documentation. But, I found community answer which says so https://community.databricks.com/s/question/0D58Y00009VZwZoSAL/what-is-the-timeout-for-dbutilsnotebookrun-timeout-0-

I also tested that the job does not timeout immediately when set to 0 and succeeds until completion.

cosmos/providers/databricks/notebook.py
pankajkoti: I believe it should not take 10s for setup but ideally a few microseconds. Also, I don't know why but not totally convinced in having this manipulation here and the user would be confused as to why is the Databricks task timing out before the specified timeout value. As per Airflow, the task might fail at the first attempt due to timeout, but when it retries it marks the latest status of the task in the retried attempt.

tatiana (approved):

Issue #121, opened 2023-02-05
Expose email and webhook notifications


dimberman:
Databricks Jobs offers the ability to send webhook or email notifications on the failure of a job,

While Airflow also has this ability, it's apparent that the databricks notification would have more detailed information as to why the job failed internally. We've received requests from users to expose slack and email notifications via webhooks

On a DAG level this would look something like this

```
with dag:
    # [START howto_databricks_workflow_notebook]
    task_group = DatabricksWorkflowTaskGroup(
        group_id="test_workflow",
        databricks_conn_id=DATABRICKS_CONN_ID,
        job_clusters=job_cluster_spec,
        notebook_params=[],
        webhook_notifications=[{
            "on_failure": [
                {
                    "id": "b0aea8ab-ea8c-4a45-a2e9-9a26753fd702"
                }
            ],
        }]
    )
```

and the modified json would look like this:

```
    "job_id": 843165100284734,
    "creator_user_name": "daniel@astronomer.io",
    "run_as_user_name": "daniel@astronomer.io",
    "run_as_owner": true,
    "settings": {
        "name": "example_databricks_workflow.test_workflow",
        "email_notifications": {
            "no_alert_for_skipped_runs": false
        },
        "webhook_notifications": {
            "on_failure": [
                {
                    "id": "b0aea8ab-ea8c-4a45-a2e9-9a26753fd702"
                }
            ]
        },
```

Another option (though not sure how difficult this would be), would be to create an airflow webhook s.t. the databricks notification would notify airflow which would allow airflow to log/expose the message. This might be extra complexity, so I'd say lets not spend more than a few hours spiking for this.

<img width="735" alt="Screenshot 2023-02-05 at 10 09 37 AM" src="https://user-images.githubusercontent.com/2644098/216837110-04cd5dad-0ec1-49e9-9ba0-b6ab0af52c24.png">

jlaneve:
does web hook here include slack?

dimberman:
@jlaneve yes if you look at the UI screenshot I took you'll see that you can do email, slack, etc with one of these webhooks.

tatiana:
I changed the original proposed implementation of this ticket to use `extra_job_params`, introduced in #153.

So this is what we have at the moment:
```
    task_group = DatabricksWorkflowTaskGroup(
        group_id="test_workflow",
        databricks_conn_id=DATABRICKS_CONN_ID,
        job_clusters=job_cluster_spec,
        notebook_params=[],
[...]

PR #169, opened 2023-03-02
Add example to using webhooks and email notifications on Databricks Workflow

tatiana:
As part of `extra_job_params`, we already supported `email_notifications` and `webhooks`.

Closes: #121

Example:
```
   task_group = DatabricksWorkflowTaskGroup(
        group_id="test_workflow",
        databricks_conn_id=DATABRICKS_CONN_ID,
        job_clusters=job_cluster_spec,
        notebook_params=[],
        notebook_packages=[
            {
                "pypi": {
                    "package": "simplejson==3.18.0",  # Pin specification version of a package like this.
                    "repo": "https://pypi.org/simple",  # You can specify your required Pypi index here.
                }
            },
        ],
        extra_job_params={
            "email_notifications": {
                "on_start": [DATABRICKS_NOTIFICATION_EMAIL],
            },
            "webhook_notifications": {
                "on_start": [{"id": DATABRICKS_DESTINATION_ID}],
            },
        },
    )
```


pankajkoti (commented):
The ticket #121  also suggests adding support for `email_notifications` which is another param as can be seen in the payload screenshot. Shouldn't we also allow users to configure this along with the `webhook_notifcations` parameter?

Screenshot 2023-03-02 at 8 50 41 PM

dimberman (dismissed):
Ahh good catch @pankajkoti. Yes let's add emails as well.

tatiana:
@dimberman @pankajkoti I realised that email and webhooks were already supported by using `extra_job_params`, which I introduced in #153 . I simplified the code, updated the example DAG and saw things working locally.

cosmos/providers/databricks/workflow.py
```diff
@@ -112,8 +112,8 @@ def create_workflow_json(self) -> dict[str, object]:
             "job_clusters": self.job_clusters,
             "max_concurrent_runs": self.max_concurrent_runs,
         }
-        full_json = merge(full_json, self.extra_job_params)
-        return full_json
+        merged_json = merge(default_json, self.extra_job_params)
```
pankajkoti, line 115: Looking at the constructor `__init__ ()`, `self.extra_job_params` could be `None` if it is not passed. We should initialise it to empty dictionary `{}` if it is `None`. Or else merge would give a `TypeError` here.

pankajkoti (changes requested):

cosmos/providers/databricks/workflow.py
tatiana: @pankajkoti the class `DatabricksWorkflowTaskGroup` was already handling that - but I'll add it to the Operator so we're on the safe side.

cosmos/providers/databricks/workflow.py
```diff
@@ -131,7 +131,9 @@ def execute(self, context: Context) -> Any:
         job_id = job["job_id"] if job else None
         current_job_spec = self.create_workflow_json()
         if not isinstance(self.task_group, DatabricksWorkflowTaskGroup):
-            raise AirflowException("Task group must be a DatabricksWorkflowTaskGroup")
+            raise AirflowException(
+                "Task group must be a Databricthat'sksWorkflowTaskGroup"
```
pankajkoti, line 135:
```suggestion
                "Task group must be a DatabricksWorkflowTaskGroup"
```

pankajkoti (approved):
This looks a lot cleaner and easier to support API params. Amazing! ❤️

dimberman (approved):

Issue #174, opened 2023-03-06
Add Python 3.9 and Airflow 2.2.4 to the Databricks CI test matrix


tatiana:
At the moment, the customer who requested the Databricks work uses:
* Python 3.9
* Airflow 2.2.4
(based on Sigma)

However, we are only testing Python 3.8 and Airflow 2.5. We should extend the current CI setup https://github.com/astronomer/astronomer-cosmos/issues/161 (PR https://github.com/astronomer/astronomer-cosmos/pull/167) to also validate against this version of Python and Airflow.

PR #177, opened 2023-03-07
Support older Airflow versions and run example DAGs on them in CI

pankajkoti:
While testing the Databricks workflow and notebook operators across various versions of Airflow, it was observed that quite a few attributes are not available as per their current usage available in the latest Airflow version. 
To support these limitations with older Airflow versions, the PR adds the following fixes:
1. `DatabricksNotebookOperator`: The `task_group` attribute is not available with the `BaseOperator`. For that, we fetch the task group from the `TaskGroupContext`
2. `DatabricksWorkflowTaskGroup`: Like in `DatabricksNotebookOperator`, we need to fetch the task group from the `TaskGroupContext` and additionally, the `dag` attribute is not available with the `TaskGroup` base class which we need to fetch from the `DagContext`
3. `plugin.py` module: The `airflow_flask_app` utility is unavailable. So, we use the `current_app` from the flask context to get the Airflow flask instance
4. `conftest.py`: The `EmptyOperator` used here is unavailable. So we make use of the `DummyOperator` here
5. `dag.test()` is unavailable. So, we introduce local utilities to leverage the `test_dag` utility for us to run example DAGs without the scheduler

Additionally, in order to support running the example DAGs i.e. running concurrent Databricks jobs with the same notebooks across multiple Airflow and Python versions we have introduced the following changes:
1. Provider unique `job_cluster_key` to the Databricks jobs
2. Add the Airflow (`2.2.4`, `2.3`, `2.4`, `2.5`) and Python (`3.8`, `3.9`) versions for the matrix in `ci.yml`

closes: #174


.github/workflows/ci.yml
```diff
@@ -115,7 +115,7 @@ jobs:
           path: |
             ~/.cache/pip
```
tatiana, line 116: Would it make sense to have separate caches by (python, airflow)?

.github/workflows/ci.yml
```diff
@@ -125,7 +125,12 @@ jobs:
           name: coverage-${{ matrix.python }}-${{ matrix.airflow }}
           path: ./.coverage
 
-  Run-example-dag-tests-Airflow-2-5:
+  Run-example-dag-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        version: [ '3.8', '3.9']
```
tatiana, line 132: In some areas it seems we are calling this `{{matrix.version}}` (e.g line 111), in others `{{matrix.python}}`, should we normalise, perhaps sticking to `matrix.python`?

tests/example_dags/.airflowignore
```diff
@@ -0,0 +1,2 @@
+utils.py
+test_example_dags.py
```
tatiana, line 2: How hard would it be to move these files outside the `example_dags` folder? It would be cleaner to have a directory only with example DAGs.

tatiana (approved):
@pankajkoti I made some minor comments, but it looks great - really happy to see we are supporting these versions of Airflow now.

.github/workflows/ci.yml
pankajkoti: yes, a great catch, thank you!

.github/workflows/ci.yml
pankajkoti: If I'm understanding your comment correctly, we do maintain separate caches based on the (python version)-(airflow version) combination with the below `key` attribute.

Did you mean something else?

tests/example_dags/.airflowignore
pankajkoti: yes, I agree that is cleaner. Moved them out.

pyproject.toml
```diff
@@ -35,7 +35,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 dependencies = [
-    "apache-airflow>=2.4",
+    "apache-airflow>=2.2.4",
```
pankajkoti, line 38: I am trying to spin Airflow with runtime image 4.2.1 which is being used by customer B and while building the image in the examples Astro project it complains about conflicting Airflow versions. I am able to get Airflow up and running now but facing a failure while running the DAG and the error log is here: https://astronomer.slack.com/archives/C03HRT0MRPD/p1678198413109119?thread_ts=1678198362.881859&cid=C03HRT0MRPD.

We might also want to check if this affects the DBT provider and if yes, we can only change this in the newer Databricks providers repository.

cosmos/providers/databricks/plugin.py
```diff
@@ -12,7 +12,13 @@
 from airflow.plugins_manager import AirflowPlugin
 from airflow.providers.databricks.hooks.databricks import DatabricksHook
 from airflow.security import permissions
-from airflow.utils.airflow_flask_app import get_airflow_app
+
+try:
+    from airflow.utils.airflow_flask_app import get_airflow_app
```
dimberman, line 17: Same here

noxfile.py
```diff
@@ -32,17 +32,19 @@ def _expand_env_vars(file_path: Path):
...
     env = {
         "AIRFLOW_HOME": f"~/airflow-{airflow}-python-{session.python}",
         "AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES": "airflow\\.* astro\\.* cosmos\\.*",
     }
 
-    session.install(f"apache-airflow=={airflow}")
     session.install("-e", ".[all,tests]")
+    session.install(f"apache-airflow=={airflow}")
+
+    if airflow in ("2.2.4", "2.3", "2.4"):
```
dimberman, line 46: Leave a comment here mentioning why we no longer need this 2.5+. My hope is in the future as we no longer support these versions we can remove these lines.

tests/test_example_dags.py
```diff
@@ -0,0 +1,43 @@
...
+    return get_session()
+
+
+def get_dag_bag() -> DagBag:
+    """Create a DagBag by adding the files that are not supported to .airflowignore"""
+    example_dags_dir = Path(__file__).parent / "example_dags"
+    airflow_ignore_file = example_dags_dir / ".airflowignore"
+
+    print(".airflowignore contents: ")
+    print(airflow_ignore_file.read_text())
```
dimberman, line 30: Why are we doing this?

tests/utils.py
```diff
@@ -0,0 +1,174 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def run_dag(dag: DAG, conn_file_path: str) -> DagRun:
+    return test_dag(dag=dag, conn_file_path=conn_file_path)
+
+
+@provide_session
+def test_dag(
```
dimberman, line 28: Please add a note mentioning that we have to copy DAG.test for older airflow versions

dimberman (commented):
Looking good overall! Added a few comments

cosmos/providers/databricks/notebook.py
```diff
@@ -98,6 +98,10 @@ def __init__(
         self.new_cluster = new_cluster or {}
         self.existing_cluster_id = existing_cluster_id or ""
         super().__init__(**kwargs)
+        if not hasattr(self, "task_group"):
```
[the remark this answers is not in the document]
pankajkoti: added a comment for the same that we need this for Airflow versions <2.3. Tested that this is not needed 2.3 onward.

cosmos/providers/databricks/plugin.py
pankajkoti: added a note that this is needed for Airflow versions < 2.3.3. Tested that this is not needed for >= 2.3.3

cosmos/providers/databricks/plugin.py
```diff
@@ -272,20 +389,45 @@ def get_link(
...
+                key="return_value",
+            )
+
+        except AttributeError:
+            if not ti:
+                raise TaskInstanceNotFound()
+            self.log.info("Getting XCOM")
+            self.log.info("Getting task_id %s", ti_key.task_id)
+            self.log.info("Getting dag_id %s", ti_key.dag_id)
+            self.log.info("Getting execution_date %s", ti.execution_date)
```
pankajkoti, line 422:
```suggestion
```

cosmos/providers/databricks/plugin.py
```diff
@@ -272,20 +389,45 @@ def get_link(
...
+            metadata = XCom.get_one(
+                task_id=ti_key.task_id,
+                dag_id=ti_key.dag_id,
+                execution_date=ti.execution_date,
+                key="return_value",
+            )
+
+        self.log.info("metadata %s", metadata)
```
pankajkoti, line 430:
```suggestion
```

cosmos/providers/databricks/plugin.py
```diff
@@ -165,19 +250,42 @@ def get_link(
...
+            self.log.info("Getting task_id %s", ti_key.task_id)
+            self.log.info("Getting dag_id %s", ti_key.dag_id)
+            self.log.info("Getting execution_date %s", ti.execution_date)
+            metadata = XCom.get_one(
+                task_id=ti_key.task_id,
+                dag_id=ti_key.dag_id,
+                execution_date=ti.execution_date,
+                key="return_value",
+            )
+        self.log.info("metadata %s", metadata)
```
pankajkoti, line 287:
```suggestion
```

cosmos/providers/databricks/plugin.py
```diff
@@ -165,19 +250,42 @@ def get_link(
...
+                ti_key=ti_key,
+                key="return_value",
+            )
+        except AttributeError:
+            if not ti:
+                raise TaskInstanceNotFound()
+            self.log.info("Getting XCOM")
+            self.log.info("Getting task_id %s", ti_key.task_id)
+            self.log.info("Getting dag_id %s", ti_key.dag_id)
+            self.log.info("Getting execution_date %s", ti.execution_date)
```
pankajkoti, line 280:
```suggestion
```

pankajkoti:
Closing this PR as we have created the PR in the new repo: https://github.com/astronomer/astro-providers-databricks/pull/4

PR #133, opened 2023-02-08
Add KubernetesPodOperator & DockerOperator management

juldrixx:
This pull request creates two operators, `DbtKubernetesBaseOperator` and `DbtDockerBaseOperator`, to clone the logic in the `DbtBaseOperator`  with the same subclasses (`DbtLSOperator`, `DbtSeedOperator`, `DbtRunOperator`, ...) but to use a `KubernetesPodOperator` or a `DockerOperator`.

I'm trying to meet a community need (see https://github.com/astronomer/astronomer-cosmos/issues/128 or https://github.com/astronomer/astronomer-cosmos/issues/97).

My PR is not perfect at all I think, but I will open it to start discussions and make it evolve to improve it as these discussions go on.

PR #194, opened 2023-03-22
Add KubernetesPodOperator & DockerOperator management

juldrixx:
As https://github.com/astronomer/astronomer-cosmos/pull/133 has been reverted I'm reopenning it to restart the discussion around this feature.


jlaneve:
Hey @juldrixx, thanks for re-opening this. Before releasing from main I tried to use it a bit and couldn't quite get it working. I know the PR is already quite big, but I wanted to get your general thoughts/expectations on a couple items:
- this was merged without docs, which made it difficult to understand how to use. Can we include those as part of this PR?
- how do you think connection/profile sharing should work? We could/should set the env vars in the KPO from the Airflow connection, but how do we ensure the image has the templated profiles.yml file?
- similarly, the image needs the dbt projects on it, right? And you need those projects in Airflow too, right? Can we make that explicit in the docs?

juldrixx:
> Hey @juldrixx, thanks for re-opening this. Before releasing from main I tried to use it a bit and couldn't quite get it working. I know the PR is already quite big, but I wanted to get your general thoughts/expectations on a couple items:
> 
> * this was merged without docs, which made it difficult to understand how to use. Can we include those as part of this PR?
> * how do you think connection/profile sharing should work? We could/should set the env vars in the KPO from the Airflow connection, but how do we ensure the image has the templated profiles.yml file?
> * similarly, the image needs the dbt projects on it, right? And you need those projects in Airflow too, right? Can we make that explicit in the docs?
> * should we set the `self.cmds` instead of `self.arguments` in KPO? Setting `self.arguments` didn't work for me, because it was trying to run a command like `dbt dbt run ...` since the base command from the image was `dbt`

1- Sure, is there a specific way to do it or I can just follow the existing one format?
2- I don't think you should pass the env variables as it was done locally because you don't want your credentials exposed in the description of your pods for example. And it would be quite tricky, I think, to manage the profiles.yaml as it is done locally. Personally, I'm currently using these features on GCP Composer/Airflow, and the profiles.yaml file in our docker image is configured to use BigQuery oauth authentication and we just use worload identity to pass the BigQuery right to the pod. In the same way as with the DockerOperator, I simply set the env variable `GOOGLE_APPLICATION_CREDENTIALS`. The problem is still that you can't manage the `profiles.yaml` profiles in the container. You could create secret in Kubernetes and mount them as env variables, but you won't be able to 100% ensure that the `profiles.yaml` references those env variables.
3- Yep the dbt project is needed in both the image and the Airflow.
4- Yes we could image doing it, I did that because I had an ENTRYPOINT in my image and my base CMD was `/bin/bash`.

If you have any other questions, I'll be happy to answer them. And if you have any other tips on what should be improved, I'll be happy to discuss them (I'm not really an airflow expert like all of you so I'm sure there is room for improvement).

cosmos/providers/dbt/core/operators/kubernetes.py
```diff
@@ -0,0 +1,176 @@
...
+            flags.append("--args")
+            flags.append(yaml.dump(self.args))
+        return flags
+
+    def execute(self, context: Context):
+        cmd_flags = self.add_cmd_flags()
+        return self.build_and_run_cmd(context=context, cmd_flags=cmd_flags)
+
+
+class DbtDepsKubernetesOperator(DbtKubernetesBaseOperator):
```
tatiana, line 164: @juldrixx does the `DepsOperator` make sense in the case of `Docker` and `Kubernetes`? Perhaps it would make more sense to have this as an argument of the `Run` and `Seed` operators, WDYT?

juldrixx: I see what you mean, it's true that doing it in isolation has no use. In our case we execute it during the construction of the Docker image because we have the dbt project in it. That's why I didn't pay attention to it and just copied the logic of the local way.

And in some cases (mine for example), if we install packages in a container at runtime, it triggers a security alert so maybe it should be optional.

but it is certain that if we inject the dbt project into the container, we will have to install these dependencies at runtime. And maybe if you don't want to do it you should provide your own image with them in it.

tatiana: Even in the context of Airflow, the Deps operator should be removed. It would not work if the Deps task runs in a different Airflow worker from where the Run and Seed operators are being run...! But this is a matter for another PR.

Suggestions for the current PR:
1. Remove `DbtDepsKubernetesOperator` and `DbtDepsDockerOperator`
2. We'll need to add to the documentation that `deps` should be handled within the container image-building process

juldrixx: It's done, I also added the re-export for the `DbtDepsLocalOperator` because it wasn't present for some reason

tatiana: Thank you very much!

tatiana: BTW: just as a reference, I logged this ticket: #218

cosmos/providers/dbt/task_group.py
```diff
@@ -28,6 +28,9 @@ class DbtTaskGroup(CosmosTaskGroup):
         Defaults to "after_each"
     :param select: A dict of dbt selector arguments (i.e., {"tags": ["tag_1", "tag_2"]})
     :param exclude: A dict of dbt exclude arguments (i.e., {"tags": ["tag_1", "tag_2"]})
+    :param execution_mode: The execution mode in which the dbt project should be run.
+        Options are "local", "docker", and "kubernetes".
+        Defaults to "local"
```
tatiana, line 33: @juldrixx While testing, I realised that if the user passes something like:
```
    run_models = DbtTaskGroup(
        group_id="jaffle_shop",
        dbt_root_path="./dags/dbt/",
        conn_id="postgres_default",
        execution_mode="kubernetes",
        dbt_args={
            "schema": "public",
            "image": "dbt-jaffle-shop:0.0.6",
            "project_dir": PROJECT_DIR,
            "get_logs": True,
            "conn_id": "postgres_default",
            "image": "dbt-jaffle-shop:0.0.6",
            "is_delete_operator_pod": False,
        }
    )
```
The `project_dir` being set inside the inner DBT tasks, created by the task group, is not  `PROJECT_DIR`, but actually  `dbt_root_path` +  `dbt_project_name`. Whereas this a fair assumption for local execution, I don't think it is for the Dpckerfile - since at the moment, the user controls how the Dockerfile is created, and they may want to organise the files in a different way there. Could we adjust the code so that if the user gives `project_dir` in the `dbt_args`, it takes priority?

juldrixx: I added a parameter `dbt_cmd_flags` in which you can put any parameter of the dbt command that will be used in it.
So it should be more like:

```python
    run_models = DbtTaskGroup(
        group_id="jaffle_shop",
        dbt_root_path="./dags/dbt/",
        conn_id="postgres_default",
        execution_mode="kubernetes",
        dbt_args={
            "schema": "public",
            "image": "dbt-jaffle-shop:0.0.6",
            "get_logs": True,
            "conn_id": "postgres_default",
            "image": "dbt-jaffle-shop:0.0.6",
            "is_delete_operator_pod": False,
            "dbt_cmd_flags": {                
                "project_dir": PROJECT_DIR,
            }
        }
    )
```

([here](https://github.com/astronomer/astronomer-cosmos/pull/194/files#diff-568541a54f2c6b233d209889aba3b3199918a4c06eac3d722fe4ae805cb42962R184))

tatiana: @juldrixx I found it a bit confusing `dbt_cmd_flags` inside `dbt_cmd_flags`. What do you think of changing to:
```
  run_models = DbtTaskGroup(
        group_id="jaffle_shop",
        dbt_root_path="./dags/dbt/",
        conn_id="postgres_default",
        execution_mode="kubernetes",
        operator_args={
            "image": "dbt-jaffle-shop:0.0.6",
            "get_logs": True,
            "is_delete_operator_pod": False,
        },
        dbt_args={
            "schema": "public",
            "project_dir": PROJECT_DIR,
        }
    )
```

If you're up for it, please, go ahead with the change and I'll approve the PR!

juldrixx: I made the change, I'll let you confirm it if it's the way you want it. For now, you can put the parameters in `operator_args` or in `dbt_args`, it will make no difference.

tatiana (approved):

cosmos/providers/dbt/task_group.py
tatiana: Thank you very much, @juldrixx , it all works!

juldrixx:
> Hey @juldrixx - if you can resolve the merge conflicts, this looks ready to merge!

sure

juldrixx:
> > Hey @juldrixx - if you can resolve the merge conflicts, this looks ready to merge!
> 
> sure

Should be good.

jlaneve (approved):

PR #197, opened 2023-03-24
Change the key of the GCP service account keyfile

mmenalla:
Airflow version 2.4.2 connection generates a key_file with the key named "extra__google_cloud_platform__keyfile_dict" while the create_profile_vars_google_cloud_platform() function tries to access it with the keyword "keyfile_dict", which was used in older Airflow versions.

### DIFF
--- a/cosmos/providers/dbt/core/profiles/bigquery.py
+++ b/cosmos/providers/dbt/core/profiles/bigquery.py
@@ -40,7 +40,7 @@ def create_profile_vars_google_cloud_platform(
     https://docs.getdbt.com/reference/warehouse-setups/bigquery-setup
     https://airflow.apache.org/docs/apache-airflow-providers-google/stable/connections/gcp.html
     """
-    bigquery_key_file = json.loads(conn.extra_dejson.get("keyfile_dict"))
+    bigquery_key_file = json.loads(conn.extra_dejson.get("extra__google_cloud_platform__keyfile_dict"))
 
     if not schema_override:
         raise ValueError(


cosmos/providers/dbt/core/profiles/bigquery.py
```diff
@@ -40,7 +40,9 @@ def create_profile_vars_google_cloud_platform(
     https://docs.getdbt.com/reference/warehouse-setups/bigquery-setup
     https://airflow.apache.org/docs/apache-airflow-providers-google/stable/connections/gcp.html
     """
-    bigquery_key_file = json.loads(conn.extra_dejson.get("keyfile_dict"))
```
pohek321, line 43: @mmenalla thanks for opening this! Can we add a check for those running various versions of Airflow? Something like:

```python
bigquery_key_file = conn.extra_dejson.get("extra__google_cloud_platform__keyfile_dict") or conn.extra_dejson.get("keyfile_dict")
bigquery_key_file = json.loads(bigquery_key_file or "{}")

if not bigquery_key_file:
    raise ValueError("BigQuery key file not found in connection parameters")
```

*Disclaimer* I haven't tested this yet, but the idea is that we should accommodate as many versions of Airflow as possible.

pohek321 (changes requested):

cosmos/providers/dbt/core/profiles/bigquery.py
mmenalla: @pohek321 I checked and it works so i pushed the latest changes.

pohek321 (approved):

pohek321:
@mmenalla, tysm for this contribution! I'm going to go ahead and merge it and it'll be included on our next release (which is hopefully going out today).

Issue #150, opened 2023-02-17
Implement dbt snapshot nodes


MikeWallis42:
The classes exist for it but they're not implemented in the rendering part of the project.

pohek321:
@MikeWallis42 are you envisioning that we make a DbtSnapshotOperator that unlocks [this functionality](https://docs.getdbt.com/docs/build/snapshots) to our end users?

MikeWallis42:
@pohek321 yes I'm suggesting that a DbtSnapshotOperator be created and then added as part of the rendering process so that the cosmos rendering process offers support for a broader range of dbt models.

PR #200, opened 2023-03-25
Dbt snapshot and seed support

szecsip:
#150 

Add DbtSnapshot operator:
* cosmos/providers/dbt/core/operators.py

Extend model handling for snapshots:
* cosmos/providers/dbt/parser/project.py

Extend rendering for snapshot:
* cosmos/providers/dbt/render.py

Add dbt_snapshot_dir parameter to TaskGroup:
* cosmos/providers/dbt/task_group.py

And add docs for DbtSnapshot operator:
* docs/reference/dbt-reference.rst
* docs/reference/_generated/cosmos.providers.dbt.DbtSnapshotOperator.rst

### DIFF
--- a/cosmos/providers/dbt/render.py
+++ b/cosmos/providers/dbt/render.py
@@ -15,6 +15,7 @@ from airflow.exceptions import AirflowException
 from cosmos.core.graph.entities import CosmosEntity, Group, Task
 from cosmos.providers.dbt.core.utils.data_aware_scheduling import get_dbt_dataset
 from cosmos.providers.dbt.parser.project import DbtProject
+from cosmos.providers.dbt.parser.project import DbtModelType
 
 logger = logging.getLogger(__name__)
 
@@ -23,6 +24,7 @@ def render_project(
     dbt_project_name: str,
     dbt_root_path: str = "/usr/local/airflow/dbt",
     dbt_models_dir: str = "models",
+    dbt_snapshots_dir: str = "snapshots",
     task_args: Dict[str, Any] = {},
     test_behavior: Literal["none", "after_each", "after_all"] = "after_each",
     emit_datasets: bool = True,
@@ -47,6 +49,7 @@ def render_project(
     project = DbtProject(
         dbt_root_path=dbt_root_path,
         dbt_models_dir=dbt_models_dir,
+        dbt_snapshots_dir=dbt_snapshots_dir,
         project_name=dbt_project_name,
     )
 
@@ -111,12 +114,24 @@ def render_project(
             else:
                 run_args["outlets"] = outlets
 
-        # make the run task
-        run_task = Task(
-            id=f"{model_name}_run",
-            operator_class="cosmos.providers.dbt.core.operators.DbtRunOperator",
-            arguments=run_args,
-        )
+        if model.type == DbtModelType.DBT_MODEL:
+            # make the run task for model
+            run_task = Task(
+                id=f"{model_name}_run",
+                operator_class="cosmos.providers.dbt.core.operators.DbtRunOperator",
+                arguments=run_args,
+            )
+        elif model.type == DbtModelType.DBT_SNAPSHOT:
+            # make the run task for snapshot
+            run_task = Task(
+                id=f"{model_name}_snapshot",
+                operator_class="cosmos.providers.dbt.core.operators.DbtSnapshotOperator",
+                arguments=run_args,
+            )
+        else:
+            logger.error(
+                f"Unknown DBT type."
+            )
 
         # if test_behavior isn't "after_each", we can just add the task to the
         # base group and do nothing else for now
--- a/cosmos/providers/dbt/render.py
+++ b/cosmos/providers/dbt/render.py
@@ -14,8 +14,7 @@ from airflow.exceptions import AirflowException
 
 from cosmos.core.graph.entities import CosmosEntity, Group, Task
 from cosmos.providers.dbt.core.utils.data_aware_scheduling import get_dbt_dataset
-from cosmos.providers.dbt.parser.project import DbtProject
-from cosmos.providers.dbt.parser.project import DbtModelType
+from cosmos.providers.dbt.parser.project import DbtModelType, DbtProject
 
 logger = logging.getLogger(__name__)
 
@@ -129,9 +128,7 @@ def render_project(
                 arguments=run_args,
             )
         else:
-            logger.error(
-                f"Unknown DBT type."
-            )
+            logger.error("Unknown DBT type.")
 
         # if test_behavior isn't "after_each", we can just add the task to the
         # base group and do nothing else for now
--- a/cosmos/providers/dbt/render.py
+++ b/cosmos/providers/dbt/render.py
@@ -2,6 +2,7 @@
 This module contains a function to render a dbt project into Cosmos entities.
 """
 import logging
+import itertools
 
 try:
     from typing import Literal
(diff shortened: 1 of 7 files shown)


cosmos/providers/dbt/render.py
```diff
@@ -111,12 +113,22 @@ def render_project(
...
-            arguments=run_args,
-        )
+        if model.type == DbtModelType.DBT_MODEL:
+            # make the run task for model
+            run_task = Task(
+                id=f"{model_name}_run",
+                operator_class="cosmos.providers.dbt.core.operators.DbtRunOperator",
+                arguments=run_args,
+            )
+        elif model.type == DbtModelType.DBT_SNAPSHOT:
```
pohek321, line 123: @szecsip I think it's safe to assume that people might name their snapshot models the same as their base model (based on dbt's docs [here](https://docs.getdbt.com/docs/build/snapshots#example) they name the snapshot `orders` just like the base model `orders`). With the logic that you are proposing here, if there was an `orders` dbt model, then it would get overridden by the `orders` dbt snapshot. 

To add a visualization see what happens when I add `orders.sql` to my `/snapshots` directory:

```sql
{% snapshot orders_snapshot %}

{{
    config(
      target_schema='sandbox_chronek',
      unique_key='id',
      strategy='timestamp',
      updated_at='updated_at',
    )
}}

select * from {{ source('jaffle_shop', 'orders') }}

{% endsnapshot %}
```

Screenshot 2023-03-27 at 10 58 33 AM

when `jaffle_shop` looked like this previously:

Screenshot 2023-03-27 at 11 01 14 AM



**Can we change this so that both the dbt model and snapshot will parse if they are named the same thing?**

pohek321 (changes requested):
@szecsip thank you so much for opening this PR. I tested and everything is looking really good besides the one comment that I made. Please let me know if you have any questions!

cosmos/providers/dbt/render.py
szecsip: @pohek321 Thank you for the quick testing. It sounds weird to named a modell and snapshot to same, but I could reproduce the issue. I try to find a solution which can handle this case too.

szecsip:
Currently it can handle same model and snapshot filename. Moreover I also add seed supporting. I tested with my Airflow environment and here is the printsreen about it:

image

Seed:
* raw_customers
* raw_orders
* raw_payments

Model:
* stg_customers
* stg_orders
* stg_payments
* customers
* orders

Snapshot:
* orders_snapshot

It was important to the Airflow to give unique task id for each tasks. Therefore I used snapshot's name between {% snapshot **orders_snapshot** %} instead of snapshot's file name (orders).

cosmos/providers/dbt/render.py
pohek321: @szecsip your solution is excellent! Was just telling @jlaneve that it's a really good pattern to follow. I did however get it to fail when changing the `test_behavior` parameter to `after_each` (see more details [here](https://astronomer.github.io/astronomer-cosmos/dbt/configuration.html#testing)). Mainly because cosmos got confused and tried to add a `test` step after the `snapshot`.

Screenshot 2023-04-07 at 3 37 26 PM



I'll put together a commit so that your changes would work with `test_behavior` set to `after_all` or `after_each`

pohek321: @szecsip can I get a 👍/👎 on the latest commit [here](https://github.com/astronomer/astronomer-cosmos/pull/200/commits/98f3414db234f4ba2087ba1a2c3e192b02908048). This will (hopefully) resolve the previous comment that I made above 👆

szecsip: @pohek321 Thank you really much for the quick fix implementation. I would also do some really similar things.

pohek321 (approved):

cosmos/providers/dbt/render.py
```diff
@@ -74,7 +77,9 @@ def render_project(
             )
 
     # iterate over each model once to create the initial tasks
-    for model_name, model in project.models.items():
+    for model_name, model in itertools.chain(
+        project.models.items(), project.snapshots.items(), project.seeds.items()
+    ):
```
DucNgn, line 82: Hi

I wanted to link this to issue #235, which I opened a few days ago. It seems that data is not being seeded before the DbtTaskGroup collects the DBT models. Even after updating to release 0.6.0 (which includes this commit), the problem is still present.

After reviewing the code, I'm wondering if we should consider reordering this chain so that seeding tasks are created first, as shown below:

```
for model_name, model in itertools.chain(
    project.snapshots.items(), 
    project.seeds.items(),
    project.models.items(), 
):
```

I would appreciate your thoughts on this potential solution, and any other insights you may have to address this issue.

Thank you!

szecsip: Hi @DucNgn,

Unfortunatelly I couldn't reproduce the issue #235 in release 0.6.0.
It build dependency and task order from DBT metadata and not from the running order of the for loop. 

I tested with slightly modificated jaffle demo project, which has got the following structure:
├── models
│   ├── customers.sql
│   ├── docs.md
│   ├── orders.sql
│   ├── overview.md
│   ├── schema.yml
│   ├── sources.yml
│   └── staging
│       ├── schema.yml
│       ├── stg_customers.sql
│       ├── stg_orders.sql
│       └── stg_payments.sql
├── seeds
│   ├── raw_customers.csv
│   ├── raw_orders.csv
│   └── raw_payments.csv
├── snapshots
   └── orders.sql

And it generated the following Airflow task group:
image

I used the **ref** keyword in **stg_customers.sql** to reference the **raw_customers.csv** seed:
`select * from {{ ref('raw_customers') }}`
How do you reference to your seed?

If you see "Dependency abc not found for model DbtModel..." message then probably entities Dict doesn't contain the neccessary dependency: https://github.com/astronomer/astronomer-cosmos/blob/astronomer-cosmos-v0.6.0/cosmos/providers/dbt/render.py#L209
It probably means you failed to read your seeds files. If these .csv files doesn't located in your DBT project snapshots subfolder then you have to specify with dbt_snapshots_dir param in DbtTaskGroup. For example: 
```
    dbt_tg = DbtTaskGroup(
        group_id="dbt_tg",
        dbt_root_path="your_root_path_location",
        dbt_snapshots_dir="your_snapshot_dir_location",
        conn_id="your_pg_connection",
        dbt_args={
            "schema": "public",
        },
    )
```
Where are your DBT seed files located? 

I hope I could help you. Please provide more details about the structure of the project and dependencies in case the error still persists.

DucNgn: Hi @szecsip,

I wanted to thank you for your detailed explanation in response to my issue with `DbtTaskGroup`. After reviewing the code again, I believe the problem is that Cosmos is not currently reading `dbt_project.yml` to determine the path to the seed files. This is likely the reason for the errors I'm experiencing.

While we don't currently have a way for clients to change the path to a different seeding folder, I'm planning to open a PR that adds an argument to specify the location of the seed files.

szecsip: Hi @DucNgn ,

I forgot to take out seed_dir argument in few places, but I have already prepared it. If you don't mind I will send a PR to #235 in a day.

DucNgn: Sorry, I didn't see this message earlier. I opened #241 to tackle this already. 
It's just a small change, so no big deals.

szecsip: No problem, thank you for the supplement and congrats on the PR.

Issue #208, opened 2023-03-28
Project parsing fails with empty yaml 


dh-racheldean:
The `_handle_config_file` function in `DbtProject` class will exit silently if no `models` key is found in the specified yaml file - however if the file is completely empty, `config_dict` is loaded as `None`, and the dict lookup fails and throws an exception. The function should handle empty files.

jlaneve:
looks like this was fixed in #209!

PR #209, opened 2023-03-28
Handle possible empty yml file

dh-racheldean:
° Handles the case of empty yml config files in the project directory.
° Close issue #208


pohek321 (approved):
@dh-racheldean tysm for your contribution! Just tested it and it works as expected. Will merge and include in our next release.

PR #214, opened 2023-03-29
Make `shutil.copytree` call compatible with python3.7

jbandoro:
This is another bug we ran into when trying to use v0.5.1 with astronomer-comos[dbt-bigquery]. The `dirs_exist_ok` arg for `shutil.copytree` was [added](https://docs.python.org/3/whatsnew/3.8.html#shutil) in python 3.8.

This PR has an update to support python 3.7


pohek321 (approved):
@jbandoro thanks for catching this! Approved. Will merge and include in a patch release shortly.

PR #230, opened 2023-04-12
Start using tilt for local dev

jlaneve:
This PR adds Tilt for local dev. To use:
- run `tilt up`
- open localhost:8080 to see the Airflow UI

There are lots of optimizations and example DAGs that can be added, but I wanted to start small and improve the experience over time! Note that most of the diff here comes from adding the jaffle_shop project, which we want a local copy of in case we want to make modifications to highlight Cosmos functionality.


iancmoritz (approved):
quick skim mostly honor system for @jlaneve

pohek321 (approved):
LGTM!

PR #236, opened 2023-04-14
Add flag to install deps on execution

### LINKED ISSUE #234: after merging #233 some dbt tasks are failing
After merging #233 there is a problem in running some dbt tasks that rely on 3rd party packages

Screenshot 2023-04-14 at 19 13 09

```
INFO - Compilation Error
INFO -   dbt found 1 package(s) specified in packages.yml, but only 0 package(s) installed in dbt_packages. Run "dbt deps" to install package dependencies.
```

probably this is because of the new isolated nature of the temp folder. maybe by closing #218 this issue is automatically resolved

jlaneve:
This PR updates the local operator to run `dbt deps` if you pass in the `install_deps` flag to any of the operators (or the TaskGroup's `dbt_args`), i.e.

```python
            DbtSeedOperator(
                ...,
                install_deps=True,
            )
```

closes #234

### DIFF
--- a/cosmos/providers/dbt/core/operators/base.py
+++ b/cosmos/providers/dbt/core/operators/base.py
@@ -61,6 +61,7 @@ class DbtBaseOperator(BaseOperator):
     :param dbt_executable_path: Path to dbt executable can be used with venv
         (i.e. /home/astro/.pyenv/versions/dbt_venv/bin/dbt)
     :param dbt_cmd_flags: Flags passed to dbt command override those that are calculated.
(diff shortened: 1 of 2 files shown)


pohek321 (approved):
Love it! Maybe we should remove the DbtDepsOperator now?

ashishmgofficial:
@pohek321 Yes we should. I was stuck with this as I was trying 0.6.1 and had to reach here to identify the solution. I was using DBTDepsOperator and wondering why others are not picking the external packages.

PR #241, opened 2023-04-18
Add support for specifying custom Dbt Seeds Dir

### LINKED ISSUE #235: ERROR [cosmos.providers.dbt.render] Dependency not found
Hi there,

I'm experiencing an issue with my Airflow project that uses `DbtTaskGroup`. Specifically, I keep getting an error message that reads:

```
ERROR [cosmos.providers.dbt.render] Dependency abc not found for model DbtModel(name='xyz'...)
```

I believe the missing dependency is a data file in `.csv` format.

My understanding is that Cosmos should seed the data before collecting the DBT tasks. However, if this is not the case, I would appreciate any guidance on the proper way to run seeding before DbtTaskGroup collects DBT models.

Thank you for your help!

DucNgn:
Closes #235
Clients can specify the location of the seeds folder with an argument in `DbtTaskGroup`

### DIFF
--- a/cosmos/providers/dbt/dag.py
+++ b/cosmos/providers/dbt/dag.py
@@ -21,6 +21,7 @@ class DbtDag(CosmosDag):
     :param dbt_project_name: The name of the dbt project
     :param dbt_root_path: The path to the dbt root directory
     :param dbt_models_dir: The path to the dbt models directory within the project
+    :param dbt_seeds_dir: The path to the dbt seeds directory within the project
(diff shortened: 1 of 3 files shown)


jlaneve (commented):
thank you, and congrats on the first PR! do you mind adding this to the `DbtDag` too?

DucNgn:
@jlaneve Just did! Thanks for the suggestion!

jlaneve (approved):

PR #243, opened 2023-04-18
Lazy load Docker and Kubernetes better for backwards compatibility

jlaneve:
This PR changes how we lazy load the Kubernetes and Docker required packages so that old imports (`from cosmos.providers.dbt import ...`) still work. It also adds the optional dependencies to the docs.


tatiana (approved):
@jlaneve, thanks a lot for improving this. It helps with backward compatibility.

The one thing that needs to be added is tests. I'm happy for this change to be merged without tests for now, but let's ensure we add tests covering this feature and future PRs.

PR #252, opened 2023-04-19
Measure test coverage and increase overall project test coverage

jlaneve:
* Add Github workflow to run the project's unit tests
* Run tests against multiple versions of Python
* Increase test coverage to 86%

Allow developers to check the tests coverage locally with one environment by running:
```
hatch run tests.py3.10-2.5:test-cov
```

or with all environments by running:
```
hatch run tests:test-cov
```


cosmos/providers/dbt/core/profiles/tests/profiles_test.py
```diff
@@ -211,7 +203,7 @@ def test_create_profile_vars_postgres(airflow_connection: Connection) -> None:
         "POSTGRES_HOST": host,
         "POSTGRES_USER": login,
         "POSTGRES_PASSWORD": password,
-        "POSTGRES_DATABASE": schema,
+        "POSTGRES_DATABASE": database,
```
tatiana, line 206: Currently, there is an issue in the Airflow Postgres Hook where schema references the database. Was this change accidental?

pyproject.toml
```diff
@@ -113,6 +113,37 @@ include = [
...
+    "apache-airflow-providers-docker>=3.5.0",
+    "apache-airflow-providers-cncf-kubernetes>=5.1.1",
+]
+
+[[tool.hatch.envs.tests.matrix]]
+python = ["3.7", "3.8", "3.9", "3.10"]
+
+[tool.hatch.envs.tests.scripts]
+test = "pytest ."
+test-cov = "pytest --cov=cosmos --cov-report=term-missing --cov-report=xml ."
```
tatiana, line 141: One of the benefits of having separate directories for tests and code is that we could check the coverage only in the actual non-test modules. Currently, we're checking the coverage both in tests and code, which can be very misleading.

pyproject.toml
```diff
@@ -113,6 +113,37 @@ include = [
...
+    "types-requests",
+    "mypy",
+    "sqlalchemy-stubs", # Change when sqlalchemy is upgraded https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html
+
+    "apache-airflow-providers-docker>=3.5.0",
+    "apache-airflow-providers-cncf-kubernetes>=5.1.1",
+]
+
+[[tool.hatch.envs.tests.matrix]]
+python = ["3.7", "3.8", "3.9", "3.10"]
```
tatiana, line 137: @jlaneve, were you able to identify a hatch native way for us to test against multiple versions of Airflow? Or should we follow, as part of a separate PR, the `nox` approach we adopted in the Databricks providers package (https://github.com/astronomer/astro-provider-databricks)?

pyproject.toml
jlaneve: I just fixed this issue - good catch!

pyproject.toml
jlaneve: yes - this now works locally and in the actions! see the `pyproject.toml` file and the `.github/workflows/test.yaml` files, respectively.

cosmos/providers/dbt/core/profiles/tests/profiles_test.py
jlaneve: good catch, I added a test specifically for that here: https://github.com/astronomer/astronomer-cosmos/pull/252/commits/64d99ba10af26e5b25a8acce507b2936fe8c1586

pohek321 (approved):
@jlaneve thanks for the walk through! Approved.

cosmos/providers/dbt/core/profiles/tests/test_profiles.py
```diff
@@ -205,13 +197,14 @@ def test_create_profile_vars_postgres(airflow_connection: Connection) -> None:
     airflow_connection.host = host
     airflow_connection.login = login
     airflow_connection.password = password
+    airflow_connection.schema = database
```
tatiana, line 200: It may be worth adding a link to the Airflow Postgres hook:
https://github.com/apache/airflow/blob/main/airflow/providers/postgres/hooks/postgres.py#L73

And, perhaps, if the user uses newer Airflow versions to set the `database` parameter.

tatiana (approved):

PR #253, opened 2023-04-19
fixes #203 enable project tags

### LINKED ISSUE #203: Update dbt select/exclude documentation
This [section](https://github.com/astronomer/astronomer-cosmos/blob/main/docs/dbt/configuration.rst?plain=1#L32-L35) suggests that you should structure your select or exclude like this.

```python3
select={"configs": "tags:daily"}
```
This too did not work but at least did not result in no tasks being returned.
```python3
select={"tags": "daily"}
```
I'm setting my tags in dbt_project.yml, not schema.yml or directly inside the models like below.
Is this supported?
```yaml
models:
  jaffle_shop:
      staging:
        +tags:
          - hourly
      modelling:
        +tags:
          - daily
      mart:
        +tags:
          - daily
```

MikeWallis42:
fixes #203 enable project tags


cosmos/providers/dbt/parser/project.py
```diff
@@ -124,30 +118,28 @@ def __post_init__(self) -> None:
...
         ast = env.parse(code)
-
         # iterate over the jinja nodes to extract info
-        for base_node in ast.find_all(jinja2.nodes.Call):
-            if hasattr(base_node.node, "name"):
+        for jinja_call_node in ast.find_all(Call):
+            if hasattr(jinja_call_node.node, "name"):
                 # check we have a ref - this indicates a dependency
-                if base_node.node.name == "ref" or base_node.node.name == "source":
+                # TODO Do we care about source references? (unless we add source freshness) also a source takes 2 args.
```
jlaneve, line 130: yeah, we probably need to do a full spike on sources - there are lots of ways to handle them. I think we have a few issues opened. I'd say let's keep the behavior as-is for now and do a deeper dive into sources as a follow-on.

cosmos/providers/dbt/parser/project.py
```diff
@@ -157,28 +149,24 @@ def __post_init__(self) -> None:
...
-            except Exception as e:
-                # if we can't convert it to a constant, we can't do anything with it
+            except Impossible as e:
                 logger.warning(
                     f"Could not parse {config_name} from config in {self.path}: {e}"
                 )
                 pass
+            else:
+                parsed_value = None
+                if isinstance(value, List):
```
jlaneve, line 165: I think this should be `list`, right?
```suggestion
                if isinstance(value, list):
```

cosmos/providers/dbt/parser/project.py
```diff
@@ -213,37 +199,38 @@ class DbtProject:
...
             self._handle_csv_file(file_name)
-
-        # crawl the config files in the project
-        for file_name in self.models_dir.rglob("*.yml"):
-            self._handle_config_file(file_name)
+        for root, dirs, files in os.walk(self.models_dir, topdown=False):
+            for file in files:
+                if file.endswith(".yml"):
+                    path = Path(root) / file
+                    self._handle_config_file(path)
```
jlaneve, line 224: curious why the logic was changed here - does the previous version not work properly?

cosmos/providers/dbt/parser/project.py
```diff
@@ -310,19 +291,78 @@ def _handle_config_file(self, path: Path) -> None:
...
+                selectors,
+                Path(self.dbt_root_path) / project_name / "models" / model_folder,
+            )
+
+    def _handle_project_config_recursive(
+        self, project_config: dict, selectors: list[str], project_path: Path
+    ) -> None:
+        """Goes bottom up to get tags against models (leaves to root).
+
+        This is because the most specific wins for some types.
```
jlaneve, line 323: I'm not sure I understand what this comment means, do you mind explaining more/rewording?

pyproject.toml
```diff
@@ -113,6 +113,20 @@ include = [
...
+    "pytest-split",
+    "pytest-dotenv",
+    "requests-mock",
+    "pytest-cov",
+    "pytest-describe",
+    "types-requests",
+    "mypy",
+    "sqlalchemy-stubs", # Change when sqlalchemy is upgraded https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html
+]
+
```
jlaneve, line 129: note: we have a [PR open](https://github.com/astronomer/astronomer-cosmos/pull/252) to make the testing experience better

cosmos/providers/dbt/parser/project.py
```diff
@@ -286,21 +271,17 @@ def _handle_config_file(self, path: Path) -> None:
         """
         # parse the yml file
         config_dict = yaml.safe_load(path.read_text())
-
         # iterate over the models in the config
-        if not (config_dict and config_dict.get("models")):
+        if not config_dict.get("models"):
```
tatiana, line 275: Why did we change this line? There are circumstances when `config_dict` may be `None`, based on the `yaml.safe_load` implementation:
https://github.com/yaml/pyyaml/blob/master/lib/yaml/constructor.py#L52
https://github.com/yaml/pyyaml/blob/master/lib/yaml/__init__.py#L81

tatiana (commented):
@MikeWallis42 Thanks a lot for the contribution - I left some comments in-line; could you help get the checks (tests, coverage) passing?

jlaneve:
this should be fixed with a combination of https://github.com/astronomer/astronomer-cosmos/pull/385 and https://github.com/astronomer/astronomer-cosmos/pull/360

PR #256, opened 2023-04-26
Update the default value for the dbt_root_path

pohek321:
This has been throwing a lot of customers off ever since we changed the dbt parent to be within `/dags`. Proposing that we change the default to match our docs.


jlaneve:
Nice! Do we need any docs changes here?

pohek321:
> Nice! Do we need any docs changes here?

I don't think so!

jlaneve (approved):

PR #270, opened 2023-05-12
explicitly ignore user-supplied profiles file for now

### LINKED ISSUE #257: Ignore profiles.yml in dbt project directories
If customers have `profiles.yml` files in their dbt project directories, it will collide with the one that cosmos generates under the hood by default. We need to add logic so that it gets ignored.

jlaneve:
## Description

Previously, you could put a `profiles.yml` file in a dbt project, leading to unintended behaviors. For now, all profiles are autogenerated by Cosmos. https://github.com/astronomer/astronomer-cosmos/issues/269 deals with user-supplied profiles.yml files.

## Related Issue(s)

closes #257

## Breaking Change?

There may be cases where a user did supply a `profiles.yml` file, and now this could potentially break that setup. But it's unlikely it would break things, because Cosmos enforces that you pass a connection.

## Checklist

### DIFF
--- a/cosmos/providers/dbt/core/operators/local.py
+++ b/cosmos/providers/dbt/core/operators/local.py
@@ -129,7 +129,9 @@ class DbtLocalBaseOperator(DbtBaseOperator):
             profiles_path = os.path.join(tmp_project_dir, "profiles.yml")
             if os.path.exists(profiles_path):
                 logger.warning(
-                    "Cosmos doesn't support user-supplied profiles.yml files. Removing..."
+                    "Cosmos doesn't support user-supplied profiles.yml files. \
+                    Please use Airflow connections instead. \
+                    Ignoring profiles.yml…"
                 )
                 os.remove(profiles_path)
 
--- a/cosmos/providers/dbt/core/operators/local.py
+++ b/cosmos/providers/dbt/core/operators/local.py
@@ -128,7 +128,9 @@ class DbtLocalBaseOperator(DbtBaseOperator):
             # ensure there's no profiles.yml file in the project dir
             profiles_path = os.path.join(tmp_project_dir, "profiles.yml")
             if os.path.exists(profiles_path):
-                logger.warning("Cosmos doesn't support user-supplied profiles.yml files. Removing...")
+                logger.warning(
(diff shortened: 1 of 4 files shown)


cosmos/providers/dbt/core/operators/local.py
```diff
@@ -125,6 +125,14 @@ def run_command(
                 tmp_project_dir,
             )
 
+            # ensure there's no profiles.yml file in the project dir
+            profiles_path = os.path.join(tmp_project_dir, "profiles.yml")
+            if os.path.exists(profiles_path):
+                logger.warning(
+                    "Cosmos doesn't support user-supplied profiles.yml files. Removing..."
```
tatiana, line 132: It may be worth to mention this is a limitation of the local execution mode:
```suggestion
                    "Cosmos doesn't support user-supplied profiles.yml files when using local execution mode."
```

cosmos/providers/dbt/core/operators/local.py
```diff
@@ -125,6 +125,14 @@ def run_command(
...
+                )
+                os.remove(profiles_path)
```
tatiana, line 134: It's probably worth not deleting the `profiles_path` since users may be using it to build the Docker image for the docker or K8s modes.

cosmos/providers/dbt/core/operators/local.py
```diff
@@ -111,6 +111,20 @@ def store_compiled_sql(
...
+        user-supplied profiles.yml files.
+        """
+        profiles_path = os.path.join(project_dir, "profiles.yml")
+        if os.path.exists(profiles_path):
+            logger.warning(
+                "Cosmos doesn't support user-supplied profiles.yml files. \
+                    Please use Airflow connections instead. \
+                    Ignoring profiles.yml…"
+            )
+            os.remove(profiles_path)
```
tatiana, line 126: @jlaneve, couldn't we ignore the `profiles_path` without deleting it..? I'm slightly worried about us removing a file which users wouldn't expect to be removed. Let's say they sometimes use DBT directly to troubleshoot things, etc. Or they are building the docker container from this folder so that they can use the K8s/Docker Cosmos `execution mode` (and still need `profiles.yml`).

Issue #267, opened 2023-05-09
Key-based auth for Snowflake


jschenz-chwy:
Hello all.  I have been looking at `astronomer-cosmos` recently, but am a bit stuck on the auth piece of it.  We are using key-based auth, and I see that it is not supported.  Hoping for a bit of guidance as I start digging into this to see if I can add that option.  It looks like there a few things to consider:

- There is a collection of profiles, from each of the files under `cosmos.providers.dbt.core.profiles`.  
- These profiles are all specified as using password-based auth.
- There is currently no way to update the profile structure based on the Airflow connection. 

My initial concept is to replace the current process of creating the `profiles.yaml` file with one that would interrogate the available connections and generate the corresponding format for the profile.  To do that I would need to do a few things:
1. Get a list of supported connections
2. For each connection, create a connection object and interrogate that to understand what type of auth is being applied.
3. Modify the yaml in `profiles.yaml` for that connection so that it uses the correct auth type.

Curious if others are already working on some solution for key-based auth already.  If so, I am happy to help out.  If not, I would love to get some feedback on my thought process for this.

Thanks!

jlaneve:
Hey @jschenz-chwy, the profiles have been cleaned up and now should be very easy to contribute to. Check out [the docs](https://astronomer.github.io/astronomer-cosmos/dbt/connections-profiles.html#connections-and-profiles) and please let me know if you have any questions! If you don't plan on contributing this, also let us know and we can try to open a PR.

HybridNeos:
Commenting for visibility. This feature would be much appreciated as some organizations only allow private key authentication.

ivanstillfront:
We are also using key-based auth. I see https://github.com/dbt-labs/dbt-snowflake/pull/671 needs a rebase? Anything else that I can help with?

jschenz-chwy:
@ivanstillfront I just merged main into my branch on the dbt-snowflake PR.  Still waiting on a reviewer though.

patawan:
Is there any way to get some traction on the PR open in dbt-snowflake? Seems like it's just been chilling for a while

jschenz-chwy:
Just merged `main` into my branch again.  Not sure how to get attention over there.  No doubt everyone is focused on other things, but it would be nice to get some feedback at least.

patawan:
Thanks for getting that merged @jschenz-chwy! I opened a PR here #378  for the cosmos functionality. I need to add some docs but please let me know what all needs cleaned up.

Issue #266, opened 2023-05-09
--target=dev in all environments


eldan-dario:
When running task (within DbtTaskGroup), when checking the logs both in dev and prod, the target doesn't seem to change (it stays target='dev'). 

![Screenshot 2023-05-09 at 11 38 53](https://user-images.githubusercontent.com/119860732/237043136-1c1cdbe6-0062-4a01-a697-fa83e9dbf510.png)

I believe it's to do with astronomer-cosmos generating a snowflake profile (aka "snowflake_profile") based on the Airflow connection, but is there a possibility to set this as a dbt_arg?

jlaneve:
Hey @eldan-dario - you're correct that astronomer-cosmos takes care of translating Airflow connections to profiles for you. This means the profiles should stay the same across environments, but the values in the profile should change depending on the Airflow connection. To ensure this is a bug, can you confirm:
- you're using different Airflow connections for dev and prod (where, for example, the schema is different)
- the dbt models _are_ being run against the same connections (e.g. the schema remains the same)

If you _are_ providing different Airflow connections but the target values remain the same, then this is a bug and I can look into it!

dschultz5197:
I would advocate for target being an option that can actually be used. We're using macros which reference the target name to help determine the database to be written to, or the sources to read from. While we can use different Airflow connections to change the database depending upon the environment, this will only apply to the single database set as the default in the connection, and will not apply to all of the databases referenced in our dbt project.

For reference, we're using target.name in a similar manner to the examples in the following dbt documentation: https://docs.getdbt.com/reference/dbt-jinja-functions/target#examples

jlaneve:
@dschultz5197 I agree here - do you think the solution is as simple as passing a `target_name`? Then Cosmos could ensure the profile name gets translated properly.

eldan-dario:
Hey @jlaneve, following your requests on confirming:
**1. _Using different Airflow connections for dev and prod (where, for example, the schema is different)_**

- Prod connection:

![conn prod test](https://github.com/astronomer/astronomer-cosmos/assets/119860732/f58a7799-6325-46f6-a739-3d7dc68a4cc2)

- Prod connection dbt dag run  (logs `--target='dev'`):

![dag prod dbt](https://github.com/astronomer/astronomer-cosmos/assets/119860732/c75cb800-175c-4809-a895-55a93d2b8625)


- Dev connection:

[...]

PR #271, opened 2023-05-12
[looking for feedback] refactor profiles for maintainability

jlaneve:
## Description

This PR refactors how we deal with the Airflow connection -> dbt profile translation. Previously, there was no standard, meaning maintaining and contributing profiles was really difficult. This PR fixes that issue by ensuring there's a consistent way to contribute profiles: a class that inherits from `BaseProfileMapping` and defines at least:

- `airflow_connection_type`: the Airflow connection type that this profile mapping is for.
- `required_fields`: a list of required fields for the profile. This refers to the field as it's named in the dbt profile.
- `secret_fields`: a list of fields that are secret. These fields will be passed to dbt as environment variables.
- `airflow_param_mapping`: a dictionary that maps the Airflow connection fields to the dbt profile fields. The keys are the Airflow connection fields and the values are the dbt profile fields.
- Optionally, a profile mapping can specify a `transform_{dbt_field_name}` function for each dbt profile field. This function will be called on the value of the Airflow connection field before it is passed to dbt. This is useful for transforming the value of a field before it is passed to dbt. For example, sometimes host fields need to be passed to dbt without the http:// prefix.

This PR also lays the groundwork for a few user-requested features:
- supplying your own profile.yml instead of translating Airflow connections
- supplying additional arguments to the profile that aren't present in the Airflow connection (e.g. `threads`)
- modifying the target name (https://github.com/astronomer/astronomer-cosmos/issues/266)
- key-based auth (https://github.com/astronomer/astronomer-cosmos/issues/267)
- various updates to existing connections and new connection requests (lots of issues here, not linking)

**For reviewers...**

There was no easy way to avoid how large this PR is, because it required changing every aspect of how we manage profiles. I recommend reviewing a few files in particular:
- https://github.com/astronomer/astronomer-cosmos/blob/refactor-profiles/cosmos/providers/dbt/core/profiles/base.py - this contains logic shared by all new profile mappings
- each of the profiles in [the profiles dir](https://github.com/astronomer/astronomer-cosmos/tree/refactor-profiles/cosmos/providers/dbt/core/profiles) that you care about
- the [new docs page](https://github.com/astronomer/astronomer-cosmos/blob/refactor-profiles/docs/dbt/connections-profiles.rst)

## Related Issue(s)

nothing explicitly closed by this PR, but this will help with future issues quite a bit

## Breaking Change?

There is a breaking change, but a workaround is included in the code with a deprecation warning.

Previously, a user would have to specify a `schema` like so:

```python
basic_cosmos_dag = DbtDag(
    dbt_args={"schema": "public"},
    ...,
)
```

Now, you specify it like so:

```python
basic_cosmos_dag = DbtDag(
    ...,
)
```

## Checklist


jlaneve:
I've tagged a prerelease of this because I can't test all of the connection types myself unfortunately. If anyone uses Cosmos for connections outside Postgres + Snowflake, I'd love your help testing! Should be as simple as bumping the version to `0.7.0rc1`.

https://pypi.org/project/astronomer-cosmos/0.7.0rc1/

Tagging a few folks below who have helped contribute to the various profiles:
@MikeWallis42 can you help with Trino and/or Databricks?
@jensenity can you help test Spark + Thrift?
@nepux can you help with Exasol?
@mmenalla can you help with BigQuery?

cosmos/providers/dbt/core/profiles/trino/jwt.py
```diff
@@ -0,0 +1,44 @@
...
+        common_profile_vars = super().get_profile()
+
+        # need to remove jwt from profile_args because it will be set as an environment variable
+        profile_args = self.profile_args.copy()
+        profile_args.pop("jwt", None)
+
+        profile_vars = {
+            **common_profile_vars,
+            "method": "jwt",
+            "jwt": self.get_env_var_format("jwt"),
```
matthewzahar, line 39: wrong name for jwt token in trino profile
```suggestion
            "jwt_token": self.get_env_var_format("jwt"),
```

jlaneve: Ah yep great catch @matthewzahar. Tagged `v0.7.0rc2` with this fix!

https://pypi.org/project/astronomer-cosmos/0.7.0rc2/

pyproject.toml
```diff
@@ -154,6 +154,14 @@ freeze = "pip freeze"
 test = "pytest ."
 test-cov = "pytest --cov=cosmos --cov-report=term-missing --cov-report=xml ."
 
+######################################
+# PYTEST
+######################################
+[tool.pytest.ini_options]
+filterwarnings = [
+    "ignore::DeprecationWarning",
```
tatiana, line 162: Why are we adding this..?

jlaneve: There were lots of deprecation warnings from packages used by Airflow and other dependencies that grew to be quite noisy. Open to other ideas to make the test output more readable, but if you run them with and without this line you'll see the difference.

cosmos/__init__.py
```diff
@@ -2,4 +2,4 @@
 Astronomer Cosmos is a library for rendering 3rd party workflows in Airflow.
 """
 
-__version__ = "0.6.7"
+__version__ = "0.7.0rc2"
```
tatiana, line 5: It is probably worth having a separate PR with the release tag, so we have independent commits

cosmos/providers/dbt/core/operators/base.py
```diff
@@ -139,6 +126,8 @@ def __init__(
         dbt_ol_path = shutil.which("dbt-ol")
         if dbt_executable_path == "dbt" and shutil.which("dbt-ol"):
             self.dbt_executable_path = dbt_ol_path
+        elif dbt_executable_path == "dbt":
```
tatiana, line 129: How does this change relate to the profile refactor?
Let's introduce it in a separate PR.

cosmos/providers/dbt/core/operators/base.py
```diff
@@ -139,6 +126,8 @@ def __init__(
...
+            self.dbt_executable_path = shutil.which("dbt")
```
tatiana, line 130: How does this change relate to the profile refactor?
Perhaps we should introduce it in a different PR.

README.rst
```diff
@@ -59,7 +59,7 @@ You can render an Airflow Task Group using the ``DbtTaskGroup`` class. Here's an
         dbt_tg = DbtTaskGroup(
             dbt_project_name="jaffle_shop",
             conn_id="airflow_db",
-            dbt_args={
+            profile_args={
```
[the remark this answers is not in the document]
jlaneve: Good question, I left it as `profile_args` because these are explicitly placed into the dbt profile. In vanilla dbt you can define multiple profiles, just as with Airflow you can have multiple connections. But you'd typically only use one per dbt project, just as you'd typically only use Airflow connection per operator. 

Before 1.0 we should also look at the params a user has to pass into the cosmos objects and make sure they make sense. Lumping everything under a single untyped `dbt_args` doesn't make a whole lot of sense to me because the arguments are used in different ways. This is just the first opportunity to do that.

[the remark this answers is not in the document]
jlaneve: If you feel strongly that this should be `profile_target_args` I'd be open to that, but I don't think it should cause that much confusion as-is because you only specify a single target/profile. We don't really have "single profile with many targets" like in vanilla dbt.

cosmos/__init__.py
jlaneve: Yeah, just have this in there for now to be able to tag pre-releases for testing.

cosmos/providers/dbt/core/operators/base.py
```diff
@@ -71,8 +63,6 @@ class DbtBaseOperator(BaseOperator):
         "selector",
         "vars",
         "models",
-        "profiles_dir",
-        "profile",
```
tatiana, line 75: This looks like a braking change. Why are we removing these parameters from the `Operator` level?

cosmos/providers/dbt/core/operators/local.py
```diff
@@ -128,6 +130,29 @@ def run_command(
                 tmp_project_dir,
             )
 
+            # get the profile name from the dbt_project.yml file
+            dbt_project_path = os.path.join(tmp_project_dir, "dbt_project.yml")
+            with open(dbt_project_path, encoding="utf-8") as f:
+                dbt_project = yaml.safe_load(f)
```
tatiana, line 136: It would be great if we could add tests to cover this part of the code.

cosmos/providers/dbt/core/operators/base.py
```diff
@@ -82,6 +72,7 @@ class DbtBaseOperator(BaseOperator):
         "warn_error",
     )
 
+    include_system_env = False
```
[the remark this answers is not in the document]
jlaneve: good catch, old code - removed

cosmos/providers/dbt/core/operators/local.py
jlaneve: Not sure I understand what you want tested here - opening the file? I added another check to ensure the file exists, let me know if you still want more.

jensenity (approved):

cosmos/providers/dbt/core/operators/base.py
jlaneve: This does relate to the env change - there was [incorrect code](https://github.com/astronomer/astronomer-cosmos/blob/1bce508f1f71553d33c72bee6aafb6169b3a4d0d/cosmos/providers/dbt/core/operators/base.py#L158-L160) on main that would append the system env if there was no `env` argument explicitly passed to the operator. When removing that, if you don't specify a dbt executable path, it would fail because it wasn't in the `PATH`.

cosmos/providers/dbt/core/profiles/base.py
```diff
@@ -0,0 +1,167 @@
...
+import yaml
+from typing_extensions import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from airflow.models import Connection
+
+logger = getLogger(__name__)
+
+
+class BaseProfileMapping:
```
tatiana, line 19: WDYT if we made this an [abstract class](https://docs.python.org/3/library/abc.html) inheriting from `abc.ABC`?

cosmos/providers/dbt/core/profiles/base.py
```diff
@@ -0,0 +1,167 @@
...
+
+        return True
+
+    def get_profile(self) -> dict[str, Any]:
+        """
+        Return a dbt profile based on the Airflow connection.
+        """
+        raise NotImplementedError
+
+    def get_env_vars(self) -> dict[str, str]:
```
tatiana, line 68: WDYT if we had used Python `@property` for getters which don't need a parameter, such as:
* `get_profile` -> `profile`
* `get_env_vars` -> `env_vars`

tatiana:
@jlaneve I rebased to try out things, but I didn't want to override the current PR before we aligned. For now, I pushed in a separate branch: `refactor-profiles-rebase`.

I left some last minor comments in-line.

Once the checks are passing, including the integration tests which run our example DAGs, I'm happy for this PR to be merged.

tatiana (approved):

PR #274, opened 2023-05-17
Update contributing guidelines with hatch test cmd

tatiana:
Document a lightweight way of running tests in a local development environment.

This is how the Sphinx documentation renders with this change:

Screenshot 2023-05-17 at 14 39 38


jlaneve (approved):
This is awesome! Can we also add a line about running against the matrix of python and Airflow versions with hatch run tests:test-cov?

Preemptively approving

PR #275, opened 2023-05-17
Use hatch to run tests in the CI

tatiana:
Have a consistent way of installing dependencies when installing and running tests locally and in the CI, by using `hatch`.


jlaneve (approved):
Nice!

Issue #222, opened 2023-04-02
Manage virtual env automatically


iglikoxha:
Instead of manually having to create a virtual env just to install dbt and then specify the path to dbt_executable_path. Wouldn't it be neat if this was done easier and automatically. Something similar to PythonVirtualenvOperator.

This would make this operator also runnable in Google Cloud Composer where installing dbt in the same environment as Airflow is a headache since it conflicts with a lot of packages, also having a custom Airflow image with the dbt_venv environment in there is impossible.

Also maybe this should be implemented in the CosmosDag level since I suppose is a feature that will be needed in the future for other authoring interfaces since we can't guarantee that they will not have any conflicts with airflow python packages.

CorsettiS:
This is actually a very good idea and not that cumbersome to implement since the skeleton is already defined [here](https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/operators/python.html#PythonVirtualenvOperator) for the PythonVirtualenvOperator. It also makes it possible for teams that dont have direct access to the airflow infrastructure to run this library. Can you create a PR for that ? otherwise I may do it once my workload reduces

jlaneve:
Great idea! There's another PR that's already separating out the execution methods (to rename local and add docker & Kubernetes). It could be worth holding off until that PR's merged, because that'll establish how to add operators for different execution methods.

CorsettiS:
@jlaneve alright, once the new release with this new change is deployed, I'll work on this then ! Thanks for telling us

PR #279, opened 2023-05-18
Add `virtualenv` execution mode to run DBT commands

tatiana:
Introduce a new Cosmos execution mode: `virtualenv`, as an alternative to the existing execution modes: `local`, `kubernetes` and `docker`. 

This aims to simplify the user journey, so they don't need to manually create a Python virtual environment and define the argument `dbt_executable_path`.

The `execution_mode="virtualenv"` allows users of several cloud-managed Airflow services, such as Google Cloud Composer, to use Cosmos/DBT in an isolated way without dealing with Python dependency conflict hell - without the overhead of dealing with containers, Docker or Kubernetes.

Closes: #222

### DIFF
--- a/cosmos/providers/dbt/core/operators/local.py
+++ b/cosmos/providers/dbt/core/operators/local.py
@@ -36,10 +36,6 @@ class DbtLocalBaseOperator(DbtBaseOperator):
 
     :param install_deps: If true, install dependencies before running the command
     :param callback: A callback function called on after a dbt run with a path to the dbt project directory.
-    :param venv_requirements: Creates a virtual environment with the specified arguments
-    :param venv_system_site_packages: Whether or not all the Python packages from the Airflow instance, will be
-           accessible within the virtual environment (if py_requirements argument is specified).
-           Avoid using unless the DBT job requires it.
     """
 
     template_fields: Sequence[str] = DbtBaseOperator.template_fields + ("compiled_sql",)
--- a/cosmos/providers/dbt/core/operators/local.py
+++ b/cosmos/providers/dbt/core/operators/local.py
@@ -36,6 +36,10 @@ class DbtLocalBaseOperator(DbtBaseOperator):
 
     :param install_deps: If true, install dependencies before running the command
     :param callback: A callback function called on after a dbt run with a path to the dbt project directory.
+    :param venv_requirements: Creates a virtual environment with the specified arguments
+    :param venv_system_site_packages: Whether or not all the Python packages from the Airflow instance, will be
+           accessible within the virtual environment (if py_requirements argument is specified).
+           Avoid using unless the DBT job requires it.
     """
 
     template_fields: Sequence[str] = DbtBaseOperator.template_fields + ("compiled_sql",)
@@ -120,6 +124,9 @@ class DbtLocalBaseOperator(DbtBaseOperator):
             )
             os.remove(profiles_path)
 
+    def run_subprocess(self, *args, **kwargs):
+        return self.subprocess_hook.run_command(*args, **kwargs)
+
     def run_command(
         self,
         cmd: list[str],
@@ -147,7 +154,7 @@ class DbtLocalBaseOperator(DbtBaseOperator):
 
             # if we need to install deps, do so
             if self.install_deps:
-                self.subprocess_hook.run_command(
+                self.run_subprocess(
                     command=[self.dbt_executable_path, "deps"],
                     env=env,
                     output_encoding=self.output_encoding,
@@ -156,7 +163,7 @@ class DbtLocalBaseOperator(DbtBaseOperator):
 
             logger.info(f"Trying to run the command:\n {cmd}\nFrom {tmp_project_dir}")
 
-            result = self.subprocess_hook.run_command(
+            result = self.run_subprocess(
                 command=cmd,
                 env=env,
                 output_encoding=self.output_encoding,
--- a/cosmos/providers/dbt/core/operators/virtualenv.py
+++ b/cosmos/providers/dbt/core/operators/virtualenv.py
@@ -3,9 +3,10 @@ from __future__ import annotations
 import logging
 from pathlib import Path
 from tempfile import TemporaryDirectory
+from typing import TYPE_CHECKING
+
 
 from airflow.compat.functools import cached_property
-from airflow.utils.context import Context
 from airflow.utils.python_virtualenv import prepare_virtualenv
 
 from cosmos.providers.dbt.core.operators.local import (
@@ -18,6 +19,9 @@ from cosmos.providers.dbt.core.operators.local import (
     DbtSnapshotLocalOperator,
     DbtTestLocalOperator,
 )
+
+if TYPE_CHECKING:
+    from airflow.utils.context import Context
 
 logger = logging.getLogger(__name__)
 
--- a/cosmos/providers/dbt/core/operators/virtualenv.py
+++ b/cosmos/providers/dbt/core/operators/virtualenv.py
@@ -43,8 +43,8 @@ class DbtVirtualenvBaseOperator(DbtLocalBaseOperator):
         py_system_site_packages: bool = False,
         **kwargs,
     ) -> None:
-        self.py_requirements = py_requirements
-        self.py_system_site_packages = py_system_site_packages or []
+        self.py_requirements = py_requirements or []
+        self.py_system_site_packages = py_system_site_packages
         super().__init__(**kwargs)
         self._venv_tmp_dir = ""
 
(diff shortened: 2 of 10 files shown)


cosmos/providers/dbt/core/operators/local.py
```diff
@@ -38,6 +38,10 @@ class DbtLocalBaseOperator(DbtBaseOperator):
 
     :param install_deps: If true, install dependencies before running the command
     :param callback: A callback function called on after a dbt run with a path to the dbt project directory.
+    :param venv_requirements: Creates a virtual environment with the specified arguments
+    :param venv_system_site_packages: Whether or not all the Python packages from the Airflow instance, will be
```
tatiana, line 42: Delete this for now

cosmos/providers/dbt/core/operators/virtualenv.py
```diff
@@ -0,0 +1,106 @@
...
+    DbtRunLocalOperator,
+    DbtRunOperationLocalOperator,
+    DbtSeedLocalOperator,
+    DbtSnapshotLocalOperator,
+    DbtTestLocalOperator,
+)
+
+logger = logging.getLogger(__name__)
+
+FullOutputSubprocessResult = namedtuple(
```
tatiana, line 19: This can be imported from `local`

jlaneve (commented):
Do you mind adding docs on this?

jlaneve (commented):
I think we should wait to merge this until the profile changes are merged. Once the profile changes are in, we can define the dbt package requirements in each of the profiles and a user won't have to define the python requirements by hand. We can just provide an interface for them to override if they need to use a specific version of dbt.

dev/dags/.airflowignore
```diff
@@ -0,0 +1,2 @@
+dbt_docs.py
+cosmos_seed_dag.py
```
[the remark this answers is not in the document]
tatiana: https://github.com/astronomer/astronomer-cosmos/issues/282

tatiana:
@jlaneve I disagree this PR should be dependent upon to the profiles PR, which didn't have a ticket associated with it and changes a huge part of the Cosmos project codebase.

This is an independent execution mode, just like Kubernetes and Docker, and it meets all the criteria described in the ticket.

pyproject.toml
```diff
@@ -36,8 +36,10 @@ classifiers = [
 ]
 dependencies = [
     "apache-airflow>=2.3.0",
+    "importlib-metadata; python_version < '3.8'",
     "Jinja2>=3.0.0",
     "typing-extensions; python_version < '3.8'",
+    "virtualenv",
```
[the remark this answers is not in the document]
tatiana: `import-metadata` is not a dev dependency:
https://github.com/astronomer/astronomer-cosmos/pull/279/files#diff-3ac91a8ab337ad9986f47dfe301b0dba83c5318e360cf1727c57aef7155422f5R76

I believe both `virtualenv` and `local` modes should be the default modes in the Cosmos project. I'm planning to contribute to Airflow so we replace `virtualenv` by the standard library `venv` and we'll be able to remove this dependency once this work is done.

pyproject.toml
```diff
@@ -154,7 +156,7 @@ matrix.airflow.dependencies = [
 freeze = "pip freeze"
 test = 'pytest  -vv . -m "not integration"'
 test-cov = 'pytest -vv --cov=cosmos --cov-report=term-missing --cov-report=xml . -m "not integration"'
-test-integration = "pytest -vv cosmos/tests/ --cov=cosmos --cov-report=term-missing --cov-report=xml -m integration"
+test-integration = "pytest -vv cosmos/tests/test_example_dags.py::test_example_dag[example_virtualenv] --cov=cosmos --cov-report=term-missing --cov-report=xml -m integration"
```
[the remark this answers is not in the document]
tatiana: Left over code, just deleted

dev/dags/.airflowignore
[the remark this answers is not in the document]
jlaneve: I don't think we need the import there - dbt already prints out the CLI version when you're running commands.

pyproject.toml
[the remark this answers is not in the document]
jlaneve: I don't think we need the `importlib-metadata` - dbt already prints out the CLI version whenever you run things.

Screen Shot 2023-05-31 at 1 11 55 PM

jlaneve (approved):

cosmos/providers/dbt/core/operators/local.py
```diff
@@ -156,7 +159,7 @@ def run_command(
 
             logger.info(f"Trying to run the command:\n {cmd}\nFrom {tmp_project_dir}")
```
josh-fell, line 160: WDYT about moving this logging line to the `run_subprocess` method so it's consistently logged?

cosmos/providers/dbt/core/operators/virtualenv.py
```diff
@@ -0,0 +1,146 @@
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+from airflow.compat.functools import cached_property
+from airflow.utils.context import Context
```
josh-fell, line 8: Might be nice to move this import under `TYPE_CHECKING`. One fewer import at runtime. WDYT?

cosmos/providers/dbt/core/operators/virtualenv.py
```diff
@@ -0,0 +1,146 @@
...
+class DbtVirtualenvBaseOperator(DbtLocalBaseOperator):
+    """
+    Executes a dbt core cli command within a Python Virtual Environment, that is created before running the DBT command
+    and deleted at the end of the operator execution.
+
+    :param py_requirements: If defined, creates a virtual environment with the specified dependencies. Example:
+           ["dbt-postgres==1.5.0"]
+    :param py_system_site_packages: Whether or not all the Python packages from the Airflow instance will be accessible
+           within the virtual environment (if py_requirements argument is specified).
+           Avoid using unless the DBT job requires it.
```
josh-fell, line 37:
```suggestion
    Executes a dbt core cli command within a Python Virtual Environment, that is created before running the dbt command
    and deleted at the end of the operator execution.

    :param py_requirements: If defined, creates a virtual environment with the specified dependencies. Example:
           ["dbt-postgres==1.5.0"]
    :param py_system_site_packages: Whether or not all the Python packages from the Airflow instance will be accessible
           within the virtual environment (if py_requirements argument is specified).
           Avoid using unless the dbt job requires it.
```
Nit for consistency of "dbt" in the docstring. There are a few other places "DBT" is used in the changes, but up to you to implement whichever formatting. I don't have strong opinions on the format -- I know dbt folks do 🙂 -- but consistency of choice is good I think.

cosmos/providers/dbt/core/operators/virtualenv.py
```diff
@@ -0,0 +1,146 @@
...
+                "-c",
+                "from importlib.metadata import version; print(version('dbt-core'))",
+            ]
+        )
+        dbt_version = cmd_output.output
+        self.log.info("Using DBT version %s available at %s", dbt_version, dbt_binary)
+        return str(dbt_binary)
+
+    def run_subprocess(self, command, *args, **kwargs):
+        if self.py_requirements:
```
josh-fell, line 84: Should there logging for the command being executed here like there is in the DbtLocalBaseOperator?

josh-fell (approved):
Looks great! Only some trivial suggestions.

cosmos/providers/dbt/core/operators/local.py
tatiana: That's a good suggestion, @josh-fell !

I started making this change, @josh-fell, but that would require extending the named arguments on both the `local` and `virtualenv` `BaseOperators` `run_subprocess` methods and adding logging logic to this method in both classes - so it felt we were adding more complexity than needed. Moving forward, as part of another PR, we can remove this log message and change the logging in the `subprocess_hook.run_command`.

cosmos/providers/dbt/core/operators/virtualenv.py
tatiana: Since the log is currently in `run_command` and the `VirtualenvOperator` is not overriding it, it is printed in both cases. I noticed some duplication in logging the command since the Subprocess hook also logs the command from `run_command` (but without `the `cwd`). I'll be improving this in a separate PR!

cosmos/providers/dbt/core/operators/virtualenv.py
tatiana: Thanks, @josh-fell , it was a copy & paste issue - just fixed in the `virtualenv.py` module. I'm adding a ticket to make this consistent across the project.

tatiana: This is the follow up ticket: #306

PR #280, opened 2023-05-19
Automate running example DAGs in the CI

tatiana:
Introduce a new hatch script, `test-integration`, which automates the execution of the example DAGs from the command line.

This change increases the project test coverage from 74.53% to 83.00%.

Example of how to use it, from `astronomer-cosmos`: 
```bash
export PYTHONPATH=`pwd`:$PYTHONPATH
export AIRFLOW_HOME=`pwd` 

hatch run airflow db init
hatch run tests.py3.10-2.5:test-integration
```

Until this point, developers were expected to manually run the DAGs through the Airflow UI in order to validate the existing example DAGs.

Fixing the previously broken example DAG  `dtb_docs.py` is outside this commit/PR. A follow-up ticket was logged to address this: #282.


cosmos/providers/dbt/core/operators/local.py
```diff
@@ -66,9 +66,9 @@ def exception_handling(self, result: FullOutputSubprocessResult):
             raise AirflowSkipException(
                 f"dbt command returned exit code {self.skip_exit_code}. Skipping."
             )
-        elif result.exit_code != 0:
             raise AirflowException(
-                f"dbt command failed. The command returned a non-zero exit code {result.exit_code}."
+                f"dbt command failed. The command returned a non-zero exit code {result.exit_code}. Details: "
```
joedborg, line 70: Does the "SkipException" do some magic?  I'm wondering how we can raise 2 exceptions in a row like this.

joedborg (approved):
One question added.  Apart from that, looks good to me!

cosmos/providers/dbt/core/operators/local.py
tatiana: Well spotted, @joedborg ! I deleted the `elif` line by mistake, and we had no tests covering the other branch.  Thanks a lot for catching this.

Issue #286, opened 2023-05-23
Extend DBT docs operator to upload S3 and Azure Blob Storage (static website)


szecsip:
Implement individual operators to upload docs to S3 or to Azure Blob Storage which base class would be DbtDocsLocalOperator. In this case the upload process would be much easier to only call DbtDocsS3Operator / DbtDocsAzureStorageOperator and the end users don't need to implement the uploading callback.

PR #287, opened 2023-05-23
Extend DBT docs operator to upload S3 and Azure Blob Storage (static website)

szecsip:
## Description

This PR extend DBT docs operators:
- adds the `DbtDocsS3LocalOperator` and `DbtDocsAzureStorageLocalOperator` to the base DbtDocsLocalOperator
operator. This lets a user generate docs and also take action with them. These operators handle the appropriate callback invocation automatically in the background. The user's task is simply to fill in the parameters required by the operators, and there is no need for them to write a callback for task execution.
- adds a modified docs page with an examples
- adds a modified DAG in the `dev` directory

## Related Issue(s)

Closes: #286

## Breaking Change?

No.

## Checklist


cosmos/providers/dbt/core/operators/local.py
```diff
@@ -365,6 +365,120 @@ def execute(self, context: Context):
...
+            logger.error("ERROR: the S3Hook isn't installed")
+        # if there's a botocore.exceptions.NoCredentialsError, print a warning and just copy the docs locally
+        except Exception as exc:
+            if "NoCredentialsError" in str(exc):
+                logger.error(
+                    "ERROR: No AWS credentials found.\
+                    To upload docs to S3, install the S3Hook and configure an S3 connection."
+                )
+            else:
+                logger.error("ERROR: " + str(exc))
```
jlaneve, line 419: We can probably get rid of this now, right? This was originally meant moreso for something that would always work locally

cosmos/providers/dbt/core/operators/local.py
```diff
@@ -365,6 +365,120 @@ def execute(self, context: Context):
...
+                    )
+        # if there's a botocore.exceptions.NoCredentialsError, print a warning and just copy the docs locally
+        except Exception as exc:
+            if "NoCredentialsError" in str(exc):
+                logger.error(
+                    "ERROR: No Azure credentials found.\
+                    To upload docs to Azure Blob Storage, install the WasbHook and configure a connection."
+                )
+            else:
+                logger.error("ERROR: " + str(exc))
```
jlaneve, line 477: same here

jlaneve:
This is looking great so far, thank you for opening @szecsip!

cosmos/providers/dbt/core/operators/local.py
szecsip: Thank you for your feedback, I have modified the exception handling.

cosmos/providers/dbt/core/operators/local.py
```diff
@@ -365,6 +365,118 @@ def execute(self, context: Context):
...
+        self.bucket_name = bucket_name
+        self.folder_dir = folder_dir
+        super().__init__(**kwargs)
+        self.callback = self.upload_to_s3
+        self.s3_hook = S3Hook(
+            aws_conn_id=self.target_conn_id, extra_args={"ContentType": "text/html"}
+        )
+        if self.s3_hook.test_connection()[0] is not True:
+            logger.error("ERROR: Failed to connect S3")
+            raise AirflowException("Failed to connect S3")
```
jlaneve, line 395: Do you know when this gets run? Is it parse time or task runtime? I don't know if we need it on `__init__`, I think we can just let it fail in the callback function itself

cosmos/providers/dbt/core/operators/local.py
```diff
@@ -365,6 +365,118 @@ def execute(self, context: Context):
...
+                        bucket_name=self.bucket_name,
+                        key=key_path,
+                        replace=True,
+                    )
+        except ImportError:
+            logger.error("ERROR: the S3Hook isn't installed")
+        except Exception as exc:
+            logger.error("ERROR: " + str(exc))
+
+        return
```
jlaneve, line 419: I think we can get rid of the try/except here since the import is now local, but I could be wrong. Do you mind testing that assumption? At the very least we should remove the last `except Exception as exc` and let the task just fail

jlaneve (dismissed):

jlaneve (approved):

PR #289, opened 2023-05-24
Improve execution mode documentation

tatiana:
The execution mode documentation was spread across several parts of our documentation, and the installation instructions were also mixed up with some how-to-run tutorials. This PR/commit aims to improve it.

This PR relies on the changes from the PR #279, and should be merged only after the other one is approved/merged.


jlaneve (changes requested):
I don't think the intent of these docs is correct - we designed Cosmos' local execution option to allow a user to install dbt into a virtualenv in a manner of their choosing. That's still going to be the recommended option: create a virtualenv either by modifying the base image (hopefully using the astro venv provider 🙂) or adding a startup script, and then providing the path to `dbt` to the Cosmos operators. 

Separately, a couple notes:
- I wouldn't say the virtualenv execution mode is "fast". Unless my understanding is incorrect, it re-creates the virtualenv each time it executes a model, which can be quite slow.
- The execution option pages feel more like tutorials than docs. Can we update them to be documentation focused explicitly on the execution option (and not stuff like setting up a database, etc)
- The new docs hierarchy feels incorrect - IMO it should go installation, usage, then other things. This PR inserts the tutorials above everything else.
- There are multiple issues and typos in the docs, do you mind taking another detailed pass to make sure they look correct? examples: `pip install astronomer-cosmos[kubernetes]` needs to be `pip install 'astronomer-cosmos[kubernetes]'`, in the execution modes page under virtualenv the code block doesn't show up, we have duplicated text in the virtualenv section in execution mode page, there are hanging `:` at the beginning of lots of links, etc.

tatiana:
Thanks for the review, @jlaneve , I'll address the comments!

tatiana:
@jlaneve When this PR was raised, I made assumptions since there was no other Cosmos maintainer to discuss overall intentions. As with any PR, it was up to review so we could align. The Cosmos installation documentation was confusing and I tried to improve - but if we think this PR does not bring improvements when compared to the previous documentation, I'm happy for us to close it - and for someone specilised in documentation to take care of this.

I've addressed the feedback as follows:

> I don't think the intent of these docs is correct - we designed Cosmos' local execution option to allow a user to install dbt into a virtualenv in a manner of their choosing. That's still going to be the recommended option: create a virtualenv either by modifying the base image (hopefully using the astro venv provider 🙂) or adding a startup script, and then providing the path to dbt to the Cosmos operators.

I removed references saying the virtualenv is the overall recommended approach to run Cosmos.

In the case of the [Airflow Dataflow Operator](https://airflow.apache.org/docs/apache-airflow-providers-google/stable/_api/airflow/providers/google/cloud/operators/dataflow/index.html#airflow.providers.google.cloud.operators.dataflow.DataflowCreatePythonJobOperator), that is the standard. In the case of Cosmos, this may not be ideal since we have multiple tasks to run a single DBT pipeline.

The advantage of having the virtualenv mode as default is that we'd have simple installation instructions to end-users, which would work in any Cloud provider or Airflow deployment.

That said, I've tried to rollback any changes on recommendations in this regard - and local remains the default mode.

> I wouldn't say the virtualenv execution mode is "fast". Unless my understanding is incorrect, it re-creates the virtualenv each time it executes a model, which can be quite slow.

It is slower than `local`, but potentially faster than `docker` and `kubernetes`, assuming the docker container is built during run-time. I've updated the docs to clarify this.

> The execution option pages feel more like tutorials than docs. Can we update them to be documentation focused explicitly on the execution option (and not stuff like setting up a database, etc)

It is a reference material with examples. There are no step-by-step in the main execution mode page. I'm happy for it to be reviewed, refactored and redone by native speakers and people with more experience in building documentation.

> The new docs hierarchy feels incorrect - IMO it should go installation, usage, and other things. This PR inserts the tutorials above everything else.

I cleaned up the Sphinx toctree, this is how it looks now:
Screenshot 2023-06-02 at 13 06 55

> There are multiple issues and typos in the doc. Doo you mind taking another detailed pass to make sure they look correct? 

I have used Grammarly to review the two pages which were mostly affected.

> examples: pip install astronomer-cosmos[kubernetes] needs to be pip install 'astronomer-cosmos[kubernetes]',

This is not accurate for bash and Linux - the single quotes are not necessary. But I've updated so MacOS users are happy to copy-and-paste.

> in the execution modes page under virtualenv the code block doesn't show up, we have duplicated text in the virtualenv section in execution mode page, there are hanging: at the beginning of lots of links, etc.

I tried to resolve all the issues reported, please, let me know if there is anything pending. There is still some duplication - but I think this can be reviewed by Astronomer's documentation team, together with the entire Cosmos documentation.

jlaneve (approved):

Issue #282, opened 2023-05-22
Fix the example DAGs dbt_docs.py and cosmos_seed_dag.py


tatiana:
The `dtb_docs.py` example DAG seems to be misbehaving. This issue was identified as part of  #280.

Detailed stack traces generated by this DAG can be found below. They were obtained by running this task with Tilt, plain Python, and in the CI (Github Actions).

**Acceptance criteria**
* [ ] Tilt can run the DAGs `dtb_docs.py` and `cosmos_seed_dag.py` successfully and there are no stack traces in the task logs
* [ ] Delete the line https://github.com/astronomer/astronomer-cosmos/blob/f447874affe03991a6008a7d4e79c2442e26f8ca/cosmos/tests/test_example_dags.py#L20
* [ ]  The command `hatch run tests.py3.8-2.5:test-integration` runs the `dtb_docs.py` and `cosmos_seed_dag.py` DAGs without issues
* [ ] The CI runs the `dtb_docs.py` and `cosmos_seed_dag.py` DAGs without issues, as part of the jobs `Run-Integration-Tests`

## Exception raised with Tilt

When using `tilt up`, the task `generate_dbt_docs` logs the following:
```
[2023-05-22, 10:44:53 UTC] {adapted_subprocesshook.py:68} INFO - Running command: ['dbt', 'docs', 'generate', '--profile', 'postgres_profile']
[2023-05-22, 10:44:53 UTC] {adapted_subprocesshook.py:79} INFO - Output:
[2023-05-22, 10:44:54 UTC] {adapted_subprocesshook.py:91} INFO - [0m10:44:54  Running with dbt=1.5.0
[2023-05-22, 10:44:54 UTC] {adapted_subprocesshook.py:91} INFO - [0m10:44:54  Encountered an error:
[2023-05-22, 10:44:54 UTC] {adapted_subprocesshook.py:91} INFO - Runtime Error
[2023-05-22, 10:44:54 UTC] {adapted_subprocesshook.py:91} INFO -   No dbt_project.yml found at expected path /tmp/tmpsaevo38a/dbt_project/dbt_project.yml
[2023-05-22, 10:44:54 UTC] {adapted_subprocesshook.py:91} INFO -   Verify that each entry within packages.yml (and their transitive dependencies) contains a file named dbt_project.yml
[2023-05-22, 10:44:54 UTC] {adapted_subprocesshook.py:91} INFO - 
[...]

PR #296, opened 2023-05-31
Fix docs dag

jlaneve:
## Description

This PR fixes the docs DAG to only run specific operators when the relevant connection is available.

## Related Issue(s)

Begins to address https://github.com/astronomer/astronomer-cosmos/issues/282

## Breaking Change?

No

## Checklist


harels (approved):
nice!

PR #298, opened 2023-05-31
fix select & exclude behavior to be and, not or

jlaneve:
## Description

This PR fixes an issue where the `select` logic would include any models that meet _any_ of the filters, which results in many more models selected than expected. Instead, this PR updates the logic to only include/exclude things where _all_ conditions are met.

Separately, we should revisit this in the future - the current behavior is pretty confusing, and I myself had to read through the source code to remember. Filed follow-up issue https://github.com/astronomer/astronomer-cosmos/issues/299.

## Breaking Change?

Yes, but this corrects previous incorrect behavior.

## Checklist

### DIFF
--- a/cosmos/providers/dbt/render.py
+++ b/cosmos/providers/dbt/render.py
@@ -111,7 +111,7 @@ def render_project(
 
         if "configs" in select:
             # TODO: coverme
-            if not set(select["configs"]).intersection(model.config.config_selectors):
+            if not set(select["configs"]).intersection(model.config.config_selectors) == set(select["configs"]):
                 continue
 
         if "configs" in exclude:
--- a/cosmos/providers/dbt/render.py
+++ b/cosmos/providers/dbt/render.py
@@ -116,7 +116,7 @@ def render_project(
 
         if "configs" in exclude:
             # TODO: coverme
-            if set(exclude["configs"]).intersection(model.config.config_selectors):
+            if set(exclude["configs"]).intersection(model.config.config_selectors) == set(exclude["configs"]):
                 continue
 
         run_args: Dict[str, Any] = {**task_args, **operator_args, "models": model_name}
--- a/dev/dags/cosmos_filtering_dag.py
+++ b/dev/dags/cosmos_filtering_dag.py
@@ -1,0 +1,26 @@
+"""
+An example DAG that uses Cosmos to render a dbt project. It selects a subset of models to run.
+"""
+import os
+from datetime import datetime
+from pathlib import Path
+
+from cosmos.providers.dbt.dag import DbtDag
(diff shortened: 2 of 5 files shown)


cosmos/providers/dbt/render.py
```diff
@@ -111,12 +111,12 @@ def render_project(
...
         if "configs" in select:
             # TODO: coverme
-            if not set(select["configs"]).intersection(model.config.config_selectors):
+            if not set(select["configs"]).intersection(model.config.config_selectors) == set(select["configs"]):
                 continue
 
         if "configs" in exclude:
             # TODO: coverme
-            if set(exclude["configs"]).intersection(model.config.config_selectors):
+            if set(exclude["configs"]).intersection(model.config.config_selectors) == set(exclude["configs"]):
```
harels, line 119: this should be OR

dev/dags/cosmos_filtering_dag.py
```diff
@@ -0,0 +1,26 @@
+"""
+An example DAG that uses Cosmos to render a dbt project. It selects a subset of models to run.
```
harels, line 2: update description here

docs/dbt/configuration.rst
```diff
@@ -89,7 +89,9 @@ When at least one WARN message is present, the function passed to ``on_warning_c
 Selecting and Excluding
 ----------------------
 
-Cosmos allows you to filter by configs (e.g. ``materialized``, ``tags``) using the ``select`` and ``exclude`` parameters. If a model contains any of the configs in the ``select``, it gets included as part of the DAG/Task Group. Similarly, if a model contains any of the configs in the ``exclude``, it gets excluded from the DAG/Task Group.
+Cosmos allows you to filter by configs (e.g. ``materialized``, ``tags``) using the ``select`` and ``exclude`` parameters.
+
+If a model contains ALL of the configs in the ``select``, it gets included as part of the DAG/Task Group. Similarly, if a model contains ALL of the configs in the ``exclude``, it gets excluded from the DAG/Task Group.
```
harels, line 94:
```suggestion
If a model contains ALL of the configs in the ``select``, it gets included as part of the DAG/Task Group. Similarly, if a model contains ANY of the configs in the ``exclude``, it gets excluded from the DAG/Task Group.
```

cosmos/providers/dbt/render.py
tatiana: It feels like a great opportunity to add tests to cover this part of the code, which handn't been covered yet!

jlaneve:
closing this for now as we should address https://github.com/astronomer/astronomer-cosmos/issues/308 first

PR #314, opened 2023-06-06
Profile name override

jlaneve:
## Description

This PR allows a user to specify a `profile_name_override`. When specified, the profile Cosmos generates will be called the value passed.

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/266

## Checklist


cosmos/providers/dbt/core/operators/tests/test_local.py
```diff
@@ -83,3 +83,36 @@ def test_dbt_base_operator_get_env(p_context_to_airflow_vars: MagicMock) -> None
...
+@patch("os.path.exists")
+def test_get_profile_name(mock_os_path_exists) -> None:
+    mock_os_path_exists.return_value = True
+
+    # check that a user-specified profile name is returned when specified
+    dbt_base_operator = DbtLocalBaseOperator(
+        conn_id="my_airflow_connection",
+        task_id="my-task",
+        project_dir="my/dir",
+        profile_name="default",
```
tatiana, line 97: It may be worth giving a different name since "default" is already the default behaviour if this parameter isn't given

cosmos/providers/dbt/core/operators/local.py
```diff
@@ -112,6 +116,38 @@ def store_compiled_sql(self, tmp_project_dir: str, context: Context, session: Se
...
+        # and need to raise an error
+        if not os.path.exists(dbt_project_path):
+            raise AirflowException(f"dbt project directory {self.project_dir} does not contain a dbt_project.yml file.")
+
+        # get file contents using path
+        dbt_project = yaml.safe_load(Path(dbt_project_path).read_text(encoding="utf-8")) or {}
+
+        profile_name = dbt_project.get("profile", "cosmos_profile")
+
+        if not isinstance(profile_name, str):
```
tatiana, line 142: In which scenarios wouldn't the `profile_name` be a string?

cosmos/providers/dbt/core/operators/local.py
```diff
@@ -112,6 +116,38 @@ def store_compiled_sql(self, tmp_project_dir: str, context: Context, session: Se
...
+        """
+        if self.profile_name:
+            return self.profile_name
+
+        # get the profile name from the dbt_project.yml file
+        dbt_project_path = os.path.join(project_dir, "dbt_project.yml")
+
+        # if there's no dbt_project.yml file, we're not in a dbt project
+        # and need to raise an error
+        if not os.path.exists(dbt_project_path):
```
tatiana, line 134: It would be great to see a test case covering this

tatiana (approved):
Thanks a lot, @jlaneve!

Issue #319, opened 2023-06-09
Duplicated line in `execution-modes.rst`


pgoslatara:
There is a duplicated line in `execution-modes.rst` [here](https://github.com/astronomer/astronomer-cosmos/blob/4e38f7b22b6f288857cfc2b00823fd963f30d287/docs/dbt/execution-modes.rst?plain=1#L78):

```
...

In this case, users are responsible for declaring which version of ``dbt`` they
want to use using the argument ``py_requirements``.

In this case, users are responsible for declaring which version of ``dbt`` they want to use by giving the argument ``py_requirements``. 

...
```

PR #320, opened 2023-06-09
Removing duplicated line from docs

pgoslatara:
## Description

Removing a duplicated line from `execution-modes.rst` [here](https://github.com/astronomer/astronomer-cosmos/blob/4e38f7b22b6f288857cfc2b00823fd963f30d287/docs/dbt/execution-modes.rst?plain=1#L78).

## Related Issue(s)

Addresses #319 .

## Breaking Change?

No.

## Checklist


harels (approved):
thanks for the contribution and for improving our docs!

PR #329, opened 2023-06-16
Add AWS glue profile mapping.

binhnq94:
## Description

Add AWS glue profile mapping.

## Breaking Change?

It support create `DbtDag` and `DbtTaskGroup` with AWS Connection.

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,22 @@
 Changelog
 =========
 
+1.0.5 (2023-08-09)
+------------------
+
+Enhancements
+* Improve logs to include astornomer-cosmos identifier by @tatiana in #450
+* Support OAuth authentication for Big Query by @MonideepDe in #431
+
+Bug fixes
+* Fix selector for config tags by @javihernovoa in #441
+* Fix BigQuery keyfile_dict mapping for connection created from webserver UI by @jbandoro in #449
+
+Others
+* [pre-commit.ci] pre-commit autoupdate by @pre-commit-ci in #446
+* Resolve MyPy errors when adding Airflow pre-commit dependency by @abhi12mohan in #434
+
+
 1.0.0 (2022-12-14)
 -------------------
 
--- a/cosmos/__init__.py
+++ b/cosmos/__init__.py
@@ -5,7 +5,7 @@ Astronomer Cosmos is a library for rendering dbt workflows in Airflow.
 
 Contains dags, task groups, and operators.
 """
-__version__ = "1.0.4"
+__version__ = "1.0.5"
 
 from cosmos.airflow.dag import DbtDag
 from cosmos.airflow.task_group import DbtTaskGroup


cosmos/providers/dbt/core/profiles/glue/aws_role_arn.py
```diff
@@ -0,0 +1,52 @@
...
+    is_community: bool = True
+
+    # Reference link:
+    required_fields = [
+        "role_arn",
+        "region",
+        "workers",
+        "worker_type",
+        "schema",
+        "session_provisioning_timeout_in_seconds",
```
tatiana, line 31: Can we define some default values instead of expecting the user to define all these arguments? This could apply to `session_provisioning_timeout_in_seconds` (e.g. 120), `workers` (e.g. 1) and `worker_type` (e.g. Standard).

jlaneve: I don't think we should provide defaults here - these are the required fields according to the [dbt-glue package](https://github.com/aws-samples/dbt-glue#example-config) (which doesn't provide default values), so for consistency's sake we should do the same.

cosmos/providers/dbt/core/profiles/glue/aws_role_arn.py
```diff
@@ -0,0 +1,52 @@
...
+class AWSRoleARNProfileMapping(BaseProfileMapping):
+    """
+    Maps Airflow Aws connections to Glue dbt profiles.
+
+    https://docs.getdbt.com/docs/core/connect-data-platform/glue-setup
+    https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/connections/aws.html
+    https://github.com/aws-samples/dbt-glue
+    """
+
+    airflow_connection_type: str = "aws"
```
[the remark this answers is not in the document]
jlaneve: I wonder if we need to let the user specify which dbt profile type they want - right now we assume the mapping based on whether or not the required fields are there, but that may not be sufficient if there are cases where there's on airflow connection -> many dbt profiles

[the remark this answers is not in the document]
tatiana: I agree, @jlaneve, AWS connection is just one of these scenarios.

I believe it is important addressing this concern as part of this PR, otherwise, in future, we'll likely have to introduce a breaking change to handle this scenario.

cosmos/providers/dbt/core/profiles/glue/aws_role_arn.py
tatiana: @jlaneve I think it depends if we want to be Airflow-oriented or DBT-oriented.

The DBT profile aims to centralise all the database connection information in a single place. Airflow connections have the same purpose. However, in most cases, the mandatory information in Airflow connections is a subset of the required information in DBT.

If we want to be consistent with DBT required parameters, this will mean that Airflow Cosmos users will have to define "database connection" properties in two places: in the Airflow connection and as an argument of the Cosmos operator, task group or DAG. It feels like we are defying the original purpose of this centralised "connection" representation.

To have default values for DBT profiles required values which are not required by Airflow connections would simplify the usage for those who do not want to customise these values - avoiding having to define mandatory properties in two different places.

@jlaneve, if you think this is good enough for 1.0, we can stick to this pattern of forcing users to define profile properties in multiple places, and we can review this in future  Cosmos versions.

cosmos/providers/dbt/render.py
```diff
@@ -247,7 +247,14 @@ def render_project(
                 execution_mode=execution_mode,
                 dbt_class="DbtTest",
             ),
-            arguments={**task_args, **operator_args},
```
binhnq94, line 250: I also fix argments for test after all.

cosmos/providers/dbt/core/profiles/glue/aws_secret_access_key.py
```diff
@@ -0,0 +1,74 @@
...
+    ]
+
+    @property
+    def env_vars(self) -> dict[str, str]:
+        "Returns a dictionary of environment variables that should be set based on self.secret_fields."
+        env_vars = {}
+
+        for field in self.secret_fields:
+            # I re-write env_var_name for BashOperator can understand.
+            env_var_name = f"{self.airflow_connection_type.upper()}_{field.upper()}"
```
binhnq94, line 54: @tatiana @jlaneve I just updated the code where we can pass 2 connection attributes `login` and `password` to `DbtLocalBaseOperator` object as 2 env vars `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` respoding?

cosmos/providers/dbt/parser/project.py
```diff
@@ -132,12 +132,12 @@ def __post_init__(self) -> None:
         for base_node in ast.find_all(jinja2.nodes.Call):
             if hasattr(base_node.node, "name"):
                 # check we have a ref - this indicates a dependency
-                if base_node.node.name == "ref" or base_node.node.name == "source":
```
binhnq94, line 135: @tatiana @jlaneve  Ignore source because we not load source node from `yml` file. We can add it again after load source `at _handle_config_file`.

cosmos/providers/dbt/render.py
```diff
@@ -196,79 +197,78 @@ def render_project(
             logger.error("Unknown dbt type.")
             continue
 
+        # Add model to list models that will render.
+        models_to_render[model_name] = model
+
         # if test_behavior isn't "after_each", we can just add the task to the
         # base group and do nothing else for now
-        if test_behavior != "after_each":
+        if test_behavior != "after_each" or model.type != DbtModelType.DBT_MODEL:
```
binhnq94, line 205: Type != DbtModelType.DBT_MODEL have no test.

cosmos/providers/dbt/render.py
```diff
@@ -196,79 +197,78 @@ def render_project(
...
+        entities[model_name] = model_group
+        base_group.add_entity(entity=model_group)
 
     # add dependencies now that we have all the entities
-    for model_name, model in itertools.chain(project.models.items(), project.snapshots.items(), project.seeds.items()):
+    for model_name, model in models_to_render.items():
+        # Only add dependencies if model in `entities`
         upstream_deps = model.config.upstream_models
         for upstream_model_name in upstream_deps:
-            try:
```
binhnq94, line 237: This change made KeyError raise because `model_name` not in `entities`. It occurs when we set select or exclude.

cosmos/providers/dbt/render.py
```diff
@@ -196,79 +197,78 @@ def render_project(
...
+                "on_warning_callback": on_warning_callback,
+            },
         )
         entities[test_task.id] = test_task
 
         # add it to the base group
         base_group.add_entity(test_task)
 
         # add it as an upstream to all the models that don't have downstream tasks
-        # since we don't have downstream info readily available, we have to iterate
```
binhnq94, line 258: This change fix the bug: no dependence to test_task when we set select or exclude.

cosmos/providers/dbt/core/profiles/glue/aws_role_arn.py
binhnq94: @tatiana Understood, so in next version, I think we need Comos profile. It initiate with connection id an custom values.
For each kind of data flatform we will have a class.

cosmos/providers/dbt/render.py
```diff
@@ -196,82 +197,79 @@ def render_project(
...
                 entities[model_name].add_upstream(dep_task)
-            except KeyError:
+            else:
                 logger.error(f"Dependency {upstream_model_name} not found for model {model}")
+
     if test_behavior == "after_all":
         # make a test task
         test_task = Task(
-            id=f"{dbt_project_name}_test",
+            id="dbt_test",
```
binhnq94, line 244: dbt_project_name_test only fit for CosmosDag. For CosmosTaskGroup, it should be another prefix.

cosmos/providers/dbt/parser/project.py
tatiana: @binhnq94 I think we actually need `source` for the open lineage support (8df9ee4) - I noticed another community member made a PR just with this change: #338. WDYT if we remove this change from this PR and have the discussion in that dedicated PR?

cosmos/providers/dbt/render.py
tatiana: There's lots of room to improve the coverage! Any contributions in this sense would be very welcome

cosmos/providers/dbt/parser/project.py
binhnq94: Yeah, remove help avoid the logs here: https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/render.py#L241

binhnq94: I will remove it.

harels:
@binhnq94 there's a bunch of conflicts since we made changes with the 1.0 release.
Would you prefer if we take over and open up a new PR for this?

binhnq94:
> @binhnq94 there's a bunch of conflicts since we made changes with the 1.0 release. Would you prefer if we take over and open up a new PR for this?

Ok.
Thank you!

PR #334, opened 2023-06-22
Improve running tests locally

tatiana:
Previously, developers new to Cosmos could struggle to run integration tests locally.

Summary of improvements:
- Unify where tests dependencies are declared (hatch)
- Unify command to run integration tests (hatch)
- Have a single way of defining connections for tests (environment variable)
- Clean up CI workflow to be compatible with the cleanup
- Improve documentation on how to run integration tests locally

How to test:

Run Postgres locally:
```
    docker run --name postgres -p 5432:5432 -p 5433:5433 -e POSTGRES_PASSWORD=postgres postgres
```

Run integration tests from the `astronomer-cosmos` directory:
```
    export AIRFLOW_HOME=`pwd`
    export AIRFLOW_CONN_AIRFLOW_DB=postgres://postgres:postgres@0.0.0.0:5432/postgres
    hatch run tests.py3.8-2.5:test-integration
```


jlaneve (commented):
curious why we don't use sqlite or duckdb here? something in-memory that doesn't require a container vs running Postgres in a container

tatiana:
@jlaneve We can change as part of another PR, ATM. This is just consistent with our initial example DAGs which were using Postgres!

pyproject.toml
```diff
@@ -156,7 +149,14 @@ matrix.airflow.dependencies = [
...
 test = 'pytest -vv --durations=0 . -m "not integration"'
 test-cov = 'pytest -vv --cov=cosmos --cov-report=term-missing --cov-report=xml --durations=0 -m "not integration"'
-test-integration = "pytest -vv --cov=cosmos --cov-report=term-missing --cov-report=xml --durations=0 -m integration"
+# we install using the following workaround to overcome installation conflicts, such as:
+# apache-airflow 2.3.0 and dbt-core [0.13.0 - 1.5.2] and jinja2>=3.0.0 because these package versions have conflicting dependencies
+test-integration = """\
+pip uninstall dbt-postgres; \
+rm -rf airflow.*; \
+airflow db init; \
+pip install dbt-postgres; \
```
jlaneve, line 158: does this mean every time we run the tests, we're installing pypi packages and re-initializing the airflow db? is there any better way to do this? just worried this makes the local iteration speed much slower because we're not really re-using an environment

tatiana: Yes, @jlaneve - and that's what we're doing in the CI now.

Given the conflict errors between `dbt` & `airflow,` I didn't manage to put the `dbt-postgres` dependency together with the other `pyproject.toml` dependencies.

I can happily remove these lines. Would you prefer that?
```
rm -rf airflow.*; \
airflow db init; \
```

The only consequence is that the developer doing the local setup will have to manage their Airflow Home and avoid conflicts if they are running multiple versions of Airflow.

jlaneve: ah interesting - maybe we break it out into two separate scripts? one for `init-db` and one for `run-integration-tests`

tatiana: Good idea, @jlaneve , I'll break this up early tomorrow BST!

tatiana: @jlaneve please, check if e4420083235f4c31cbcc6733a0ed45b3b6b84ab5 is okay!

jlaneve (approved):

PR #344, opened 2023-06-26
remove dependency on redis and celery

jlaneve:
## Description

This PR removes Redis, flower, and the local celery worker in favor of using Airflow's LocalExecuter from tilt for local dev.

## Related Issue(s)

n/a

## Breaking Change?

No

## Checklist


dev/docker-compose.yaml
```diff
@@ -10,10 +10,8 @@ x-airflow-common:
   environment:
     &airflow-common-env
     DB_BACKEND: postgres
-    AIRFLOW__CORE__EXECUTOR: CeleryExecutor
+    AIRFLOW__CORE__EXECUTOR: LocalExecutor
```
tatiana, line 13: I think time will tell us if `LocalExecutor` will limit the tests we run with `Tilt` or not - we can always revert this change if we face issues in the future.

tatiana (approved):

tatiana (commented):
@jlaneve I'm happy for us to simplify the tilt setup. If we notice jobs being stuck or not properly scheduled, we can revert to using the `CeleryExecutor` again.

dev/docker-compose.yaml
jlaneve: yeah that's what I'm thinking too - worth noting that I checked and the Astro CLI uses the `LocalExecutor`, so this is consistent with an end user's local dev experience

PR #349, opened 2023-06-27
ensure dbt_cmd_flags is used

jlaneve:
## Description

This PR ensures we're using the `dbt_cmd_flags` argument. It's documented, but not used in the code anywhere.

## Checklist


harels (approved):
good catch!

Issue #328, opened 2023-06-16
Make store_compiled_sql Optional 


npezman:
Using: astronomer-cosmos==0.7.1
Problem: Sometimes, if the number of models is big and we have too many lines of code on each one, we experience Airflow UI crashes after the task fails in these. 
I wanted to propose to make the rendering optional, keeping the templated field anyway.

jlaneve:
Yes, I like this idea! Maybe we should make it globally disabled with an Airflow config or environment variable? What do you think?

PR #351, opened 2023-06-28
Add option to disable storing the compiled SQL

jlaneve:
## Description

This PR adds a new config option, `should_store_compiled_sql`, that configures the behavior of the local operator. When set to True (default), there's no change in behavior - Cosmos will write dbt's compiled SQL to Airflow's templated field. When set to False, nothing happens.

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/328

## Breaking Change?

No breaking changes

## Checklist


harels (approved):
improved usability, nice! thanks @jlaneve

Issue #350, opened 2023-06-28
GCP BigQuery service account json not supported in v0.7


jbandoro:
We ran into this problem upgrading from v0.6.X -> v0.7.X using the default `google_cloud_platform` [airflow connection ](https://airflow.apache.org/docs/apache-airflow-providers-google/stable/connections/gcp.html#configuring-the-connection) that uses a keyfile json and not the keyfile path that is currently [supported](https://github.com/astronomer/astronomer-cosmos/blob/1ebee498b8ccc3a295b6c737a81d336ddcca6626/cosmos/profiles/bigquery/service_account_file.py#L9) in astronomer cosmos.

All of the profile var mapping of the json service account can be found in an older tag before the refactor of profiles in 0.7 here: https://github.com/astronomer/astronomer-cosmos/blob/astronomer-cosmos-v0.6.8/cosmos/providers/dbt/core/profiles/bigquery.py as a place to start from.

When I have time I'd be happy to take a stab at this, but opening up this issue in case others also run into this during an upgrade.

JoeSham:
Hi, I was just going to open a similar issue :) We were not using the older version, but I guess we have the same problem: we store our Service Account JSON file as a secret in Google Secret Manager, so when we retrieve it, it is not a path to a file, but the actual file/JSON. I started testing with Cosmos 0.7.3 and so far did not manage getting it to somehow digest the SA keyfile as a json...

JoeSham:
@tatiana Hi, I see you did the refactor probably, could you plz check this? Is there really no way to read SA keyfile_dict currently, or is there some other way? Thank you

tatiana:
@JoeSham @jbandoro I believe this may have been an issue introduced by the PR where we refactored profiles: https://github.com/astronomer/astronomer-cosmos/pull/271

To confirm if I understood the issue, Cosmos works as expected if you create a [Google Cloud Platform](https://airflow.apache.org/docs/apache-airflow-providers-google/stable/connections/gcp.html) connection specifying the `key_path` property but not if you specify `keyfile_dict` - is that the case?

JoeSham:
[...]

PR #352, opened 2023-06-28
Add support to GCP connections that define `keyfile_dict` instead of `keyfile`

JoeSham:
Add support to Google Cloud Platform connections that define `keyfile_dict` (actual value) instead of `keyfile` (path).

A design decision for this implementation was to not add `keyfile_json` to  `secret_fields`. This was not done because this property is originally a JSON. While storing it as an environment variable is simple, we'd need more significant changes to our profile parsing to correctly render this to the `profile.yml`  generated by Cosmos. 

This used to work in Cosmos 0.6.x and stopped working in 0.7.x as part of a previous profile refactors #271.

Closes: https://github.com/astronomer/astronomer-cosmos/issues/350

Co-authored-by:  Tatiana Al-Chueyr <tatiana.alchueyr@gmail.com>

**How to validate this change**

1. Have a GCP BQ service account with the `BigQuery Data Editor` role
2. Create a namespace that can be accessed by the service account (1)
3. Create an Airflow GCP connection that uses `keyfile` (path to the service account credentials saved locally). Example (replace `some-namespace` (2) and `key_path`(1)):
```
export AIRFLOW_CONN_GOOGLE_CLOUD_DEFAULT='{"conn_type": "google_cloud_platform", "extra": {"key_path": "/home/some-user/key.json", "scope": "https://www.googleapis.com/auth/cloud-platform", "project": "astronomer-dag-authoring", "dataset": "some-namespace" , "num_retries": 5}}'
```
4. Change the `basic_cosmos_dag.py` with the following lines, making sure it references the Airflow connection created in (3) and the dataset created in (2):
 ```
    profile_args={
        "dataset": "some-namespace",
    },
```

5. Run the DAG, for instance:
```
 PYTHONPATH=`pwd` AIRFLOW_HOME=`pwd` AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT=20000 AIRFLOW__CORE__DAG_FILE_PROCESSOR_TIMEOUT=20000 airflow dags test basic_cosmos_dag `date -Iseconds`
```

6.  Change the Airflow GCP connection to use `keyfile_dict` (hard-code the `keyfile` content in the Airflow connection, replacing `<your keyfile content here>`)
```
export AIRFLOW_CONN_GOOGLE_CLOUD_DEFAULT='{"conn_type": "google_cloud_platform", "extra": {"keyfile_dict": <your keyfile content here>, "scope": "https://www.googleapis.com/auth/cloud-platform", "project": "astronomer-dag-authoring", "dataset": "cosmos" , "num_retries": 5}}'
```

7. Run the previously created Cosmos-powered DAG (5) that confirms (6) works

### DIFF
--- a/cosmos/profiles/bigquery/service_account_keyfile_dict.py
+++ b/cosmos/profiles/bigquery/service_account_keyfile_dict.py
@@ -37,7 +37,6 @@ class GoogleCloudServiceAccountDictProfileMapping(BaseProfileMapping):
         Even though the Airflow connection contains hard-coded Service account credentials,
         we generate a temporary file and the DBT profile uses it.
         """
-        # keyfile_path = self.dump_credentials_to_disk()
         return {
             "type": "bigquery",
             "method": "service-account-json",
--- a/cosmos/profiles/bigquery/service_account_keyfile_dict.py
+++ b/cosmos/profiles/bigquery/service_account_keyfile_dict.py
@@ -22,6 +22,10 @@ class GoogleCloudServiceAccountDictProfileMapping(BaseProfileMapping):
         "keyfile_dict",
     ]
 
+    secret_fields = [
+        "keyfile_dict",
+    ]
+
     airflow_param_mapping = {
         "project": "extra.project",
         # multiple options for dataset because of older Airflow versions
@@ -39,6 +43,6 @@ class GoogleCloudServiceAccountDictProfileMapping(BaseProfileMapping):
             "project": self.project,
             "dataset": self.dataset,
             "threads": self.profile_args.get("threads") or 1,
-            "keyfile_json": self.keyfile_dict,
+            "keyfile_json": self.get_env_var_format("keyfile_dict"),
             **self.profile_args,
         }
--- a/cosmos/profiles/bigquery/service_account_keyfile_dict.py
+++ b/cosmos/profiles/bigquery/service_account_keyfile_dict.py
@@ -1,0 +1,43 @@
+"Maps Airflow GCP connections to dbt BigQuery profiles if they use a service account keyfile dict/json."
+from __future__ import annotations
+
+from typing import Any
+
+from cosmos.profiles.base import BaseProfileMapping
+
+
+class GoogleCloudServiceAccountDictProfileMapping(BaseProfileMapping):
+    """
+    Maps Airflow GCP connections to dbt BigQuery profiles if they use a service account keyfile dict/json.
+
+    https://docs.getdbt.com/reference/warehouse-setups/bigquery-setup#service-account-file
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by tatiana** [comment]
@JoeSham I wrote some tests for this feature in 615a256. Feel free to cherry-pick them!

File: cosmos/profiles/bigquery/service_account_keyfile_dict.py [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+    @property
+    def profile(self) -> dict[str, Any | None]:
+        """Generates profile. Defaults `threads` to 1."""
+        return {
+            "type": "bigquery",
+            "method": "service-account-json",
+            "project": self.project,
+            "dataset": self.dataset,
+            "threads": self.profile_args.get("threads") or 1,
+            "keyfile_json": self.keyfile_dict,
```
tatiana, line 41: @JoeSham WDYT of adding this to `secret_fields`, so we use the environment variables to represent it and not have it injected into the `profiles.yml` created by Cosmos? cc @jlaneve

tatiana: I wrote down a code to store the `keyfile_json` as an environment variable:
b0ab173594bb74ccc3cb148ea621e4aeaf645747

@JoeSham, if you confirm this works on your setup, we could cherry-pick it and merge it to the main branch sooner than later

tatiana (approved):
Thanks a lot, @JoeSham ! This looks great.
I triggered Cosmos Airflow DAGs with both ways of creating a BQ profile and confirmed it works as expected.

joppedh:
@JoeSham @tatiana  Out of curiosity. How does this work?
I added the connection and it get's rendered like expected:
```yaml
bigquery:
    outputs:
        dev:
            dataset: my_bigquery_dataset
            keyfile_json: '{   "type": "service_account",   "project_id": "dhh-ncr-stg",   "universe_domain":
                "googleapis.com" }'
            method: service-account-json
            threads: 1
            type: bigquery

```
But that is not supported by DBT. It does not compile. Where am I making a mistake?

joppedh:
@jlaneve Exactly. I think we still need to unpack the JSON and create yaml fields of it.

jlaneve:
> @jlaneve Exactly. I think we still need to unpack the JSON and create yaml fields of it.

I agree. Do you mind opening an issue for this? We're also very open to contributions in case you'd like to open a PR 🙂

joppedh:
Hey @tatiana I does work when I create connection from the command line like `airflow connection add...`
Where it does not work is if you add credentials manual from the webserver. This happens because of how Airflow stores it JSON. When we read it out it will be a string and not a dictionary.

Issue #342, opened 2023-06-26
tilt up not working for local development


CorsettiS:
When I tried to follow the guidelines of local development using tilt, I got the following error 

```
Loading Tiltfile at: ***/astronomer-cosmos/Tiltfile
Traceback (most recent call last):
  ***/astronomer-cosmos/Tiltfile:1:15: in <toplevel>
Error in docker_compose: command ["docker" "compose" "--project-name" "dev" "--project-directory" "***/astronomer-cosmos/dev" "-f" "***/astronomer-cosmos/dev/docker-compose.yaml" "config"] failed.
error: exit status 14
stdout: ""
stderr: 'open ***/astronomer-cosmos/dev/.env: no such file or directory
```

jlaneve:
Can you try creating an empty .env file and see if that fixes the issue?

CorsettiS:
@jlaneve it fixes it partially, however there are still few errors coming up
```

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
       → google-cloud-workflows 1.7.1 requires protobuf<4.0.0dev,>=3.19.0, but you have protobuf 4.23.3 which is incompatible.
       → google-cloud-vision 1.0.2 requires protobuf<4.0.0dev, but you have protobuf 4.23.3 which is incompatible.
       → google-cloud-videointelligence 1.16.3 requires protobuf<4.0.0dev, but you have protobuf 4.23.3 which is incompatible.
       → google-cloud-translate 1.7.2 requires protobuf<4.0.0dev, but you have protobuf 4.23.3 which is incompatible.
       → google-cloud-texttospeech 1.0.3 requires protobuf<4.0.0dev, but you have protobuf 4.23.3 which is incompatible.
       → google-cloud-tasks 2.10.1 requires protobuf<4.0.0dev,>=3.19.0, but you have protobuf 4.23.3 which is incompatible.
...
```

```
STEP 2/2 — Deploying
Container dev-airflow-init-1  Creating
Container dev-airflow-init-1  Created
no such service: redis

got unexpected error during build/deploy: command ["docker" "compose" "--project-name" "dev" "--project-directory" "***/astronomer-cosmos/dev" "-f" "***/astronomer-cosmos/dev/docker-compose.yaml" "up" "--no-deps" "-d" "airflow-init"] failed.
error: exit status 1

[...]

PR #354, opened 2023-06-30
add .env for a straightforward dev deployment

CorsettiS:
## Description

In the process of setting up the dev environment, an empty .env file is requested to build the image. Since there is no visible impact on having it not having it, for consistency it is better to remove this not used dependency from the docker compose yml file.

## Related Issue(s)

somehow related to #342 

## Breaking Change?

No major changes are going to happen

## Checklist


dev/docker-compose.yaml
```diff
@@ -6,7 +6,6 @@ x-airflow-common:
   build:
     context: ..
     dockerfile: dev/Dockerfile
-  env_file: .env
```
tatiana, line 9: @CorsettiS ATM, the `.env` file exposes sensitive information to the container, such as passwords and API keys. That's why it is not committed in the main repo, but we expect users to be able to use it.

CorsettiS: but for local development, if not needed, is there any advantage in keeping it in the docker compose yaml file?

tatiana: What would be the alternative for the developers who want to use the `.env`? Would they need to change the `docker-compose.yaml` file?

CorsettiS: I would advocate for that, since from my perspective the ideal scenario for local development is the one that you face zero problems to set it up.   The idea of the `docker-compose.yaml` file is to simply create a test dev for contributors, and If someone wants to use the `.env`, it makes sense for me that the person should set it up starting from the minimum provided setup.
This is just my opinion as a dev tho.

tatiana: I agree with zero problems during setup.

I wonder if creating an empty `.env` would be better than completely removing it. There may also be other alternatives - do you have any other suggestions?

As a dev, considering you have sensitive connection information (e.g., DB password), how do you currently make it available in `docker-compose-yml`?

I've seen several developers using the `.env` approach. If we removed this support, several developers would have problems with the local setup, which this PR aims to avoid.

CorsettiS: Creating an empty `.env` sounds like a reasonable approach for me, although it would be needed to remove it from the .gitignore file.

If I have any sensitive info I have that I want to test out in airflow (such as DB passwords) I usually pass it directly to the airflow UI, which I thought was the default way that devs were doing it.

tatiana: @CorsettiS We find the `.env` file particularly useful when testing against multiple versions of Python and Airflow - it makes it much quicker to set sensitive information without worrying about the UI.

A proposal:
- have an empty `.env` in the repo (we'd remove it from `.gitignore`)
- have a pre-commit hook to block users from committing a non-empty `.env`

CorsettiS: @tatiana  done. I have tested locally and it works fine.

harels (approved):
thanks for addressing the feedback, @CorsettiS!

PR #355, opened 2023-06-30
Replace tilt for docker compose in the contributing docs

CorsettiS:
## Description

From my personal experience using tilt, it did not work locally, while the classic docker compose command worked fine. I would suggest to change it to docker compose to simplify development, although it is understandable if the project wants to stick with tilt since it should, in theory simplify some of the overhead. 

## Related Issue(s)

this is a fix for a issue I was facing #342 

## Breaking Change?

No

## Checklist

### DIFF
--- a/dev/docker-compose.yaml
+++ b/dev/docker-compose.yaml
@@ -6,7 +6,6 @@ x-airflow-common:
   build:
     context: ..
     dockerfile: dev/Dockerfile
-  env_file: .env
   environment:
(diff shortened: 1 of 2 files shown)


jlaneve (approved):
Nice, thank you! The big thing tilt gives us is the ability to hot rebuild images when you change requirements, etc. But I think the trade off in complexity isn’t worth it, so thank you for putting this PR together.

PR #360, opened 2023-07-07
Refactor dbt project parsing and Airflow DAG generation

tatiana:
This is a major refactor related to how we render `dbt` projects in Airflow.

Some of the changes:
* Create a common class to define the share interface between `DbtDag` and `DbtTaskGroup`
* Define data classes to exchange information between the different steps of the Cosmos (parse dbt project, validate arguments, filter, create Airflow DAG)
* Split the logic of parsing `dbt` projects from the actual Airflow DAG generation
* Add support to parse dbt projects using `dbt ls`
* Add support to parse dbt projects using their `manifest.json`

Breaking changes:
* `from cosmos.dag import DbtDag` is now `from cosmos.airflow.dag import DbtDag` or `from cosmos import DbtDag`
* `from cosmos.dag import DbtTaskGroup` is now `from cosmos.airflow.dag import DbtTaskGroup` or `from cosmos import DbtTaskGroup`
* `select` and `exclude` arguments, passed to `DbtDag` or `DbtTaskGroup` used to be dictionaries, now they are lists - so we are compatible with the representation in dbt

**DISCLAIMER**: This PR is in draft mode because I'm improving the following:

(features)
* [x] expose automatic load mode and manifest in the `DbtToAirflowConverter` class
* [x] decide if we want to use Enum or Literals on the interface with end-users
* [x] the `dbt ls` expects the user to have a valid `profiles.yml` - we can extract the logic from the local executor. It feels we could do this in a follow-up PR

(test features we lack integration tests)
* [ ] validate with K8s operators
* [ ] validate with docker operators

(good practices)
* [x] test coverage
* [x] docstrings
* [x] type hints
* [x] improve times for running tests (because automatic is the default load mode, it seems the parsing time increased - both for unit and integration)


cosmos/airflow/group.py
```diff
@@ -0,0 +1,167 @@
...
+                f"{set(select['paths']).intersection(exclude['paths'])}"
+            )
+
+    # if task_args has a schema, add it to the profile args and add a deprecated warning
+    if "schema" in task_args:
+        profile_args["schema"] = task_args["schema"]
+        logger.warning("Specifying a schema in the task_args is deprecated. Please use the profile_args instead.")
+
+
+class AirflowGroup:
```
tatiana, line 74: Find a better name - perhaps `AirflowDbtGroup`? Feedback from @harels

cosmos/dbt/graph.py
```diff
@@ -0,0 +1,130 @@
...
+        self.dbt_cmd = dbt_cmd
+
+    # TODO: implement smart load, which tries other load modes if the desired fails
+    def load(self, method=LoadMode.CUSTOM):
+        load_method = {
+            LoadMode.CUSTOM: self.load_via_custom_parser,
+            LoadMode.DBT_LS: self.load_via_dbt_ls,
+            LoadMode.DBT_MANIFEST: self.load_from_dbt_manifest,
+        }
+        load_method[LoadMode.CUSTOM]()
```
tatiana, line 63: Change from `LoadMode.CUSTOM` to `method`

cosmos/dbt/graph.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def load_via_dbt_ls(self):
+        command = [self.dbt_cmd, "ls", "--output", "json", "--profiles-dir", self.project.dir]
+        if self.exclude:
+            command.extend(["--exclude", ",".join(self.exclude)])
+        if self.select:
+            command.extend(["--select", ",".join(self.select)])
+
+        process = Popen(command, stdout=PIPE, stderr=PIPE, cwd=self.project.dir)
+        stdout, stderr = process.communicate()
```
tatiana, line 73: Handle errors more gracefully, logging if there are any errors

cosmos/dbt/graph.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class DbtGraph:
+    """
+    Support loading a dbt project graph (represented by nodes) using different strategies.
+    """
+
+    nodes: list[DbtNode] = []
+    filtered_nodes: list[DbtNode] = []
+
```
tatiana, line 47: These are actually dictionaries

cosmos/airflow/group.py
tatiana: Renamed to `DbtToAirflowConverter` and placed in `cosmos/converter.py`

jlaneve (approved):
approving conditional on unit tests passing

PR #367, opened 2023-07-14
Fix BQ connection parsing with dataset info

### LINKED ISSUE #365: Cosmos does not parse BigQuery connections correctly
Hi there!

I might be mistaken, but it looks like there is an issue with cosmos version `0.7.4` around parsing of BiqQuery connection information.

Looking at the code [here](https://github.com/astronomer/astronomer-cosmos/blob/c919e2043b07d7e06e8a8663ffcbf84f18a34b2e/cosmos/profiles/bigquery/service_account_file.py#L25), the mapping of parameters from airflow connection to seems to expect `dataset` as a top level parameter in connection json.

```
airflow_param_mapping = {
        "project": "extra.project",
        "dataset": "dataset",
        "keyfile": "extra.key_path",
    }
```

However, based on the documentation of `apache-airflow-providers-google` [here](https://airflow.apache.org/docs/apache-airflow-providers-google/stable/connections/gcp.html#configuring-the-connection), `dataset` can not appear  as a top level parameter in connection json.

To verify this further, if I try to set dataset as a top level parameter in connection json, I get error as shown below:

```bash
$ airflow connections add 'my_cosmos_connection' --conn-json '{
	"conn_type": "google-cloud-platform",
	"dataset": "my_dataset",
	"extra": {
		"project": "my_gcp_project", 
		"num_retries": "5",
		"key_path": "/path/to/key.json"
	}
}'
...
...
...
TypeError: __init__() got an unexpected keyword argument 'dataset'
```

Hence, if I am not missing something, I think the mapping of dataset parameter should be with `extra.dataset` and this should be modified at the line which I linked above. I could make this work, after making the suggested change locally.

If this sounds good, I can try and submit a PR.

MonideepDe:
## Description

Cosmos version `0.7.4` currently does not correctly parse BigQuery connection information with dataset information.
Cosmos expects dataset as a top level parameter in connection json.

However, based on the documentation of apache-airflow-providers-google [here](https://airflow.apache.org/docs/apache-airflow-providers-google/stable/connections/gcp.html#configuring-the-connection), dataset can not appear as a top level parameter in connection json.

This PR fixes this by updating the mapping in cosmos, so that it expects dataset parameter in `extra` field of connection.

## Related Issue(s)
closes #365 

## Breaking Change?

None that I am aware of

## Checklist

### DIFF
--- a/cosmos/profiles/bigquery/service_account_file.py
+++ b/cosmos/profiles/bigquery/service_account_file.py
@@ -24,7 +24,7 @@ class GoogleCloudServiceAccountFileProfileMapping(BaseProfileMapping):
 
     airflow_param_mapping = {
         "project": "extra.project",
-        "dataset": "dataset",
+        "dataset": "extra.dataset",
         "keyfile": "extra.key_path",
     }
 
--- a/tests/profiles/bigquery/test_bq_service_account_file.py
+++ b/tests/profiles/bigquery/test_bq_service_account_file.py
@@ -70,6 +70,16 @@ def test_connection_claiming() -> None:
     profile_mapping = GoogleCloudServiceAccountFileProfileMapping(conn, {})
     assert not profile_mapping.can_claim_connection()
 
+    # if we have dataset specified in extra, it should claim
+    dataset_dict = {"dataset": "my_dataset"}
+    conn = Connection(
+        conn_id="my_bigquery_connection",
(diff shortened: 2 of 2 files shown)


jlaneve (approved):

jlaneve:
@MonideepDe thank you for the contribution!

Issue #316, opened 2023-06-06
Update CONTRIBUTING.md to document creating a fork


jlaneve:
The contributing docs only work if you have write access to the repo (which is only applicable to the maintainers). We should update the docs to reflect the fact that you first have to fork the repo.

Issue #307, opened 2023-06-02
Address warnings when building docs


tatiana:
When running:
```
hatch -e docs run build
```

We see 207 warnings during the documentation generation:

```
generating indices... genindex py-modindex done
writing additional pages... search done
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded, 207 warnings.
```

The goal with this ticket is to reduce to zero warnings, if possible.

PR #374, opened 2023-07-18
Update docs structure and add getting started guides

jlaneve:
## Description

This PR updates the docs site with a better structure. It also adds getting started guides for each of the popular ways of running Airflow.

## Related Issue(s)

Closes: #219
Closes: #316 
Closes: #307 

## Checklist


docs/conf.py
```diff
@@ -20,11 +20,9 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
 
 extensions = [
-    "autoapi.extension",
+    # "autoapi.extension",
```
tatiana, line 23: Curious, do you know why this was commented before?

docs/configuration/parsing-methods.rst
```diff
@@ -0,0 +1,57 @@
+.. _parsing-methods:
+
+Parsing Methods
```
tatiana, line 3: @jlaneve would it make sense for us to rename from `load_mode` to `parse_method` and `LoadMode` to `ParseMethod`, so we are consistent with the terminology used in the docs and code?

docs/configuration/parsing-methods.rst
```diff
@@ -0,0 +1,57 @@
...
+You can supply a ``manifest_path`` parameter on the DbtDag / DbtTaskGroup with a path to a ``manifest.json`` file. For example:
+
+.. code-block:: python
+
+    DbtDag(
+        manifest_path="/path/to/manifest.json"
+        ...,
+    )
+
+``dbt_ls``
```
tatiana, line 37: I believe the key advantage of using this parsing mode is the fact that it filters using `dbt` itself, as opposed to our custom filtering logic. I believe it would be worth to explain the filtering strategy used for each parsing method, WDYT?

docs/configuration/parsing-methods.rst
```diff
@@ -0,0 +1,57 @@
...
+----------
+
+.. note::
+
+    This only works for the ``local`` and ``virtualenv`` execution modes.
+
+If you don't have a ``manifest.json`` file, Cosmos will attempt to generate one from your dbt project. It does this by running ``dbt ls`` and parsing the output.
+
+When Cosmos runs ``dbt ls``, it also passes your ``select`` and ``exclude`` arguments to the command. This means that Cosmos will only generate a manifest for the models you want to run.
```
tatiana, line 46: The manifest contains all nodes, it disregards anything withing the ``select`` and ``exclude`` clauses.

The difference here is that we are using the `stdout` of the `dbt ls` command, which applies those filtering clauses.

docs/configuration/parsing-methods.rst
```diff
@@ -0,0 +1,57 @@
+.. _parsing-methods:
+
+Parsing Methods
+===============
+
```
tatiana, line 5: Should we give a high-level example of how to import LoadMethod and how to use it within a `DbtDag` or `DbtTaskGroup`?

docs/configuration/selecting-excluding.rst
```diff
@@ -0,0 +1,43 @@
+.. _selecting-excluding:
+
+Selecting & Excluding
+=======================
+
+Cosmos allows you to filter by configs (e.g. ``materialized``, ``tags``) using the ``select`` and ``exclude`` parameters. If a model contains any of the configs in the ``select``, it gets included as part of the DAG/Task Group. Similarly, if a model contains any of the configs in the ``exclude``, it gets excluded from the DAG/Task Group.
```
tatiana, line 6: When using `load_mod=LoadMode.DBT_LS,` we support any `dbt` type of filtering.

docs/configuration/selecting-excluding.rst
```diff
@@ -0,0 +1,43 @@
...
+
+The ``select`` and ``exclude`` parameters are dictionaries with the following keys:
```
tatiana, line 8: Since #360, they are no longer dictionaries - they are lists.
An example of the syntax:
```
select=["config.materialized:table"]
```

More examples can be found at:
https://github.com/astronomer/astronomer-cosmos/blob/main/tests/dbt/test_selector.py

docs/configuration/selecting-excluding.rst
```diff
@@ -0,0 +1,43 @@
...
+
+- ``configs``: a list of configs to filter by. The configs are in the format ``key:value``. For example, ``tags:daily`` or ``materialized:table``.
```
tatiana, line 10: Need to be updated to the new format. Examples:
```
select=["config.materialized:table"]
exclude=["tag:has_child"]
```

docs/configuration/selecting-excluding.rst
```diff
@@ -0,0 +1,43 @@
...
+- ``paths``: a list of paths to filter by. The paths are in the format ``path/to/dir``. For example, ``analytics`` or ``analytics/tables``.
```
tatiana, line 11: Example of path-based selector:
```
select=["path:gen2/models"]
```

docs/configuration/selecting-excluding.rst
```diff
@@ -0,0 +1,43 @@
...
+
+.. note::
+    Cosmos currently reads from (1) config calls in the model code and (2) .yml files in the models directory for tags. It does not read from the dbt_project.yml file.
```
tatiana, line 14: > It does not read from the dbt_project.yml file.
Doesn't it take into account the `dbt_project.yml` for `LoadMode.DBT_MANIFEST` and `LoadMode.DBT_LS`?

docs/configuration/selecting-excluding.rst
```diff
@@ -0,0 +1,43 @@
...
+
+Examples:
+
+.. code-block:: python
+
+    from cosmos import DbtDag
+
+    jaffle_shop = DbtDag(
+        # ...
+        select={"configs": ["tags:daily"]},
```
tatiana, line 24:
```suggestion
        select=["config.tags:daily"]
```

docs/configuration/selecting-excluding.rst
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+.. code-block:: python
+
+    from cosmos import DbtDag
+
+    jaffle_shop = DbtDag(
+        # ...
+        select={"configs": ["schema:prod"]},
```
tatiana, line 33:
```suggestion
        select=["config.schema:prod"]
```

docs/configuration/selecting-excluding.rst
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+.. code-block:: python
+
+    from cosmos import DbtDag
+
+    jaffle_shop = DbtDag(
+        # ...
+        select={"paths": ["analytics/tables"]},
```
tatiana, line 42:
```suggestion
        select=["path:analytics/tables"]
```

tatiana (commented):
This is looking great, @jlaneve ! It will be super helpful for Cosmos users. I'm adding a few comments in-line.

docs/configuration/testing-behavior.rst
```diff
@@ -1,10 +1,10 @@
...
-Cosmos offers a few different configuration options for how your dbt project is run and structured. This page describes the available options and how to configure them.
+Testing Behavior
+================
 
-Testing
-----------------------
+Testing Configuration
+---------------------
 
 By default, Cosmos will add a test after each model. This can be overridden using the ``test_behavior`` field. The options are:
```
tatiana, line 9: It is probably worth stating that the default behavior of Cosmos is different from the default behaviour from dbt (running each test individually) and explain why.

docs/getting_started/gcc.rst
```diff
@@ -0,0 +1,66 @@
+.. _gcc:
+
+Getting Started on GCC
```
tatiana, line 3: It's probably worth say Google Cloud Composer somwhere

docs/conf.py
jlaneve: I think it may have been my fault - was playing around with different options for reference doc generation. We definitely don't need it.

docs/configuration/parsing-methods.rst
jlaneve: we talked about this - going to leave this as-is in the code and just document it as parsing methods

harels (approved):

Issue #182, opened 2023-03-09
Support Python models in Cosmos


aiyaswamy-tgam:
Background:
DBT support Python models as well, we would like to make sure Cosmos supports this functionality in DBT.
DbtTaskGroup supports only the SQL files in the render class, Does not support python dbt models

Acceptance Criteria:
We can run Python models in Cosmos (in all execution modes?)

aiyaswamy-tgam:
> @aiyaswamy-tgam by Python dbt models, you mean something like [this](https://docs.getdbt.com/docs/build/python-models), correct?

Yes, that's right!

PR #375, opened 2023-07-19
Add support to run dbt Python models

tatiana:
**Disclaimer**: I'm fixing running the integration tests in the CI, but would love feedback. This PR won't be merged until all checks pass.

Add support to run dbt Python models, as described in the official documentation: https://docs.getdbt.com/docs/build/python-models

Add an example of Cosmos running Python models on Databricks (Postgres is not supported as of dbt 1.6).

The dbt example can be run by itself from the directory `dev/dags/dbt/jaffle_shop_python`  by exporting the following environment variables:
* DATABRICKS_HOST: Databricks host, similar to: dbc-some-id.cloud.databricks.com
* DATABRICKS_WAREHOUSE_ID: Databricks SQL Warehouse ID from connection HTTP path (example: `ca312a2206dfb361`)
* DATABRICKS_TOKEN: User Databricks access token
* DATABRICKS_CLUSTER_ID: Databricks cluster ID (example: 0201-094352-wab833sb)

And running:
```
dbt build
```

To validate the feature from a Cosmos perspective, set up the `databrics_default` connection. One way of accomplishing this is by using environment variables:
```
export AIRFLOW_CONN_DATABRICKS_DEFAULT=databricks://@dbc-<account-id>.cloud.databricks.com?token=<access-token>&http_path=/sql/1.0/warehouses/<warehouse-id>
```

From a previously set-up Airflow environment, run the `example_cosmos_python_models` DAG. An example of how to execute it from the command line:
```
airflow dags test example_cosmos_python_models `date -Iseconds`
```

This feature was validated with `load_mode=LoadMode.DBT_LS` and `LoadMode.CUSTOM`. Example of the rendered DAG:
Screenshot 2023-07-19 at 23 46 14

Review can be simplified by checking the commits individually, especially between 796d6e888e9c0f293eb3120d7af1e7d4f6872008 and 78a7d31c1ae0e898f32d422d259ebf583386d8e1.

The downside with this change is that the integration tests are now slower to run and depend on the following environment variables being set up in the CI:
* `AIRFLOW_CONN_DATABRICKS_DEFAULT`
* `DATABRICKS_CLUSTER_ID`

Closes: #182


tatiana:
The only test failing is test coverage related to a part of the code which was already in the code base and wasn't previously tested. Since this PR is already quite extensive, I suggest we disregard this check, exceptionally.

cosmos/dbt/graph.py
```diff
@@ -136,7 +137,9 @@ def load_via_dbt_ls(self):
             command.extend(["--select", *self.select])
         logger.info(f"Running command: {command}")
         try:
-            process = Popen(command, stdout=PIPE, stderr=PIPE, cwd=self.project.dir, universal_newlines=True)
+            process = Popen(
+                command, stdout=PIPE, stderr=PIPE, cwd=self.project.dir, universal_newlines=True, env=os.environ
+            )
```
jlaneve, line 142: what's the reason for this change?

cosmos/dbt/parser/project.py
```diff
@@ -87,6 +94,30 @@ def _config_selector_ooo(
...
+
+    upstream_entities = []
+    model_function = ""
+    for node in source_code.body:
+        if isinstance(node, ast.FunctionDef) and node.name == DBT_PY_MODEL_METHOD_NAME:
+            model_function = node
+            break
+
+    for item in ast.walk(model_function):
+        if isinstance(item, ast.Call) and item.func.attr == DBT_PY_DEP_METHOD_NAME:
```
jlaneve, line 113: how about sources? https://docs.getdbt.com/docs/build/python-models#referencing-other-models

cosmos/dbt/parser/project.py
```diff
@@ -216,6 +249,10 @@ def __post_init__(self) -> None:
         for file_name in self.models_dir.rglob("*.sql"):
             self._handle_sql_file(file_name)
 
+        # crawl the models in the project
+        for file_name in self.models_dir.rglob("*.py"):
+            self._handle_sql_file(file_name)
```
jlaneve, line 254: wonder if this should be a separate `_handle_python_file`?

cosmos/dbt/graph.py
tatiana: So the user `models/schema.yml` can have references to environment variables using dbt's `env_var`.

cosmos/dbt/parser/project.py
tatiana: Ah, I missed this! Will add support now.

I think there may be a few circumstances in which we may miss references to upstream Python models, e.g. if they are in it is in an f-string. But we can iterate over this feature over time based on user's feedback!

cosmos/dbt/parser/project.py
tatiana: @jlaneve we could. I was apprehensive about the low code coverage of this part of the code and our short timelines. Would it be okay if I log a ticket and we refactor this before 1.1? This would give us more time to improve the test coverage of this part of the code.

cosmos/dbt/parser/project.py
jlaneve: Ahh right we've had conversations about this before, thanks for reminding me. This is good as-is.

cosmos/dbt/parser/project.py
jlaneve: yes, that's good with me!

tatiana: Thanks, @jlaneve , I just added the follow up ticket: #384 384

jlaneve (approved):

PR #378, opened 2023-07-20
Snowflake private key

### LINKED ISSUE #267: Key-based auth for Snowflake

patawan:
## Description

Added new profile mapping for Snowflake User/Private Key authentication. Currently only supports `private_key_content` from a Snowflake Airflow connection.

## Related Issue(s)

Closes #267 

## Breaking Change?
N/A

## Checklist

### DIFF
--- a/cosmos/profiles/snowflake/user_privatekey.py
+++ b/cosmos/profiles/snowflake/user_privatekey.py
@@ -1,0 +1,86 @@
+"Maps Airflow Snowflake connections to dbt profiles if they use a user/private key."
+from __future__ import annotations
+
+import json
+from typing import TYPE_CHECKING, Any
+
+from ..base import BaseProfileMapping
+
+if TYPE_CHECKING:
+    from airflow.models import Connection
+
+
+class SnowflakePrivateKeyPemProfileMapping(BaseProfileMapping):
+    """
+    Maps Airflow Snowflake connections to dbt profiles if they use a user/private key.
+    https://docs.getdbt.com/docs/core/connect-data-platform/snowflake-setup#key-pair-authentication
+    https://airflow.apache.org/docs/apache-airflow-providers-snowflake/stable/connections/snowflake.html
+    """
+
+    airflow_connection_type: str = "snowflake"
+
+    required_fields = [
+        "account",
+        "user",
+        "database",
+        "warehouse",
+        "schema",
+        "private_key_content",
+    ]
+    secret_fields = [
+        "private_key_content",
+    ]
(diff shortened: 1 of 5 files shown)


patawan:
Hmmm trying to figure out how/why codecov dropped so much

cosmos/profiles/snowflake/user_privatekey.py
```diff
@@ -0,0 +1,86 @@
...
+    from airflow.models import Connection
+
+
+class SnowflakePrivateKeyPemProfileMapping(BaseProfileMapping):
+    """
+    Maps Airflow Snowflake connections to dbt profiles if they use a user/private key.
+    https://docs.getdbt.com/docs/core/connect-data-platform/snowflake-setup#key-pair-authentication
+    https://airflow.apache.org/docs/apache-airflow-providers-snowflake/stable/connections/snowflake.html
+    """
+
```
tatiana, line 19: @patawan pleas,e could you add:
```
is_community: bool = True
```

tatiana (commented):
It looks great! Thanks a lot, @patawan . One minor suggestion and we can merge.

cosmos/profiles/snowflake/user_privatekey.py
patawan: Thanks @tatiana! I've got that flag added

tatiana (approved):

cosmos/profiles/snowflake/user_privatekey.py
tatiana: Thank you very much, @patawan !

PR #379, opened 2023-07-21
Change `test_behavior` and `execution_mode` to use `Enum` as opposed to Literals

### LINKED ISSUE #371: convert literals to enums
Right now, we use literal types to designate which values are allowed for certain fields. Where we use literals, we should use enums instead as they're safer, easier to understand, and easier to use.

tatiana:
Change `DbtDag` and `DbtTaskGroup` so they use `Enum` as opposed to `Literal` strings to represent the arguments:
* `test_behavior`
* `execution_mode`

This is a breaking change. Before:
```
DbtDag(..., execution_mode="kubernetes", test_behavior="after_all")
```

Now:
```
from cosmos import ExecutionMode, TestBehavior

DbtDag(..., execution_mode=ExecutionMode.kubernetes, test_behavior=TestBehavior.AFTER_ALL)
```

Closes #371

### DIFF
--- a/cosmos/__init__.py
+++ b/cosmos/__init__.py
@@ -6,13 +6,12 @@ Contains dags, task groups, and operators.
 
 __version__ = "0.7.5"
 
-from cosmos.dataset import get_dbt_dataset
-
-# re-export the dag and task group
 from cosmos.airflow.dag import DbtDag
 from cosmos.airflow.task_group import DbtTaskGroup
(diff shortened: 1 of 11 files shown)


jlaneve:
only comment here is to see if we can convert strings to enums in/before the converter so this isn't a breaking change. all current code looks good though!

tatiana:
Thanks, @jlaneve , I'll address it!

Issue #372, opened 2023-07-17
Unify `dbt_project_name` and `dbt_root_path`


tatiana:
Unify in a single parameter the current parameters `dbt_project_name` and `dbt_root_path` - > `dbt_project_path`

At the moment, users are very often confused and don't specify the `dbt_root_path`, and end up with a DAG that has no tasks.

We also currently set the default value of the path to `/usr/local/airflow/dags/dbt`. This should be a mandatory property without any default values.

Add support for relative paths as well, that is, relative to the location of the DAG file.

Issue #369, opened 2023-07-14
Rendered SQL is not available for failed runs.


saikotcharlakota:
Airflow Rendered template is empty if the task if failed. I wish to see the rendered query/test that was executed resulting in the failure.

IMO, this is because `store_compiled_sql` is called after the error/exception check [local.py](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/local.py#L237) file. 

`exception_handling` methods simply raises AirflowException in case of error resulting rendered template as empty.

saikotcharlakota:
Hello,
I tested this issue and it still exists. 

DBT: 1.5.1
Cosmos: 1.2.0
Airflow: 2.5.1

agreenburg:
@tatiana I think this would be a very helpful thing to address and I would be happy to provide a submission.

tatiana:
Just read this issue, thanks for looping me in @agreenburg , and thanks for reporting it @saikotcharlakota !

It is a relevant improvement to Cosmos, and we'd appreciate if you'd like to contribute to fix this, @agreenburg !

Issue #336, opened 2023-06-22
Using normal dbt profiles instead of airflow connections


P1et1e:
Add a way to add normal dbt profiles to support a broader variety of different source and target systems.

Is there a workaround to use dbt-sqlserver with the current version of astronomer-cosmos ?

jlaneve:
Hey @P1et1e, thanks for opening this issue! We have another issue to track this: https://github.com/astronomer/astronomer-cosmos/issues/269

Out of curiosity, how are you running Airflow and how would you get the dbt profile into Airflow? Do you have a profile.yml file that you'd load into the same filesystem as Airflow, or would you want to supply the values via `profile_args` from your Airflow DAG code?

magdagultekin:
Hi @jlaneve, I also would love to be able to use the `profiles.yml` file :) I could go both ways - use the `profile_args` in the DAG code or use the yml file that I store in my `dbt` directory.

Issue #304, opened 2023-06-01
Review and refactor code to avoid polluting Xcom


tatiana:
At least one of the operators implemented by Cosmos is dumping unnecessary data to XCom:

https://github.com/astronomer/astronomer-cosmos/blob/5e0b2216f901ab14a79e94f2af5e42898bb28576/cosmos/providers/dbt/core/operators/local.py#L186-L187

The goal of this ticket is to:
* [ ] Identify parts of the code that have this behaviour
* [ ] Remove this behaviour (we shouldn't add anything to XCom unless we need it)

PR #385, opened 2023-07-24
Simplify interfaces everywhere to use new Config objects

jlaneve:
## Description

_Note: while this PR looks huge, it's mostly because we were passing around parameters in so many places. This is mostly a lift-and-shift from individual parameters everywhere to the 4 config classes described below._

Over time, we've continued adding parameters everywhere without giving much thought to the interface. This PR aims to simplify the interface for working with Cosmos DAGs/Task Groups and operators by collapsing the arguments into four config objects:
- `ProjectConfig`: parameters that Cosmos needs to understand where your dbt project is
- `RenderConfig`: parameters to tell Cosmos how to render your dbt project as a DAG
- `ExecutionConfig`: parameters to tell Cosmos how to execute your dbt project
- `ProfileConfig`: parameters to tell Cosmos what to run your dbt against

You can see the exact parameters each config takes here: https://github.com/astronomer/astronomer-cosmos/blob/simplify-interface/cosmos/config.py

With this, there are a few minor changes:
- profile name and dbt root path are collapsed into a single argument
- profile mappings are now optional - instead of passing a connection ID, you can now supply a path to a dbt profile.yml file to use
- somewhat related, I'm removing the automatic profile mapping selection - it was too hard to debug when you pass in a connection and the profile mapping doesn't pick up on it
- fixes a bug where the rendered SQL would only get set if the task was successful

Docs updates will be addressed in my open docs PR.

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/372
closes https://github.com/astronomer/astronomer-cosmos/issues/369
closes https://github.com/astronomer/astronomer-cosmos/issues/336
closes https://github.com/astronomer/astronomer-cosmos/issues/304
closes https://github.com/astronomer/astronomer-cosmos/issues/269
closes https://github.com/astronomer/astronomer-cosmos/issues/141

## Breaking Change?

Yes, the parameters to the DAG/Task Group and operators change. This is an intentional breaking change before the 1.0 release

## Checklist


cosmos/converter.py
```diff
@@ -50,149 +42,91 @@ def airflow_kwargs(**kwargs: dict[str, Any]) -> dict[str, Any]:
...
-            raise AirflowException(f"Can't specify the same {field[:-1]} in `select` and `exclude`: " f"{intersection}")
+    "Validates all configuration to ensure that they are compatible with each other."
+    project_config.validate_project()
 
-    # if task_args has a schema, add it to the profile args and add a deprecated warning
-    if "schema" in task_args:
-        profile_args["schema"] = task_args["schema"]
-        logger.warning("Specifying a schema in the `task_args` is deprecated. Please use the `profile_args` instead.")
+    # if we're in local or venv mode, make sure we have a profile
+    if execution_config.execution_mode == "local" or execution_config.execution_mode == "venv":
```
tatiana, line 59:
```suggestion
    if execution_config.execution_mode in (ExecutionMode.LOCAL or ExecutionMode.VIRTUALENV):
```

dev/dags/dbt/jaffle_shop/profiles.yml
```diff
@@ -1,13 +1,12 @@
...
-      user: postgres
-      password: postgres
+      host: postgres
+      user: airflow
+      password: pg_password
       port: 5432
-      dbname: public
-      schema: dbt_example
+      dbname: airflow
+      schema: public
```
tatiana, line 11: Let's use environment variables so each developer can configure this in the way it makes sense for them, and it would work for local development and remote.

tatiana (commented):
@jlaneve, it's great to see the interfaces simplified. Thanks a lot for working on this!

I'm still looking at the code, overall the changes look great. I am apprehensive about this last-minute breaking change:
> somewhat related, I'm removing the automatic profile mapping selection - it was too hard to debug when you pass in a connection and the profile mapping doesn't pick up on it

My suggestion is that we work together to support this in 1.0.

We can make debugging clear. I feel this is an important feature in the project for `ExecutionMode.LOCAL` and `ExecutionMode.VIRTUALENV` users - which consist of [64% of current Cosmos users](https://apache-airflow.slack.com/archives/C059CC42E9W/p1688140783397969).

cosmos/airflow/graph.py
```diff
@@ -78,15 +86,14 @@ def create_task_metadata(node: DbtNode, execution_mode: ExecutionMode, args: dic
             arguments=args,
         )
         return task_metadata
-    else:
-        logger.error(f"Unsupported resource type {node.resource_type} (node {node.unique_id}).")
+
+    raise ValueError(f"Unsupported resource type {node.resource_type} (node {node.unique_id}).")
```
tatiana, line 90: It could be more helpful to end-users if we were using a custom exception, as opposed to a Python one

cosmos/airflow/graph.py
```diff
@@ -78,15 +86,14 @@ def create_task_metadata(node: DbtNode, execution_mode: ExecutionMode, args: dic
...
 
 
 def create_test_task_metadata(
     test_task_name: str,
     execution_mode: ExecutionMode,
-    task_args: dict,
-    on_warning_callback: callable,
```
tatiana, line 89: Why did we remove the callback?

cosmos/config.py
```diff
@@ -0,0 +1,210 @@
...
+    :param append_env: If True, append the env dictionary to the existing environment. If False,
+    replace the existing environment with the env dictionary. Defaults to False
+    :param cancel_query_on_kill: If True, cancel the query when the dbt process is killed. If False,
+    do not cancel the query when the dbt process is killed. Defaults to True
+    :param install_deps: If True, install dbt dependencies before running dbt. Defaults to False
+    :param skip_exit_code: If the dbt process exits with this exit code, do not raise an exception.
+    Defaults to None
+    """
+
+    execution_mode: str | ExecutionMode = ExecutionMode.LOCAL
```
tatiana, line 188: I really don't think we should support strings in our representation of `execution_mode`, `load_method`, and `test_behavior`, now that we are exposing these configs to end-users and breaking many Cosmos interfaces.

Using strings to represent constants/enums is error-prone - for both us and end-users.

dev/dags/dbt/jaffle_shop_python/.user.yml
```diff
@@ -0,0 +1 @@
+id: 052feccc-57bc-4f49-9d9b-4d09a08ebf97
```
tatiana, line 1: We probably should add this to `.gitignore`

dev/dags/cosmos_seed_dag.py
```diff
@@ -1,53 +1,53 @@
...
-    jaffle_shop_seed = DbtSeedOperator(
-        task_id="seed_jaffle_shop",
-        conn_id="airflow_db",
-        project_dir=DBT_ROOT_PATH / "jaffle_shop",
-        profile_args={"schema": "public"},
-        outlets=[Dataset("SEED://JAFFLE_SHOP")],
-    )
-
-    drop_seeds >> jaffle_shop_seed
+# """
```
tatiana, line 1: Why did we comment this out?

cosmos/airflow/graph.py
jlaneve: callback is part of the task_args!

cosmos/config.py
```diff
@@ -0,0 +1,203 @@
...
+    seeds: str | Path = "seeds"
+    snapshots: str | Path = "snapshots"
+    manifest: str | Path | None = None
+
+    def __post_init__(self) -> None:
+        "Converts paths to `Path` objects."
+        self.dbt_project_path = Path(self.dbt_project)
+        self.models_path = self.dbt_project_path / Path(self.models)
+        self.seeds_path = self.dbt_project_path / Path(self.seeds)
+        self.snapshots_path = self.dbt_project_path / Path(self.snapshots)
```
tatiana, line 64: These need to be corrected. Let's say the user has:
```
>>> dbt_project = Path("/tmp/a")
>>> models =  Path("a/b")
>>> dbt_project / models
PosixPath('/tmp/a/a/b')
```
Which is not what the users may be expecting in the first place.

The previous implementation handles this type of scenario:
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/dbt/project.py#L10-L28

cosmos/dbt/graph.py
```diff
@@ -215,24 +209,30 @@ def load_from_dbt_manifest(self):
...
         * self.filtered_nodes
         """
+        if not self.project_config.is_manifest_available():
+            raise CosmosValueError(f"Unable to load manifest using {self.project_config.manifest_path}")
+
         logger.info("Trying to parse the dbt project using a dbt manifest...")
         nodes = {}
-        with open(self.project.manifest_path) as fp:
-            manifest = json.load(fp)
+        with self.project_config.manifest_path.open() as manifest_file:
```
tatiana, line 217: This is less readable than what we have before, I suggest to what we had before.

cosmos/dbt/graph.py
```diff
@@ -122,24 +87,47 @@ def load_via_dbt_ls(self):
...
             process = Popen(
-                command, stdout=PIPE, stderr=PIPE, cwd=self.project.dir, universal_newlines=True, env=os.environ
+                command,
+                stdout=PIPE,
+                stderr=PIPE,
+                cwd=self.project_config.dbt_project_path,
+                env=env,
+                universal_newlines=True,
             )
-        except FileNotFoundError as exception:
```
tatiana, line 135: This exception matters. It happens when the `dbt` binary path is not found in the system. We should not delete it.

cosmos/config.py
```diff
@@ -0,0 +1,203 @@
...
+
+@dataclass
+class CosmosConfig:
+    """
+    Contains all Cosmos config.
+    """
+
+    project_config: ProjectConfig
+    profile_config: ProfileConfig
+    render_config: RenderConfig = field(default_factory=RenderConfig)
```
tatiana, line 202: `render_config` is a very confusing name because it looks like a function name that renders a configuration.

cosmos/config.py
jlaneve: I don't think I understand, the implementation looks the same.

```
>>> dbt_project = Path("/tmp/a")
>>> models =  Path("a/b")
>>> dbt_project / models
PosixPath('/tmp/a/a/b')
```

This is what I'd expect to happen, actually. The models path is relative to the root of the directory. And, looking at the previous code, it'd do the same?

cosmos/dbt/graph.py
jlaneve: this felt like the more pathlib-native way of doing it. I don't think there's a reason to change it back unless you feel very strongly.

cosmos/dbt/graph.py
jlaneve: what do you think about adding an explicit check that the `dbt` binary is found? it really wasn't clear to me what this exception was supposed to be doing

cosmos/config.py
jlaneve: how about `renderer_config`? I don't think it's that confusing given it's one of the 4 configs

tests/dbt/test_graph.py
```diff
@@ -39,45 +54,116 @@ def test_load_via_manifest_with_exclude(pipeline_name, manifest_filepath, model_
 
 @patch("cosmos.dbt.graph.DbtGraph.load_from_dbt_manifest", return_value=None)
 def test_load_automatic_manifest_is_available(mock_load_from_dbt_manifest):
-    dbt_project = DbtProject(name="jaffle_shop", root_dir=DBT_PROJECTS_ROOT_DIR, manifest_path=SAMPLE_MANIFEST)
-    dbt_graph = DbtGraph(project=dbt_project)
-    dbt_graph.load(execution_mode=ExecutionMode.LOCAL)
+    dbt_graph = DbtGraph(
```
tatiana, line 57: It feels that we are overusing the overall config abstraction. The previous interface of `DbtGraph` was clear, and simpler to understand and test.

cosmos/dbt/graph.py
```diff
@@ -51,68 +35,49 @@ class DbtGraph:
...
+                    return self.load_via_dbt_ls()
                 except FileNotFoundError:
-                    self.load_via_custom_parser()
-                    return
-            else:
-                self.load_via_custom_parser()
-                return
+                    return self.load_via_custom_parser()
 
-        if method == LoadMode.DBT_MANIFEST and not self.project.is_manifest_available():
```
tatiana, line 110: Why was this removed?

tests/dbt/test_graph.py
```diff
@@ -139,29 +257,66 @@ def test_load_via_dbt_ls_with_exclude():
...
-    with pytest.raises(CosmosLoadDbtException) as err_info:
+    dbt_graph = DbtGraph(
+        cosmos_config=CosmosConfig(
+            project_config=ProjectConfig(
+                dbt_project=DBT_PROJECTS_ROOT_DIR / "jaffle_shop",
+            ),
+            profile_config=ProfileConfig(
+                profile_name="default",
+                target_name="dev",
+                path_to_profiles_yml=DBT_PROJECTS_ROOT_DIR / "jaffle_shop/profiles.yml",
```
tatiana, line 287:
```suggestion
                path_to_profiles_yml=DBT_PROJECTS_ROOT_DIR / pipeline_name / "profiles.yml",
```

tests/dbt/test_graph.py
```diff
@@ -171,12 +326,21 @@ def test_load_via_dbt_ls_with_runtime_error_in_stdout(mock_popen_communicate):
...
-        root_dir=DBT_PROJECTS_ROOT_DIR,
+    dbt_graph = DbtGraph(
+        cosmos_config=CosmosConfig(
+            project_config=ProjectConfig(
+                dbt_project=DBT_PROJECTS_ROOT_DIR / pipeline_name,
+            ),
+            profile_config=ProfileConfig(
+                profile_name="default",
+                target_name="dev",
+                path_to_profiles_yml=DBT_PROJECTS_ROOT_DIR / "jaffle_shop/profiles.yml",
```
tatiana, line 337:
```suggestion
                path_to_profiles_yml=DBT_PROJECTS_ROOT_DIR / pipeline_name / "profiles.yml",
```

tatiana:
@jlaneve I rebased `main` in a separate branch:
https://github.com/astronomer/astronomer-cosmos/tree/simplify-interface-rebase-wip

I ran the unit and integration tests locally for Py 3.8 and Airflow 2.6 and they passed. MyPy checks also seem to be happy.

Since it would override the history of this PR, I did it separately. 

If you're happy with the rebase, feel free to push to the branch of this PR using something like:
```
git push origin simplify-interface-rebase-wip:simplify-interface
```

PR #389, opened 2023-07-25
Initial implementation of simplifying config interfaces

jlaneve:
## Description

Over time, we've continued adding parameters everywhere without giving much thought to the interface. This PR aims to simplify the interface for working with Cosmos DAGs/Task Groups and operators by collapsing the arguments into four config objects:

- `ProjectConfig`: parameters that Cosmos needs to understand where your dbt project is
- `RenderConfig`: parameters to tell Cosmos how to render your dbt project as a DAG
- `ExecutionConfig`: parameters to tell Cosmos how to execute your dbt project
- `ProfileConfig`: parameters to tell Cosmos what to run your dbt against

You can see the exact parameters each config takes here: https://github.com/astronomer/astronomer-cosmos/blob/simplify-interface/cosmos/config.py

It also:
- simplifies the `dbt_project_name` and `dbt_root_path` arguments into a single arg
- removes `dbt_args`, which was a duplicate of `operator_args`

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/372

## Breaking Change?

Yes, the parameters to the DAG/Task Group and operators change. This is an intentional breaking change before the 1.0 release.

## Checklist

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -70,7 +70,6 @@ class ProjectConfig:
         mandatory_paths = {
             "dbt_project.yml": project_yml_path,
             "models directory ": self.models_relative_path,
-            "manifest": self.parsed_manifest_path,
         }
         for name, path in mandatory_paths.items():
             if path is None or not path.exists():
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -20,7 +20,7 @@ class RenderConfig:
 
     :param emit_datasets: If enabled model nodes emit Airflow Datasets for downstream cross-DAG
     dependencies. Defaults to True
-    :param test_behavior: The behavior for running tests. Defaults to after each
+    :param test_behavior: The behavior for running tests. Defaults to after each (model)
     :param load_method: The parsing method for loading the dbt model. Defaults to AUTOMATIC
     :param select: A list of dbt select arguments (e.g. 'config.materialized:incremental')
     :param exclude: A list of dbt exclude arguments (e.g. 'tag:nightly')
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -19,9 +19,9 @@ class RenderConfig:
     Class for setting general Cosmos config.
 
     :param emit_datasets: If enabled test nodes emit Airflow Datasets for downstream cross-DAG
-    dependencies
+    dependencies. Defaults to True
     :param test_behavior: The behavior for running tests. Defaults to after each
-    :param execution_mode: The execution mode for dbt. Defaults to local
+    :param load_method: The parsing method for loading the dbt model. Defaults to AUTOMATIC
     :param select: A list of dbt select arguments (e.g. 'config.materialized:incremental')
     :param exclude: A list of dbt exclude arguments (e.g. 'tag:nightly')
     """
@@ -43,7 +43,7 @@ class ProjectConfig:
     :param seeds_dir: The path to the dbt seeds directory within the project. Defaults to seeds
     :param snapshots_dir: The path to the dbt snapshots directory within the project. Defaults to
     snapshots
-    :param manifest_path: The path to the dbt manifest file. Defaults to None
+    :param manifest: The path to the dbt manifest file. Defaults to None
     """
 
     dbt_project: str | Path
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -18,7 +18,7 @@ class RenderConfig:
     """
     Class for setting general Cosmos config.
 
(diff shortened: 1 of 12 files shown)


dev/Dockerfile
```diff
@@ -7,7 +7,7 @@ COPY ./README.rst  ${AIRFLOW_HOME}/astronomer_cosmos/
 COPY ./cosmos/  ${AIRFLOW_HOME}/astronomer_cosmos/cosmos/
 
 # install the package in editable mode
-RUN pip install -e "${AIRFLOW_HOME}/astronomer_cosmos"[dbt-postgres]
+RUN pip install -e "${AIRFLOW_HOME}/astronomer_cosmos"[dbt-postgres,dbt-databricks]
```
harels, line 10: why did we add databricks here?

jlaneve: The python models require Databricks!

harels (approved):
this interface looks good to me, pushed some minor fixes

harels:
@tatiana wdyt?

cosmos/config.py
```diff
@@ -0,0 +1,122 @@
...
+
+logger = getLogger(__name__)
+
+
+@dataclass
+class RenderConfig:
+    """
+    Class for setting general Cosmos config.
+
+    :param emit_datasets: If enabled test nodes emit Airflow Datasets for downstream cross-DAG
```
tatiana, line 21:
```suggestion
    :param emit_datasets: If enabled model nodes emit Airflow Datasets for downstream cross-DAG
```

cosmos/config.py
```diff
@@ -0,0 +1,122 @@
...
+    dependencies. Defaults to True
+    :param test_behavior: The behavior for running tests. Defaults to after each
```
tatiana, line 23:
```suggestion
    :param test_behavior: The behavior for running tests. Defaults to after each (model)
```

cosmos/config.py
```diff
@@ -0,0 +1,122 @@
...
+    manifest: str | Path | None = None
+
+    manifest_path: Path | None = None
+
+    def __post_init__(self) -> None:
+        "Converts paths to `Path` objects."
+        self.dbt_project_path = Path(self.dbt_project)
+        self.models_path = self.dbt_project_path / Path(self.models)
+        self.seeds_path = self.dbt_project_path / Path(self.seeds)
+        self.snapshots_path = self.dbt_project_path / Path(self.snapshots)
```
tatiana, line 62: Following the [discussion](https://github.com/astronomer/astronomer-cosmos/pull/385/files#r1272817393) from the previous PR:

I believe we should:
(a) document that models/seeds/snapshots must be relative to the `dbt` project
(b) validate if they are valid, raising an exception if they don't exist

cosmos/config.py
tatiana: We can log a ticket and assign it to a future milestone to change where the datasets are emitted. I do think we need to review how we're supporting tests overall, including running a test per node, if the user wants to.

cosmos/config.py
```diff
@@ -0,0 +1,122 @@
...
+    :param snapshots_dir: The path to the dbt snapshots directory within the project. Defaults to
+    snapshots
+    :param manifest: The path to the dbt manifest file. Defaults to None
+    """
+
+    dbt_project: str | Path
+    models: str | Path = "models"
+    seeds: str | Path = "seeds"
+    snapshots: str | Path = "snapshots"
+    manifest: str | Path | None = None
```
tatiana, line 53: It would be great if our attribute names make a distinction between what is a folder and what is a file path

cosmos/config.py
```diff
@@ -0,0 +1,122 @@
...
+    :param seeds_dir: The path to the dbt seeds directory within the project. Defaults to seeds
+    :param snapshots_dir: The path to the dbt snapshots directory within the project. Defaults to
+    snapshots
+    :param manifest: The path to the dbt manifest file. Defaults to None
+    """
+
+    dbt_project: str | Path
+    models: str | Path = "models"
+    seeds: str | Path = "seeds"
+    snapshots: str | Path = "snapshots"
```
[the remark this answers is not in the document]
harels: https://github.com/astronomer/astronomer-cosmos/issues/392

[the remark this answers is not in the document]
harels: the can be paths within the dbt projects, not just folders directly under the root folder of the project

cosmos/config.py
harels: clarified the name so that there's no ambiguity

PR #402, opened 2023-07-27
update graph parsing mode when virtualenv is selected

### LINKED ISSUE #401: Unable to find the dbt executable while trying to run cosmos with virtual env
I have tried to call a DbtDag using cosmos v1.0.0 in my custom airflow docker image and I got the following error message

```
Broken DAG: [/usr/local/airflow/dags/dag_dbt.py] Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/cosmos/dbt/graph.py", line 110, in load
    self.load_via_dbt_ls()
  File "/usr/local/lib/python3.7/site-packages/cosmos/dbt/graph.py", line 139, in load_via_dbt_ls
    raise CosmosLoadDbtException(f"Unable to find the dbt executable: {self.dbt_cmd}")
cosmos.dbt.graph.CosmosLoadDbtException: Unable to find the dbt executable: dbt

```

Which is a very strange error since dbt executable should not even be searched for in the early stages if a virtual env will be called to run it.

Code
```
from airflow import DAG
from pendulum import datetime
from cosmos import DbtDag, ExecutionMode, ExecutionConfig, ProjectConfig, ProfileConfig
from cosmos.profiles import GoogleCloudServiceAccountFileProfileMapping
from pathlib import Path

DBT_ROOT_PATH = Path(__file__).parent / "dbt"

    profile_name="default",
    target_name="dev",
    profile_mapping=GoogleCloudServiceAccountFileProfileMapping(
        conn_id="bigquery_default",
        profile_args={
            "project" : "zattoo-dataeng",
            "dataset": "temp"},
    ),
)

my_cosmos_dag = DbtDag(
    project_config=ProjectConfig(
    ),
        execution_mode=ExecutionMode.VIRTUALENV,
    ),
    operator_args={
        "py_system_site_packages": False,
        "py_requirements": ["dbt-bigquery==1.5.0"],
    },
    # normal dag parameters
    schedule_interval="@daily",
    catchup=False,
    dag_id="my_cosmos_dag",
)
```

I checked that in the cosmos dev sandbox, the virtualenv works fine with very similar configs, but the image _quay.io/astronomer/astro-runtime:7.3.0-base_ is being used into it, which I believe that maybe has dbt installed into it by default?

CorsettiS:
## Description

When using a virtualenv to run our dbt tasks, we are working under the hypothesis that dbt is not installed in the machine. However as of now, if we choose virtualenv as the execution mode, dbt tries to parse the graph of the dbt tasks with **load_via_dbt_ls()**, while it should point out to **load_via_custom_parser()** due to the fact aforementioned. we can accomplish it by removing the virutalenv executor from the _elif_ condition for **load_via_dbt_ls()**, so it will be picked up automatically by the _else_ condition.

## Related Issue(s)

closes #401

## Breaking Change?

No. it worked fine in the dev sandbox, although I did not have the opportunity to test it with any image without dbt pre-installed.
Obs: I had to change also the tests for this change

## Checklist


jlaneve (approved):
nice, thank you! Integration tests are failing because external contributors don't have access to GH actions secrets, but this looks good

PR #407, opened 2023-07-27
Ensure safer profile mappings by removing __getattr__

jlaneve:
## Description

Before, profile mappings would override `__getattr__` to get the dbt profile value, primarily for convenience. However, this was causing unintended side effects - infinite recursive loops. This PR fixes the issue by removing the `__getattr__` override in favor of being more explicit.

## Breaking Change?

No, this is all behind-the-scenes.

## Checklist


harels (approved):
Approved, pending integration tests passing

Issue #411, opened 2023-07-27
Write logs to a temp dir when using `dbt_ls` parsing mode


jlaneve:
In Slack, a user was getting an issue where the `dbt ls` command for generating the manifest was trying to write logs to the same directory, which is typically read-only.

```
20:43:06  Encountered an error:
[Errno 30] Read-only file system: '/usr/local/airflow/dags/dbt_grindr/logs/dbt.log'
```

We should instead have dbt write to a temp dir and paste the logs to stdout if there are any written to disk

tatiana:
Users could resolve this by setting the environment variable `DBT_LOG_PATH` (https://docs.getdbt.com/reference/project-configs/log-path)

tatiana:
I just found the original bug report by Adam Underwood:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1690490672020429?thread_ts=1690473607.733569&cid=C059CC42E9W

The problem goes beyond the logs directory since the `dbt ls` command also creates a `target` directory, where several files are attempted to be written, including: `target/partial_parse.msgpack`. This could be overcome by setting `DBT_TARGET_PATH`.

tatiana:
A solution to fix both issues if the user does not set DBT-specific environment variables can be to create a temporary directory from where we'll run `dbt ls`.

This would assume that the user/process running cosmos has to write permissions to create/write temporary directories.

tatiana:
@jlaneve let's align here on the approach, given the feedback you gave in:
https://github.com/astronomer/astronomer-cosmos/pull/414

A proposal:
1. If the user defines `DBT_TARGET_PATH` and/or `DBT_LOG_PATH`, we should use them when running `dbt ls`.
2. If the user doesn't define one/both of them, we'll be creating a `tmp` directory where we'll output both target and logs
3. By the end of running `dbt ls`, we'll log everything that is in the given dbt logs file (we can start by logging it using the `debug` mode, in future we can use the Airflow log level to decide which dbt log levels we want to print)

Are you happy with this approach?

jlaneve:
> @jlaneve let's align here on the approach, given the feedback you gave in: #414
> 
> A proposal:
> 
> 1. If the user defines `DBT_TARGET_PATH` and/or `DBT_LOG_PATH`, we should use them when running `dbt ls`.
> 2. If the user doesn't define one/both of them, we'll be creating a `tmp` directory where we'll output both target and logs
[...]

PR #414, opened 2023-07-28
Refactor `dbt ls` to run from a temporary directory

tatiana:
As of Cosmos 1.0.0, `LoadMode.DBT_LS` runs `dbt ls` from within the original dbt project directory.

The `dbt ls` outputs files to the directory it's running from unless the environment variables `DBT_LOG_PATH` and `DBT_TARGET_PATH` are specified (as of dbt 1.6). 

Depending on the deployment, the Airflow worker does not have write permissions to the dbt project directory. This can lead to an error message similar to the following:

```
20:43:06  Encountered an error:
[Errno 30] Read-only file system: '/usr/local/airflow/dags/dbt_grindr/logs/dbt.log'
```

This PR changes the behavior of `dbt ls` to try to make the `dbt ls` artifacts (logs and target directory) not be written to the original project directory.

In addition to the introduced test, this change was validated using airflow 2.6 and dbt 1.6, by following these steps:
(1) Delete folders `logs` and `target` from `astronomer-cosmos/dev/dags/dbt/jaffle_shop`
(2) Add a breakpoint after `stdout, stderr = process.communicate()` in `dbt/graph.py`
(3) Run a DAG that uses  `astronomer-cosmos/dev/dags/dbt/jaffle_shop`, e.g.:
```
airflow dags test basic_cosmos_dag `date -Iseconds`
```
(4) When the breakpoint happens, check that no `target` or `logs` folder was created after running `dbt ls` in `astronomer-cosmos/dev/dags/dbt/jaffle_shop`

A limitation with the current approach is that, although `dbt ls` is not creating these directories in the given circumstances, if the user is using the local executor or an earlier version of `dbt`, the files will still be written to the project directory.

Closes: #411

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
       - id: rst-backticks
       - id: python-check-mock-methods
   - repo: https://github.com/Lucas-C/pre-commit-hooks
-    rev: v1.5.3
+    rev: v1.5.1
     hooks:
       - id: remove-crlf
       - id: remove-tabs
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -145,7 +145,10 @@ class DbtGraph:
             command.extend(["--select", *self.select])
 
         with tempfile.TemporaryDirectory() as tmpdir:
-            shutil.copytree(self.project.dir, tmpdir, dirs_exist_ok=True)
+            # when we drop support to Python 3.7, we can use:
+            # shutil.copytree(self.project.dir, tmpdir, dirs_exist_ok=True)
+            cwd = Path(tmpdir) / self.project.name
+            shutil.copytree(self.project.dir, cwd)
 
             with self.profile_config.ensure_profile() as (profile_path, env_vars):
                 command.extend(
@@ -168,7 +171,7 @@ class DbtGraph:
                     command,
                     stdout=PIPE,
                     stderr=PIPE,
-                    cwd=tmpdir,
+                    cwd=cwd,
                     universal_newlines=True,
                     env=env,
                 )
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -15,6 +15,8 @@ from cosmos.constants import (
     ExecutionMode,
     LoadMode,
     DBT_LOG_FILENAME,
+    DBT_LOG_PATH_ENVVAR,
+    DBT_TARGET_PATH_ENVVAR,
 )
 from cosmos.dbt.executable import get_system_dbt
 from cosmos.dbt.parser.project import DbtProject as LegacyDbtProject
@@ -169,6 +171,11 @@ class DbtGraph:
             with tempfile.TemporaryDirectory() as tmpdir:
                 logger.info("Running command: `%s`", " ".join(command))
                 logger.info("Environment variable keys: %s", env.keys())
+                log_dir = Path(env.get(DBT_LOG_PATH_ENVVAR) or tmpdir)
+                target_dir = Path(env.get(DBT_TARGET_PATH_ENVVAR) or tmpdir)
+                env[DBT_LOG_PATH_ENVVAR] = str(log_dir)
+                env[DBT_TARGET_PATH_ENVVAR] = str(target_dir)
+
                 process = Popen(
                     command,
                     stdout=PIPE,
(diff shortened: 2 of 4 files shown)


tests/dbt/test_graph.py
```diff
@@ -108,6 +108,29 @@ def test_load(
...
+                conn_id="airflow_db",
+                profile_args={"schema": "public"},
+            ),
+        ),
+    )
+
+    dbt_graph.load_via_dbt_ls()
+    used_cwd = Path(mock_popen.call_args[0][0][-5])
+    assert used_cwd != dbt_project.dir
+    assert not used_cwd.exists()
```
jlaneve, line 131: this doesn't look like it's confirming that files are written to the temp directory, right?

cosmos/dbt/graph.py
```diff
@@ -143,33 +144,39 @@ def load_via_dbt_ls(self) -> None:
...
-                    self.profile_config.profile_name,
-                    "--target",
-                    self.profile_config.target_name,
-                ]
-            )
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # when we drop support to Python 3.7, we can use:
+            # shutil.copytree(self.project.dir, tmpdir, dirs_exist_ok=True)
+            cwd = Path(tmpdir) / self.project.name
+            shutil.copytree(self.project.dir, cwd)
```
jlaneve, line 151: it's been raised elsewhere that copying the entire directory isn't a great idea with larger projects. curious why we don't just use the log/target path args?

https://docs.getdbt.com/reference/project-configs/log-path#configuration
https://docs.getdbt.com/reference/project-configs/target-path

cosmos/dbt/graph.py
```diff
@@ -143,33 +144,39 @@ def load_via_dbt_ls(self) -> None:
...
+                    command,
+                    stdout=PIPE,
+                    stderr=PIPE,
+                    cwd=cwd,
+                    universal_newlines=True,
+                    env=env,
+                )
 
-            stdout, stderr = process.communicate()
+                stdout, stderr = process.communicate()
```
jlaneve, line 179: if logs do get written, how would a user access them?

tests/dbt/test_graph.py
tatiana: Just that they are not being written to the project dir and that the directory they were written to no longer exists.
Temp directories naming can vary depending on the OS, so I thought this was safer. Do you recommend any alternative?

cosmos/dbt/graph.py
tatiana: That's a fair point, I'll change to pass the arguments using the value of `DBT_LOG_PATH`  and `DBT_TARGET_PATH`, if they weren't set, and setting those to the temp dir otherwise

cosmos/dbt/graph.py
tatiana: If users care about logs, they should define `DBT_LOG_PATH`

jlaneve: I disagree - logs are written when the parsing fails. If I'm setting up Cosmos, and the parsing fails, I won't have set up my own `DBT_LOG_PATH` in advance. Instead, I would expect to see the logs of _what_ exactly failed so I could troubleshoot.

IMO the desired behavior is: we set the log and target path for the user; if, at the end of execution, there's something in the log folder (i.e. the command failed and dbt wrote logs), we should print those logs so a user can troubleshoot on their own. If we don't print the logs, a user won't be able to access them because they get cleaned up via the temp dir.

tatiana: Just implemented this logic, @jlaneve

cosmos/dbt/graph.py
tatiana: Unfortunately, neither the flags nor the env vars necessarily work with versions before dbt 1.6 😢 as discussed here:
https://github.com/astronomer/astronomer-cosmos/issues/411#issuecomment-1668259407

However, we've changed the approach and we are no longer copying the project dir 🎉

tatiana:
Thanks, @jlaneve , I think I addressed all the feedback!

tatiana:
A different approach we could adopt to solve this issue is not to pass `project-dir` to `dbt-ls` and to create symbolic links for the original `project-dir` files to the temp-dir. This way, the target & logs will be written to the temp dir.

Still, this change will not solve the limitation that ATM local operators run things locally. @jlaneve @harels is the scope of this ticket only `dbt ls` or all local operator executions?

.pre-commit-config.yaml
```diff
@@ -37,7 +37,7 @@ repos:
       - id: rst-backticks
       - id: python-check-mock-methods
   - repo: https://github.com/Lucas-C/pre-commit-hooks
-    rev: v1.5.3
+    rev: v1.5.1
```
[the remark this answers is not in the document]
harels: It's ok to try in the PR

cosmos/dbt/graph.py
```diff
@@ -158,46 +167,58 @@ def load_via_dbt_ls(self) -> None:
...
-        self.filtered_nodes = nodes
+                stdout, stderr = process.communicate()
+
+                logger.debug("dbt output: %s", stdout)
+                log_filepath = log_dir / DBT_LOG_FILENAME
+                logger.debug("dbt logs available in: %s", log_filepath)
+                if log_filepath.exists():
+                    with open(log_filepath) as logfile:
+                        for line in logfile:
+                            logger.debug(line.strip())
```
harels, line 195: Nice!

harels (approved):

PR #419, opened 2023-08-01
Allow profiles_yml_filepath to be a str type 

CorsettiS:
## Description

In the documentation of astronomer, user-supplied profiles are defined as follows

```

    profile_name="my_snowflake_profile",
    target_name="dev",
    profiles_yml_filepath="/path/to/profiles.yml",
)

dag = DbtDag(profile_config=profile_config, ...)
```

however in the code the _profiles_yml_filepath_ crashes if a string is supplied to it, since it assumes a Path type instead of a str type. In the code logic, **self.profiles_yml_filepath.exists()** crashes if the profiles_yml_filepath is a str. this PR just makes a few adjustment to allow users to supply a custom filepath in a string format without having to wrap it over a Path()

## Related Issue(s)

---

## Breaking Change?

No
## Checklist


harels (approved):
good catch! thanks @CorsettiS!

PR #423, opened 2023-08-01
Resolving "Crash when adding sources.yml"

abhi12mohan:
## Description

Simply adds "source" as a DbtResourceType. 

## Checklist


jlaneve (approved):
tested locally by adding a source, can confirm it works

tatiana (commented):
@abhi12mohan, we need to add some automated tests to make sure we don't re-introduce this issue in the future (having a `sources.yml` file). I see the change was already merged. Could you add this in a separate PR, please?

PR #426, opened 2023-08-01
Fix integration tests

jlaneve:
## Description

The Kubernetes Airflow provider and dbt Core released new versions in the last 24 hours, both of which broke Cosmos. This puts an upper version limit on the packages while we work to address those breaking changes.

## Breaking Change?

No

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,38 +46,38 @@ dependencies = [
 
 [project.optional-dependencies]
 dbt-all = [
-    "dbt-bigquery",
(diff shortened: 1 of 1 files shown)


harels (approved):

tatiana:
This was released as part of 1.1.3:
https://pypi.org/project/astronomer-cosmos/1.1.3/
https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.1.3

PR #429, opened 2023-08-01
Implement prefix + DBT selector graph operator for model selection

tseruga:
## Description

This PR adds a commonly used graph operator for model selection in DBT.
https://docs.getdbt.com/reference/node-selection/graph-operators#the-plus-operator

This is just a first stab at implementing (what I assume is) the most commonly used graph operator. It just so happens that it's also the easiest to implement.

This change allows users to specify the DBT model they wish to select, as well as its upstream dependencies.

`select=["+my_model"]` will return the `my_model` as well as any of the nodes that it `depends_on`, and any nodes _those_ nodes depend on, etc.

Haven't written up the docs because I want to make sure folks are happy with handling the graph operators this way. It might limit future implementation of the remaining operators, so happy to hear feedback on the topic.

This is just what I hacked together while trying to play around with Cosmos post-1.0.0 :)

NOTE: I also cherry-picked #426 in order to get the tests to work, but will rebase when required.

## Checklist


tseruga:
Added another test to hopefully make Coverage happy -- and rebased off of main.

tatiana:
@tseruga, this work was looking really good. It would be great if we could rebase it and get to the finishing line! Please let us know if you need any support.

tatiana:
Thanks a lot, @tseruga ! Please let us know when the PR is ready for review - I've assigned it for the 1.3 release, and we can review it if necessary.

tatiana:
@tseruga Since this PR was stale, and we had recent customer requests for this feature, I followed up on the missing bits and created a new PR: #728. You're added as a co-author - please feel free to review and give feedback!

PR #434, opened 2023-08-02
Resolved MyPy errors when adding Airflow pre-commit dependency.

abhi12mohan:
## Description

Provided different solutions for new MyPy errors that were the result of adding "apache-airflow" as a dependency within the pre-commit config.

## Checklist

### DIFF
--- a/cosmos/dataset.py
+++ b/cosmos/dataset.py
@@ -15,8 +15,10 @@ except (ImportError, ModuleNotFoundError):
             self.id = id
             logger.warning("Datasets are not supported in Airflow < 2.5.0")
 
-        def __eq__(self, other: "Dataset") -> bool:
-            return bool(self.id == other.id)
+        def __eq__(self, other: object) -> bool:
+            if not isinstance(other, Dataset):
+                return NotImplemented
+            return bool(self.id == other.id)  # type: ignore[attr-defined]
 
 
 def get_dbt_dataset(connection_id: str, project_name: str, model_name: str) -> Dataset:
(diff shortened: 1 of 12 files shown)


cosmos/dataset.py
```diff
@@ -15,8 +15,10 @@ def __init__(self, id: str, *args: Tuple[Any], **kwargs: str):
             self.id = id
             logger.warning("Datasets are not supported in Airflow < 2.5.0")
 
-        def __eq__(self, other: "Dataset") -> bool:
-            return bool(self.id == other.id)
+        def __eq__(self, other: object) -> bool:
+            if not isinstance(other, Dataset):
+                return NotImplemented
+            return bool(self.id == other.id)  # type: ignore[attr-defined]
```
jlaneve, line 21: see if swapping the cases fixes this

cosmos/operators/local.py
```diff
@@ -110,18 +111,22 @@ def store_compiled_sql(self, tmp_project_dir: str, context: Context, session: Se
...
-        session.query(RenderedTaskInstanceFields).filter(
-            RenderedTaskInstanceFields.dag_id == self.dag_id,
-            RenderedTaskInstanceFields.task_id == self.task_id,
-            RenderedTaskInstanceFields.run_id == ti.run_id,
-        ).delete()
-        session.add(rtif)
-
-    def run_subprocess(self, *args: Tuple[Any], **kwargs: Any) -> FullOutputSubprocessResult:
+
+        if isinstance(ti, TaskInstance):
```
jlaneve, line 115: add comment about what else it could be / why we need this

cosmos/dataset.py
abhi12mohan: it doesn't, seems like ignoring is the least invasive solution rn

jlaneve (approved):
Changes look good, CI is passing, and I can run things locally. Nice!

PR #438, opened 2023-08-03
renaming on_warning_callback argument to on_test_warning_callback

CorsettiS:
## Description

 The arg **on_warning_callback** is renamed to **on_test_warning_callback**. This change is being done to prepare the groundwork for the **on_test_error_callback** argument that I plan to implement soon for the local execution mode and then expand both args to other execution modes.

## Breaking Change?

Yes.  **on_warning_callback** became **on_test_warning_callback**, therefore users should rename it accordingly 

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -206,6 +206,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
       - name: Set up Python 3.10
         uses: actions/setup-python@v3
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,12 +5,23 @@ on:
     branches: [main]
   pull_request: # Run on pull requests targeting the default branch
     branches: [main]
+  pull_request_target: # Run on pull requests originated from forks
+    branches: [main]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:
+
+  Authorize:
+    environment:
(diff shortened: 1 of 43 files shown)


jlaneve:
Thanks @CorsettiS! Two quick things:
- this is a breaking change, users who supply the `on_warning_callback` today will need to update their code to `on_test_warning_callback` (to be clear: I'm fine with the breaking change, just want to call this out for any users looking at this PR)
- mind fixing the [pre-commit checks](https://results.pre-commit.ci/run/github/577757880/1691093425.hkkSJSZ3QI-ECezowTtFkA)? once they're fixed, this looks good to me!

CorsettiS:
> Thanks @CorsettiS! Two quick things:
> 
> * this is a breaking change, users who supply the `on_warning_callback` today will need to update their code to `on_test_warning_callback` (to be clear: I'm fine with the breaking change, just want to call this out for any users looking at this PR)
> * mind fixing the [pre-commit checks](https://results.pre-commit.ci/run/github/577757880/1691093425.hkkSJSZ3QI-ECezowTtFkA)? once they're fixed, this looks good to me!

@jlaneve thanks for the clarification ! just changed it !

tatiana (changes requested):
Let's deprecate the previous parameter, having a deprecation warning. Since this is a breaking change, we should not release it in minor/micro releases.

CorsettiS:
> [@tatiana](/tatiana)
> ** tatiana ** requested changes

I have done as requested. now everytime that on_warning_callback is called instead of on_test_warning_callback a logging.warn will pop up

CorsettiS:
I have seen that some tests failed. I will have to take a look into that later, although recently I am busy with some personal projects. I will add updates here once I have more time

CorsettiS:
I will close this PR for now and reopen it once I deploy the on_test_failure_callback method

Issue #432, opened 2023-08-02
Don’t run integration tests if PR is coming from a forked repo


jlaneve:

tatiana:
An alternative to not running the integration tests from forks is only to run integration tests if the maintainers add labels to those PRs originated from forks, like this: https://github.com/astronomer/astro-sdk/pull/546

PR #441, opened 2023-08-04
Fix selector for config tags

### LINKED ISSUE #440: dbt/selectors doesn't work for config tags
The filtering done on select for config.tags checks if a config from the select is a subset of the node's config but this will always fail as it will try to check if a key: string is a subset of a key: list.

So for example, using `select="config.tags:has_child"` on a node with its config tags equal to `['has_child']`, the process would look something like this:
```
config = { 'tags': 'has_child' }
supported_node_config = { 'tags': ['has_child'] }
config.items() <= supported_node_config.items()
```
Which would always result to `False`.

I considered modifying the `load_from_statement` function to parse the config tags into a list but that would only work if the tags match exactly and would fail in a different scenario like:
```
config = { 'tags': ['has_child'] }
supported_node_config = { 'tags': ['has_child', 'nightly'] }
config.items() <= supported_node_config.items()
```
Which would return `False`.

My suggestion is to check for the config tags and then do the current check for all other configs. I'll add a PR in a bit.

javihernovoa:
## Description

Fixing an issue where the check in `selector.py`, specifically `select_nodes_ids_by_intersection`, didn't validate tags in the node's config. The function will now check if the tag specified in config is contained in the node's config and check all other configs to be filtered for are a subset of the node's config, as previously.  

## Related Issue(s)

closes #440 
## Checklist

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -89,8 +89,17 @@ def select_nodes_ids_by_intersection(nodes: dict[str, DbtNode], config: Selector
             continue
 
         supported_node_config = {key: value for key, value in node.config.items() if key in SUPPORTED_CONFIG}
-        if config.config and not (config.config.items() <= supported_node_config.items()):
-            continue
+        if config.config:
+            config_tag = config.config.get('tags')
+            if config_tag and not config_tag in supported_node_config.get("tags", []):
+                continue
+
+            # Remove 'tags' as we've already filtered for them
+            config.config.pop("tags", None)
+            supported_node_config.pop("tags", None)
+
+            if not (config.config.items() <= supported_node_config.items()):
+                continue
 
         if config.paths and not (set(config.paths).issubset(set(node.file_path.parents))):
             continue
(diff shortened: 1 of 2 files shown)


cosmos/dbt/selector.py
```diff
@@ -89,8 +89,17 @@ def select_nodes_ids_by_intersection(nodes: dict[str, DbtNode], config: Selector
...
 
         supported_node_config = {key: value for key, value in node.config.items() if key in SUPPORTED_CONFIG}
-        if config.config and not (config.config.items() <= supported_node_config.items()):
-            continue
+        if config.config:
+            config_tag = config.config.get("tags")
+            if config_tag and config_tag not in supported_node_config.get("tags", []):
+                continue
+
+            # Remove 'tags' as we've already filtered for them
```
javihernovoa, line 97:
```suggestion
            # Remove 'tags' as they've already been filtered for
```

tatiana (approved):
Thank you very much, @javihernovoa , it is looking great!

We still need to fix our CI to run tests from forks as part of #432. In the meantime, I ran the tests locally, and they are passing.

Issue #443, opened 2023-08-07
GoogleCloudServiceAccountDictProfileMapping should create yaml fields


joppedh:
The `GoogleCloudServiceAccountDictProfileMapping` should extend the fields as yaml and not as JSON as it does now. This is not supported by DBT. See https://docs.getdbt.com/docs/core/connect-data-platform/bigquery-setup#service-account-json

Currently it creates the profile like so:
```yaml
bigquery:
    outputs:
        dev:
            dataset: my_bigquery_dataset
            keyfile_json: '{   "type": "service_account",   "project_id": "stg",   "universe_domain":
                "googleapis.com" }'
            method: service-account-json
            threads: 1
            type: bigquery
```
It should be like so:
```yaml
bigquery:
    outputs:
        dev:
            dataset: my_bigquery_dataset
            keyfile_json: 
              type: xxx
              project_id: xxx
              private_key_id: xxx
            method: service-account-json
            threads: 1
            type: bigquery
```

joppedh:
@tatiana I created an issue here

jbandoro:
I'm seeing this as well in 1.0.4. In the last cosmos version that worked with keyfile json, `json.loads()` was used after getting the keyfile dict: https://github.com/astronomer/astronomer-cosmos/blob/1bce508f1f71553d33c72bee6aafb6169b3a4d0d/cosmos/providers/dbt/core/profiles/bigquery.py#L43-L46

@tatiana  would the best fix for this be to add a `transform_keyfile_json` method to `GoogleCloudServiceAccountDictProfileMapping` where it does `json.loads`?

Issue #436, opened 2023-08-03
Add support to create the profile from Google Cloud Airflow connection (Keyfile JSON)


manuelgblanco:
Hi there!

we are using Astronomer Cosmos 🚀, specifically the `DbtTaskGroup` operator with the profile for BigQuery (`GoogleCloudServiceAccountDictProfileMapping`):

```python
bq_dbt_profile_config = ProfileConfig(
    profile_name="bigquery_profile",
    target_name="bigquery_profile",
    profile_mapping=GoogleCloudServiceAccountDictProfileMapping(
        conn_id="google_cloud_default",
        profile_args={
            "project": "my_project_id",
            "dataset": "my_dataset_name"
        }
    )
)
```

The issue that we have is when we set the Google Cloud Airflow connector via _`Keyfile JSON`_, like this:

![image](https://github.com/astronomer/astronomer-cosmos/assets/93135676/2e09fd59-dc13-491a-b0ad-a5b9e23b4d67)

where the content is the result of paste the SA json file, being for instance, something like:
```json
{
  "type": "service_account",
  "project_id": "my_project_id",
  "private_key_id": "xxxx",
  "private_key": "-----BEGIN PRIVATE KEY-----\nblabla\nblabla\n-----END PRIVATE KEY-----\n",
  "client_email": "xxx@my_project_id.iam.gserviceaccount.com",
  "client_id": "my_client_id",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://oauth2.googleapis.com/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/xxxx.iam.gserviceaccount.com"
}
```

The result parsed is:
![image](https://github.com/astronomer/astronomer-cosmos/assets/93135676/08403d0f-59eb-43af-8134-c1b488b86c90)

![image](https://github.com/astronomer/astronomer-cosmos/assets/93135676/993bbf02-8801-4e6b-9483-0f2c2895f801)

From my humble opinion, I think that this is due to not processing as json [here](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/profiles/base.py#L150), only it is parsed when the keyword start with `extra.`

Right now, we are setting the Airflow connection as `Google Bigquery` so that this connection type allows extra parameters:
[...]

PR #449, opened 2023-08-09
Fix BigQuery `keyfile_dict` mapping for connection created from webserver

jbandoro:
## Description

GCP connections created with a keyfile json from the webserver UI were not previously working as documented in the related issues below. The keyfile_dict created in the webserver UI is stored as a string and is the reason why in the GoogleBaseHook [here](https://github.com/apache/airflow/blob/a752c902861d9a829d7451cc69d1343cbe5b54dc/airflow/providers/google/common/hooks/base_google.py#L250-L254) it uses `json.loads` if the keyfile_dict isn't a dict.

## Related Issue(s)

#436 #443

## Breaking Change?

None
## Checklist

### DIFF
--- a/cosmos/profiles/bigquery/service_account_keyfile_dict.py
+++ b/cosmos/profiles/bigquery/service_account_keyfile_dict.py
@@ -46,7 +46,7 @@ class GoogleCloudServiceAccountDictProfileMapping(BaseProfileMapping):
             **self.profile_args,
         }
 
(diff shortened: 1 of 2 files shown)


tatiana:
Thank you very much, @jbandoro, for reporting and fixing this issue! The integration tests are failing due to a misconfiguration in our CI. I successfully ran the integration tests locally, and they passed.

tatiana (approved):

PR #461, opened 2023-08-14
Change CI to run coverage after tests pass

tatiana:
There are two benefits to this change:
1. The CI will wait for all the coverage to be collected, avoiding failing a build incorrectly because of code coverage that was not collected yet
2. Fewer requests to the third-party `codecov` service - we only worry about code coverage when/if tests pass


harels (approved):
Nice improvement! Only drawback I can think of is that external committers will have to wait for us to approve the integration test workflow before they see codecov stats - but that feels like a good tradeoff.

PR #463, opened 2023-08-14
Fix bug on select node add exclude selector subset ids logic 

jensenity:
## Description

1. Add Skip case if all configs are None
2. Exclude subset id to not be overwrite as it iterates
3. Add more test cases

## Related Issue(s)

It closes: https://github.com/astronomer/astronomer-cosmos/issues/464

## Checklist

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -42,6 +42,10 @@ class SelectorConfig:
         self.config: dict[str, str] = {}
         self.other: list[str] = []
         self.load_from_statement(statement)
+        self.is_config_empty: bool = False
+
+        if not self.paths and not self.tags and not self.config and not self.other:
+            self.is_config_empty = True
 
     def load_from_statement(self, statement: str) -> None:
         """
@@ -85,7 +89,7 @@ def select_nodes_ids_by_intersection(nodes: dict[str, DbtNode], config: Selector
     """
     selected_nodes = set()
     for node_id, node in nodes.items():
-        if not config.paths and not config.tags and not config.config:
+        if config.is_config_empty:
             continue
 
         if config.tags and not (sorted(node.tags) == sorted(config.tags)):
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -88,10 +88,8 @@ def select_nodes_ids_by_intersection(nodes: dict[str, DbtNode], config: Selector
     https://docs.getdbt.com/reference/node-selection/yaml-selectors
     """
     selected_nodes = set()
-    for node_id, node in nodes.items():
-        if config.is_empty_config:
-            continue
-
+    if not config.is_empty_config:    
+        for node_id, node in nodes.items():
         if config.tags and not (sorted(node.tags) == sorted(config.tags)):
             continue
 
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -88,8 +88,8 @@ def select_nodes_ids_by_intersection(nodes: dict[str, DbtNode], config: Selector
     https://docs.getdbt.com/reference/node-selection/yaml-selectors
     """
     selected_nodes = set()
-    if not config.is_empty:
-        for node_id, node in nodes.items():
+    for node_id, node in nodes.items():
+        if not config.is_empty:
             if config.tags and not (sorted(node.tags) == sorted(config.tags)):
                 continue
 
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from pathlib import Path
+import copy
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


cosmos/dbt/selector.py
```diff
@@ -85,6 +85,9 @@ def select_nodes_ids_by_intersection(nodes: dict[str, DbtNode], config: Selector
     """
     selected_nodes = set()
     for node_id, node in nodes.items():
+        if not config.paths and not config.tags and not config.config:
```
tatiana, line 88: @jensenity What do you think of bringing this statement to the top of the method and using it as a condition before entering the `for`  loop? This would reduce the cyclomatic complexity from O(n) to O(1).

What do you think if we created a method in `SelectorConfig`, `is_empty`, which would return (`config.paths and config.tags and config.config`)?

Also, which select statement led to this uncovering this issue in production?

tatiana (approved):
Thank you very much, @jensenity! This looks great.
I left some minor comments inline. I would love your thoughts on them. Also, if you could rebase, we can aim to merge this change today.

cosmos/dbt/selector.py
jensenity: [0200aef](https://github.com/astronomer/astronomer-cosmos/pull/463/commits/0200aef4cc99c8a8d17bbf0d2f60583e8764bfa7) @tatiana is this better?

tatiana: It is, @jensenity , thank you! I think that the inclusion of `self.other` is breaking tests

jensenity: >Also, which select statement led to this uncovering this issue in production?

So I discovered this when I was testing the exclude test case. And when I ran the test case, for instance, if the test case is only selecting "is_child"..It doesnt not return on the child node, however, the grandparent & parent is all return
Screen Shot 2023-08-15 at 8 57 28 PM

`test_select_nodes_by_select_union_config_tag`
And for this test, it returns all 5 nodes. Thats how i discovered it.

tatiana: Thanks, @jensenity ! Happy to merge once the tests are passing.

jensenity: [9985698](https://github.com/astronomer/astronomer-cosmos/pull/463/commits/998569811e2f2ea8e964b0f94823e05f414d971e)

cosmos/dbt/selector.py
```diff
@@ -166,9 +169,10 @@ def select_nodes(
 
     nodes_ids = set(nodes.keys())
 
+    exclude_ids: set[str] = set()
     for statement in exclude:
         config = SelectorConfig(project_dir, statement)
-        exclude_ids = select_nodes_ids_by_intersection(nodes, config)
+        exclude_ids = exclude_ids.union(set(select_nodes_ids_by_intersection(nodes, config)))
```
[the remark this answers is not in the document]
jensenity: 🙇 
My pleasure.

cosmos/dbt/selector.py
```diff
@@ -43,6 +43,10 @@ def __init__(self, project_dir: Path, statement: str):
         self.other: list[str] = []
         self.load_from_statement(statement)
 
+    @property
+    def is_empty_config(self) -> bool:
+        return not self.paths and not self.tags and not self.config and not self.other
```
tatiana, line 48:
```suggestion
        return not (self.paths and self.tags and self.config and self.other)
```

cosmos/dbt/selector.py
```diff
@@ -85,6 +89,9 @@ def select_nodes_ids_by_intersection(nodes: dict[str, DbtNode], config: Selector
     """
     selected_nodes = set()
     for node_id, node in nodes.items():
+        if config.is_empty_config:
+            continue
+
```
tatiana, line 94: We don't need to iterate over all nodes and continue since the config is the same for all of them:

```suggestion
    if not config.is_empty_config:    
        for node_id, node in nodes.items():
```
This change will also require the rest of the "for" block to be indented.

cosmos/dbt/selector.py
```diff
@@ -43,6 +43,10 @@ def __init__(self, project_dir: Path, statement: str):
         self.other: list[str] = []
         self.load_from_statement(statement)
 
+    @property
+    def is_empty_config(self) -> bool:
```
tatiana, line 47: Since this property is within the config class, the method could be called `is_empty`.

cosmos/dbt/selector.py
```diff
@@ -85,26 +89,27 @@ def select_nodes_ids_by_intersection(nodes: dict[str, DbtNode], config: Selector
...
-            if not (config.config.items() <= supported_node_config.items()):
-                continue
+                # Remove 'tags' as they've already been filtered for
+                config.config.pop("tags", None)
+                supported_node_config.pop("tags", None)
 
-        if config.paths and not (set(config.paths).issubset(set(node.file_path.parents))):
-            continue
+                if not (config.config.items() <= supported_node_config.items()):
+                    continue
```
tatiana, line 107: After the first match, we'd lose this configuration when we iterate over the following nodes, and nodes that should not match, may match because of the missing of config. Suggestion:
```
                config_copy = copy.deepcopy(config.config)
                config_copy.pop("tags", None)
                supported_node_config.pop("tags", None)

                if not (config_copy.items() <= supported_node_config.items()):
                    continue
```
Sample code:
https://github.com/astronomer/astronomer-cosmos/commit/5b961a5590cc7f512ea8e8e3959c732687cc30e4

I ran:
```
hatch run tests.py3.8-2.6:test-cov
```

And things seem to work as expected 🤞

jensenity: [9138cf7](https://github.com/astronomer/astronomer-cosmos/pull/463/commits/9138cf7013df9153ac680b5de3c2f4fba7bdacb3)

tatiana (approved):
Thanks a lot for the contribution and for addressing all the feedback, @jensenity !

Issue #413, opened 2023-07-28
error when using `cache_selected_only` set to True


bawehkacrut:
When i set argument `cache_selected_only` to True, there is an error when running the actual dbt command. dbt command will be like this `dbt run --select<model name> --cache-selected-only`. based on documentation on dbt https://docs.getdbt.com/reference/global-configs/cache, the command should be `dbt --cache-selected-only run --select <model name>`

tatiana:
This issue was also raised by @jensenity in the slack channel: https://apache-airflow.slack.com/archives/C059CC42E9W/p1686028084902759

tatiana:
@bawehkacrut this is solved as part of the 1.1 release:
https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.1.0

Thanks for sharing this bug!

PR #469, opened 2023-08-15
Support dbt global flags (via `dbt_cmd_global_flags` in `operator_args`)

tatiana:
Allow users to define `dbt_cmd_global_flags` inside the `operator_args` dictionary to set global flags such as `--cache-selected-only`. These flags will be added between the dbt binary and the dbt subcommand being run.

This PR is not a breaking change since it introduces a new option/parameter.

An alternative implementation would be to hard-code the current 48 dbt (1.6) global flags into Cosmos source code, deciding from within Cosmos which of the flags given to `dbt_cmd_flags` are global or which are not. This alternative implementation would represent more maintainability to Cosmos developers and is less future-proof to newer versions of dbt.

Closes:  #413

### DIFF
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -55,6 +55,7 @@ class DbtBaseOperator(BaseOperator):
     :param dbt_executable_path: Path to dbt executable can be used with venv
         (i.e. /home/astro/.pyenv/versions/dbt_venv/bin/dbt)
     :param dbt_cmd_flags: List of flags to pass to dbt command
+    :param dbt_cmd_global_flags: List of dbt global flags to be passed to the dbt command
     """
 
     template_fields: Sequence[str] = ("env", "vars")
@@ -100,6 +101,7 @@ class DbtBaseOperator(BaseOperator):
         cancel_query_on_kill: bool = True,
         dbt_executable_path: str = "dbt",
         dbt_cmd_flags: list[str] | None = None,
+        dbt_cmd_global_flags: list[str] | None = None,
         **kwargs: Any,
     ) -> None:
         self.project_dir = project_dir
@@ -132,6 +134,7 @@ class DbtBaseOperator(BaseOperator):
         else:
             self.dbt_executable_path = dbt_executable_path
         self.dbt_cmd_flags = dbt_cmd_flags
+        self.dbt_cmd_global_flags = dbt_cmd_global_flags or []
         super().__init__(**kwargs)
 
     def get_env(self, context: Context) -> dict[str, str | bytes | os.PathLike[Any]]:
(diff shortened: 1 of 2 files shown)


cosmos/operators/base.py
```diff
@@ -100,6 +101,7 @@ def __init__(
         cancel_query_on_kill: bool = True,
         dbt_executable_path: str = "dbt",
         dbt_cmd_flags: list[str] | None = None,
+        dbt_cmd_global_flags: list[str] | None = None,
```
harels, line 104:
```suggestion
        dbt_global_flags: list[str] | None = None,
```

cosmos/operators/base.py
```diff
@@ -55,6 +55,7 @@ class DbtBaseOperator(BaseOperator):
     :param dbt_executable_path: Path to dbt executable can be used with venv
         (i.e. /home/astro/.pyenv/versions/dbt_venv/bin/dbt)
     :param dbt_cmd_flags: List of flags to pass to dbt command
+    :param dbt_cmd_global_flags: List of dbt global flags to be passed to the dbt command
```
harels, line 58:
```suggestion
    :param dbt_global_flags: List of dbt global flags to be passed to the dbt command
```

cosmos/operators/base.py
```diff
@@ -132,6 +134,7 @@ def __init__(
         else:
             self.dbt_executable_path = dbt_executable_path
         self.dbt_cmd_flags = dbt_cmd_flags
+        self.dbt_cmd_global_flags = dbt_cmd_global_flags or []
```
harels, line 137:
```suggestion
        self.dbt_global_flags = dbt_global_flags or []
```

cosmos/operators/base.py
```diff
@@ -210,6 +213,8 @@ def build_cmd(
     ) -> Tuple[list[str | None], dict[str, str | bytes | os.PathLike[Any]]]:
         dbt_cmd = [self.dbt_executable_path]
 
+        dbt_cmd.extend(self.dbt_cmd_global_flags)
```
harels, line 216:
```suggestion
        dbt_cmd.extend(self.dbt_global_flags)
```

tests/operators/test_local.py
```diff
@@ -48,7 +48,24 @@ def test_dbt_base_operator_add_user_supplied_flags() -> None:
...
+    assert cmd[-1] == "--full-refresh"
+
+
+def test_dbt_base_operator_add_user_supplied_global_flags() -> None:
+    dbt_base_operator = DbtLocalBaseOperator(
+        profile_config=profile_config,
+        task_id="my-task",
+        project_dir="my/dir",
+        base_cmd=["run"],
+        dbt_cmd_global_flags=["--cache-selected-only"],
```
harels, line 61:
```suggestion
        dbt_global_flags=["--cache-selected-only"],
```

harels (approved):
Given that there are many global flags, this makes sense.

harels:
Can we add a followup PR to document this new param?

cosmos/operators/base.py
tatiana: @harels we spoke about this and I talked to @jlaneve about this to get a third opinion on the naming, and he voted for us to keep the originally proposed - so I won't be applying the rename.

PR #471, opened 2023-08-16
BigQuery keyfile_dict profile mapping use env vars for sensitive fields

jbandoro:
## Description
I noticed that if `GoogleCloudServiceAccountDictProfileMapping` is used for profile mapping, the full keyfile json is logged in the UI. See example below:
image

This PR hides the sensitive private key fields (`private_key` and `private_key_id`) as env vars so that they will not show up in the logs, see below:
image

Would appreciate feedback since unlike other profiles, the `secret_fields` here are nested within the `keyfile_json` field so I had to override `BaseProfileMapping.env_vars` here to make it work.
## Related Issue(s)

None

## Breaking Change?

None

## Checklist

### DIFF
--- a/cosmos/profiles/bigquery/service_account_keyfile_dict.py
+++ b/cosmos/profiles/bigquery/service_account_keyfile_dict.py
@@ -61,7 +61,7 @@ class GoogleCloudServiceAccountDictProfileMapping(BaseProfileMapping):
         else:
             keyfile_json_dict = json.loads(keyfile_json)
             if not isinstance(keyfile_json_dict, dict):
-                raise ValueError("keyfile_json cannot be loaded as a dict.")
+                raise CosmosValueError("keyfile_json cannot be loaded as a dict.")
 
         for field in self.secret_fields:
             value = keyfile_json_dict.get(field)
--- a/cosmos/profiles/bigquery/service_account_keyfile_dict.py
+++ b/cosmos/profiles/bigquery/service_account_keyfile_dict.py
@@ -53,8 +53,8 @@ class GoogleCloudServiceAccountDictProfileMapping(BaseProfileMapping):
 
     def transform_keyfile_json(self, keyfile_json: str | dict[str, str]) -> dict[str, str]:
         """
-        Transforms the keyfile_json param to a dict if it is a string, and
-        sets environment variables for the service account credentials.
+        Transforms the keyfile_json param to a dict if it is a string, and sets environment
+        variables for the service account json secret fields.
         """
         if isinstance(keyfile_json, dict):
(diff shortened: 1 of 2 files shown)


cosmos/profiles/bigquery/service_account_keyfile_dict.py
```diff
@@ -48,12 +53,27 @@ def profile(self) -> dict[str, Any | None]:
...
-            return keyfile_json
-        keyfile_json = json.loads(keyfile_json)
-        if isinstance(keyfile_json, dict):
-            return keyfile_json
+            keyfile_json_dict = keyfile_json
         else:
-            raise ValueError("keyfile_json cannot be loaded as a dict.")
+            keyfile_json_dict = json.loads(keyfile_json)
+            if not isinstance(keyfile_json_dict, dict):
+                raise ValueError("keyfile_json cannot be loaded as a dict.")
```
tatiana, line 64: Could we add a test case covering this scenario?

cosmos/profiles/bigquery/service_account_keyfile_dict.py
```diff
@@ -48,12 +53,27 @@ def profile(self) -> dict[str, Any | None]:
...
+
+        for field in self.secret_fields:
+            value = keyfile_json_dict.get(field)
+            if value is None:
+                raise CosmosValueError(f"Could not find a value in service account json field: {field}.")
```
tatiana, line 69: Could we add a test case covering this scenario?

tatiana (commented):
This looks great, @jbandoro ! Thanks a lot for your contribution.

Sorry for the misleading coverage comments (PR #768 should fix this).

Can we add tests covering the exceptions being raised? I'm happy to approve/merge the PR once this is done.

cosmos/profiles/bigquery/service_account_keyfile_dict.py
jbandoro: Added a test to check that the error is raised as expected.

cosmos/profiles/bigquery/service_account_keyfile_dict.py
jbandoro: Added a test covering this, and updated it raise a `CosmosValueError` to be consistent.

cosmos/profiles/bigquery/service_account_keyfile_dict.py
tatiana: Thanks for addressing this so quickly, @jbandoro !

tatiana (approved):
Thank you very much for addressing the feedback, @jbandoro , and fixing this issue.

PR #481, opened 2023-08-21
Run dbt deps before running the dbt ls command

### LINKED ISSUE #445: Run `dbt deps` when using `load_mode=LoadMode.DBT_LS`
As of Cosmos 1.0, when we run `dbt ls` as part of `load_mode=LoadMode.DBT_LS`, we don't pre-run `dbt deps`.

This would be a useful feature based on the community feedback, including this thread in the Airflow Community Slack.
https://apache-airflow.slack.com/archives/C059CC42E9W/p1691413162622129?thread_ts=1691398106.912769&cid=C059CC42E9W

The goal with this ticket is to run `dbt deps` before `dbt ls`.

DanMawdsleyBA:
## Description

Runs the dbt deps command before running the dbt ls command

## Related Issue(s)

Closes  #445

## Checklist


tatiana (commented):
This looks great. Thanks a lot for your first contribution, @DanMawdsleyBA ! I'm hopeful this could be part of the upcoming 1.1 release.

I can see two areas for improvement:
1. A test that represents the original issue this PR solves, that breaks before the change and passes after this modification. Ideally, it would highlight what happens when a profile depends on a specific dbt adapter that is not installed and which error would be raised with `dbt ls`.
2. Do you think all users would be happy for, during DAG parsing time, to have `dbt deps` installed? What would happen if they had a read-only filesystem? A possibility to address this could be to have a new parameter added to `RenderConfig`, something like `install_deps`, set to `True` by default, but users could opt out (@jlaneve  may have other ideas/suggestions)

Please do reach out if you'd like support writing the test.

DanMawdsleyBA:
One thing I have realised is that this fix won't work on systems that have the read only issue as by default the packages are stored in dbt project path/ packages . There is a variable that can be set in the dbt_project.yml `packages_install_path` which may help however I don't think there is a way to specify that via a flag or env_var (Although happy to be wrong on that). It might be possible to do some kind of project.yml update to that variable and override the current but feels a janky solution. Not sure if @harels or @tatiana do you have any better ideas?
I'm starting to feel it might be easier to just copy the whole dbt project into a new directory for read only systems and have everything point to that new directory.

tatiana:
> One thing I have realised is that this fix won't work on systems that have the read only issue as by default the packages are stored in dbt project path/ packages . There is a variable that can be set in the dbt_project.yml `packages_install_path` which may help however I don't think there is a way to specify that via a flag or env_var (Although happy to be wrong on that). It might be possible to do some kind of project.yml update to that variable and override the current but feels a janky solution. Not sure if @harels or @tatiana do you have any better ideas? I'm starting to feel it might be easier to just copy the whole dbt project into a new directory for read only systems and have everything point to that new directory.

Hi @DanMawdsleyBA , that's a very important point.
I refactored a bit the logic in the `dbt ls` and hopefully it will be able to address the issue of running `dbt deps` on project directories that were originally read only: #488

cosmos/dbt/graph.py
```diff
@@ -194,9 +219,12 @@ def load_via_dbt_ls(self) -> None:
                         for line in logfile:
                             logger.debug(line.strip())
 
-                if stderr or "Runtime Error" in stdout:
+                if stderr or "Runtime Error" or "Compilation Error" in stdout:
```
tatiana, line 222: It may be better to capture any "Error" (check #488), what do you think?

dev/dags/dbt/jaffle_shop/packages.yml
```diff
@@ -0,0 +1,3 @@
+packages:
```
tatiana, line 1: By adding this, the previously passing tests will already start failing - unless the `dbt deps` code is introduced, right?

tests/dbt/test_graph.py
```diff
@@ -251,6 +251,30 @@ def test_load_via_dbt_ls_with_invalid_dbt_path():
     assert err_info.value.args[0] == expected
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize("pipeline_name", ("jaffle_shop", "jaffle_shop_python"))
```
tatiana, line 255: Thanks for adding the test, @DanMawdsleyBA !
```suggestion
@pytest.mark.parametrize("pipeline_name", ("jaffle_shop"))
```
Since `jaffle_shop_python` doesn't have any `packages.yml` - which is also a valid use-case

cosmos/dbt/graph.py
DanMawdsleyBA: I have changed to just "Error"

dev/dags/dbt/jaffle_shop/packages.yml
DanMawdsleyBA: Yeah it works when the dbt deps code is added but if that was removed but packages kept it would fail. Could create another project with a package? Or just ensure when we have the package we have the dbt deps up to you. 

The test deliberately sets the dbt_deps to False such that it would fail due to package error.

tatiana: It looks good, @DanMawdsleyBA , thank you!

tatiana:
Hi @DanMawdsleyBA! Sorry for the delay. The PR https://github.com/astronomer/astronomer-cosmos/pull/488 was merged into the main. Any chance we could rebase this PR based on the latest changes? This way the changes you introduced (installing deps before running `dbt ls` would work on deployments where the dbt project directory read-only  🤞

DanMawdsleyBA:
> Hi @DanMawdsleyBA! Sorry for the delay. The PR #488 was merged into the main. Any chance we could rebase this PR based on the latest changes? This way the changes you introduced (installing deps before running `dbt ls` would work on deployments where the dbt project directory read-only 🤞

@tatiana I've rebased it now so have a look and yes fingers crossed it works. I think one of your other tests (test_example_dag) may need to be updated but don't think it has any direct impact on the change. The other one test_load_via_dbt_ls_with_runtime_error_in_stdout is just asserting a different error.

tatiana:
@DanMawdsleyBA thanks, we are super close! A few integration tests are failing, e.g.:
```
  - Unable to run the command due to the error:
  ?                ^
  + Unable to run dbt deps command due to the error:
  ?               ++ ^^ ++
    Some Runtime Error
```

If you'd like help fixing them, please add me as a contributor to your fork.

DanMawdsleyBA:
> @DanMawdsleyBA thanks, we are super close! A few integration tests are failing, e.g.:
> 
> ```
>   - Unable to run the command due to the error:
>   ?                ^
>   + Unable to run dbt deps command due to the error:
>   ?               ++ ^^ ++
>     Some Runtime Error
> ```
> 
> If you'd like help fixing them, please add me as a contributor to your fork.

I've added you as a collaborator hopefully you've received an invite

tatiana:
@DanMawdsleyBA We managed to get all the tests to pass 🎉  Let's see after the rebase. Once this is passing, we'll merge this. It will be released as part of 1.1.

tatiana (approved):
Happy for these changes to be merged, I could test it locally.

It will help many Cosmos users with DAGs with no nodes due to dependencies not being installed. We'll improve the coverage in a follow up PR.

PR #433, opened 2023-08-02
Ensure datasets are emitted from the right task

abhi12mohan:
## Description

Converts "emit_datasets" into the "emit_datasets_from" Enum, with corresponding support for emitting datasets from the model based on whether tests run.

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

Issue #305, opened 2023-06-01
Improve Dataset URIs


tatiana:
Cosmos, at the moment, has a naive approach to creating Dataset URIs:

https://github.com/astronomer/astronomer-cosmos/blob/2aacff5d33fc729614c6ca88095a068646c9e844/cosmos/providers/dbt/core/utils/data_aware_scheduling.py#L4-L5

We should revisit this and adopt the patterns described at:
https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md

mobuchowski:
Yes - the dataset names should represent database that `dbt` is connecting to.

In the future, OL will provide dataset classes that will ensure that dataset name is constructed up to spec.

tatiana:
As of 1.0, Cosmos implements datasets only from an outlet perspective (it doesn't implement inlets). Since the output of seeds, snapshots, and models are "tables," this is the only type of dataset we need to worry about. In other words, we don't need to worry about creating dataset URIs for files in the meantime.

As of 1.6, dbt allows tables to be materialized in four different ways:
- table
- view
- CTE
- increment

This means Cosmos should not output datasets for resources that generate CTE (materialized=ephemeral). We also should not render these nodes as Airflow tasks.

A second part of this ticket is the Dataset URI definition itself. For most of the currently supported profiles, we'll need the following information to create a Dataset URI:
i. database type (e.g. Postgres, BigQuery, Redshift, MSSQL)
ii. database (e.g. "public")
iii. schema (e.g. "myschema")
iv. table name (dbt resource name)

We already have (iv) as part of the `DbtNode` definition.

Regarding the schema and database, they can come from three different sources:

1) In the SQL/Python resource file itself, such as:
```
{{ config(schema='marketing') }}

select ...
```

2) As part of the profile file, for example:
```
jaffle_shop:
  target: dev
  outputs:
    dev:
      type: postgres
      host: localhost
      user: postgres
      password: postgres
      port: 5432
      dbname: public
      schema: dbt_example
      threads: 4
```

3) Within the `dbt_project.yml` file, such as:
```
models:
  my_project:
    marketing:
      +schema: marketing
```

To retrieve the schema (iii) or database (ii), we can do the following:
a)  Enrich the `DbtNode` definition to contain the schema and database. If the project is loaded with `dbt ls` or `manifest`, and if the user sets this information in the `dbt_project.yml` or in the model itself, we solve the problem.
[...]

PR #485, opened 2023-08-22
Fix Dataset URI by using openlineage-integration-common as lib

tatiana:
Use `openlineage-dbt` to create outlets Dataset URIs from within Cosmos.

Closes: #305
Closes: #497
Closes: #433 (only emits outlet events from the model - the same behaviour as openlineage-dbt)

Validation
This change was tested by running Marquez locally and triggering the dag `basic_cosmos_dag` using Airflow 2.6.1 and Python 3.10.10.

The output generated by this version of Cosmos can be seen in the following screenshots:

Screenshot 2023-09-04 at 22 08 32

Screenshot 2023-09-04 at 22 09 01

Screenshot 2023-09-04 at 22 09 31

Tasks
- [x] Fix pre-commit checks
- [x] Add test
- [x]  Inlets support
- [x]  ~Emit open lineage events~
- [x]  ~Support Docker/K8s~ (deverred to issue: #496)
- [x]  ~Create a PR on openlineage-dbt to remove the dbt dependency~ Not needed since the code depends on `openlineage-integration-common` and not `openlineage-dbt` ([more info](https://github.com/OpenLineage/OpenLineage/blob/81372ca2bc2afecab369eab4a54cc6380dda49d0/integration/common/setup.py#L15))
- [x]  Understand which dataset is being emitted from test tasks (only inlets, no outlets)


tests/operators/test_local.py
```diff
@@ -124,6 +139,68 @@ def test_dbt_base_operator_get_env(p_context_to_airflow_vars: MagicMock) -> None
     assert env == expected_env
 
 
+# TATI IS HERE
```
tatiana, line 142: Remove this comment 🙈

cosmos/operators/local.py
```diff
@@ -132,9 +155,10 @@ def run_subprocess(self, *args: Any, **kwargs: Any) -> FullOutputSubprocessResul
 
     def run_command(
         self,
-        cmd: list[str | None],
+        cmd: list[str],
         env: dict[str, str | bytes | os.PathLike[Any]],
         context: Context,
+        emit_openlineage: bool = False,
```
mobuchowski, line 161: Why `emit_openlineage`? 

If OpenLineage integration exists and is properly configured, it will call `get_openlineage_facets_on_complete` itself. I don't think there's need to have additional explicit boolean check for it, although I don't fully know the context.

cosmos/operators/local.py
```diff
@@ -5,14 +5,34 @@
...
+try:
+    from airflow.providers.openlineage.extractors.base import OperatorLineage
+except ImportError:
+    from openlineage.airflow.extractors.base import OperatorLineage
+
+try:
+    from openlineage.common.provider.dbt.local import DbtLocalArtifactProcessor
+    from airflow.datasets import Dataset
+except ModuleNotFoundError:
+    is_open_lineage_available = False
```
JDarDagran, line 29: small thing but `is_openlineage_available` to be consistent with naming

cosmos/operators/local.py
tatiana: By default, it seems the common processor would raise an exception if it is an unsupported dbt node:
https://github.com/OpenLineage/OpenLineage/blob/main/integration/common/openlineage/common/provider/dbt/processor.py#L141

I'll change our code to set:
```
    should_raise_on_unsupported_command = False
```

cosmos/operators/local.py
```diff
@@ -189,13 +220,97 @@ def run_command(
...
+        Return a list of RunEvents
+        """
+        # Since openlineage-integration-common relies on the profiles definition, we need to make these newly introduced
+        # environment variables to the library. As of 1.0.0, DbtLocalArtifactProcessor did not allow passing environment
+        # variables as an argument, so we need to inject them to the system environment variables.
+        for key, value in env.items():
+            os.environ[key] = str(value)
+
+        openlineage_processor = DbtLocalArtifactProcessor(
+            producer="https://github.com/astronomer/astronomer-cosmos/",
```
harels, line 242: should this come from project consts?

harels (approved):
looks good, thanks @tatiana.
Left some comments.
2 questions: 
1. do we have tests that verify the output of the openlineage events themselves? 
2. do we need to update docs as well?

tatiana:
Thanks a lot for the review and feedback, @harels @mobuchowski and @JDarDagran ! I've addressed most of the feedback. As @harels mentioned, one point that is missing is to have a better automated testing strategy to make sure Cosmos is emitting the open lineage events correctly. I've logged a follow up ticket for us to improve this: #509 509

PR #487, opened 2023-08-23
Add Airflow 2.7 to the testing matrix

tatiana:
Make sure Cosmos is compliant with the recently released Apache Airflow 2.7.


harels (approved):
good idea to start testing 2.7 now

PR #489, opened 2023-08-24
Parse DAGs with no connections (offline)

jlaneve:
## Description

Cosmos requires connections to be defined when using the `dbt_ls` parsing method. This is not ideal for 2 reasons:
1. Requiring connections for parsing DAGs is generally bad Airflow practice, as it requires reaching out to a secrets backend/Airflow metadata DB on every parse
2. Connections aren't set up in all environments, i.e. CI/CD

This PR fixes the issue by generating a mock profile when Cosmos is parsing (not executing) the DAG. This works for all but the automatic_profile_mapping, which requires a live connection.

## Checklist

### DIFF
--- a/cosmos/profiles/base.py
+++ b/cosmos/profiles/base.py
@@ -144,7 +144,7 @@ class BaseProfileMapping(ABC):
             logger.info("Using mock values for profile %s")
         else:
             profile_vars = self.profile
-            logger.info("Using real values for profile %s")
+            logger.info("Using real values for profile %s", profile_name)
 
         # filter out any null values
         profile_vars = {k: v for k, v in profile_vars.items() if v is not None}
--- a/cosmos/profiles/base.py
+++ b/cosmos/profiles/base.py
@@ -141,7 +141,7 @@ class BaseProfileMapping(ABC):
         """
         if use_mock_values:
             profile_vars = self.mock_profile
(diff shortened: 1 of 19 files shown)


cosmos/profiles/base.py
```diff
@@ -107,11 +133,18 @@ def env_vars(self) -> dict[str, str]:
...
+    def get_profile_file_contents(
+        self, profile_name: str, target_name: str = "cosmos_target", use_mock_values: bool = False
+    ) -> str:
         """
         Translates the profile into a string that can be written to a profiles.yml file.
         """
-        profile_vars = self.profile
+        if use_mock_values:
+            profile_vars = self.mock_profile
+            logger.info("Using mock values for profile %s")
```
harels, line 144:
```suggestion
            logger.info("Using mock values for profile %s", profile_name)
```

cosmos/profiles/base.py
```diff
@@ -107,11 +133,18 @@ def env_vars(self) -> dict[str, str]:
...
+        else:
+            profile_vars = self.profile
+            logger.info("Using real values for profile %s")
```
harels, line 147:
```suggestion
            logger.info("Using real values for profile %s", profile_name)
```

harels (approved):
made comments, applied changes and fixed tests. approving pending all tests passing

PR #490, opened 2023-08-24
drop support for python 3.7 in tests

harels:
## Description

Python 3.7 is EOL, dropping support and removing from tests.

## Breaking Change?

partially, support is dropped for an older python version.

## Checklist


tatiana (approved):
Thanks, @harels ! Sorry for the unintended side effects of the previous change

Issue #476, opened 2023-08-18
Logs are not properly propagated on 1.0.5 on Airflow


nacpacheco:
Upon testing 1.0.5 I noticed that logs from cosmos operators are no longer visible on 1.0.5 on Airflow.
I have these settings:

airflow version: 2.6.3
dbt-core: 1.5.2
execution mode: local

My Airflow logs are outputed to S3 [using](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/logging/s3-task-handler.html#enabling-remote-logging)

Speaking to @tatiana on slack on this [thread](https://apache-airflow.slack.com/archives/C059CC42E9W/p1692198850850209) she pointed out that this [propagate flag](https://github.com/astronomer/astronomer-cosmos/blob/1af694c3f7b86eeae16c2a428445999ed364a58b/cosmos/log.py#L27) could be causing the issue

PR #498, opened 2023-08-30
Fix log propagation issue

tatiana:
We were not propagating logs, which led to people who had custom log handlers to miss Cosmos events.

Closes: #476


harels (approved):
good catch!

PR #502, opened 2023-08-31
Render vars when parsing DbtModel and appending upstream_models

### LINKED ISSUE #495: Render vars when parsing DbtModel
When parsing DbtModel if I have vars in sql file and I fail to get valid upstream_models. 
```
-- sql file
SELECT
    column_1,
	column_2,
FROM {{ ref('stg_users_'~ var('country_code')) }}
```  
```
-- model
- name: stg_users_{{ var("country_code") }}
  columns:
	...
```

For instance, if I have sql file and model file as above during `__post_init__` of `DbtModel` method, the referenced model is not added to config.upstream_models since it is not `jinja2.nodes.Const` type (If I have vars it is `Concat` type). I use vars in ref because I define single model configuration for multiple country codes.

```
-- something like this
 Concat(nodes=[Const(value='stg_users_'), Call(node=Name(name='var', ctx='load'), args=[Const(value='country_code')], kwargs=[], dyn_args=None, dyn_kwargs=None)]))
```
https://github.com/astronomer/astronomer-cosmos/blob/56d38b9a38ccc7fbe17cdcb2debb71e04f518c8b/cosmos/dbt/parser/project.py#L166-L175

So is it possible to receive vars in `DbtModel` and render ref value and add it to upstream_models? I want something like the following logic added

```
if isinstance(first_arg, jinja2.nodes.Concat):
    value = ''
    for node in base_node.args[0].nodes:
        if isinstance(node, jinja2.nodes.Const):
            value += node.value
        elif isinstance(node, jinja2.nodes.Call):
            if node.node.name == 'var':
                value += self.task_args[node.args[0].value]
```

dojinkimm:
## Description

When parsing DbtModel if I have vars in sql file, I fail to get valid upstream models (vars were not rendered as expected). This is because when add a model to upstream_models only `Const` types are allowed and `Concat` types are skipped.
```
-- sql file
SELECT
    column_1,
	column_2,
FROM {{ ref('stg_users_'~ var('country_code')) }} # ex) `stg_users_us` is added to upstream_models
```

I passed operator_args and used `vars` from operator_args when adding upstream_models. if variable exist but it is not defined in `vars` it raises KeyError and skips as before. If correct value exist for `vars` it the value is added up to upstream_model 

## Related Issue(s)
closes #495 

## Checklist


tatiana (approved):
Hi @dojinkimm , this is looking great. Thanks for the fix and the test!
Out of curiosity, why are you using `load_method=LoadMode.CUSTOM`?

dojinkimm:
> Out of curiosity, why are you using load_method=LoadMode.CUSTOM?

I am currently using astronomer-cosmos 0.6.8 due to some package conflicts in my environment, and load_method does not exist in that version. So, I was using default parsing logic

tatiana:
Hi @dojinkimm , I see - is there any chance we could help you with the dependency conflict? Please feel free to log a ticket.
We were aiming to release this as part of 1.1.1 - and I'd love to see if we can help you move to the newer version.

Issue #493, opened 2023-08-25
dbt with kubernetes


rtooker:
Similar to #491  but I think subtly different.

I'm using 
- apache-airflow==2.6.3
- astronomer-cosmos==1.0.5

DbtTaskGroup has a mandatory `profile_config` arg, but when set this seems to be passed through to the DbtRunKubernetesPodOperator causing in exception

```pyton
with DAG(
    Path(__file__).stem,
    description="*****",
    schedule=None,
    start_date=datetime(2023,1,1),
    catchup=False,
) as dag:
    dbt_task_group = DbtTaskGroup(
        project_config=ProjectConfig(
            dbt_project_path = Path(__file__).parent / 'dbt',
            manifest_path=Path(__file__).parent / 'manifests' / 'manifest.json',
        ),
        render_config=RenderConfig(
            load_method=LoadMode.DBT_MANIFEST,
        ),
        execution_config=ExecutionConfig(
            execution_mode=ExecutionMode.KUBERNETES,
        ),
        profile_config=ProfileConfig(
            profile_name='*****',
            target_name='*****'
        ),
        operator_args={
           ...
        }
    )



if __name__ == "__main__":
    dag.test()

```

yields
```
airflow.exceptions.AirflowException: Invalid arguments were passed to DbtRunKubernetesOperator (task_id: ******). Invalid arguments were:
**kwargs: {'profile_config': ProfileConfig(profile_name='****', target_name='****', profiles_yml_filepath=None, profile_mapping=None)}
```

adityasharma1685:
After moving my DBT project under Airflow Dags folder, I am getting same error.

david-mag:
I've got the same issue. The step-by-step guide in the documentation doesn't look like its up to date either. It uses imports (`from cosmos.providers.dbt.task_group import DbtTaskGroup`) that don't work with the latest version of cosmos anymore.

david-mag:
I've added a pull request for an attempt to fix this issue [here](https://github.com/astronomer/astronomer-cosmos/pull/505)

rtooker:
Thanks @david-mag . 
I'm now getting the same issue as reported in  https://github.com/astronomer/astronomer-cosmos/pull/505#issuecomment-1711707263 

```
airflow.exceptions.AirflowException: Invalid arguments were passed to DbtRunKubernetesOperator (task_id: XXXX). Invalid arguments were:
**kwargs: {'emit_datasets': True}
```

[...]

PR #505, opened 2023-09-02
Add profile_config arg to DbtKubernetesBaseOperator

david-mag:
## Description

When creating `DbtTaskGroup`s and `DbtDag`s, a `ProfileConfig` needs to be provided, since it's a required positional argument of the `DbtAirflowConverter` class. However, when providing a `profile_config` in `execution_mode="kubernetes"`, the following Exception is thrown:
```
AirflowException: Invalid arguments were passed to DbtRunKubernetesOperator (task_id: my_first_dbt_model_run). Invalid arguments were:
**kwargs: {'profile_config': ProfileConfig(profile_name='jaffle_shop', target_name='dev', profiles_yml_filepath='jaffle_shop/profiles.yml', profile_mapping=None)}
```

In my humble attempt of a solution, I make the `profile_config` an optional argument of the `DbtKubernetesBaseOperator` class (mainly to satisfy the tests ... since its a required argument in the previous part of the "pipeline", having it being a required argument would make sense imho) and parse it inside the `build_kube_args` method (similar, but not identically) to how it is done in the `run_command` method of the `DbtLocalBaseOperator` class.

This solution fixes the thrown exception and has the added benefit of giving the user that want to run dbt inside kubernetes the option to specify `--profile` and `--target`, which is a useful feature when wanting to schedule the same model in airflow, pointing to dev and prod targets.

I've tested this solution in my GCC environment.

## Related Issue(s)

This PR resolves the issue in issue #493

## Breaking Change?

This should fix something that is broken atm, as far as I can see.

## Checklist


cosmos/operators/kubernetes.py
```diff
@@ -59,16 +59,14 @@ def build_kube_args(self, context: Context, cmd_flags: list[str] | None = None)
         # to add that in the future
         self.dbt_executable_path = "dbt"
         dbt_cmd, env_vars = self.build_cmd(context=context, cmd_flags=cmd_flags)
-        
+
         # Parse ProfileConfig and add additional arguments to the dbt_cmd
-        if self.profile_config.profiles_yml_filepath:
-            with self.profile_config.ensure_profile() as (profile_path, _):
-                dbt_cmd.extend(["--profiles-dir", str(profile_path.parent)])
```
david-mag: Since `profile_yml_filepath` is tested against the local airflow `profiles.yml` file's location, and the docker image's environment most likely will differ, including this argument makes little sense.

jlaneve (approved):
nice, this looks great! we should add the same support in the Docker base operator. I've filed https://github.com/astronomer/astronomer-cosmos/issues/512 as a follow up!

agneku:
Hi @david-mag, @jlaneve,  I'm now trying to create a DbtDag with execution_mode='kubernetes'. Even though I don't receive the profile_config warning anymore (since the 1.1.0 release), I got a new warning: 
`airflow.exceptions.AirflowException: Invalid arguments were passed to DbtRunKubernetesOperator (task_id: final_model2_run). Invalid arguments were:
**kwargs: {'emit_datasets': True}`

The emit_datasets arg is not something I set on my own. Here's the DbtDag for reference:
`transform_data = DbtDag(
    project_config=ProjectConfig(DBT_PROJECT_PATH),
        execution_mode=ExecutionMode.KUBERNETES,
    ),
    operator_args={
        "queue": "kubernetes",
        "image": DBT_IMAGE,
        "image_pull_policy": "Always",
        "get_logs": True,
        "is_delete_operator_pod": False,
    },
    dag_id = "dbt_test",
    start_date=datetime(2023, 8, 1),
    schedule=None,
    catchup=False,
)`

is this a related issue?

david-mag:
As far as I can see, with the release of this fix, more "unknown" arguments have been added, which broke the kubernetes DAGs and TaskGroups again right away.

qimumu9406:
@david-mag I also encountered the same problem. If using cosmos on kubernetes, it seems this is a fatal problem.

PR #507, opened 2023-09-04
Update selector.py closes #506

### LINKED ISSUE #506: Not all selectors are allowed when using manifest load mode.
I have dag group I am creating using following DbtTaskGroup:
```
    DbtTaskGroup(
        group_id="dbt-core",
        project_config=ProjectConfig(
            DBT_ROOT_PATH / "test",
            manifest_path=DBT_ROOT_PATH  / "target"/ "manifest.json"
        ),

        render_config=RenderConfig(
            select=[selector],
            exclude=["config.materialized:ephemeral"],
            test_behavior=TestBehavior.AFTER_EACH,
            dbt_deps=True
        ),
        operator_args={"install_deps": True},
        profile_config=profile_config
    )
```
I then for selector when I pass only model name, or use 1+ModelName it throws error. When I used ls mode it was allowed and was only giving warnings. This selector should be treated as allowed one since dbt can select appropriate models based on this.

Culprit for this behavious is here in Selector.py:
https://github.com/astronomer/astronomer-cosmos/commit/dd553dcf8d6e71e33a17e866dff6156dbd9e204d

Tarjei400:
## Description

Only log from not supported filter, utill you figure how you would like to validate incomming selectors.
This PR at least allows selectors that are not taken into account by a validator like in case of LS LoadMode

## Checklist

- [-] I have made corresponding changes to the documentation (if required)
- [-] I have added tests that prove my fix is effective or that my feature works - tested on my own airflow instance

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -160,7 +160,7 @@ def select_nodes(
             elif any([filter_parameter.startswith(CONFIG_SELECTOR + config + ":") for config in SUPPORTED_CONFIG]):
                 continue
             else:
-                raise CosmosValueError(f"Invalid {filter_type} filter: {filter_parameter}")
+                logger.warning(f"Invalid {filter_type} filter: {filter_parameter}")
 
     subset_ids: set[str] = set()
 


odanu (approved):
Good catch 👍

tatiana:
Hey @Tarjei400, would you like to move forward with this change? Just noticed the PR was prematurely closed!

Tarjei400:
@tatiana I closed pr, as it turned out my change was utterly wrong, seems that cosmos has its own selector implementation if mode other than DBT_LS is used, hence that change would be wrong. The only hope of using selectors at the moment is by using DBT_LS mode, although scanning of dags gets severly extended due to this, but I was unable to find any other way.

tatiana:
Thank you very much @Tarjei400. Even though it wasn't merged, we appreciate your contribution!

If you could share some examples of selectors you're trying to use and did not work with Cosmos selectors, we could have dedicated issues - so they can be worked on.

tatiana:
Thank you very much @Tarjei400. Even though it wasn't merged, we appreciate your contribution!

If you could share some examples of selectors you're trying to use and did not work with Cosmos selectors, we could have dedicated issues - so they can be worked on.

PR #510, opened 2023-09-05
Enable type checks (MyPy) in the CI

tatiana:
## Description

We had to disable MyPy checks from the pre-commit hook because we were reaching a quota from the pre-commit service:
```
build of https://github.com/pre-commit/mirrors-mypy:types-PyYAML,types-attrs,attrs,types-requests,types-python-dateutil,apache-airflow@v1.5.1 for python@python3 exceeds tier max size 250MiB: 262.2MiB
```

As observed in https://github.com/astronomer/astronomer-cosmos/pull/482.

This PR re-enables MyPy checks but in the CI.


pyproject.toml
```diff
@@ -150,22 +154,23 @@ matrix.airflow.dependencies = [
...
 test = 'pytest -vv --durations=0 . -m "not integration" --ignore=tests/test_example_dags.py --ignore=tests/test_example_dags_no_connections.py'
-test-cov = 'pytest -vv --cov=cosmos --cov-report=term-missing --cov-report=xml --durations=0 -m "not integration" --ignore=tests/test_example_dags.py --ignore=tests/test_example_dags_no_connections.py'
+test-cov = """pytest -vv --cov=cosmos --cov-report=term-missing --cov-report=xml --durations=0 -m "not integration" --ignore=tests/test_example_dags.py --ignore=tests/test_example_dags_no_connections.py"""
 # we install using the following workaround to overcome installation conflicts, such as:
 # apache-airflow 2.3.0 and dbt-core [0.13.0 - 1.5.2] and jinja2>=3.0.0 because these package versions have conflicting dependencies
 test-integration-setup = """pip uninstall dbt-postgres dbt-databricks; \
 rm -rf airflow.*; \
 airflow db init; \
-pip install 'dbt-postgres<=1.5.4' 'dbt-databricks<=1.5.4'"""
+pip install  'openlineage-airflow' 'dbt-postgres<=1.5.4' 'dbt-databricks<=1.5.4'"""
```
tatiana, line 165:
```suggestion
pip install 'dbt-databricks<=1.4' 'dbt-postgres<=1.4' 'dbt-sqlite<=1.4' openlineage-airflow"""
```

cosmos/operators/local.py
```diff
@@ -248,6 +251,12 @@ def calculate_openlineage_events_completes(
         for key, value in env.items():
             os.environ[key] = str(value)
 
+        lineage_namespace = os.getenv("OPENLINEAGE_NAMESPACE", "default")
```
harels, line 254: perhaps the namespace here could be cosmos?

tatiana: @harel we already set the producer URL to be cosmos-specific, would it also make sense to segregate the namespace depending on the tool used to process the data? I honestly don't know best lineage best practices and will follow what you recommend

harels: Ideally, the namespace is a logic grouping of the data itself.
I would encourage the users to define that with the ENV_VARS, but I think that if by default it shows up as cosmos, it helps users have a better understanding of where it's coming from and then change as this see fit.

harels: producer URL is pretty much internal implementation detail for the OL consumer

tatiana: Changed to `cosmos`, @harels !

PR #511, opened 2023-09-05
Fetch port from connection

### LINKED ISSUE #508: Retrieve Postgres port from connection when using `PostgresUserPasswordProfileMapping`
Cosmos <=1.0 `PostgresUserPasswordProfileMapping` enforces the port to be 5452, regardless of the value set in the Airflow connection. We should retrieve this value from the Airflow connection.

jlaneve:
## Description

This PR fixes a bug that was introduced in Cosmos 1.0: the port was misplaced in the dictionary, causing the port to _always_ be fixed.

## Related Issue(s)

closes #508

## Checklist


harels (approved):

tatiana (commented):
Thanks, @jlaneve, this looks good - thanks for being so quick at writing a fix for it. What do you think if we added a test to confirm that if a user set as part of the Airflow connection a different port, the Profile will contain that custom port?

PR #513, opened 2023-09-06
Change operators to log ``dbt`` commands output as opposed to recording to XCom

tatiana:
All Cosmos operators were dumping unnecessary data to XCom.
Change the behavior to log the information as opposed to polluting XCom.

Closes: #304

### DIFF
--- a/cosmos/operators/docker.py
+++ b/cosmos/operators/docker.py
@@ -52,6 +52,10 @@ class DbtDockerBaseOperator(DockerOperator, DbtBaseOperator):  # type: ignore[mi
         self.environment = {**env_vars, **self.environment}  # type: ignore[has-type]
         self.command = dbt_cmd
 
+    def execute(self, context: Context) -> None:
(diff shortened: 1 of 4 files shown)


cosmos/operators/docker.py
```diff
@@ -52,6 +52,10 @@ def build_command(self, context: Context, cmd_flags: list[str] | None = None) ->
         self.environment = {**env_vars, **self.environment}  # type: ignore[has-type]
         self.command = dbt_cmd
 
+    def execute(self, context: Context) -> None:
+        result = self.build_and_run_cmd(context=context)
+        logger.info(result.output)
```
jlaneve, line 57: wdyt about adding the `logger.info` in `build_and_run_cmd()`? it seems we log the output every time the function is called, so this could help with code duplication

tatiana: Good idea, I'll do this, @jlaneve !

harels (approved):

Issue #260, opened 2023-05-02
Add support dbt project render for models contract


jensenity:
Starting from dbt 1.5.0, models now supports both model contract and model version.
https://docs.getdbt.com/docs/collaborate/govern/model-contracts.

we can define just the model name with the version argument separately. 

```
models:
  - name: dim_customers
    latest_version: 1
    config:
      materialized: table
      contract: {enforced: true}
    columns:
      - name: customer_id
        description: This is the primary key
        data_type: int
      - name: country_name
        description: Where this customer lives
        data_type: varchar
    
    # Declare the versions, and highlight the diffs
    versions:
    
      - v: 1
        # Matches what's above -- nothing more needed
    
      - v: 2
        # Removed a column -- this is the breaking change!
        columns:
          # This means: use the 'columns' list from above, but exclude country_name
          - include: all
            exclude: [country_name]
```
https://docs.getdbt.com/docs/collaborate/govern/model-versions

But the rendering doesn't not fully support the versions in config yaml so there are some minor logic need to add in the DbtProject class while handling sql files and config files.

tatiana:
This discussion has been raised recently in the Airflow slack:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1693472093058619

PR #516, opened 2023-09-06
Add support render for versioned dbt models.

binhnq94:
## Description

[Model version](https://docs.getdbt.com/docs/collaborate/govern/model-versions) is supported by dbt-core >= 1.5.
As cosmos v1.1.0a1, it not support to render task for model have versions.  

In this PR I propose to use `node_dict["alias"]`  as `DbtNode.name` instead of `node_dict["name"]` if `node_dict["resource_type"]` == `model`.

This PR also help cosmos render the names of nodes to be correctly the same as `identifier` if users config alias or custom the `generate_alias_name` macro. Refer: [alias document](https://docs.getdbt.com/reference/resource-configs/alias) and [the `generate_alias_name` implementation](https://github.com/dbt-labs/dbt-core/blob/ada8860e48b32ac712d92e8b0977b2c3c9749981/core/dbt/include/global_project/macros/get_custom_name/get_custom_alias.sql#L26-L30).

## Related Issue(s)

#260 

## Breaking Change?

When model have version config, each model version will create a database relation with alias `<model_name>_v<v>`, for examples: `dim_customers_v1`,  `dim_customers_v2`... 

So I expected that cosmos will created run/test tasks for each versions, for example: dim_customers_v1_run, dim_customers_v2_run, dim_customers_v1_test, dim_customers_v2_test...

The second expectation is the dag will render right depedences, for example: stg_customers_v1 >> customers_v1,  stg_customers_v2 >> customer_v2...

## Checklist


tatiana (commented):
@binhnq94, thanks a lot for your contribution!

I will still read the docs to understand more about aliases and versions, but I have two initial questions:
1. Would there be any benefit in introducing a new attribute to `DbtNode` called `alias` instead of overloading the property `name` with two possible sources? We could also introduce an `identifier` if it helped. What are the pros and cons of overriding `name`?
2. When we render dbt nodes as Airflow nodes, what would be the benefit of rendering models using aliases instead of their names?

cosmos/dbt/graph.py
```diff
@@ -324,7 +327,9 @@ def load_from_dbt_manifest(self) -> None:
 
             for unique_id, node_dict in manifest.get("nodes", {}).items():
                 node = DbtNode(
-                    name=node_dict["name"],
+                    name=node_dict["alias"]
```
tatiana, line 330: Are aliases available in older versions of `dbt`, or would it make sense to use perhaps isolate the logic of naming a `DbtNode` in a method and use `node_dict.get("alias")`?

binhnq94: `alias` is required attributes of 2 resources `model` and `seed` from dbt v1.0.
Refer: [Manifest version 4](https://schemas.getdbt.com/dbt/manifest/v4/index.html#nodes_additionalProperties_oneOf_i12_alias) and [Manifest Json file document](https://docs.getdbt.com/reference/artifacts/manifest-json).

binhnq94:
> @binhnq94, thanks a lot for your contribution!
> 
> I will still read the docs to understand more about aliases and versions, but I have two initial questions:
> 
> 1. Would there be any benefit in introducing a new attribute to `DbtNode` called `alias` instead of overloading the property `name` with two possible sources? We could also introduce an `identifier` if it helped. What are the pros and cons of overriding `name`?
> 2. When we render dbt nodes as Airflow nodes, what would be the benefit of rendering models using aliases instead of their names?

1. No, because if you use model versions in dbt project, name is not unique any more. `dbt run with --models <name>` will build all model versions leading to select not work as expected.
2. Use alias, cosmos will render the DAG that have the same lineage as in the docs created by `dbt docs generate` cli. . It help users to double check easier and it support model versions.

dev/dags/dbt/model_version/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
tatiana, line 1: We currently have the license. here:
https://github.com/astronomer/astronomer-cosmos/blob/main/LICENSE

So we don't need this one

dev/dags/dbt/model_version/README.md
```diff
@@ -0,0 +1,11 @@
+## `jaffle_shop`
```
tatiana, line 1: Could we have a smaller dbt sample project to illustrate the version feature, or do we need to stick to the jaffle shop one?

tests/dbt/test_graph.py
```diff
@@ -312,3 +313,40 @@ def test_load_via_load_via_custom_parser(pipeline_name):
     assert dbt_graph.nodes == dbt_graph.filtered_nodes
     # the custom parser does not add dbt test nodes
     assert len(dbt_graph.nodes) == 8
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("load_method", ["load_via_dbt_ls", "load_from_dbt_manifest"])
+def test_load_dbt_ls_and_manifest_with_model_version(load_method):
```
tatiana, line 320: @binhnq94, what do you think we are having some tests to confirm  `dbt run with --models <name>` runs as expected?
ATM, we don't have a test matrix for `dbt`, but this type of change makes me wonder if we should.

dev/dags/dbt/model_version/LICENSE
binhnq94: Ok.

dev/dags/dbt/model_version/README.md
binhnq94: It big because, I want to make sure this PR can cover the use case:
> The second expectation is the dag will render right depedences, for example: stg_customers_v1 >> customers_v1, stg_customers_v2 >> customer_v2...
Oh, it is lack of test for those dependencies. => Will add soon.

About stick to the `jaffle_shop` idea, I think we will get dbt version problem for existing tests? Because this feature only work with dbt>=1.6.0. It is my opinion, if I am wrong, please correct me :'(

tests/dbt/test_graph.py
binhnq94: Do you give me sample kind of tests ? I will add it as the same.

tatiana: @ binhnq94 BTW: I suggest we do this in a separate PR. You've already done great with the alias support here!

I thought about something along the lines of what we did here, but without extending the `test` environment, but using `test_dbt` or something similar:
https://github.com/astronomer/astronomer-cosmos/blob/f2dd7e129423a5a4ed0000835951809a74e08577/pyproject.toml#L142-L153

We could decorate a subset of relevant tests with this "dbt" marker, running only the marked dbt tests using something similar to what we did for the integration tests, validating behaviours that will change depending on the version of dbt (such as node name selection):
https://github.com/astronomer/astronomer-cosmos/blob/main/pyproject.toml#L181

tatiana:
@binhnq94 let's aim to have this change released as part of 1.2. Please, let me know if you'd like any further support on anything - I'm looking forward to merging this change.

binhnq94:
@tatiana The current cosmos suport dbt<=1.5.4 not dbt >= 1.6.0.
https://github.com/astronomer/astronomer-cosmos/blob/main/pyproject.toml#L48
Should we need another PR for upgrade dbt version?

tests/dbt/test_graph.py
binhnq94: Yeah, I think we should mark some tests for dbt>=1.6.0 only.

tatiana:
Hi @binhnq94 , really sorry for not following up earlier. I'm hopeful we'll get this change merged this or next week so that it can be part of the 1.2 release.

We just merged a PR with the dbt 1.6 support: #564

Could you rebase this branch, please? Please do let us know if you'd like any support to get this to the finishing line.

binhnq94:
> Could you rebase this branch, please? Please do let us know if you'd like any support to get this to the finishing line.

I rebased this branch. Test-integration is ok in laptop.

binhnq94:
@tatiana I fixed the error: improper relation name (too many dotted names) in [commit](https://github.com/astronomer/astronomer-cosmos/pull/516/commits/62ec55396e8b3a448ac5165951b9383e6fc312d1)
by deleting the macro file `drop_table.sql`.
So why we need custom drop_table macro? The macro not work with dbt-postgres @@

binhnq94:
@tatiana I have already fixed the fail in integrate tests (sorry, last time I only run model_version dag only.)
image
Let trigger integration tests :'(

tatiana (approved):

tatiana:
This was released as part of 1.2:
* https://pypi.org/project/astronomer-cosmos/1.2.0/
* https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.2.0
cc @binhnq94

PR #524, opened 2023-09-08
Remove prefix of run and test tasks if test_behavior = TestBehavior.AFTER_EACH

binhnq94:
## Description

If test_behavior = TestBehavior.AFTER_EACH, DbtNode.name in task_id is not necessary because the parent task group is named as `DbtNode.name`

## Related Issue(s)

[Slack thread](https://apache-airflow.slack.com/archives/C059CC42E9W/p1692776042134929)

## Breaking Change?

Remove prefix of run and test tasks if test_behavior = TestBehavior.AFTER_EACH.

Olds rendered tasks before this PR:
image

New rendered tasks:
image

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -2,16 +2,15 @@ from __future__ import annotations
 
 from typing import Any, Callable
 
+from airflow.models import BaseOperator
 from airflow.models.dag import DAG
 from airflow.utils.task_group import TaskGroup
 
-from cosmos.constants import DbtResourceType, TestBehavior, ExecutionMode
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Thanks for pointing out the redundancy in the `TaskGroup` and the inner `Task` names - this looks cleaner! 

I just checked how it looked in the bigger view of the DAG, and it makes more sense:

Screenshot 2023-09-11 at 15 14 27

Screenshot 2023-09-11 at 15 14 51

I also confirmed that things looked good in the Grid and Task Tries tabs.

PR #529, opened 2023-09-12
Fix  DbtRunOperationLocalOperator missing flags

tatiana:
@jensenity reported an issue in Slack related to Cosmos 1.1:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1694488079896779

"for some reason, the the dbtrunoperation operator doesn’t use the cmd_flags….. it didnt have the --args flags when i look into my logs."
```
DbtRunOperationOperator(
                            task_id="run_operation",
                            macro_name="stage_external_sources",
                            args=dict(select='source_a'),
                            project_dir=DBT_PROJECT_PATH,
                            install_deps=True,
                            env={
                                …
                            },
                            dbt_executable_path=DBT_EXECUTION_PATH,
                            queue="kubernetes",
                        )
```

This PR adds tests to all local operators to confirm if they are being executed with flags or not to avoid regressions.
It increases the project test coverage from 91.80% to 92.33%.

As a bonus, this PR also enables the `--full-refresh` flag when running dbt models, as requested by Monideep De in the Slack channel:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1694523446939829

This change makes sense since this flag is supported both in Seeds and Models (https://docs.getdbt.com/reference/resource-configs/full_refresh), but we only exposed it in Seeds.

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -485,6 +485,10 @@ class DbtRunOperationLocalOperator(DbtLocalBaseOperator):
             flags.append(yaml.dump(self.args))
         return flags
 
+    def execute(self, context: Context) -> None:
+        cmd_flags = self.add_cmd_flags()
+        self.build_and_run_cmd(context=context, cmd_flags=cmd_flags)
+
 
 class DbtDocsLocalOperator(DbtLocalBaseOperator):
     """
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -405,9 +405,20 @@ class DbtRunLocalOperator(DbtLocalBaseOperator):
     ui_color = "#7352BA"
     ui_fgcolor = "#F4F2FC"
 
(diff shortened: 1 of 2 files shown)


cosmos/operators/local.py
```diff
@@ -405,10 +405,21 @@ class DbtRunLocalOperator(DbtLocalBaseOperator):
...
+    def __init__(self, full_refresh: bool = False, **kwargs: Any) -> None:
+        self.full_refresh = full_refresh
         super().__init__(**kwargs)
         self.base_cmd = ["run"]
 
+    def add_cmd_flags(self) -> list[str]:
+        flags = []
+        if self.full_refresh is True:
+            flags.append("--full-refresh")
+        return flags
```
jensenity, line 417: nice! this would be helpful!!

harels (approved):
nice!

PR #532, opened 2023-09-13
fix possible typo in config.py

meyobagero:
## Description

Just appropriate argument to the error message in the function defined at line 73 of the config.py file
## Breaking Change?

Not a breaking change


tatiana (commented):
Thanks for the contribution, @meyobagero !
What do you think if we added a test to make sure we don't regress to having the incorrect path in the future?

tatiana (approved):

Issue #447, opened 2023-08-08
Improve cosmos logs to always include "astronomer-cosmos"


tatiana:
As of 1.0, there is no easy way of isolating logs created by Cosmos from other logs printed in Airflow.

Edgaras Navickas, a community member, proposed in [slack](https://apache-airflow.slack.com/archives/C059CC42E9W/p1691483255763189?thread_ts=1691398106.912769&cid=C059CC42E9W) that we improve this.

The suggestion is to prefix the project name.

Example of current log message:
```
Trying to parse the dbt project ……
```

Suggested log message:
```
[astronomer-cosmos] Trying to parse the dbt project ……
```

PR #537, opened 2023-09-14
Use Airflow task logger

jlaneve:
## Description

This PR updates the code within the Cosmos operators to use Airflow's `self.log` that gets exposed via the BaseOperator instead of the custom Cosmos logger. This ensures the task execution logs always show up properly in Airflow's task logs.

## Checklist

### DIFF
--- a/cosmos/__init__.py
+++ b/cosmos/__init__.py
@@ -5,7 +5,7 @@ Astronomer Cosmos is a library for rendering dbt workflows in Airflow.
 
 Contains dags, task groups, and operators.
 """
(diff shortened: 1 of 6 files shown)


tatiana (commented):
Hey @jlaneve, thank you!

Are there scenarios when the logs are not displayed in the operator/tasks logs? It would be great to see the steps to reproduce this.

An issue in Cosmos 1.0.x related to log propagation was fixed as part of the 1.1 (#476).

If we have to change the logging strategy, it would be great to try to keep this behaviour: #447

PR #542, opened 2023-09-15
Separate columns for dbt and Airflow links in the concept comparison Doc 

TJaniF:
## Description

This was a suggestion from @ReadytoRocc in Slack. cc: @tatiana :) 
The only downside is that the table is now side scroll... I don't know enough CSS to fix that 😅


tatiana (commented):
Thanks a lot for working on this, @TJaniF !

I just double-checked both in the GitHub UI and built the docs locally, and it seems the reference columns are not showing up:

Screenshot 2023-09-19 at 09 20 54

Screenshot 2023-09-19 at 09 22 35

Would you like help troubleshooting this?

To build the docs locally, you can use:
```
hatch run docs:build
```

TJaniF:
Hey @tatiana 

That the reference column is gone was intended, I took Rocco's suggestion and instead linked the references directly in the first two columns. 

But if you prefer a reference column I can quickly redo the table I had in an earlier commit. In that case do you want the full link link in the original or a hyperlinked word? :)

tatiana (approved):
Thanks, @TJaniF , much better to have the link in the first two columns, I completely missed it!

PR #543, opened 2023-09-15
[WIP] Fix only create task group and test task only if model has test

### LINKED ISSUE #531: Do not use task group and create a test task for a model that does not have tests
As of Cosmos 1.1, if `test_behavior=TestBehavior.AFTER_EACH` (default), we always convert models into a task group containing a model run task and a test task, regardless of actual tests for that model.

@DanMawdsleyBA mentioned this issue in https://github.com/astronomer/astronomer-cosmos/issues/519#issuecomment-1714399354  and also in slack:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1694606935594769?thread_ts=1694551447.136999&cid=C059CC42E9W

The goal of this ticket is to change the Airflow DAG generation, so we only create a task group and the test task if there are test nodes associated with the model of interest.

raffifu:
## Description

This PR will only create task group if `dbt` test exist by adding propery `has_test` on `DbtNode`. The test dependency will be updated after loading the dbt project into `DbtGraph`

## Related Issue(s)

closes #531

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works (Currently working on tests)


cosmos/airflow/graph.py
```diff
@@ -171,10 +171,14 @@ def build_airflow_graph(
             node=node,
             execution_mode=execution_mode,
             args=task_args,
-            use_name_as_task_id_prefix=test_behavior != TestBehavior.AFTER_EACH,
+            use_name_as_task_id_prefix=not (test_behavior == TestBehavior.AFTER_EACH and node.has_test),
```
tatiana, line 174: It feels we should perhaps rename this argument (`use_name_as_task_id_prefix`) to `should_use_task_group` or something similar since not only is it now affecting the name, but also the graph topology (single node x a task group with two nodes), what do you think?

cosmos/airflow/graph.py
```diff
@@ -171,10 +171,14 @@ def build_airflow_graph(
...
         )
         if task_meta and node.resource_type != DbtResourceType.TEST:
-            if node.resource_type == DbtResourceType.MODEL and test_behavior == TestBehavior.AFTER_EACH:
+            if (
+                node.resource_type == DbtResourceType.MODEL
+                and node.has_test is True
```
tatiana, line 179:
```suggestion
                and node.has_test
```

tatiana (commented):
This looks amazing, @raffifu , thank you very much for your contribution!
I added two minor comments, and we can aim to release this as part of 1.2.

cosmos/airflow/graph.py
raffifu: Thank you for the review, i agree that the argument must be changed because for some people the argument can be confusing. I think it's more clearly if we can add a property `should_use_task_group` on TaskMetadata. So we can simplify the branching on the code like this
```py
# Before Change
task_meta = create_task_metadata(...)
if task_meta and node.resource_type != DbtResourceType.TEST:
  if (
      node.resource_type == DbtResourceType.MODEL
      and node.has_test is True
      and test_behavior == TestBehavior.AFTER_EACH
  ):

# After change
task_meta = create_task_metadata(...)
if task_meta.should_use_task_group is True:
```

cosmos/airflow/graph.py
raffifu: I'll commit the suggestion with all revision

cosmos/airflow/graph.py
raffifu: Resolved on the newest commit

tatiana (approved):

tatiana:
This was released as part of 1.1.3 🎉 
https://pypi.org/project/astronomer-cosmos/1.1.3/
https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.1.3

PR #551, opened 2023-09-25
Adding emit_datasets param to DbtBaseOperator

pgoslatara:
## Description

When using the kubernetes execution mode I get this error:
image

This can be reproduced by following the README [here](https://github.com/pgoslatara/dynamic_task_mapping_for_dbt/blob/cosmos-to-k8s/README.md), the relevant DAG file is [here](https://github.com/pgoslatara/dynamic_task_mapping_for_dbt/blob/cosmos-to-k8s/dags/astronomer_cosmos_dag.py).

It appears that `emit_datasets` is not set in `DbtBaseOperator`, only in `DbtLocalBaseOperator`.

## Checklist

### DIFF
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -56,6 +56,8 @@ class DbtBaseOperator(BaseOperator):
         (i.e. /home/astro/.pyenv/versions/dbt_venv/bin/dbt)
     :param dbt_cmd_flags: List of flags to pass to dbt command
     :param dbt_cmd_global_flags: List of dbt global flags to be passed to the dbt command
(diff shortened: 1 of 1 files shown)


tatiana (commented):
Thanks a lot for the fix, @pgoslatara !

Would you be up for cherry-picking a test-case that covers this change so we avoid regression on this bug? I had created a test in 79f348b3e44360bf13d2ffcd368dcc208426c125. If you'd rather not cherry-pick and create your own test, that would be fine as well!

tatiana:
Thanks a lot for working on this, @pgoslatara ! I've merged your change as part of #554, since it includes tests and a few more bits. You are credited as a co-author!

PR #574, opened 2023-10-02
Replacing hardcoded port for a connection attr reference

### LINKED ISSUE #573: PostgresUserPasswordProfileMapping have hardcoded port
In `cosmos/profiles/postgres/user_pass.py` we have a `profile` property with a hardcoded port.

*profile property*
```python
        profile = {
            "port": 5432, # hardcoded port
            **self.mapped_params,
            **self.profile_args,
            # password should always get set as env var
            "password": self.get_env_var_format("password"),
        }
```

The port should be retrieved from the connection.

farovictor:
## Description

Replaced hardcoded port for a connection reference.

## Related Issue(s)

closes #573 

## Breaking Change?

No breaking changes.

## Checklist

### DIFF
--- a/cosmos/profiles/postgres/user_pass.py
+++ b/cosmos/profiles/postgres/user_pass.py
@@ -40,7 +40,7 @@ class PostgresUserPasswordProfileMapping(BaseProfileMapping):
     def profile(self) -> dict[str, Any | None]:
         "Gets profile. The password is stored in an environment variable."
         profile = {
-            "port": 5432,
+            "port": self.conn.port,
             **self.mapped_params,
             **self.profile_args,
             # password should always get set as env var


cosmos/profiles/postgres/user_pass.py
```diff
@@ -40,7 +40,7 @@ class PostgresUserPasswordProfileMapping(BaseProfileMapping):
     def profile(self) -> dict[str, Any | None]:
         "Gets profile. The password is stored in an environment variable."
         profile = {
-            "port": 5432,
```
tatiana, line 43: Hi @farovictor , thanks for your contribution!

There was a recent change which makes the connection information override the hard-coded default value, and we had a test for it: #511 

Could you confirm if this is not working and add a change to your PR if it is solving a different scenario?

farovictor: It does work. My bad, I was using a non-up-to-date version. Closing this PR.

PR #576, opened 2023-10-05
Ensure filtering with manifest loading works with single model

jlaneve:
## Description

This PR fixes an issue with Cosmos' custom filtering methods. Previously, when using the manifest loading method (and thus Cosmos' custom filtering), a user could not filter by path to a specific model (only a directory). This PR fixes that issue and makes some minor refactors to the selecting logic to make it more intuitive and performant.

## Breaking Change?

No breaking changes.

## Checklist


tatiana (commented):
Thanks for fixing and improving this, @jlaneve !

Happy to merge it once the checks pass.

harels (approved):
nice!

PR #581, opened 2023-10-09
Make the arg dbt_project_path in the ProjectConfig optional

### LINKED ISSUE #569: Make the arg `dbt_project_path` set in the `ProjectConfig` init optional
**Context**

Currently, users must declare the `dbt_project_path` argument when instantiating `ProjectConfig` (https://github.com/astronomer/astronomer-cosmos/blob/0123c3e19422bf665eca6a489a1248cc7ad17de9/cosmos/config.py#L57).

However, this variable should not be needed when rendering DAGs using a dbt manifest (`LoadMode.DBT_MANIFEST`). This is particularly relevant if users do not want to have their dbt pipelines together with the Airflow DAGs code, which should already be possible if they are using `ExecutionMode.KUBERNETES`.

**Proposal**

Make the `dbt_project_path` argument optional when instantiating `ProjectConfig`.

This ticket relates to #568 and it is also a follow-up to a conversation with @pgoslatara

MrBones757:
## Description

This change allows the dbt_project_path argument from ProjectConfig to be optional, and adds the ability to provide manifest_path alone. It also adds the ability for the user to (optionally) define project_name when dbt_project_path is defined, and requires project_name to be defined when dbt_project_path is not defined

## Related Issue(s)

closes #569

## Checklist


MrBones757:
hello @tatiana 
Thanks for your feedback on slack.

I have this MR to a point where i think it is just about ready to go, and would like to seek some review from you and / or your team.

I have read though the contributing guide, and believe i've done whats needed based on my understanding.
As far as local testing goes: 

The version of code currently pushed to this MR has passed all coverage tests (hatch run tests:test-cov)
I have also attempted to run the integration tests locally, most of which pass with the exception of the databricks test because i don't have that setup locally (ill check that one against CI)

I think the only item left here is to complete the docs updates, however before i complete that i want to make sure the logic is acceptable. Please let me know if you have any feedback.

cosmos/config.py
```diff
@@ -46,39 +46,67 @@ class ProjectConfig:
...
     """
 
-    dbt_project_path: str | Path
+    dbt_project_path: str | Path | None = None
     models_relative_path: str | Path = "models"
     seeds_relative_path: str | Path = "seeds"
     snapshots_relative_path: str | Path = "snapshots"
     manifest_path: str | Path | None = None
 
+    parsed_dbt_project_path: Path | None = None
```
tatiana, line 65: It feels like this would you think if this was a property of the class, as opposed to an attribute?

We could cache it using:
```
if sys.version_info >= (3, 9):
    from functools import cache
else:
    from functools import lru_cache
```

tests/test_config.py
```diff
@@ -13,25 +13,65 @@
...
     assert project_config.seeds_relative_path == Path("path/to/dbt/project/seeds")
     assert project_config.snapshots_relative_path == Path("path/to/dbt/project/snapshots")
     assert project_config.manifest_path is None
 
 
-def test_init_with_manifest():
+# Since dbt_project_path is now an optional parameter, we should test each combination for init and validation
+
+
+# Passing a manifest AND project together should succeed, as previous
```
tatiana, line 26: I think there are two better alternatives to comments in this case:
- improve the test method name (`test_init_with_manifest_and_project_succeeds`), or
- add this comment as a docstring of the test function

Happy with either of the alternatives! This applies to most of the tests added

tatiana (commented):
This looks great, @MrBones757. Thank you very much for improving this! It will be especially useful to K8s users.

I left minor comments in the code/docs.

It would also be great if you could update the correspondent docs, which previously stated this property was required:
https://github.com/astronomer/astronomer-cosmos/blob/main/docs/configuration/project-config.rst

cosmos/config.py
MrBones757: I have attempted to implement this based on my understanding, however, i found an annotation called cached_property which seems to achieve what we want and integrates more nicely with a dataclass.

tests/test_config.py
MrBones757: For this one, i have done a combination of both - preserved the comments as docstrings, and tried to standardize the function names so that they are more clear. Let me know if this is what you expected :)

cosmos/config.py
tatiana: Sounds great!

tests/test_config.py
tatiana: Thank you very much!

MrBones757:
Please let me know if there is anything else that needs doing on this card!
look forward to the release of this so we can leverage this and other related changes!

tatiana (approved):
Thank you very much for working on this and addressing all the feedback, @MrBones757 !

tatiana:
@MrBones757 This was released as part of 1.2:
* https://pypi.org/project/astronomer-cosmos/1.2.0/
* https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.2.0

tatiana:
@MrBones757 we'll probably need further changes in `coverter.py` and perhaps `dbt/graph.py` so users can benefit from this feature from a DAG perspective.
ATM if a user does not give the `dbt_project_path`, they will probably face an exception similar to the one described in #601

PR #591, opened 2023-10-10
Update contributing guide docs

### LINKED ISSUE #588: docs: Improvement on Contributing Guide
## Problem
1. Currently there are duplicates steps on the contributing guide file which can confused for some people:
image

2. I also encountered an error when trying to run `docker compose -f dev/docker-compose.yaml up -d --build` because of permission error on linux. The created folder for volume on linux is owned by root and need manual action to change permission of created folder
image

## Proposal
I want to delete the duplicate steps and add more steps for change permission of folder, so user will not encountered permission error because of lack of permission

raffifu:
## Description

This PR update contribution guide with the following details:
- Remove repetitive instructions
- Add instruction for linux to prevent permission error when running docker compose command

## Related Issue(s)

closes #588

## Checklist

Note: Currently i enrolled on hacktoberfest program, if you don't mind could you please add label `hacktoberfest-accepted` before merging this PR. Thank you


docs/contributing.rst
```diff
@@ -57,26 +57,6 @@ If testing for the same Airflow and Python version, next runs of the integration
...
-    export AIRFLOW_HOME=`pwd`
-    export AIRFLOW_CONN_AIRFLOW_DB=postgres://postgres:postgres@0.0.0.0:5432/postgres
-    hatch run tests.py3.8-2.5:test-integration-setup
-    hatch run tests.py3.8-2.5:test-integration
-
-If testing for the same Airflow and Python version, next runs of the integration tests can be:
-
-.. code-block:: bash
-
-    hatch run tests.py3.8-2.5:test-integration
```
tatiana, line 77: @raffifu I just realised this is not necessarily a duplicate from the block below but an alternative.
People may opt to develop Cosmos without Docker Compose - since it is easier to troubleshoot and use debuggers..! We can improve the docs to make this clear, but I think we should still keep documented both approaches.

raffifu: @tatiana The block code is already present on the previous line, so i think it's a duplicate

https://github.com/astronomer/astronomer-cosmos/blob/dac117a36209729519add1ad99fe3418572c6aa7/docs/contributing.rst?plain=1#L56-L58

Anyway i have an idea to add instruction `astronomer-cosmos` without docker compose (directly on host machine). Is it possible to add the instruction on docs? but i need to try first to see if my approach is worked

tatiana: Absolutely, would love to see that, @raffifu !

raffifu: @tatiana Already pushed the change on the new commit

docs/contributing.rst
```diff
@@ -16,93 +16,125 @@ To contribute to the cosmos project:
...
 
 .. code-block:: bash
 
-    export AIRFLOW_HOME=`pwd`
-    export AIRFLOW_CONN_AIRFLOW_DB=postgres://postgres:postgres@0.0.0.0:5432/postgres
-    hatch run tests.py3.8-2.5:test-integration-setup
-    hatch run tests.py3.8-2.5:test-integration
+    airflow webserver
+    airflow scheduler
+    airflow triggerer
```
tatiana, line 63: WDYT of recommending `airflow standalone`, it should run all of this? There would be a minor difference on the `admin` password, but it may be worth it.

tatiana (approved):
This is looking much better, @raffifu , thank you very much for addressing the feedback!
I left a minor comment, let me know what you think - we'll release this as part of 1.2.1.

docs/contributing.rst
raffifu: I think `airlfow standalone` will simplify all of that step. I'll update on the next commit

Anyway, i forgot to add note in the last commit, There's an error while parsing dags `example_cosmos_sources.py` because that dags will try to run dbt and cannot find adapter for sqlite (probably because the setup is use sqlite)
image

Actually there're 4 DAGs import errors, but i think it's out of scope from this PR because it's related to setup env var for databrick, etc. WDYT about it, should i add some note on the guide?

raffifu: Hii @tatiana, i've push new changes for fix typo and change to airflow standalone mode

tatiana:
@raffifu  We just published the release 1.2.1:

Issue #474, opened 2023-08-16
Add tests after snapshots


dhazan-raptive:
Currently using v1.0.x and noticed that while my models create task groups that contain the running and testing of the model, my snapshots are just a task that run the snapshot (no tests).

I'd like to see a feature added where snapshots are treated similar to models in the Airflow dag and get a task group with a run and test command.

tatiana:
Thanks for adding this, @dhazan-raptive ! This is a good first issue if anyone from the community wants to pick up, otherwise we may be able to squeeze in our next sprint at Astronomer!

PR #599, opened 2023-10-13
Add tests to sources, snapshots and seeds when using TestBehavior.AFTER_EACH

tatiana:
Previously Cosmos would only create a task group when using `TestBehavior.AFTER_EACH` for nodes of the type `DbtResourceType.MODEL`. This change adds the same behavior to snapshots and seeds.

For this to work as expected with sources, we would need to create a default operator to handle `DbtResourceType.SOURCE`, which is outside the scope of the current ticket. Once this operator exists, sources will also lead to creating a task group.

All the test selectors were tested successfully with dbt 1.6.

This screenshot illustrates the validation of this feature, with an adapted version of Jaffle Shop:
Screenshot 2023-10-13 at 19 43 52

The modifications that were done to jaffle_shop were:

Appended the following lines to `dev/dags/dbt/jaffle_shop/models/schema.yml`:
```
seeds:
  - name: raw_customers
    description: Raw data from customers
    columns:
      - name: id
        tests:
          - unique
          - not_null

snapshots:
  - name: orders_snapshot
    description: Snapshot of orders
    columns:
      - name: orders_snapshot.order_id
        tests:
          - unique
          - not_null
```

And created the file `dev/dags/dbt/jaffle_shop/snapshots/orders_snapshot.sql` with:
```
{% snapshot orders_snapshot %}

{{
    config(
      target_database='postgres',
      target_schema='public',
      unique_key='order_id',

      strategy='timestamp',
      updated_at='order_date',
    )
}}

select * from {{ ref('jaffle_shop', 'orders') }}

{% endsnapshot %}
```

Closes: #474

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -56,7 +56,7 @@ def create_test_task_metadata(
     execution_mode: ExecutionMode,
     task_args: dict[str, Any],
     on_warning_callback: Callable[..., Any] | None = None,
-    model_name: str | None = None,
+    node: DbtNode | None = None,
 ) -> TaskMetadata:
     """
     Create the metadata that will be used to instantiate the Airflow Task that will be used to run the Dbt test node.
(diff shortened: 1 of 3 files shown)


cosmos/airflow/graph.py
```diff
@@ -66,13 +66,18 @@ def create_test_task_metadata(
...
     """
     task_args = dict(task_args)
     task_args["on_warning_callback"] = on_warning_callback
-    if model_name is not None:
-        task_args["models"] = model_name
+    if node is not None:
+        if node.resource_type == DbtResourceType.MODEL:
+            task_args["models"] = node.name
+        elif node.resource_type == DbtResourceType.SOURCE:
+            task_args["select"] = f"source:{node.unique_id[len('source.'):]}"
```
harels, line 78: do we have a test case that's covering this part?

harels (approved):
lgtm

cosmos/airflow/graph.py
tatiana: @harels, we have a unit that covers the path formation for all these variations:
`test_create_test_task_metadata,` and I manually played with dbt 1.6 to confirm all these variations work.

ATM, we don't have an integration test for `DbtResourceType.SOURCE` since our only example DAG that validates this type of node overrides any default behavior. For us to have an end-to-end version of the code going through this, Cosmos would need a default behavior for SOURCE nodes, something we don't currently have.

@harels @jlaneve Should we add a ticket to define a default behavior for Source nodes? What do you think that should look like?  Should it check for freshness?

tatiana:
I'm merging since the PR was approved, and the open discussion should only be closed once we decide whether to have a follow-up ticket.

Issue #597, opened 2023-10-12
Dissolve `ProjectConfig` into `RenderConfig` and `ExecutionConfig` supporting different dbt project paths for rendering and execution


david-mag:
Solution/proposal related to issue #568 that has possible backwards compatibility issues.

**Context**

As of Cosmos 1.1.x, the argument project_dir (passed in the `ProjectConfig` constructor) is used for two purposes:

1. To interpret the dbt project & build Airflow DAG
2. To execute the dbt project nodes

Although this makes sense for the `ExecutionMode.LOCAL`, it would make more sense to declare different paths when using `ExecutionMode.DOCKER` and `ExecutionMode.KUBERNETES`.

**Proposal**

With the `ProjectConfig` serving information for two purposes (rendering and execution), and purpose specific configs already existing (`ExecutionConfig` and `RenderConfig`). One approach could be to dissolve `ProjectConfig` into `ExecutionConfig` and `RenderConfig`. 

Currently the `ExecutionConfig` serves no other purpose than to specify which execution mode to run. Arguments that are execution specific that are currently stored in the `ProjectConfig` such as `models_relative_path`, `seeds_relative_path`, and `snapshots_relative_path` could be moved there, as well as an execution specific `project_dir` (currently `dbt_project_path`). *[All these arguments could/should be optional btw.]* 

The `RenderConfig` which already concerns how the dbt DAG is presented in airflow could do just that and include a `dbt_project_path` argument as well as a `manifest_path` [*I guess one of them needs to be required*], which are currently part of the `ProjectConfig`.

That way we could get rid of a dual purpose configuration and have the remaining configs be clearer in their scope.

tatiana:
This was the way to go, but we may need to review this approach.

I have the impression that it does make sense to have a shared configuration between Execution and Render config - for local users' sake. It may or may not be `ProjectConfig`.

Whereas we should allow users who want to give different configurations to `RenderConfig` and `ExecutionConfig` to do so, we should also not burden the users who want the same configuration to be applied in both places to have to define them twice.

This is an example of an issue we came across while splitting the project dir and the dbt bin path in #568 : #666

MrBones757:
I agree with the statement/s above.
With the work undertaken to add dbt_project_path to RenderConfig and ExecutionConfig (and thus supporting None in projectconfig), as well as other similar changes such as allowing the dbt executable to be different between RenderConfig and Execution, i'd say we're already most of the way there.

[...]

Issue #568, opened 2023-09-29
Support using different dbt paths for DAG rendering and dbt execution


tatiana:
**Context**

As of Cosmos 1.1.x, the argument project_dir (passed in the ProjectConfig constructor) is used for two purposes:
1. To interpret the dbt project & build Airflow DAG
2. To execute the dbt project nodes

Although this makes sense for the `ExecutionMode.LOCAL`, it would make more sense to declare different paths when using `ExecutionMode.DOCKER` and  `ExecutionMode.KUBERNETES`.

This was raised as an issue by @adityasharma1685 in the Apache Airflow #airflow-dbt Slack channel:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1695930777905369

**Proposal**

Introduce two new configuration arguments to be passed in the constructor
* `RenderConfig(dbt_project_path="airflow-worker-dbt-project-path")` (optional, new)
* `ExecutionConfig(dbt_project_path="where-dbt-is-executed")` (optional, new)
To complement:
* `ProjectConfig(dbt_project_path="current-project-path-used-for-parsing-and-execution") (existing)`

For the DAG generation (1), we could:
a. Use `RenderConfig.dbt_project_path` if defined, or
b. Fallback to the existing config (backward compatible) `ProjectConfig.dbt_project_path`

For dbt execution (2), we could:
a. Use `ExecutionConfig.dbt_project_path` if defined, or
b. Fallback to the existing config (backward compatible) `ProjectConfig.dbt_project_path`

**Alternative**

I'm still trying to figure out alternatives. Any ideas are very welcome!

It would probably not work to set project_dir (https://github.com/astronomer/astronomer-cosmos/blob/0123c3e19422bf665eca6a489a1248cc7ad17de9/cosmos/operators/base.py#L83) as an Airflow templated field, because the Kubernetes Pod is instantiated at the Airflow Worker - and would likely have the same value as the Scheduler/DAG Parser.

**Acceptance criteria**
*  [ ] Forementioned changes
*  [ ] Update the manifest example DAG, as discussed in https://github.com/astronomer/astronomer-cosmos/pull/605#issuecomment-1776873441, so the project path is only set in the `ExecutionConfig`

tatiana:
Discussed in the #airflow-dbt channel:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1696308262251219

tatiana:
[...]

PR #605, opened 2023-10-17
Resolve errors occurring when dbt_project_path is str and partial support dbt_project_path=None

### LINKED ISSUE #601: Fix converting dbt projects that have ProjectConfig.dbt_project_path set as strings
Our documentation has examples of configuring a project using a project path that is a string:
```
from cosmos.config import ProjectConfig

config = ProjectConfig(
    dbt_project_path="/path/to/dbt/project",
    models_relative_path="models",
    seeds_relative_path="data",
    snapshots_relative_path="snapshots",
    manifest_path="/path/to/manifests",
)
```

However, if users attempt to use this, they will face an error:
```
Broken DAG: [/home/airflow/Sites/data-airflow/dags/high_freq_core_dbt.py] Traceback (most recent call last):
  File "/home/airflow/python-venvs/data-pipelines/lib/python3.10/site-packages/cosmos/airflow/task_group.py", line 26, in __init__
    DbtToAirflowConverter.__init__(self, *args, **specific_kwargs(**kwargs))
  File "/home/airflow/python-venvs/data-pipelines/lib/python3.10/site-packages/cosmos/converter.py", line 109, in __init__
    dbt_root_path = project_config.dbt_project_path.parent
AttributeError: 'str' object has no attribute 'parent'
```

Because of:
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/converter.py#L109-L110

This issue was originally reported in the #airflow-dbt slack channel by Sai:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1697234106298499?thread_ts=1697209691.193499&cid=C059CC42E9W

**Acceptance criteria**
As part of the fix, we should improve the converter tests by adding unit tests similar to:
https://github.com/astronomer/astronomer-cosmos/blob/main/tests/test_converter.py#L49-L69

We should validate at least the two following use-cases:
1. `ProjectConfig.dbt_project_path` as string
2. No `ProjectConfig.dbt_project_path`, to have a broader test of the improvement #581

MrBones757:
## Description

As part of the changes made #581, some downstream logic was missed relating to the handling of a None and String based project dir. This MR attempts to remedy this issue, by adding down steam support for project dir being None (including generation of exceptions and guarding), as well as some property reference changes in converter.

## Related Issue(s)

closes #601 

## Checklist


MrBones757:
@tatiana 

I'm not 100% confident with some of the changes i made downstream - assuming test coverage is all good we should be fine, however, given i'm still not 100% familiar with the codebase here some more detailed review may be required.

MrBones757:
I have just noticed the above errors in some tests - i will check those out shortly

tatiana (commented):
This looks excellent. Thanks a lot for quickly picking up and solving the issue, @MrBones757 !

I left some minor comments in-line, and I strongly suggest we change the following example DAGs, which are used in the integration tests:
- `dev/dags/cosmos_manifest_example.py`: to have `dbt_project_path ` as None
- `dev/dags/basic_cosmos_task_group.py`: to have `dbt_project_path ` as a string
This will complement the unit tests you created to help us verify the change works e2e.

We'll release this change as part of Cosmos 1.2.1, at the latest, early next week.

MrBones757:
I think the latest commit should address all your feedback :) let me know if theres anything else that needs updating

tatiana (commented):
@MrBones757, we're super close to merging this. Thanks for addressing all the comments and the relevant refactoring on how we represent the DbtProjects within Cosmos.

While validating, I found an issue while running the DAG `cosmos_manifest_example` using:

```
AIRFLOW_CONN_AIRFLOW_DB="postgres://postgres:postgres@0.0.0.0:5432/postgres" PYTHONPATH=`pwd` AIRFLOW_HOME=`pwd` AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT=20000 AIRFLOW__CORE__DAG_FILE_PROCESSOR_TIMEOUT=20000 airflow dags test cosmos_manifest_example `date -Iseconds`
```

It raised the following exception:
```
[2023-10-23T15:07:16.149+0100] {dag.py:2760} ERROR - (astronomer-cosmos) - Task failed; ti=<TaskInstance: cosmos_manifest_example.customers.run manual__2023-10-23T15:07:10+01:00 [up_for_retry]>
  File "/Users/tati/Code/astronomer-cosmos/venv-airflow27/lib/python3.11/site-packages/airflow/models/dag.py", line 2758, in test
    _run_task(ti, session=session)
  File "/Users/tati/Code/astronomer-cosmos/venv-airflow27/lib/python3.11/site-packages/airflow/models/dag.py", line 3902, in _run_task
    ti._run_raw_task(session=session)
  File "/Users/tati/Code/astronomer-cosmos/venv-airflow27/lib/python3.11/site-packages/airflow/utils/session.py", line 74, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tati/Code/astronomer-cosmos/venv-airflow27/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 1516, in _run_raw_task
    self._execute_task_with_callbacks(context, test_mode, session=session)
  File "/Users/tati/Code/astronomer-cosmos/venv-airflow27/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 1679, in _execute_task_with_callbacks
    result = self._execute_task(context, task_orig)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tati/Code/astronomer-cosmos/venv-airflow27/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 1742, in _execute_task
    result = execute_callable(context=context, **execute_callable_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tati/Code/cosmos-clean/astronomer-cosmos/cosmos/operators/local.py", line 445, in execute
    self.build_and_run_cmd(context=context, cmd_flags=cmd_flags)
  File "/Users/tati/Code/cosmos-clean/astronomer-cosmos/cosmos/operators/local.py", line 357, in build_and_run_cmd
    result = self.run_command(cmd=dbt_cmd, env=env, context=context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tati/Code/cosmos-clean/astronomer-cosmos/cosmos/operators/local.py", line 198, in run_command
    shutil.copytree(
  File "/opt/homebrew/Cellar/python@3.11/3.11.6/Frameworks/Python.framework/Versions/3.11/lib/python3.11/shutil.py", line 561, in copytree
    return _copytree(entries=entries, src=src, dst=dst, symlinks=symlinks,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.6/Frameworks/Python.framework/Versions/3.11/lib/python3.11/shutil.py", line 466, in _copytree
    srcname = os.path.join(src, srcentry.name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen posixpath>", line 76, in join
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

I checked in the CI, and it seems this is the only DAG not being run in any of our integration tests - something that also happens on Cosmos `main` branch. The fix for this issue is here:
#619. Please, could you rebase and look into this issue?

MrBones757:
I have been able to replicate the issue you found above, and, If i understand the issue correctly, it looks like this is related to the operator command generation no longer having a root project path to base the dbt commands of that are run within the DAGs.

It sounds like this is starting to require the logic described in : https://github.com/astronomer/astronomer-cosmos/issues/597 to fully fix the issue, however, as a stop-gap solution we could apply a default based on the operator type.

It looks to me like we could, for now, use a default as the existing k8s and docker executors do, which is to reference $airflow_home/dags as the dbt project location base (as noted in the setup instructions: https://astronomer.github.io/astronomer-cosmos/getting_started/kubernetes.html)

I will give this a go and try to get this working to unblock this, and im noting this here as a discussion point if required, and to link it to the other issue for when that is worked on.

MrBones757:
For now, while i'm testing, i've added a new field to ExecutionConfig called dbt_project_path_root, which i'm referencing in converter.py init as:
        dbt_project_path_runtime_root = execution_config.dbt_project_path_root
This way, we can pass this to the operator's args, in combination with the project_name from ProjectConfig to get the path at runtime.

I was thinking of just adding it as a default in converter, but, exposing it as a config option seemed like a better option, though, im aware we don't want to cause any backward compatibility issues with public facing interfaces like this, so happy to remove it from ExecutionConfig, and just hard-code (lookup AIRFLOW_HOME) in converter if thats a better approach

I'm currently unable to test this locally due to some environment constraints on my development environment, though im working on this and should have more tests conducted soon to fully verify the validity of this solution.

MrBones757:
I have confirmed all unit and integration tests run successfully on a fresh vm across the required python/airflow version set/s
I think my choice to add execution_config.dbt_project_path_root in this PR is questionable, and alternate implementation recommendations would be appreciated.

tatiana:
@MrBones757, thanks for looking into this, addressing it quickly, and finding a workaround for the issue!

I also have mixed feelings about the approach of setting the path in the local operator. As you pointed out, I'm worried about edge cases and the possibility of us introducing a breaking change or unexpected behaviours.

I understand that to execute the dbt commands in the LOCAL mode, we will need the original dbt project code & path. Implementing ticket #568 would solve the issue without needing the more abrupt change proposed in #597 - I see it as a first step to solving this problem. K8s and Docker users have complained about this existing behaviour. That said, fixing #568 is outside the scope of the current PR.

My suggestion for us to wrap this PR:
1. Undo the changes from the last commit, 02ae25be2e3fd16132ea1711412aef8aeb8080bd.
2. Revert the manifest example; since it uses LOCAL, it needs the project path (I'm sorry for my misleading suggestion)
3. Rely on the test https://github.com/astronomer/astronomer-cosmos/pull/605/files#diff-f56d97c6216a37d8ca841f31c63b704d05b7e3a42d6dbe783411d6fef8204615R116 to cover the change we introduced related to supporting Manifest, no project path & a non-local Cosmos execution mode. 

Once we implement #568, we can update the Manifest example safely - adding the project path only to the execution configuration. I'm adding a comment to the ticket so we don't forget this.

WDYT?

MrBones757:
@tatiana I'm happy to revert that change later, as long as we're sure that this wont break docker / k8s based deployments - i'm not 100% clear on where those execution methods get their paths from. Based on what i could see by following the code though, this property that im looking up in env when dbt_project_path is not defined seems to also control paths for docker / k8s too - can you confirm where the k8s / docker paths are controlled? my assumption was it used the same mechanism..

my thinking here, is that since there would be no root path when only a manifest is provided, dbt would have no base folder for its operations in the containerised modes and thus resul in errors locating files, unconfirmed though - purely an assumption

MrBones757:
that sounds good to me - to confirm the above:

I will revert the change to the ExecutionConfig, and we will require the project path be defined still;

Separately to that, i will have a discussion with you about the implementation of #568  as i believe i should have bandwith to do that.

MrBones757:
I cant see any other opportunities for adding coverage in the scope of this PR. Is this slight change required to be resolved to get this merged?

tatiana (approved):
This looks great, @MrBones757 , thanks a lot for the contribution and for all the patience in addressing the feedback. The refactoring was amazing, and we're very close to allowing users to support `dbt_project_path=None`!

tatiana:
@MrBones757 We just published the release 1.2.1:

PR #606, opened 2023-10-17
Fix custom selector when select has a subset of tags of the models' tags

david-mag:
## Description
Tags filtering (select/exclude) was bugged, expecting a perfect match of *all* tags set (e.g. `model1` with tags `tag1` and `tag2` would only be selected/excluded if the select/exclude statement in the `RenderConfig` is set to `select=["tag:tag1,tag:tag2"]`, but not if it is set as `select=["tag:tag1"]` or `select=["tag:tag2"]` only.)

Tests were already existing, but the node-setup did not include intersections of tags. I've fixed that and added another test as well.

## Checklist

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -106,7 +106,7 @@ def select_nodes_ids_by_intersection(nodes: dict[str, DbtNode], config: Selector
         visited_nodes.add(node_id)
 
         if config.tags:
-            if not (set(config.tags) == set(node.tags)):
+            if not (set(config.tags) <= set(node.tags)):
                 return False
 
         node_config = {key: value for key, value in node.config.items() if key in SUPPORTED_CONFIG}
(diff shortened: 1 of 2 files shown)


tatiana (commented):
Thanks, @david-mag , it looks good - happy to merge once the checks pass!

david-mag:
@tatiana Not sure whats the issue exactly, but the error msg in the failed test returns this: 
```
result = FullOutputSubprocessResult(exit_code=2, output='      Error during request to server: : Invalid access token.', full_o...time Error', '  Runtime Error', '    Database Error', '      Error during request to server: : Invalid access token.'])
```
So could it be that the integration test itself has an issue? I'm not aware of any changes regarding access tokens in my PR. 😅

tatiana:
Hey @david-mag , yes I noticed this earlier today - it seems there is an issue with the token. I will look into this within a few hours!

tatiana (approved):

tatiana:
@david-mag We just published the release 1.2.1:

Issue #580, opened 2023-10-08
Tag selector not working with tests using the manifest json


nicohein:
There is a bug with the parsing of tests (at lease when using load_from_dbt_manifest).

Dbt does not populate the tags on the tests, still I would expected the behavior like in dbt  that tests for the models matched by a tag are selected. The current implementation in astronomer-cosmos however only considers nodes in the `filtered-notes` for tests.

https://github.com/astronomer/astronomer-cosmos/blob/2e3e4fd77db427441646a6139d5ac017b186366e/cosmos/dbt/graph.py#L326

I would hope that this can be fixed by iterating through nodes not filtered_nodes in the following line:

https://github.com/astronomer/astronomer-cosmos/blob/2e3e4fd77db427441646a6139d5ac017b186366e/cosmos/dbt/graph.py#L375

Alternatively there could be a way to do a second run through the nodes in the selector to catch all tests that depend on selected models:

https://github.com/astronomer/astronomer-cosmos/blob/2e3e4fd77db427441646a6139d5ac017b186366e/cosmos/dbt/selector.py#L84

I have not thought too much about possible side effects.

*Current Workaround*: Use a different selection method or explicitly tag tests.

--

astronomer-cosmos version 1.1.3
dbt-core vserion 1.5.7

tatiana:
@nicohein This seems to be a limitation of Cosmos custom selector, which is used with `LoadMethod.CUSTOM` or `LoadMethod.DBT_MANIFEST`. If using `load_via_dbt_ls`, the selector behaviour would be just like dbt's.

Would you like to make a contribution to support this use-case in the custom selector?

nicohein:
@tatiana while I would like to make a contribution I am currently unable to give commitments. I am happy to be around for testing also beta versions within our test environment.

nicohein:
HI @tatiana ,
I have written an alternative solution for dbt dags on airflow using offline parsing of the same. I will unfortunately not be able to contribute further.
Best
Nico

PR #615, opened 2023-10-19
resolving issues with the DBT_MANIFEST load method when has_test attr…

edgga:
resolving issues with the DBT_MANIFEST load method when the has_test attribute is not assigned to the node correctly.

## Description

has_test attribute isn't properly assigned to nodes when using DBT_MANIFEST and CUSTOM load methods. Unfortunately, I have only been able to fix issues with the DBT_MANIFEST load method and left the code, that's also used by the CUSTOM load method unchanged.

## Related Issue(s)

Closes https://github.com/astronomer/astronomer-cosmos/issues/580

## Breaking Change?

The change should not be breaking anything.

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -330,9 +330,6 @@ class DbtGraph:
         However, since the Manifest does not represent filters, it relies on the Custom Cosmos implementation
         to filter out the nodes relevant to the user (based on self.exclude and self.select).
 
-        Noted that if dbt project contains versioned models, need to use dbt>=1.6.0 instead. Because, as dbt<1.6.0,
-        dbt cli doesn't support select a specific versioned models as stg_customers_v1, customers_v1, ...
-
         Updates in-place:
         * self.nodes
         * self.filtered_nodes
@@ -341,6 +338,7 @@ class DbtGraph:
         nodes = {}
         with open(self.project.manifest_path) as fp:  # type: ignore[arg-type]
             manifest = json.load(fp)
+            node_dependencies = manifest.get("child_map", {})
 
             resources = {**manifest.get("nodes", {}), **manifest.get("sources", {}), **manifest.get("exposures", {})}
             for unique_id, node_dict in resources.items():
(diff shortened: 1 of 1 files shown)


tatiana:
Thank you very much for working on this, @edgga ! Please let us know if you'd like reviews. We'll assume that while it is in the draft, it's a work in progress, and we shouldn't review it yet.

edgga:
Hello, @tatiana, I made a minor change that assigns tags to tests based on their parent model. It works as expected but I am not sure if this is the best approach. Would love to get some feedback!

tatiana (commented):
@edgga This seems like a very pragmatic solution to the problem, and it seems it allows us to have consistent behaviour on test tag-selection when using all the supported `LoadMethod`. Do you see any downside to this approach?

It would be great if we could add a test!

edgga:
@tatiana I could not think of any downsides, but it's very possible I haven't thought about every existing use case. Right, I could add some tests, I would like to clarify - should I only add `tags` for testing purposes only in manifest.json, or should I also make changes to the underlying DBT project for consistency?

tatiana:
@edgga Since ATM, this change is meant to solve a custom selection issue, we can probably have it where you've implemented it, for now. In the future, if we see the need for it, we can move this change to the `DbtGraph` - WDYT?

edgga:
@tatiana, I have added a pretty basic test that checks the general functionality of the has_test attribute update when node selection based on a TAG is used. Not sure if that's enough though. 

If you have any pointers on what else I need to do to get this merged, I am all ears, I would love to have this merged next week.

tests/dbt/test_graph.py
```diff
@@ -504,6 +504,24 @@ def test_update_node_dependency_test_not_exist():
...
+    )
+    profile_config = ProfileConfig(
+        profile_name="test",
+        target_name="test",
+        profiles_yml_filepath=DBT_PROJECTS_ROOT_DIR / DBT_PROJECT_NAME / "profiles.yml",
+    )
+    dbt_graph = DbtGraph(project=project_config, profile_config=profile_config, select=["tag:test_tag"])
+    dbt_graph.load_from_dbt_manifest()
+
+    for _, node in dbt_graph.filtered_nodes.items():
```
tatiana, line 519: @edgga could we also do an assert in the len of `dbt_graph.filtered_nodes`, just to be on the safe side?

I usually like to avoid logic  such as for loops in tests - since that increases the chances for bugs - but it's okay for this test case, as long as we confirm the for won't be iterating in a empty list :)

tatiana (commented):
@edgga thank you very much. This is looking very good!
I left a minor comment, and we'll release this as part of 1.2.2, next week.

Issue #612, opened 2023-10-18
Jinja undefined errors when extracting openlineage events


ivanstillfront:
All my cosmos-generated tasks are failing after the dbt run because of Jinja "undefined" errors during lineage processing. I have lineage disabled with `OPENLINEAGE_DISABLED=True` env var.

I am using Astro runtime 9.1.0 with dbt in a virtualenv as recommended. This is my Dockerfile:
```
FROM quay.io/astronomer/astro-runtime:9.1.0

# install dbt into a virtual environment
RUN python -m venv dbt_venv && source dbt_venv/bin/activate && \
    pip install --no-cache-dir dbt-snowflake && deactivate

ENV OPENLINEAGE_DISABLED=True
```

My model is very simple:
```sql
{{ config(materialized = 'table') }}

select
  try_to_timestamp('{{ run_started_at }}') as run_started_at
```

Everything works fine until the Task starts extracting openlineage data.

Here is the task log:
```
...normal dbt run above
[2023-10-18, 14:49:11 PDT] {subprocess.py:94} INFO - [0m21:49:11  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
[2023-10-18, 14:49:12 PDT] {subprocess.py:98} INFO - Command exited with return code 0
[2023-10-18, 14:49:12 PDT] {taskinstance.py:1935} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/cosmos/operators/local.py", line 445, in execute
    self.build_and_run_cmd(context=context, cmd_flags=cmd_flags)
  File "/usr/local/lib/python3.11/site-packages/cosmos/operators/local.py", line 357, in build_and_run_cmd
    result = self.run_command(cmd=dbt_cmd, env=env, context=context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/cosmos/operators/local.py", line 231, in run_command
    self.calculate_openlineage_events_completes(env, Path(tmp_project_dir))
  File "/usr/local/lib/python3.11/site-packages/cosmos/operators/local.py", line 268, in calculate_openlineage_events_completes
    openlineage_processor = DbtLocalArtifactProcessor(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #622, opened 2023-10-23
Honour `OPENLINEAGE_DISABLED` env var

ivanstillfront:
## Description

This allows dbt tasks to succeed successfully when lineage has been disabled by honouring the `OPENLINEAGE_DISABLED` environment variable that Airflow uses.

## Related Issue(s)

This does not resolve, but is related to #612. It allows us to run dbt tasks without having errors described in the issue.

## Breaking Change?

This should not break anything.

## Checklist

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -74,6 +74,10 @@ except airflow.exceptions.AirflowConfigException:
     LINEAGE_NAMESPACE = os.getenv("OPENLINEAGE_NAMESPACE", DEFAULT_OPENLINEAGE_NAMESPACE)
 
 
+# In the spirit of: https://github.com/apache/airflow/blob/8531396c7c8bf1e016db10c7d32e5e19298d67e5/airflow/providers/openlineage/plugins/openlineage.py#L30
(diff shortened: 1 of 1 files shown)


cosmos/operators/local.py
```diff
@@ -227,7 +231,7 @@ def run_command(
                     output_encoding=self.output_encoding,
                     cwd=tmp_project_dir,
                 )
-                if is_openlineage_available:
+                if is_openlineage_enabled:
```
tatiana, line 234: @ivanstillfront Thanks for looking into this!

I'm not convinced about this change since this line aims to set Airflow inlets and outlets - unrelated to emitting open lineage events - but to support Airflow Data Aware scheduling.

This is being improved, but it will likely only be available in Cosmos 1.3 or 1.4.

Does the following comment address your original issue?
https://github.com/astronomer/astronomer-cosmos/issues/612#issuecomment-1776880897

tatiana:
@ivanstillfront The original issue you were facing in #612 should have been fixed as part of this release:
https://pypi.org/project/astronomer-cosmos/1.2.1/
https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.2.1

PR #623, opened 2023-10-24
Add template fields

joppevos:
## Checklist


cosmos/operators/local.py
```diff
@@ -390,6 +390,7 @@ class DbtSeedLocalOperator(DbtLocalBaseOperator):
     """
 
     ui_color = "#F58D7E"
+    template_fields: Sequence[str] = ("full_refresh",)
```
tatiana, line 393: We need to "inherit" what was defined in the parent class as well:

https://github.com/astronomer/astronomer-cosmos/blob/1bf3a3efd5d4962a29650ea7357afa77018b3018/cosmos/operators/local.py#L93

```suggestion
    template_fields: Sequence[str] = DbtLocalBaseOperator.template_fields  + ("full_refresh",)  # type: ignore[operator]
```

These comments apply to the other operators as well.

tatiana (commented):
Thanks for the change, @joppevos !
In addition to the comment inline, it would be great to add a test to confirm the value of `template_fields` for these operators.

joppevos:
@tatiana thanks, I will follow up on the weekend

tatiana (approved):
This looks great, @joppevos ! Thank you very much!

tatiana:
@joppevos, when you have a chance, please rebase the branch - and we'll get it merged.

PR #625, opened 2023-10-25
Fix `LoadMode.AUTOMATIC` behaviour to use `LoadMode.DBT_LS` when `ProfileMapping` is used

tatiana:
Since #489 was merged, the behavior of `LoadMode.AUTOMATIC` changed to generate a `profiles.yml` file if the file didn't exist. However, we forgot to remove the previously necessary condition for being able to run `LoadMode.DBT_LS` (having the `profiles.yml` file).

This leads to inconsistent behaviour in Cosmos when using `LoadMode.AUTOMATIC` and the `manifest.json` was not available:
1. If the user used a `ProfileConfig` with `profiles_yml_filepath`, it would use `LoadMode.DBT_LS`
2. If the user used a `ProfileConfig` with a ProfileMapping class, it would unnecessarily use `LoadMode.CUSTOM`

This PR fixes the behaviour to attempt to use `LoadMode.DBT_LS` regardless of how the `ProfileConfig` was set.

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -163,12 +163,6 @@ class ProfileConfig:
         if not self.profiles_yml_filepath and not self.profile_mapping:
             raise CosmosValueError("Either profiles_yml_filepath or profile_mapping must be set to render a profile")
 
-    def is_profile_yml_available(self) -> bool:
-        """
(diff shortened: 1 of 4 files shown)


harels (approved):
lgtm, pending tests passing

PR #649, opened 2023-11-04
Support for Snowflake encrypted private key environment variable

### LINKED ISSUE #632: Issue with the SnowflakeEncryptedPrivateKeyPemProfileMapping
image

With the new snowflake connection with an encrypted private key in the creation of the profile it misses out the "private_key" like how it does in in the non encrypted version.

DanMawdsleyBA:
## Description

Adds a snowflake mapping for encrypted private key using an environment variable

## Related Issue(s)
https://github.com/astronomer/astronomer-cosmos/issues/632

## Breaking Change?

This does rename the previous SnowflakeEncryptedPrivateKeyFilePemProfileMapping to SnowflakeEncryptedPrivateKeyFilePemProfileMapping but this makes it clearer as a new SnowflakeEncryptedPrivateKeyPemProfileMapping is added which supports the env variable. Also was only released as a pre-release change

## Checklist

### DIFF
--- a/cosmos/profiles/__init__.py
+++ b/cosmos/profiles/__init__.py
@@ -16,7 +16,8 @@ from .postgres.user_pass import PostgresUserPasswordProfileMapping
 from .redshift.user_pass import RedshiftUserPasswordProfileMapping
 from .snowflake.user_pass import SnowflakeUserPasswordProfileMapping
 from .snowflake.user_privatekey import SnowflakePrivateKeyPemProfileMapping
-from .snowflake.user_encrypted_privatekey import SnowflakeEncryptedPrivateKeyPemProfileMapping
+from .snowflake.user_encrypted_privatekey_file import SnowflakeEncryptedPrivateKeyFilePemProfileMapping
+from .snowflake.user_encrypted_privatekey_env_variable import SnowflakeEncryptedPrivateKeyPemProfileMapping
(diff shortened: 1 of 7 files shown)


DanMawdsleyBA:
@tatiana I've made the changes. When you get the chance can you review them? Thanks

tatiana (approved):
@DanMawdsleyBA thank you for working on this and improving the Snowflake profile mappings.

I like your change, and it's a fair point that it's a change based on a pre-release, so the fact it is a breaking change should be acceptable.

@ivanstillfront @nacpacheco, you've previously contributed to the Snowflake profile mapping, do you have any thoughts about this change? Would love your thoughts before merging.

Once we merge it, we can release 1.3.0a2 - so no one is blocked. We still don't have the date of the 1.3.0 release.

ivanstillfront:
> @DanMawdsleyBA thank you for working on this and improving the Snowflake profile mappings.
> 
> I like your change, and it's a fair point that it's a change based on a pre-release, so the fact it is a breaking change should be acceptable.
> 
> @ivanstillfront @nacpacheco, you've previously contributed to the Snowflake profile mapping, do you have any thoughts about this change? Would love your thoughts before merging.
> 
> Once we merge it, we can release 1.3.0a2 - so no one is blocked. We still don't have the date of the 1.3.0 release.

This looks good. It is consistent with the other snowflake mappings.

Issue #651, opened 2023-11-06
Pydantic


ykuc:
Hi!
Why does [astronomer-cosmos](https://github.com/astronomer/astronomer-cosmos/blob/14b309039c65be9e1d27f1ddba8481e87548dbab/pyproject.toml#L40C6-L40C14) package have pydantic in dependencies?
I didn’t find it in the code, but airflow 2.7.2 upgrade to Pydantic 2

ykuc:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1699029324253389

tatiana:
@ykuc  Thanks for raising this!

This dependency was introduced 4 months ago and documented in the code:
https://github.com/astronomer/astronomer-cosmos/blob/1fb3e23f4fe5abe7c27d03403ae3ac02f201a1bc/pyproject.toml#L37
https://github.com/astronomer/astronomer-cosmos/commit/8cd9ff30d607f3e5269310b72a273dc0588a6856

It was originally introduced because of this Airflow issue:
https://github.com/apache/airflow/issues/32311

We'll remove it once our checks pass, thanks @jlaneve for the quick fix!

tatiana:
@ykuc @jlaneve It seems the Airflow 2.6 tests are failing if we remove the Pydantic dep:
https://github.com/astronomer/astronomer-cosmos/pull/654
https://github.com/astronomer/astronomer-cosmos/actions/runs/6776766746/job/18418976543?pr=654

```
= 23 failed, 152 passed, 18 deselected, 2 xfailed, 1 warning, 75 errors in 21.08s =
```

We could consider only installing this version of Pydantic for this particular version of Airflow when running its tests.

bolkedebruin:
This should indeed be a conditional dependency or preferably a constraint. This issue was resolved in Airflow since quite some time and Airflow itself is moving rapidly to Pydantic 2. Including cosmos downgrades pydantic at the moment while it doesn't have any dependencies on it.

pixie79:
Is there any progress on this, it would appear MWAA with 2.7.2 is failing with astronomer-cosmos due to constraints on needing pydantic 2.4.2 which is conflicting with the lock in cosmos.

flinz:
I came here for the same reason as @pixie79, since including cosmos pydantic is downgraded to <2, which breaks other dependencies for me (pydanctic > 2, on airflow 2.7.* and airflow 2.8.0). 

[...]

PR #654, opened 2023-11-06
Remove Pydantic dependency

jlaneve:
## Description

This PR removes the Pydantic dependency from Cosmos. It appears we're no longer using it.

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/651

## Breaking Change?

No

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,10 +34,8 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
 ]
 dependencies = [
-    # Airflow & Pydantic issue: https://github.com/apache/airflow/issues/32311
     "aenum",
     "attrs",
-    "pydantic>=1.10.0,<2.0.0",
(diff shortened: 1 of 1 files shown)


tatiana:
@jlaneve It seems the Airflow 2.6 tests are failing if we remove the Pydantic dep:

Example:
```
= 23 failed, 152 passed, 18 deselected, 2 xfailed, 1 warning, 75 errors in 21.08s =
```
https://github.com/astronomer/astronomer-cosmos/actions/runs/6776766746/job/18418976543?pr=654

We could consider only installing this version of Pydantic for this particular version of Airflow when running its tests, WDYT?

pixie79:
@jlaneve @tatiana I think I have the correct fix for this PR - #725 It removes pydantic as a dependency except if the apache-airflow version is 2.6.* which I believe is the desired outcome?

PR #655, opened 2023-11-07
Ensure local docs operators generate necessary files for datahub integration

chrishronek:
## Description

Two changes with this PR:
1. Add the `DbtFreshnessLocalOperator` & `DbtFreshnessS3LocalOperator` to properly generate the [sources.json file](https://docs.getdbt.com/reference/artifacts/sources-json)
2. Add [run_results.json](https://docs.getdbt.com/reference/artifacts/run-results-json) to the list of required files for the `DbtDocsLocalOperator`. 

These files are necessary to use [the dbt integration](https://datahubproject.io/docs/generated/ingestion/sources/dbt/) with DataHub and will make it easier for cosmos users to emit their metadata to DataHub as needed.

## Checklist

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -572,13 +572,13 @@ class DbtFreshnessS3LocalOperator(DbtFreshnessLocalOperator):
 
     def __init__(
         self,
-        aws_conn_id: str,
+        connection_id: str,
         bucket_name: str,
         folder_dir: str | None = None,
         **kwargs: str,
     ) -> None:
         "Initializes the operator."
-        self.aws_conn_id = aws_conn_id
+        self.connection_id = connection_id
         self.bucket_name = bucket_name
         self.folder_dir = folder_dir
 
@@ -591,7 +591,7 @@ class DbtFreshnessS3LocalOperator(DbtFreshnessLocalOperator):
         "Uploads the generated documentation to S3."
         logger.info(
             'Attempting to upload generated docs to S3 using S3Hook("%s")',
-            self.aws_conn_id,
+            self.connection_id,
         )
 
         from airflow.providers.amazon.aws.hooks.s3 import S3Hook
@@ -599,7 +599,7 @@ class DbtFreshnessS3LocalOperator(DbtFreshnessLocalOperator):
         target_dir = f"{project_dir}/target"
 
         hook = S3Hook(
-            self.aws_conn_id,
+            self.connection_id,
             extra_args={
                 "ContentType": "text/html",
             },
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -538,11 +538,12 @@ class DbtDocsLocalOperator(DbtLocalBaseOperator):
 
     ui_color = "#8194E0"
 
-    required_files = ["index.html", "manifest.json", "graph.gpickle", "catalog.json"]
+    required_files = ["index.html", "manifest.json", "graph.gpickle", "catalog.json", "run_results.json"]
 
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         self.base_cmd = ["docs", "generate"]
+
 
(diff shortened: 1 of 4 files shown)


cosmos/operators/local.py
```diff
@@ -538,13 +538,85 @@ class DbtDocsLocalOperator(DbtLocalBaseOperator):
...
 
-    required_files = ["index.html", "manifest.json", "graph.gpickle", "catalog.json"]
+    required_files = ["index.html", "manifest.json", "graph.gpickle", "catalog.json", "run_results.json"]
 
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         self.base_cmd = ["docs", "generate"]
 
 
+class DbtFreshnessLocalOperator(DbtLocalBaseOperator):
```
tatiana, line 548: Minor comment: what do you think about renaming `DbtFreshnessLocalOperator` to `DbtSourceFreshnessLocalOperator`?

cosmos/operators/local.py
```diff
@@ -538,13 +538,85 @@ class DbtDocsLocalOperator(DbtLocalBaseOperator):
...
+    """
+
+    ui_color = "#8194E0"
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.base_cmd = ["source", "freshness"]
+
+
+class DbtFreshnessS3LocalOperator(DbtFreshnessLocalOperator):
```
tatiana, line 561: Similar here, I wonder if `DbtSourceFreshnessS3LocalOperator` would be more representative.

docs/configuration/generating-docs.rst
```diff
@@ -41,6 +42,15 @@ You can use the :class:`~cosmos.operators.DbtDocsS3Operator` to generate and upl
...
 
+    generate_dbt_freshness_aws = DbtFreshnessS3Operator(
+        task_id="generate_dbt_freshness_aws",
+        project_dir="path/to/jaffle_shop",
+        profile_config=profile_config,
+        # docs-specific arguments
+        connection_id="test_aws",
+        bucket_name="test_bucket",
+    )
+
```
tatiana, line 53: I just realised we're adding the code manually in this doc code.

WDYT if we added a reference to the example DAG, similar to what we did in:
https://github.com/astronomer/astronomer-cosmos/blob/6225db8570fbfd92a81e4c497b5b1f651ac0ae6e/dev/dags/basic_cosmos_dag.py#L24
https://github.com/astronomer/astronomer-cosmos/blob/188fe56994e29037eda1777974faa45a183ebc1f/docs/getting_started/execution-modes.rst?plain=1#L58-L61

This way, the documentation will be up-to-date if the operator changes without additional effort.

docs/configuration/generating-docs.rst
```diff
@@ -41,6 +42,15 @@ You can use the :class:`~cosmos.operators.DbtDocsS3Operator` to generate and upl
         bucket_name="test_bucket",
     )
 
+    generate_dbt_freshness_aws = DbtFreshnessS3Operator(
```
tatiana, line 45: It may be worth to add a comment on why someone would like to send the source freshness files to S3, and perhaps a link to your blog post: https://parakeet.solutions/ingest-cosmos-dbt-into-datahub/, so people can have more

tatiana (commented):
Thanks, @chrishronek ! Very exciting to see these new operators.

I believe not only they can be useful for Datahub integration, but also with the integration to other services and tools (IIR OpenLineage also benefits from the `run_results.json` file).

Regarding tests, would it be worth to add a unit test related to `DbtFreshnessS3LocalOperator`? I noticed you added to the example DAG, which runs as part of the integration tests suite. That said, we don't have any test validating what this operator actually does.

I left some minor comments inline.

tatiana:
@chrishronek, we're starting to close stale PRs - please reopen or create a new one once you have a chance to contribute again!

PR #659, opened 2023-11-08
extend deps command with profile flags

### LINKED ISSUE #658: Fix running `install_deps=True` when using `profile_mapping` and `ExecutionMode.LOCAL`
As of Cosmos 1.2.2, when using `profile_mapping`, `install_deps=True` and the `ExecutionMode.LOCAL` to run dbt commands from within Airflow workers, we are not setting the profile mapping path. ATM, this only works if users use `profiles_yml_filepath `.

Exception raised:
```
  File "/home/airflow/.local/lib/python3.10/site-packages/cosmos/operators/local.py", line 445, in execute
    self.build_and_run_cmd(context=context, cmd_flags=cmd_flags)
  File "/home/airflow/.local/lib/python3.10/site-packages/cosmos/operators/local.py", line 357, in build_and_run_cmd
    result = self.run_command(cmd=dbt_cmd, env=env, context=context)
  File "/home/airflow/.local/lib/python3.10/site-packages/cosmos/operators/local.py", line 243, in run_command
    self.exception_handling(result)
  File "/home/airflow/.local/lib/python3.10/site-packages/cosmos/operators/local.py", line 124, in exception_handling
    raise AirflowException(
airflow.exceptions.AirflowException: ('dbt command failed. The command returned a non-zero exit code 2. Details: ', '\x1b[0m11:56:26  Running with dbt=1.5.0', '\x1b[0m11:56:29  Encountered an error:', 'Compilation Error', '  dbt found 1 package(s) specified in packages.yml, but only 0 package(s) installed in dbt_packages. Run "dbt deps" to install package dependencies.')
```

Although we're handling this properly during DAG parsing:
https://github.com/astronomer/astronomer-cosmos/blob/8e2d5908ce89aa98813af6dfd112239e124bd69a/cosmos/dbt/graph.py#L247-L266

We are indeed not passing the additional parameters in task execution:
https://github.com/astronomer/astronomer-cosmos/blob/8e2d5908ce89aa98813af6dfd112239e124bd69a/cosmos/operators/local.py#L209-L216

The goal of this ticket is to make this logic consistent, fixing the issue in the local.py.

Joppe Vos initially reported this issue in the #airflow-dbt Airflow slack channel:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1699365889141509

joppevos:
## Description

Extends the local operator when running `dbt deps` with the provides profile flags.
This makes the logic consistent between DAG parsing and task running as referenced below

https://github.com/astronomer/astronomer-cosmos/blob/8e2d5908ce89aa98813af6dfd112239e124bd69a/cosmos/dbt/graph.py#L247-L266

## Related Issue(s)

closes #658 

## Checklist

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -206,17 +206,11 @@ class DbtLocalBaseOperator(DbtBaseOperator):
                 tmp_project_dir,
             )
 
-            # if we need to install deps, do so
-            if self.install_deps:
-                self.run_subprocess(
-                    command=[self.dbt_executable_path, "deps"],
-                    env=env,
-                    output_encoding=self.output_encoding,
-                    cwd=tmp_project_dir,
-                )
-            with self.profile_config.ensure_profile() as (profile_path, env_vars):
+            with self.profile_config.ensure_profile() as profile_values:
+                (profile_path, env_vars) = profile_values
                 env.update(env_vars)
-                full_cmd = cmd + [
+
+                flags = [
                     "--profiles-dir",
                     str(profile_path.parent),
                     "--profile",
@@ -224,6 +218,18 @@ class DbtLocalBaseOperator(DbtBaseOperator):
                     "--target",
                     self.profile_config.target_name,
                 ]
+
+                if self.install_deps:
+                    deps_command = [self.dbt_executable_path, "deps"]
+                    deps_command.extend(flags)
+                    self.run_subprocess(
+                        command=[self.dbt_executable_path, "deps"],
+                        env=env,
+                        output_encoding=self.output_encoding,
+                        cwd=tmp_project_dir,
+                    )
+
+                full_cmd = cmd + flags
 
                 logger.info("Trying to run the command:\n %s\nFrom %s", full_cmd, tmp_project_dir)
                 logger.info("Using environment variables keys: %s", env.keys())


tatiana (commented):
This was super quick, @joppevos , thank you for fixing this issue.

The only thing missing is a test. I believe we could add a unit test in `astronomer-cosmos/tests/operators/test_local.py`, calling `run_command` and mocking `run_subprocess`, and checking if we're passing the expected arguments.

There is something similar in:
https://github.com/astronomer/astronomer-cosmos/blob/188fe56994e29037eda1777974faa45a183ebc1f/tests/operators/test_local.py#L369-L373

The motivation to have a test is to make sure we don't re-introduce this bug in the future.

cosmos/operators/local.py
```diff
@@ -225,6 +219,18 @@ def run_command(
                     self.profile_config.target_name,
                 ]
 
+                if self.install_deps:
```
joppevos, line 222: @tatiana We could separate the command creation outside of `run_command`  and bring it into `build_and_run`, then it's easier to test for the arguments

tatiana: It sounds like a good plan, @joppevos !

joppevos:
> Thanks for working on this and addressing the feedback so quickly, @joppevos ! We'll merge it once the checks pass.

Sounds good! Thanks for the support @tatiana :)

tatiana:
Thanks for the contribution, @joppevos  , we released it as part of Cosmos 1.2.4 yesterday:
https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.2.4

Issue #657, opened 2023-11-08
Validate dbt resource identifier to see if it is a valid airflow task id during conversion from dbt to Airflow


tatiana:
While parsing the dbt project, we should confirm if the node names we'll use to define Airflow task IDs are valid.
If they are not valid, we should at least raise a warning (probably an error message) explaining the issue to users.

We can accomplish this by using `airflow.utils.helpers.validate_key`.

This task is related to #636, but they are not duplicates and should not conflict.

This improvement came from a conversation with @MrBones757 , regarding Cosmos 1.2.2. More details:

When we create dbtNode objects in https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/dbt/graph.py#L367

We set the nodeName == alias, with a fallback to name if it doesn't exist.

We then create the airflow graph with that value here: https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/airflow/graph.py#L125

The issue: dbt doesn't impose limitations on node names, but airflow does as such. When we go to render the airflow dag in the UI, we get:
```
The key 'my$cool_alias' has to be made of alphanumeric characters, dashes, dots and underscores exclusively
```

tatiana:
For some reason, I'm not being able to assign this issue to @MrBones757 in the Github UI, but he is planning to work on this!

MrBones757:
Before implementing this, i will wait on #636 to ensure there are no clashing changes.

tatiana:
@MrBones757  #662 has a first level of "validation" (conversion from dots to underscore)

PR #662, opened 2023-11-09
Fix running models that use alias while supporting dbt versions

### LINKED ISSUE #636: Support naming tasks after the model name and not alias
**Context**

Since Cosmos 1.2, we've introduced support to run versioned dbt models (introduced in dbt 1.6), by adopting the alias as the task names, as opposed to the model name. While this works well for versioned tasks, there are use cases when users prefer to stick to the model name as opposed to the alias.

There were recent discussions on the #airflow-dbt slack channel about this:
- by Satyam: https://apache-airflow.slack.com/archives/C059CC42E9W/p1698430202280209?thread_ts=1693472093.058619&cid=C059CC42E9W
```
In one of the scenario we are passing dynamic aliases in config block using the run time parameter named 'division'.
for eg., test_customer_USA or test_customer_CAN where test_customer is my mode file name and USA is division.
I am using manifest to render airflow task and when cosmos render the task group, it uses the alias as the node name (as suggested in this thread) using the default value passed in dbt-project file.
But as per our requirement, the task name should have the model file name instead of alias like it was in last cosmos version
To give you an idea cosmos triggers following command while triggering airflow task:
bin/dbt', 'run', '--vars', "division: CAM\nrun_id: '20231026180212'\n", '--models', 'test_customer_NAM',
here NAM is default parameter passed in dbt_project.yml file and same is used while rendering the airflow task. We need task name as 'test_customer' (same as model file name)
so, need your inputs on how to handle this scenario in cosmos latest version?
```

- by Agnit Chatterjee: https://apache-airflow.slack.com/archives/C059CC42E9W/p1698637532291949
```
Hello everyone,
Hope you are doing well. With the latest cosmos version 1.2.1 we stumbled upon a new problem.
In one of the scenarios we are passing dynamic aliases in config block using the run time parameter named 'division'.
for eg., test_customer_USA or test_customer_CAN where test_customer is the model file name and USA is the division which is passed in as a parameter from the Airflow DAG.
We are using manifest to render airflow task and when cosmos render the task group, it uses the alias as the node name using the default value passed in dbt-project file.
But as per our requirement, the task name should have the model file name instead of alias like it was in last cosmos version
For example,  cosmos triggers following command while triggering airflow task:
bin/dbt', 'run', '--vars', "division: CAM\nrun_id: '20231026180212'\n", '--models', 'test_customer_NAM',
here NAM is default parameter passed in dbt_project.yml file and same is used while rendering the airflow task. We need task name as 'test_customer' (same as model file name)
so, need your inputs on how to handle this scenario in cosmos latest version?
```

**Proposal**

The suggested solution for this problem is introducing a new parameter, `use_alias_as_name`, to `RenderConfig`. By default it will be `True` - so we don't introduce a breaking change. This would allow users to set `use_alias_as_name=False` when they don't want Cosmos to render tasks based on the alias. In those scenarios, model versioning will not work.

binhnq94:
## Description

Current version, cosmos will got bug `Not found node` because it run with alias selection as: `--models customers_abc_v1 ` and `--models customers_abc_v2` .
I propose to parsing node selection in `unique_id` instead of using `alias` .
So node selection should be: `unique_id.split('.', 2)[2]` , reference to [function](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/node_args.py#L26) and [resource-details document](https://docs.getdbt.com/reference/artifacts/manifest-json#resource-details).
In addition, with this change help cosmos also support versioned models on dbt-core `>=1.5.0` instead `>=1.6.0` as current version.

## Related Issue(s)

closes #636

## Breaking Change?

Cosmos will support dynamic alias and versioned models

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -210,9 +210,6 @@ class DbtGraph:
         This is the most accurate way of loading `dbt` projects and filtering them out, since it uses the `dbt` command
         line for both parsing and filtering the nodes.
 
-        Noted that if dbt project contains versioned models, need to use dbt>=1.6.0 instead. Because, as dbt<1.6.0,
-        dbt cli doesn't support select a specific versioned models as stg_customers_v1, customers_v1, ...
-
         Updates in-place:
         * self.nodes
         * self.filtered_nodes
@@ -338,9 +335,6 @@ class DbtGraph:
 
         However, since the Manifest does not represent filters, it relies on the Custom Cosmos implementation
         to filter out the nodes relevant to the user (based on self.exclude and self.select).
-
-        Noted that if dbt project contains versioned models, need to use dbt>=1.6.0 instead. Because, as dbt<1.6.0,
-        dbt cli doesn't support select a specific versioned models as stg_customers_v1, customers_v1, ...
 
         Updates in-place:
         * self.nodes
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -42,7 +42,6 @@ class DbtNode:
     Metadata related to a dbt node (e.g. model, seed, snapshot).
     """
 
-    name: str
     unique_id: str
     resource_type: DbtResourceType
     depends_on: list[str]
@@ -50,6 +49,23 @@ class DbtNode:
     tags: list[str] = field(default_factory=lambda: [])
     config: dict[str, Any] = field(default_factory=lambda: {})
     has_test: bool = False
+
+    @property
+    def resource_name(self) -> str:
+        """
+        Use this property to retrieve the resource name for command generation, for instance: ["dbt", "run", "--models", f"{resource_name}"].
+        The unique_id format is defined as [<resource_type>.<package>.<resource_name>](https://docs.getdbt.com/reference/artifacts/manifest-json#resource-details).
+        For a special case like a versioned model, the unique_id follows this pattern: [model.<package>.<resource_name>.<version>](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/node_args.py#L26C3-L31)
+        """
+        return self.unique_id.split(".", 2)[2]
+
+    @property
+    def name(self) -> str:
+        """
+        Use this property as the task name or task group name.
+        Replace period (.) with underscore (_) due to versioned models.
+        """
+        return self.resource_name.replace(".", "_")
 
 
 def run_command(command: list[str], tmp_dir: Path, env_vars: dict[str, str]) -> str:
@@ -89,7 +105,6 @@ def parse_dbt_ls_output(project_path: Path, ls_stdout: str) -> dict[str, DbtNode
             logger.debug("Skipped dbt ls line: %s", line)
         else:
             node = DbtNode(
-                name=node_dict.get("alias", node_dict["name"]),
                 unique_id=node_dict["unique_id"],
                 resource_type=DbtResourceType(node_dict["resource_type"]),
                 depends_on=node_dict.get("depends_on", {}).get("nodes", []),
@@ -291,8 +306,7 @@ class DbtGraph:
         for model_name, model in models:
             config = {item.split(":")[0]: item.split(":")[-1] for item in model.config.config_selectors}
             node = DbtNode(
-                name=model_name,
-                unique_id=model_name,
+                unique_id=f"{model.type.value}.{self.project.project_name}.{model_name}",
                 resource_type=DbtResourceType(model.type.value),
                 depends_on=list(model.config.upstream_models),
                 file_path=Path(
@@ -347,7 +361,6 @@ class DbtGraph:
             resources = {**manifest.get("nodes", {}), **manifest.get("sources", {}), **manifest.get("exposures", {})}
             for unique_id, node_dict in resources.items():
                 node = DbtNode(
-                    name=node_dict.get("alias", node_dict["name"]),
                     unique_id=unique_id,
                     resource_type=DbtResourceType(node_dict["resource_type"]),
                     depends_on=node_dict.get("depends_on", {}).get("nodes", []),
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -83,11 +83,11 @@ def create_test_task_metadata(
         task_args["indirect_selection"] = test_indirect_selection.value
     if node is not None:
         if node.resource_type == DbtResourceType.MODEL:
-            task_args["models"] = node.name
+            task_args["models"] = node.resource_name
         elif node.resource_type == DbtResourceType.SOURCE:
-            task_args["select"] = f"source:{node.unique_id[len('source.'):]}"
+            task_args["select"] = f"source:{node.resource_name}"
         else:  # tested with node.resource_type == DbtResourceType.SEED or DbtResourceType.SNAPSHOT
-            task_args["select"] = node.name
+            task_args["select"] = node.resource_name
     return TaskMetadata(
         id=test_task_name,
         operator_class=calculate_operator_class(
@@ -108,8 +108,8 @@ def create_task_metadata(
     :param execution_mode: Where Cosmos should run each dbt task (e.g. ExecutionMode.LOCAL, ExecutionMode.KUBERNETES).
          Default is ExecutionMode.LOCAL.
     :param args: Arguments to be used to instantiate an Airflow Task
-    :param use_name_as_task_id_prefix: If resource_type is DbtResourceType.MODEL, it determines whether
-         using name as task id prefix or not. If it is True task_id = <node.name>_run, else task_id=run.
+    :param use_task_group: It determines whether to use the name as a prefix for the task id or not.
+        If it is False, then use the name as a prefix for the task id, otherwise do not.
     :returns: The metadata necessary to instantiate the source dbt node as an Airflow task.
     """
     dbt_resource_to_class = {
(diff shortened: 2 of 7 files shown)


cosmos/dbt/graph.py
```diff
@@ -52,6 +51,14 @@ class DbtNode:
     config: dict[str, Any] = field(default_factory=lambda: {})
     has_test: bool = False
 
+    @property
+    def name(self) -> str:
+        return self.unique_id.split(".", 2)[2].replace(".", "_")
```
tatiana, line 56: It would be great if we could link the documentation we're using as a reference to do this string parsing, and explain why we are doing this, as opposed to using the original `name`

It could be helpful to comment on the `unique_id` composition as well:
`<resource_type>.<package>.<resource_name>`

What about the replacement between "." and "_" - it would be great to document these.

We may want that this property use the other property, currently "select", since it is a subset of the logic.

cosmos/dbt/graph.py
```diff
@@ -52,6 +51,14 @@ class DbtNode:
...
+
+    @property
+    def select(self) -> str:
```
tatiana, line 59: I believe it can be confusing to name this property `select`, because of the selectors logic. Can you think about alternative names?

Same as before, it would be great to add a link to dbt documentation

tatiana (commented):
Hi @binhnq94 , thank you very much for doing this work!

It is a critical part of Cosmos, and I took some time to review. I am still learning about dbt's aliasses and naming strategies - and I'm slightly concerned that this change may break users or have unintended side-effects. That said, I believe the best way will be for us to release it as part of an alpha release and collect user feedback.

Please, when you have a chance, address the comments inline - I'll be testing it with a few dbt examples and we'll aim to release this as part of an alpha 1.3 release.

tatiana:
@binhnq94 confirming the overall approach we are adopting after this PR:

We will be naming Airflow tasks based on the model and version and no longer aliases since those were designed to reference SQL tables and not as model identifiers.

When we do `dbt ls` in this PR's `example_model_version`, we have the following list of models/seeds:
```
jaffle_shop.customers.v1
jaffle_shop.customers.v2
jaffle_shop.orders
jaffle_shop.staging.stg_customers.v1
jaffle_shop.staging.stg_customers.v2
jaffle_shop.staging.stg_orders.v1
jaffle_shop.staging.stg_payments
jaffle_shop.raw_customers
jaffle_shop.raw_orders
jaffle_shop.raw_payments
```

When using Cosmos 1.2.0 - 1.2.4, we rendered based on the alias, which meant that  `customers_v1` and `customers_v2` were not correctly rendered in the Airflow DAG (they were named `customers_USA_v1` and `customers_USA_v2`:
Screenshot 2023-11-16 at 12 35 09

When Cosmos attempted to run the dbt command, it failed to find the models with those alias-based names, although the task succeeded:

```
Running command: ['/Users/tati/Code/cosmos-clean/astronomer-cosmos/venv-38/bin/dbt', 'run', '--models', 'customers_USA_v1', '--profiles-dir', '/var/folders/td/522y78v91d1f5wgh67mj3p0m0000gn/T/tmpoi48p9to', '--profile', 'default', '--target', 'dev']
```

Since dbt does not recognise the alias as a model:
```
[2023-11-16T12:32:37.420+0000] {subprocess.py:94} INFO - (astronomer-cosmos) - 12:32:37  The selection criterion 'customers_USA_v1' does not match any nodes
```

With the change introduced by this PR, we can render the `example_model_version` correctly and execute all its tasks when triggering the DAG, regardless of whether they define aliases or not:
Screenshot 2023-11-16 at 12 16 34

```
[2023-11-16T12:42:57.339+0000] {subprocess.py:73} INFO - (astronomer-cosmos) - Running command: ['/Users/tati/Code/cosmos-clean/astronomer-cosmos/venv-38/bin/dbt', 'run', '--models', 'customers.v1', '--profiles-dir', '/var/folders/td/522y78v91d1f5wgh67mj3p0m0000gn/T/tmpx23gphsc', '--profile', 'default', '--target', 'dev']
```

That matches an existing model and runs it successfully:
```
[2023-11-16T12:42:59.873+0000] {subprocess.py:94} INFO - (astronomer-cosmos) - 12:42:59  1 of 1 START sql table model public.customers_USA_v1 ........................... [RUN]
[2023-11-16T12:42:59.873+0000] {subprocess.py:94} INFO - 12:42:59  1 of 1 START sql table model public.customers_USA_v1 ........................... [RUN]
[2023-11-16T12:42:59.873+0000] {subprocess.py:94} INFO - (astronomer-cosmos) - 12:42:59  1 of 1 START sql table model public.customers_USA_v1 ........................... [RUN]
[2023-11-16T12:42:59.968+0000] {subprocess.py:94} INFO - (astronomer-cosmos) - 12:42:59  1 of 1 OK created sql table model public.customers_USA_v1 ...................... [SELECT 100 in 0.09s]
[2023-11-16T12:42:59.968+0000] {subprocess.py:94} INFO - 12:42:59  1 of 1 OK created sql table model public.customers_USA_v1 ...................... [SELECT 100 in 0.09s]
[2023-11-16T12:42:59.968+0000] {subprocess.py:94} INFO - (astronomer-cosmos) - 12:42:59  1 of 1 OK created sql table model public.customers_USA_v1 ...................... [SELECT 100 in 0.09s]
[2023-11-16T12:42:59.981+0000] {subprocess.py:94} INFO - (astronomer-cosmos) - 12:42:59
[2023-11-16T12:42:59.981+0000] {subprocess.py:94} INFO - 12:42:59
[2023-11-16T12:42:59.981+0000] {subprocess.py:94} INFO - (astronomer-cosmos) - 12:42:59
[2023-11-16T12:42:59.981+0000] {subprocess.py:94} INFO - (astronomer-cosmos) - 12:42:59  Finished running 1 table model in 0 hours 0 minutes and 0.23 seconds (0.23s).
```

I'm confident with this change, and once you address the minor feedback & rebase, we can merge it into the main. Since this is a fix, we can release it in a micro release of Cosmos (e.g. 1.2.5).

tatiana:
@binhnq94 I'd love your thoughts on this ticket I logged as a follow up: https://github.com/astronomer/astronomer-cosmos/issues/682

cosmos/dbt/graph.py
tatiana: Perhaps, we could name this as `resource_name` or `defined_in`:
https://docs.getdbt.com/reference/resource-properties/versions#defined_in

I'm also happy for us to add support for `defined_in` in the future, depending on Cosmos users' needs, since this PR fixes a bug in Cosmos 1.2.

cosmos/dbt/graph.py
binhnq94: Updated.

cosmos/dbt/graph.py
binhnq94: `defined_in` is not in outputs of db ls.
```
DIVISION='abc' dbt ls --output json | grep '"resource_type": "model"'

{"name": "customers", "resource_type": "model", "package_name": "jaffle_shop", "original_file_path": "models/customers_v1.sql", "unique_id": "model.jaffle_shop.customers.v1", "alias": "customers_abc_v1", "config": {"enabled": true, "alias": "customers_abc_v1", "schema": null, "database": null, "tags": [], "meta": {}, "group": null, "materialized": "table", "incremental_strategy": null, "persist_docs": {}, "quoting": {}, "column_types": {}, "full_refresh": null, "unique_key": null, "on_schema_change": "ignore", "on_configuration_change": "apply", "grants": {}, "packages": [], "docs": {"show": true, "node_color": null}, "contract": {"enforced": false}, "post-hook": [], "pre-hook": []}, "tags": [], "depends_on": {"macros": [], "nodes": ["model.jaffle_shop.stg_customers", "model.jaffle_shop.stg_orders", "model.jaffle_shop.stg_payments"]}}
{"name": "customers", "resource_type": "model", "package_name": "jaffle_shop", "original_file_path": "models/customers.sql", "unique_id": "model.jaffle_shop.customers.v2", "alias": "customers_abc_v2", "config": {"enabled": true, "alias": "customers_abc_v2", "schema": null, "database": null, "tags": [], "meta": {}, "group": null, "materialized": "table", "incremental_strategy": null, "persist_docs": {}, "quoting": {}, "column_types": {}, "full_refresh": null, "unique_key": null, "on_schema_change": "ignore", "on_configuration_change": "apply", "grants": {}, "packages": [], "docs": {"show": true, "node_color": null}, "contract": {"enforced": false}, "post-hook": [], "pre-hook": []}, "tags": [], "depends_on": {"macros": [], "nodes": ["model.jaffle_shop.stg_customers", "model.jaffle_shop.stg_orders", "model.jaffle_shop.stg_payments"]}}
```

cosmos/dbt/graph.py
```diff
@@ -51,6 +50,23 @@ class DbtNode:
...
+        """
+        return self.unique_id.split(".", 2)[2]
+
+    @property
+    def name(self) -> str:
+        """
+        Use this property as the task name or task group name.
+        Replace period (.) with underscore (_) due to versioned models.
+        """
+        return self.resource_name.replace(".", "_")
```
binhnq94, line 68: @tatiana It is possible to related to #657.

tatiana (approved):
Thanks for fixing this and addressing all the feedback, @binhnq94 , great work!

cosmos/dbt/graph.py
tatiana: To support this, it will require a bit more work, including parsing the manifest generated during `dbt_ls` - and we'd have additional work to support the custom load method.

I've logged a ticket so we don't lose track of this discussion:
https://github.com/astronomer/astronomer-cosmos/issues/685

tatiana:
@binhnq94 This was released in:

PR #671, opened 2023-11-14
Store compiled_sql even when task fails (fixes issue #369)

### LINKED ISSUE #369: Rendered SQL is not available for failed runs.

agreenburg:
## Description

Updated DbtLocalBaseOperator code to store compiled_sql prior to exception handling so that when a task fails the compiled_sql can still be reviewed.

In the process found and fixed a related bug where compiled_sql was being dropped on some operations due to the way that the `full_refresh` field was being added to the template_fields.

## Related Issue(s)

closes #369

Fixes bug introduced in https://github.com/astronomer/astronomer-cosmos/pull/623 where compiled_sql was being lost in `DbtSeedLocalOperator` and `DbtRunLocalOperator`

## Breaking Change?

Not a breaking change

## Checklist

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -139,7 +139,6 @@ class DbtLocalBaseOperator(DbtBaseOperator):
         Gets called after every dbt run.
         """
         if not self.should_store_compiled_sql:
-            logger.info("should_store_compiled_sql is False. Skipping")
             return
 
         compiled_queries = {}
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -139,6 +139,7 @@ class DbtLocalBaseOperator(DbtBaseOperator):
         Gets called after every dbt run.
         """
         if not self.should_store_compiled_sql:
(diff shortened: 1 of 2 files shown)


agreenburg:
I wasn't sure if additional unit tests are needed for this. I can add them if so.

tatiana (approved):
This looks great, thanks for the quick fix, @agreenburg !

tatiana:
Thanks for the contribution, @agreenburg , we released it as part of Cosmos 1.2.4 yesterday:
https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.2.4

Issue #535, opened 2023-09-14
Create ExecutionMode.KUBERNETES example DAG & setup CI


tatiana:
To avoid our documentation becoming outdated in incompatible with the latest version of Cosmos, as described in #534, we should:

* [ ] Create an example DAG, add it to the folder:
https://github.com/astronomer/astronomer-cosmos/tree/main/dev/dags

* [ ] Reference the new example DAG in the docs, using rst code block:
https://github.com/astronomer/astronomer-cosmos/blob/main/docs/getting_started/execution-modes.rst?plain=1#L58-L61
https://github.com/astronomer/astronomer-cosmos/blob/main/docs/getting_started/kubernetes.rst

* [ ] Setup credentials to a K8s cluster that the CI can use to spin up the necessary resources

* [ ] Update information so the CI runs this example DAG as part of the expensive integration tests:
https://github.com/astronomer/astronomer-cosmos/blob/bcf7714bec09caf3a3ffec9c6c2e72bf43e1c7cf/pyproject.toml#L174
https://github.com/astronomer/astronomer-cosmos/blob/bcf7714bec09caf3a3ffec9c6c2e72bf43e1c7cf/pyproject.toml#L183

qimumu9406:
@tatiana  thanks you! Outdated documentation blocks POC using Cosmos.Looking forward to updated documentation

tatiana:
An option could be to bring the [Kubernetes example DAG](https://github.com/astronomer/cosmos-example/blob/main/dags/jaffle_shop_kubernetes.py) to be part of the astronomer-cosmos repo and automate the steps we describe in our [docs](https://astronomer.github.io/astronomer-cosmos/getting_started/kubernetes.html) in Github Actions, using: https://github.com/marketplace/actions/kubernetes-kind-cluster

tatiana:
This ticket is still relevant - since approximately 30% of Cosmos users use this mode

tatiana:
It seems Github actions would allow us to spin up a KinD cluster - so we could automate running the DAG that we mention in our example (but is currently in a separate repo):
https://github.com/marketplace/actions/kubernetes-kind-cluster

pankajastro:
I drafted a PR (https://github.com/astronomer/astronomer-cosmos/issues/535) for this. The automation script works fine locally, but the Postgres instance is not healthy when running in the GitHub Action.
[...]

PR #673, opened 2023-11-14
Kubernetes on warning callback

david-mag:
## Description
In order to make `on_warning_callback` work with pod operators, we need to read the logs of the dbt test runs. This is done by making sure the pod is kept alive, and `on_success_callback` the log is read and anlayzed for warnings.
Afterwards the pod is cleaned up, based on the original settings from the user.

If `on_warning_callback` is not set, everything stays the way it always was.

I've tested the changes in my dev environment, but haven't added any new tests, since this would be something (I think) we would need to add integration tests for.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,7 +53,7 @@ repos:
           - --py37-plus
           - --keep-runtime-typing
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.5
+    rev: v0.1.6
     hooks:
       - id: ruff
         args:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,31 @@ Features
 * Add ``DbtDocsGCSOperator`` for uploading dbt docs to GCS by @jbandoro in #616
 
 
+1.2.4 (2023-11-14)
+------------------
+
+Bug fixes
+
+* Store ``compiled_sql`` even when task fails by @agreenburg in #671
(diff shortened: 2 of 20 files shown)


tatiana (commented):
Hey, @david-mag, this looks very good - it's great to have feature parity to use `on_warning_callback` on Kubernetes.

I'm apprehensive about the need for tests. An integration test would be ideal - and we could do that once #535 is completed.

In the meantime, how would you feel about adding some unit tests to the new methods and the class constructor?

david-mag:
I've added some unit-tests that (I hope) will cover my additions to the `DbtTestKubernetesOperator`. I had to, however, add a condition to skip the test for airflow version 2.3, because there the `OnFinishAction` from `airflow.providers.cncf.kubernetes.utils.pod_manager` doesn't exist yet.

So using `on_warning_callback` will only be usable from airflow version 2.4 and up.

tatiana (approved):
@david-mag thanks for the contribution!

This will improve K8s users experience!

There will be scenarios when `OnFinishAction` won't be defined. If the user sets `on_warning_callback` and uses `ExecutionMode.KUBERNETES`,  `DbtTestKubernetesOperator` will only work if users use `apache-airflow-providers-cncf-kubernetes >= 7.4.0`. I believe this is a fine compromise, considering that before this PR `on_warning_callback` wasn't working for any version of the provider when users used  `ExecutionMode.KUBERNETES`.

tatiana:
@david-mag This was released in 1.2.5 🎉 

PR #678, opened 2023-11-16
Fix ExecutionConfig.dbt_executable_path to use default_factory

jbandoro:
## Description

I was seeing unit tests failing locally for `dbt.test_graph.test_load_via_dbt_ls_with_invalid_dbt_path` because I have dbt installed on my system and after the fix in #666 this test will fail since it fallbacks to the system dbt if it exists instead of "dbt" which the test is expecting.

This PR changes `ExecutionConfig.dbt_executable_path` to use a dataclass default factory for `get_system_dbt` so it is not called when ExecutionConfig is defined and allows us to more easily patch the shutil.which in the test so it won't fail if there is a system dbt installation.

## Related Issue(s)
None

## Breaking Change?
None

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -258,7 +258,7 @@ class ExecutionConfig:
 
     execution_mode: ExecutionMode = ExecutionMode.LOCAL
     test_indirect_selection: TestIndirectSelection = TestIndirectSelection.EAGER
-    dbt_executable_path: str | Path | None = None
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Thanks for this fix, @jbandoro !

tatiana:
@jbandoro  This was released in 1.2.5 🎉 

PR #681, opened 2023-11-16
Make `profiles_yml_path` optional for `ExecutionMode.DOCKER` and `KUBERNETES`

### LINKED ISSUE #656: Raise an exception if `ProfileConfig` is instantiated with both `profile_mapping` and `profiles_yml_filepath`
As of Cosmos 1.2.2, it does not complain if users set both `profile_mapping` and `profiles_yml_filepath` in a `ProfileConfig` instance. Cosmos should raise an error in this scenario.

This issue was noticed in the Slack thread:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1699365889141509

### LINKED ISSUE #680: Running in k8s or docker mode should not require a profile.yml file on the executing node.
When running in k8s or docker execution modes, running cosmos with a profiles_yml_path defined in ProfileConfig causes dag rendering to fail due to the file not existing.

This validating is occuring in the __post_init__ of ProfileConfig.

It would appear that this validation is not necessary for loading the graph, it is only necessary when running running in Local or Venv execution modes at runtime.

when using method == manifest, as we only need the profiles path to be defined so it can be appended to runtime commands
when running load_via_dbt_ls, it looks to me like we use mock values to actually tun the LS, so the profile is not required here either untill execution time either
for load_via_custom_parser, we pass operator_args from DbtGraph, which contains a projectConfig item, but looking though DbtModel i cant see any usage of this field.

Based on some initial debugging, removing the __post_init__ validation from ProfileConfig does not cause any unit tests or integration tests to fail.

Since it is required in two execution modes though, i propose to delay the validation of this file's existance until we have some context of the execution mode we are operating in.

This would allow a docker or k8s user to embed the profiles yml in their image or mount it in using k8s secrets without failures in dag rendering.

MrBones757:
## Description

This PR moves validation of the profiles_yml_path to later in the dag generation process such that additional context can be gathered from ExecutionConfig to avoid failing unnecessarily when the file does not exist.

## Related Issue(s)

closes #680
closes #656 

## Breaking Change?

N/A

## Checklist

### DIFF
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -11,6 +11,7 @@ from airflow.models.dag import DAG
 from airflow.utils.task_group import TaskGroup
 
 from cosmos.airflow.graph import build_airflow_graph
+from cosmos.constants import ExecutionMode
 from cosmos.dbt.graph import DbtGraph
 from cosmos.dbt.selector import retrieve_by_label
 from cosmos.config import ProjectConfig, ExecutionConfig, RenderConfig, ProfileConfig
@@ -49,7 +50,11 @@ def airflow_kwargs(**kwargs: dict[str, Any]) -> dict[str, Any]:
 
 
 def validate_arguments(
-    select: list[str], exclude: list[str], profile_args: dict[str, Any], task_args: dict[str, Any]
+    select: list[str],
+    exclude: list[str],
+    profile_config: ProfileConfig,
+    task_args: dict[str, Any],
+    execution_mode: ExecutionMode,
 ) -> None:
     """
     Validate that mutually exclusive selectors filters have not been given.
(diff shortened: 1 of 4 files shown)


cosmos/converter.py
```diff
@@ -69,9 +75,12 @@ def validate_arguments(
 
     # if task_args has a schema, add it to the profile args and add a deprecated warning
     if "schema" in task_args:
-        profile_args["schema"] = task_args["schema"]
+        profile_config.profile_mapping.profile_args["schema"] = task_args["schema"]
```
tatiana, line 78: Don't we need to check here if `profile_config.profile_mapping` exists as well?

tatiana (commented):
This is looking great, @MrBones757 , one minor comment, and we're good to merge!

tatiana (approved):
Thanks, @MrBones757 , when using `ExecutionMode.KUBERNETES` and `LoadMode.DBT_MANIFEST`, we really should not enforce the `profiles_yml_path` to be set.

tatiana:
@MrBones757 This was released in 1.2.5 🎉 

Issue #538, opened 2023-09-14
Vertica adapter support


perttus:
Is it possible to add support for Vertica? There is already [dbt-vertica](https://github.com/vertica/dbt-vertica) adapter for it.

jlaneve:
Hey @perttus, yes it's definitely possible! Would you be open to contributing a profile mapping? They're meant to be fairly straightforward to contribute, here's an example 🙂 

https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/profiles/postgres/user_pass.py

perttus:
I just made the PR for Vertica

PR #688, opened 2023-11-17
FIX: vertica profile username

perttus:
## Description

Vertica requires `username` instead of `user`. Without the fix Vertica profile doesn't work, tested with version 1.3.0a1

## Related Issue(s)

#538


jlaneve (approved):
ah great catch!

tatiana:
@perttus there's a pre-release of Cosmos with this change:

* https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.3.0a2
* https://pypi.org/project/astronomer-cosmos/1.3.0a2/

PR #702, opened 2023-11-23
Prevent overriding dbt profile fields with profile args of "type" or "method"

### LINKED ISSUE #696: ProfileMapping `profile_args` should raise a warning if they override `mapped_params`
This issue was found in the Slack thread [here](https://apache-airflow.slack.com/archives/C059CC42E9W/p1700501214811659?thread_ts=1700195349.394989&cid=C059CC42E9W) where the user was using the `GoogleCloudServiceAccountDictProfileMapping` with the `profile_args` similar to:

```python
        profile_name="default",
        target_name="dev",
            conn_id="google_cloud_connection",
            profile_args={"method": "service_account"},
        ),
)
```
This would result in the [dbt_profile_method](https://github.com/astronomer/astronomer-cosmos/blob/d28d3df7eacac74d0fbb016be70434bcc30c51be/cosmos/profiles/base.py#L32) class var that is mapped to dbt profile fields below:

https://github.com/astronomer/astronomer-cosmos/blob/d28d3df7eacac74d0fbb016be70434bcc30c51be/cosmos/profiles/base.py#L199-L211

to then be replaced when the profile is created below:

https://github.com/astronomer/astronomer-cosmos/blob/d28d3df7eacac74d0fbb016be70434bcc30c51be/cosmos/profiles/bigquery/service_account_keyfile_dict.py#L48C16-L52C10

because the merging of dictionaries with `profile_args={"method": "service_account"}` will replace the `mapped_params` that has `{"method": ""service-account-json""}`.

This issue is not limited to `GoogleCloudServiceAccountDictProfileMapping` and can happen with all of the profiles where there is overlap in the mapped params and the user provided profile args.

Cosmos should raise a warning if this happens to warn the user that they are replacing a cosmos mapped parameter.

jbandoro:
## Description

The issue is described in #696 which was discovered when a user was creating a ProfileConfig with `GoogleCloudServiceAccountDictProfileMapping(profile_args={"method": "service-account"})` which was overriding the dbt profile method:
https://github.com/astronomer/astronomer-cosmos/blob/24aa38e528e299ef51ca6baf32f5a6185887d432/cosmos/profiles/bigquery/service_account_keyfile_dict.py#L21

when the profile args are mapped to the created profile below:

https://github.com/astronomer/astronomer-cosmos/blob/24aa38e528e299ef51ca6baf32f5a6185887d432/cosmos/profiles/bigquery/service_account_keyfile_dict.py#L42-L52

This is not an issue with the profile mapping example above and could happen with any profile mapping  by changing the "type" from `dbt_profile_type` or "method" (if used) from `dbt_profile_method` in the class.

The fix in this PR is to not allow args with "type" or "method" that are different from the class variables in `profile_args`.

I think this is better than logging a warning because if either of those fields are different the dbt run with the created profile will fail anyways. This also allows backwards compatibility in the case users have these already set in their profile args and it matches the class variables.

## Related Issue(s)

closes #696 

## Breaking Change?

None

## Checklist


tatiana (approved):
This looks great, thanks, @jbandoro for fixing this so quickly!

tatiana:
@jbandoro  This was released in 1.2.5 🎉 

PR #725, opened 2023-11-29
add pydantic only for apache-airflow 2.6

pixie79:
## Description

Adds pydantic as a requirement only if airflow version is 2.6.

## Related Issue(s)

closes #654

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ classifiers = [
 dependencies = [
     "aenum",
     "attrs",
+    "pydantic>=1.10.0,<2.0.0; apache-airflow==2.6.*",
     "apache-airflow>=2.3.0",
     "importlib-metadata; python_version < '3.8'",
     "Jinja2>=3.0.0",


pixie79:
First attempt failed when I ran the test. This second version passed the following sets of tests:
* hatch run tests.py3.10-2.7:test-integration
* hatch run tests.py3.10-2.6:test-integration

jlaneve:
@pixie79 I may be missing something here, but doesn't this add Pydantic as a dependency for all versions of Airflow? Or where is the logic to restrict this to just Airflow 2.6?

pixie79:
It adds it as optional rather than required which allows both sets of test to pass. Also then allows the restrictions on mwaa to work. 
As far as it would appear there is no way to list it as only optional for 2.6. I did look at adding it lower in the texts section where it would only get installed for 2.6 but that doesn’t seem correct to just have it for a test otherwise it would not be needed at all

jlaneve:
> It adds it as optional rather than required which allows both sets of test to pass. Also then allows the restrictions on mwaa to work. As far as it would appear there is no way to list it as only optional for 2.6. I did look at adding it lower in the texts section where it would only get installed for 2.6 but that doesn’t seem correct to just have it for a test otherwise it would not be needed at all

Got it, that makes sense to me. So we're essentially asking the user to decide for themself if they need Pydantic? And if they decide they do, they just change their requirement to `astronomer-cosmos[...,pydantic]`?

pixie79:
Yup, which i think is fair, what is odd is by just doing that change the tests passed and I am not really sure why they should?

tatiana:
@jlaneve I think @pixie79 's is a fair assumption - and we can update the tests that need to use the pydantic restriction.

pixie79:
Does this or need to be switched to be against main rather than the remove-pydantic branch?

tatiana:
@pixie79 this change, by itself, will break some integration tests. I think our CI is set to only run integration tests on PRs to the `main` branch; for this reason, we're not seeing it.

If you'd like to see the integration tests running and fix them - please go ahead and update your PR to be based on `main`. Once the tests in your PR to the main pass, we can close PR #654 and merge your fix.

Otherwise, you can keep your PR as it is a co-author on the original PR #654. This will mean an additional PR to fix the integration tests, which will break once your PR is merged to #654.

The first approach of making a PR in the main is probably more pragmatic; let us know your preference - we're happy either way.

Issue #684, opened 2023-11-16
Add support to select using (some) graph-operators when using `LoadMode.CUSTOM` or `LoadMode.DBT_MANIFEST`


tatiana:
**Context**

As of Cosmos 1.2.4, users can use graph selectors only when using `LoadMode.DBT_LS`. Examples of some `select` statements that were requested to also work with `LoadMode.CUSTOM` or `LoadMode.DBT_MANIFEST` (but only work with  `LoadMode.DBT_LS` ATM):

```
+model_a
model_b+
+model_c+
```

https://docs.getdbt.com/reference/node-selection/graph-operators
https://docs.getdbt.com/reference/node-selection/set-operators

**Acceptance criteria**

* Support selection of model by name
* Support the selection of models by name & their children
* Support the selection of models by name & their parents
* Follow up on https://astronomer.slack.com/archives/C0599CBNBD3/p1700154259406549

tatiana:
The only parts of the code I'd expect to be changed are:
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/dbt/selector.py
https://github.com/astronomer/astronomer-cosmos/blob/main/tests/dbt/test_selector.py

PR #728, opened 2023-11-30
Add support to select using graph-operators when using `LoadMode.CUSTOM` or `LoadMode.DBT_MANIFEST`

tatiana:
Add support for the following when using `LoadMode.CUSTOM` or `LoadMode.DBT_MANIFEST`:

* Support selection of model by name
* Support the selection of models by name & their children (with or without degrees)
* Support the selection of models by name & their parents (with or without degrees)
* Support intersections and unions involving graph selectors (with or without other supported selectors, eg. tags)

Examples of select/exclusion statements that now work regardless of the `LoadMode` being used:

```
model_a
+model_b
model_c+
+model_d+
2+model_e
model_f+3
model_f+,tag:nightly
```

Related dbt documentation:
https://docs.getdbt.com/reference/node-selection/graph-operators
https://docs.getdbt.com/reference/node-selection/set-operators

Limitations:
* The at operator is not supported yet (`@`)
* If users opt to use graph selector, it will increase the DAG parsing time and the  task execution time when using `LoadMode.CUSTOM` or `LoadMode.DBT_MANIFEST`

This PR improves and extends the original implementation proposed by @tseruga in #429. Some of the changes that were introduced on top of the original PR:
* Add support to descendants (before only precursors were supported)
* Add support to different depths/degrees of precursors/descendants
* Add support to the union between graph operators and graph/non-graph operators
* Add support to the intersection between graph operators and graph/non-graph operators

Closes: #684

Co-authored-by: Tyler Seruga <t.seruga@m1finance.com>

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 import copy
 import re
+from collections import defaultdict
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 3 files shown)


cosmos/dbt/selector.py
```diff
@@ -16,11 +18,154 @@
...
+                previous_generation = new_generation
+                depth -= 1
+
+    def select_node_descendants(self, nodes: dict[str, DbtNode], root_id: str, selected_nodes: set[str]) -> None:
+        """
+        Parse original nodes and add the descendant nodes related to this config to the selected_nodes set.
+
+        :param nodes: Original dbt nodes list
+        :param root_id: Unique identifier of self.node_name
+        :param selected_nodes: Set where precursor nodes will be added to.
```
harels, line 117:
```suggestion
        :param selected_nodes: Set where descendant nodes will be added to.
```

harels (approved):
minor doc string comment, lgtm

Issue #727, opened 2023-11-29
Use preinstalled deps from cicd instead of rerunning


DanMawdsleyBA:
Currently I use cicd to run the dbt parse and dbt deps command. Such that all the packages have been preinstalled however when the airflow dag parses it still says it's missing packages. Possibly we create the syslink copies we don't copy accross the dbt_packages folder?

![image](https://github.com/astronomer/astronomer-cosmos/assets/134394384/05028085-3ff0-45a5-a166-5c98851796b1)

DanMawdsleyBA:
I have already set the deps flag to False. I can see the cosmos recognizes the original folder and can see the dbt_packages. I think it might be something to do with the temp directory. Any ideas what might be the issue @tatiana happy to implement a fix for this

DanMawdsleyBA:
Also seems like dbt 1.7 behavoiour might of changed. When I have any form dependancy I get the below error:
![image](https://github.com/astronomer/astronomer-cosmos/assets/134394384/fc8ed838-5ed5-49eb-a734-241c08fc13d7)

PR #730, opened 2023-11-30
Adding dbt_packages when dbt_deps is False

DanMawdsleyBA:
## Related Issue(s)

https://github.com/astronomer/astronomer-cosmos/issues/727

## Checklist

### DIFF
--- a/cosmos/dbt/project.py
+++ b/cosmos/dbt/project.py
@@ -6,13 +6,17 @@ from cosmos.constants import (
     DBT_LOG_DIR_NAME,
     DBT_TARGET_DIR_NAME,
 )
+from cosmos.config import RenderConfig
 from contextlib import contextmanager
 from typing import Generator
 
 
 def create_symlinks(project_path: Path, tmp_dir: Path) -> None:
     """Helper function to create symlinks to the dbt project files."""
-    ignore_paths = (DBT_LOG_DIR_NAME, DBT_TARGET_DIR_NAME, "dbt_packages", "profiles.yml")
+    if RenderConfig.dbt_deps:
+        ignore_paths = (DBT_LOG_DIR_NAME, DBT_TARGET_DIR_NAME, "profiles.yml")
+    else:
+        ignore_paths = (DBT_LOG_DIR_NAME, DBT_TARGET_DIR_NAME, "dbt_packages", "profiles.yml")
     for child_name in os.listdir(project_path):
         if child_name not in ignore_paths:
             os.symlink(project_path / child_name, tmp_dir / child_name)
--- a/cosmos/dbt/project.py
+++ b/cosmos/dbt/project.py
@@ -6,15 +6,14 @@ from cosmos.constants import (
     DBT_LOG_DIR_NAME,
     DBT_TARGET_DIR_NAME,
 )
-from cosmos.config import RenderConfig
 from contextlib import contextmanager
 from typing import Generator
 
 
-def create_symlinks(project_path: Path, tmp_dir: Path) -> None:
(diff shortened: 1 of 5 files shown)


cosmos/dbt/project.py
```diff
@@ -6,13 +6,16 @@
...
 from contextlib import contextmanager
 from typing import Generator
 
 
 def create_symlinks(project_path: Path, tmp_dir: Path) -> None:
     """Helper function to create symlinks to the dbt project files."""
-    ignore_paths = (DBT_LOG_DIR_NAME, DBT_TARGET_DIR_NAME, "dbt_packages", "profiles.yml")
+    ignore_paths = [DBT_LOG_DIR_NAME, DBT_TARGET_DIR_NAME, "profiles.yml"]
+    if RenderConfig.dbt_deps is False:
+        ignore_paths.append("dbt_packages")
```
tatiana, line 18: @DanMawdsleyBA confirming our intended behaviour:

1. when the user specifies `dbt_deps == True`, since Cosmos installs the deps, we don't want to create the symlink to `dbt_packages`
2. if the user sets  `dbt_deps == False`, we assume the deps are pre-installed, so we want to create symlinks

Is this the case?

DanMawdsleyBA:
@tatiana Would you be able to help me fix the integration test? I'm trying to add a test that manually install dbt deps beforehand and then runs cosmos with dbt_deps is False which should mean it uses the pre-installed dbt-packages

cosmos/dbt/project.py
jbandoro: @DanMawdsleyBA I believe the integration test is failing because your check here is on the class `RenderConfig` instead of the `render_config` object. `RenderConfig.dbt_deps` will always be `True` because that's the default [here](https://github.com/DanMawdsleyBA/astronomer-cosmos/blob/049d99c01483b0689f145c1e7be00e01f62ce439/cosmos/config.py#L54). 

It might make more sense in `create_symlinks` to have an arg for `ignore_dbt_packages` or something like that when it's called here: https://github.com/astronomer/astronomer-cosmos/blob/e1f34ea424b8940f1c06660554298008876ea11b/cosmos/dbt/graph.py#L235

DanMawdsleyBA: @jbandoro that's a good spot thanks! I have updated it to pass through the actual dbt deps. I get your point on the name and it is a little confusing but wonder if adding an additional variable is necessary as the value of dbt deps is directly linked to whether we want to link the packages folder or not. Open to counter suggestion and thoughts @tatiana

tatiana: Thanks, @jbandoro and @DanMawdsleyBA, for following up on this!

I liked the suggestion. What if we renamed the argument from `dbt_deps` to `skip_dbt_packages` (only within `create_symlinks`)? From the `create_symlinks` function, this name would probably make more sense. I'm happy with `ignore_dbt_packages` as well.

The name `dbt_deps` continues to make sense in the context of `load_via_dbt_ls`.

DanMawdsleyBA: Yeah that's a good middle ground I updated the name to ignore_dbt_packages (to align it with the ignore paths name) . I added in a comment for future reference

tatiana (approved):
Thanks a lot, @DanMawdsleyBA , we'll be releasing this as part of an alpha 1.3 this week!

DanMawdsleyBA:
That's great thanks! We have a release on Thursday would be great to have it in time for that. Thanks for your help @tatiana

tatiana:
@DanMawdsleyBA It's part of 1.3.0a3:
https://pypi.org/project/astronomer-cosmos/1.3.0a3/

Please let us know if you find any issues!

PR #732, opened 2023-12-01
Speed up integration tests

jbandoro:
## Description

I was able to speed up the integration tests by caching the dag bag result.

Previously for each parametrized dag run test, it would reparse all of the dags which takes a non-trivial amount of time to parse all of the cosmos example dags. 

On my local machine the total time went from 1616s to 540s, which will save a good amount of GH minutes 😃.

## Related Issue(s)

None

## Breaking Change?

None
## Checklist

### DIFF
--- a/tests/test_example_dags.py
+++ b/tests/test_example_dags.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from pathlib import Path
-from functools import lru_cache
+from functools import cache
 
 import airflow
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_example_dags.py
```diff
@@ -37,6 +38,7 @@ def session():
     return get_session()
 
 
+@lru_cache
```
joppevos, line 41: Could `from functools import cache` be used instead of `lru_cache`?
as I understand from the [docs](https://docs.python.org/3/library/functools.html#functools.lru_cache), it's a wrapper around `lru_cache` to improve readability.

tatiana (approved):
This looks amazing, @jbandoro , thanks a lot for the improvement!
Can't wait to merge this after you reply @joppevos' feedback :)

tests/test_example_dags.py
jbandoro: Didn't know about that 💡, thanks!

PR #733, opened 2023-12-01
Add new parsing method `LoadMode.DBT_LS_FILE`

woogakoki:
## Description

Adds a new load method which is in between manifest and dbt_ls. Goal of this is to have a dbt ls file with the output prepared and only use the parsing logic of dbt_ls load method. This should increase performance compared to using dbt_ls. 
However this method needs one output file for each taskgroup build. 

## Breaking Change?

No breaking changes, it just adds a new load method on top but leaves the rest as is.

## Checklist


docs/configuration/parsing-methods.rst
```diff
@@ -75,6 +77,23 @@ To use this:
...
+----------------
+
+If you provide the output of ``dbt ls --output json`` as a file, you can use this to parse similar to  ``dbt_ls``.
+You can supply a ``dbt_ls_path`` parameter on the DbtDag / DbtTaskGroup with a path to a ``dbt_ls_output.txt`` file.
+
+To use this:
+
+.. code-block:: python
+
+    DbtDag(
```
tatiana, line 90: As mentioned in the broader comment, if we could update a DAG in `dev/dags` and add a reference to that runnable DAG code here, similar to:
https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/basic_cosmos_dag.py#L24
it would be awesome.

Since the example DAGs are always run as part of our test suite, we'd guarantee this excerpt doesn't become outdated.

tatiana:
Hey @woogakoki, we're super close to getting this merged!
Once you address the code coverage comments, we can release it as part of a 1.3 alpha.

docs/configuration/parsing-methods.rst
woogakoki: I've updated the dag you mentioned in the broader comment. Hope that is what you had in mind?

docs/configuration/parsing-methods.rst
```diff
@@ -75,6 +77,23 @@ To use this:
         # ...,
     )
 
+``dbt_ls_file``
+----------------
+
```
tatiana, line 82:
```suggestion

.. note::
   New in Cosmos 1.3.

```

tatiana (commented):
@woogakoki Just added some last comments before we can merge this feature. Can't wait to try this out. Do you by any chance have numbers on the performance improvement by using this method?

cosmos/dbt/graph.py
```diff
@@ -272,6 +273,31 @@ def load_via_dbt_ls(self) -> None:
...
+        if not self.render_config.is_dbt_ls_file_available():
+            raise CosmosLoadDbtException(f"Unable to load dbt ls file using {self.render_config.dbt_ls_path}")
+
+        if not self.execution_config.project_path:
+            raise CosmosLoadDbtException("Unable to load dbt ls file without ExecutionConfig.dbt_project_path")
+
+        with open(self.render_config.dbt_ls_path) as fp:  # type: ignore[arg-type]
+            dbt_ls_output = fp.read()
+            project_path = self.execution_config.project_path
+            nodes = parse_dbt_ls_output(project_path=project_path, ls_stdout=dbt_ls_output)
```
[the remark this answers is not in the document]
woogakoki: Ah okay I see. Changed it 👍

woogakoki:
> @woogakoki Just added some last comments before we can merge this feature. Can't wait to try this out. Do you by any chance have numbers on the performance improvement by using this method?

Unfortunatly not as I only tested it locally

woogakoki:
@tatiana The Type-Check test is failing due to some issue in cosmos/operators/kubernetes.py. Is that known and can be ignored?

tatiana:
@woogakoki Thanks for addressing the feedback. Yes, the issue with the type check was solved by @jbandoro in #766. Once the checks pass, I'll merge your PR!

tatiana (approved):

tatiana:
@woogakoki when you get a chance, please, check the failing tests:
```
FAILED tests/dbt/test_graph.py::test_load_via_dbt_ls_file - cosmos.dbt.graph.CosmosLoadDbtException: Unable to load dbt ls file without RenderConfig.project_path
===== 1 failed, 22 passed, 3 skipped, 330 deselected in 459.96s (0:07:39) ======
```

dev/dags/dbt/jaffle_shop/dbt_ls_models_staging.txt
```diff
@@ -0,0 +1,6 @@
...
+{"name": "stg_payments", "resource_type": "model", "package_name": "jaffle_shop", "original_file_path": "models/staging/stg_payments.sql", "unique_id": "model.jaffle_shop.stg_payments", "alias": "stg_payments", "config": {"enabled": true, "alias": null, "schema": null, "database": null, "tags": [], "meta": {}, "group": null, "materialized": "view", "incremental_strategy": null, "persist_docs": {}, "quoting": {}, "column_types": {}, "full_refresh": null, "unique_key": null, "on_schema_change": "ignore", "on_configuration_change": "apply", "grants": {}, "packages": [], "docs": {"show": true, "node_color": null}, "contract": {"enforced": false}, "post-hook": [], "pre-hook": []}, "tags": [], "depends_on": {"macros": [], "nodes": ["seed.jaffle_shop.raw_payments"]}}
```
tatiana, line 6: Since these nodes depend on the seeds, the example DAG will fail if we try to run it in isolation. WDYT of replacing these by the seed nodes? Or also adding the seed nodes?

woogakoki: @tatiana Ah okay yeah got it. I just created the output using the standard jaffle project. So no problem replacing it.

woogakoki: Okay, I just included the seed nodes in the end 😄

tatiana:
Amazing work, @woogakoki , thank you very much for getting this work through

PR #737, opened 2023-12-02
Add dbt docs natively in Airflow via plugin

### LINKED ISSUE #571: Render DBT Docs
https://apache-airflow.slack.com/archives/C059CC42E9W/p1696010143212889?thread_ts=1695509325.058599&cid=C059CC42E9W

It was requested to have a link to dbt docs, such as are rendered with 
```shell
dbt docs generate
dbt docs serve
```

@chrishronek mentioned that these could be generated and served statically. 
I commented that Airflow can serve static files with a plugin (which could maybe be shipped with cosmos?)

dwreeves:
## Description

This PR adds a plugin (via the Airflow plugins entrypoint) that adds a menu item inside of `Browse` that renders the dbt docs:

image

And this is what it looks like. (This example is inside the dev docker compose):

image

The docs are rendered via an iframe with some additional hacks to make the page render in a user friendly way. I chose an iframe over vendoring the `index.html` in the templates for a few reasons, but mostly to support custom `{% block __overview__ %}` text. However, extracting the text from `index.html` and rendering it in a custom page is certainly an option too.

The dbt docs are specified in the Airflow config with the following parameters:

```ini
[cosmos]
dbt_docs_dir = path/to/docs/here
dbt_docs_conn_id = my_conn_id
```

Note that the path can be a link to any of the following:

- S3
- Azure Blob Storage
- Google Cloud Storage
- HTTP/HTTPS
- Local storage

This is designed to work with the operators that dump the dbt docs, and the documentation changes I added make that clear.

Lastly, if docs are not hooked up, a message comes up telling the user that they should set their dbt docs up:

image

### Current limitations

- Most importantly, **I need help testing the S3 / Azure / GCS integrations.** I _think_ I got them right but I'll need someone to actually try them.
- **I also wouldn't mind some help testing the UI on more browsers.** I've tested both Firefox and Chrome.
- **The iframe hack is less than ideal; I would preferably want the dbt docs to have a fixed height.** So instead of using the scroll bar of the Airflow UI, use the scroll bar of the dbt docs UI. The issue is basically that I am not an HTML/CSS/JavaScript person. I don't think there is any reason this shouldn't be possible, so I can continue to look into this as the PR is reviewed, or someone else can just do it for me.
- I cannot run tests locally (lots of issues, mostly the databricks DAG in `dev/dags/` fails locally), so I actually have no idea whether the test suite works. I was just planning on letting Github Actions take a stab at it.

### API Decisions

The core maintainers of the repo should provide some feedback on a few high level API decisions:

- **Config variable names:** Let me know if `dbt_docs_dir` and `dbt_docs_conn_id` are appropriate names. Other names could be like, `dbt_docs_path`, or `dbt_docs_dir_conn_id`, or `dbt_docs_path_conn_id`, etc.
- **Location in UI:** I entertained two ideas: (a) Adding a menu button called Cosmos with dbt docs underneath. (b) Adding it under browse. Ultimately I decided on option 2.

## Related Issue(s)

Closes #571.

## Breaking Change?

This PR should not cause any breaking changes.

## Checklist


jlaneve:
I've been excited about this one! I haven't looked at this super in-depth yet, but how does the local filesystem option work? Could I add running `dbt docs` to my build process, bake it into my Airflow image, and read from there?

dwreeves:
> I've been excited about this one! I haven't looked at this super in-depth yet, but how does the local filesystem option work? Could I add running dbt docs to my build process, bake it into my Airflow image, and read from there?

Yes, exactly. For my own professional usage this is the option I am doing.

As part of my deployment process, I `dbt compile` and then specify the `target_path` in the `ProjectConfig`. I like this deployment approach because it relieves some compute pressure on the scheduler since it doesn't need to `dbt ls` from a clean slate on every heartbeat. (Actually, this deployment approach isn't documented; could be something I add to the docs...)

So when this feature is in, I would `dbt docs generate` as part of the deployment process as well. And then I would just .

This approach does have a small downside (and now that I mention it I should note it in the documentation), which is that the `count(*)` doesn't get updated in the docs regularly; it's just using whatever the count was at the last deploy.

dwreeves:
Please bear with me as I just throw some commits at Github Actions in a desperate attempt to get the tests I added working. I cannot for the life of me get the tests working on my local machine. 😓

dwreeves:
OK, sorry for the noise, but tests are working now! Again, I could not get tests set up locally.

dwreeves:
Ah, so I accidentally deleted the `manifest.json` at first because I needed to rename `dags/` to `_dags/` to test locally, then when I renamed back it got covered by `.gitignore`, but I didn't bother to then un-delete it because I wasn't aware integration tests depended on it.

I think the `.gitignore` in `dev/dags/dbt/jaffle_shop` should probably make an exception for that if it's required. I'll re-add it and make that change to the gitignore.

dwreeves:
> Would this work with MWAA?

Yes

> Personally I would prefer to have a nightly DAG scheduled to create and update the docs, so it would be great to set a docs path in the default bucket rather than have it read from the DAG directory on S3.
> I have DAGS in one directory and DBT files in another.

This is supported; the docs additions explain how to use docs in S3.

dwreeves:
@tatiana Thank you for deploying integration tests.

It looks like dbt-core's dependency lock of `jinja2~=3.1.2` does not play nicely with flask<2.3 being used by Airflow 2.3, which causes the integration test to fail when compiling the tests. Technically this dependency conflict had already existed before; this is just surfacing the issue more explicitly.

What do you want to do about this? The easiest option is to skip the plugin unit tests for Airflow 2.3 and 2.4 and lazy load the offending Flask modules.

tatiana:
> @tatiana Thank you for deploying integration tests.
> 
> It looks like dbt-core's dependency lock of `jinja2~=3.1.2` does not play nicely with flask<2.3 being used by Airflow 2.3, which causes the integration test to fail when compiling the tests. Technically this dependency conflict had already existed before; this is just surfacing the issue more explicitly.
> 
> What do you want to do about this? The easiest option is to skip the plugin unit tests for Airflow 2.3 and 2.4 and lazy load the offending Flask modules.

@dwreeves Did you consider importing `escape` from MarkupSafe instead of Jinja (`from markupsafe import escape`)? It seems this is a recommendation since 3.1:

https://jinja.palletsprojects.com/en/3.1.x/changes/#version-3-1-0

dwreeves:
I'm not importing escape, one of the dependencies is. 😬

dwreeves:
@tatiana I am trying a little hack to get integration tests working:

```python
try:
    import flask  # noqa: F401
except ImportError:
    import markupsafe
    import jinja2

    jinja2.Markup = markupsafe.Markup
    jinja2.escape = markupsafe.escape
```

😅

Thanks for bearing with me!

dwreeves:
Ah, code coverage isn't up to snuff with these changes. (I was not aware because I'm not able to test anything locally! 😓) I think I know which chunks of the code do not have coverage. I'm not 100% sure how to test them but I'll see what I can do.

dwreeves:
@tatiana I've covered as much as I reasonably can. This should get it past the coverage threshold.

image

Writing the tests was more work than writing the integration. 😅

dwreeves:
It looks like my coverage was 1 line short of meeting the threshold required to pass the coverage requirements. I'll be able to get to this tonight or tomorrow. Thank you for your patience.

dwreeves:
OK, there is no way the coverage checks don't pass this time.... right? cc @tatiana

dwreeves:
What's the status of this? 🤔 The issue we left off at was that @tatiana was having some issues with the HTTP method for retrieving files, but I was unable to replicate it or figure out what the problem was (it worked on my end...). We both were able to confirm that the local method worked, and the GCS/S3/Azure methods have not been fully integration tested.

jbandoro (approved):
> What's the status of this? 🤔 The issue we left off at was that @tatiana was having some issues with the HTTP method for retrieving files, but I was unable to replicate it or figure out what the problem was (it worked on my end...). We both were able to confirm that the local method worked, and the GCS/S3/Azure methods have not been fully integration tested.

Thanks for another great contribution @dwreeves, this is awesome!

I confirmed that the HTTP, local file, and GCS paths work with docker-compose. @tatiana mentioned to me that the HTTP path wasn't working when running [airflow standalone](https://airflow.apache.org/docs/apache-airflow/stable/start.html#quick-start), and I had an issue where the webserver process kept exiting with signal 11 when trying to view the dbt docs when I was using the example docs dir you put in the docker-compose [here](https://github.com/astronomer/astronomer-cosmos/blob/6f39181ef3f568904041022755e2f1065dcb97ca/dev/docker-compose.yaml#L13), but it did work with a local file path in airlfow standalone mode.  

Airflow recommends against using airflow standalone in production, and issues could be related to local system, and since this worked well in my testing with docker-compose, happy to approve with the minor comment below fixed before merging.

dwreeves:
Thanks for the approval. Just updated.

ms32035:
@dwreeves how does this work when there are multiple dbt projects/doc dirs?

dwreeves:
@ms32035 I was a little worried someone would ask. It is not supported directly. It could be in theory, although I cannot think of an API+UI design for multiple dbt projects' docs that wouldn't complicate things a lot for users with just one project. I'm all ears for what user-facing interface you'd think is appropriate.

Another solution, which I find reasonably elegant, is to import projects into a "docs project" so to speak. So you create a dbt project that has a `packages.yml` that looks like this:

```yaml
packages:
- local: ../project1
- local: ../project2
```

It's possible that this pattern could/should be documented. Or we just support multiple projects' docs directly, although again, I don't know how to avoid the complications for the API.

ms32035:
@dwreeves the only design idea I have is to have a list page as an entry point, where you'd have to pre-configure the folders, - a comma separated string in airflow conf. I did something similar here https://github.com/ms32035/airflow-multirepo-deploy/blob/f08d8a5863b3311bf4210d3dc2c835370dd0296d/multirepo-deploy-plugin/multirepo_deploy_plugin.py#L110

dwreeves:
I would only support that solution if, when there is only one project, the dbt docs load normally without a list. So the list is just there if you have multiple projects.

Making each attribute into a comma separated list is that gets zip()'d is something I briefly considered. It does help make things simple for single project users, although it feels weird and slightly inexplicit as a data model. That's the trade-off, basically.

PR #738, opened 2023-12-03
Reduce Mccabe code complexity to 8

joppevos:
## Description

Created this small PR to reduce the maccabe score to 8 and refactored the code flagged as too complex.

I removed flake8, since we can use ruffs maccabe linter.

The following functions have been addressed
```shell
> pre-commit run flake8 --all-files
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

cosmos/airflow/graph.py:196:1: C901 'build_airflow_graph' is too complex (9)
cosmos/converter.py:101:5: C901 'DbtToAirflowConverter.__init__' is too complex (9)
cosmos/dbt/parser/project.py:277:5: C901 'LegacyDbtProject.__post_init__' is too complex (10)
cosmos/dbt/selector.py:197:1: C901 'select_nodes' is too complex (9)

```
I picked out the functions which stood out to me as complex. If we want to reduce it to 6, there are plenty more functions to tackle. For some functions I would consider it unnecessary to drop it to 6. 
```shell
cosmos/config.py:108:9: C901 `__init__` is too complex (7 > 6)
cosmos/dbt/parser/project.py:63:9: C901 `_config_selector_ooo` is too complex (7 > 6)
cosmos/dbt/parser/project.py:98:5: C901 `extract_python_file_upstream_requirements` is too complex (7 > 6)
cosmos/dbt/parser/project.py:165:9: C901 `extract_sql_file_requirements` is too complex (7 > 6)
cosmos/dbt/selector.py:52:9: C901 `load_from_statement` is too complex (7 > 6)
cosmos/dbt/selector.py:119:9: C901 `_should_include_node` is too complex (7 > 6)
cosmos/hooks/subprocess.py:34:9: C901 `run_command` is too complex (8 > 6)
cosmos/operators/base.py:137:9: C901 `get_env` is too complex (7 > 6)
cosmos/operators/base.py:186:9: C901 `add_global_flags` is too complex (7 > 6)
cosmos/operators/local.py:136:9: C901 `store_compiled_sql` is too complex (7 > 6)
cosmos/profiles/base.py:183:9: C901 `get_dbt_value` is too complex (8 > 6)
Found 11 errors.
```

## Related Issue(s)
This PR continues on the amazing work of @jbandoro in on #525 
Related issue #641 

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #743, opened 2023-12-04
Replace flake8 for ruff

joppevos:
## Description
We are using ruff, which has drop-in parity with flake8. 
We can continue using only ruff and shave a few seconds of CI.

## Related Issue(s)

Related #738 
## Checklist


tatiana (approved):
Thanks for isolating this change in a separate PR, @joppevos !

PR #748, opened 2023-12-05
Tests: use temp dir for `test_load_via_dbt_ls_without_dbt_deps_and_preinstalled_dbt_packages`

jbandoro:
## Description

This is a follow up to #730 so that the fixture `tmp_dbt_project_dir` is used for the test instead of the actual project directory since it installs dbt packages that need to be cleaned up.

## Related Issue(s)

None

## Breaking Change?

None

## Checklist


tatiana (approved):
Thanks for fixing this, @jbandoro !

PR #751, opened 2023-12-06
Select models which have tests with additional dependencies

### LINKED ISSUE #719: load_from_dbt_manifest is de-selecting valid test nodes
`NodeSelector._should_include_node` has the following line

```python   
if node.resource_type == DbtResourceType.TEST:
    node.tags = getattr(self.nodes.get(node.depends_on[0]), "tags", [])
```
which gives the test node the tags of its parent model. Obviously this means that we cannot select tests by tag as we can in dbt-core, which would be ideal, but there is another problem.

`node.depends_on[0]` is supposed to return the id of the parent model of the test, however, it actually just selects the first dependency. We have tests with additional dependencies (sources), ie.

```yaml
models:
- name: MY_MODEL
  config:
    tags:
    - mytag
  tests:
  - comparison_to_expected_results:
      expected: source('EXPECTED', 'MY_MODEL')
```

This is valid dbt but now breaks in astronomer-cosmos because the test tags are set to the source tags, not the parent model tags.

Solution:

I have found that `node_dict['refs'][0]['name']` is more reliable for fetching the parent model of the test than `node_dict.get("depends_on", {}).get("nodes", [])` which is used in the `DbtNode` class, although I appreciate that dbt doesn't seem to offer a reliable way to do this.

Alternatively, we could try and remove non-model ids from the list at least to try and prune it down to being more likely to be the parent model rather than a stray source, and warn on lists greater than length 1.

```python
if node.resource_type == DbtResourceType.TEST:
    dependency_ids = [node_id for node_id in node.depends_on if node_id.startswith('model.')]
    parent_id = dependency_ids[-1]
    if len(dependency_ids) > 1:
        logger.warning(
            f'Test node {node.name} has more than one model dependency {dependency_ids}, selected tags from parent assumed to be {parent_id}.')
    parent_tags = getattr(self.nodes.get(parent_id), "tags", [])
    node.tags = list(set(node.tags + parent_tags))
```

Here we can additionally use and filter on the actual test tags too, improving parity with dbt-core --select.

Alternatively, I secretly suspect that `depends_on['nodes'][-1]` will always be the test parent model, instead of `depends_on['nodes'][0]`, although I haven't been through the dbt-core code to prove this.

adammarples:
## Description

Tests such as the relationship test will fail to be selected if we are selecting models with tags and using the manifest loading method.

## Related Issue(s)

closes #719 

## Breaking Change?

Almost all tests will have only one 'depends_on', but when they have two, ie. `relationship`, the true parent is the last one. 

`load_from_dbt_manifest` now considers the primary selectable parent model of a test to be the last item in its 'depends_on' list, rather than the first. 

This would break any tests being loaded by tag selection using `load_from_dbt_manifest` which somehow depended on a model being the first in that list rather than the last.

## Checklist


cosmos/dbt/selector.py
```diff
@@ -295,7 +295,14 @@ def _should_include_node(self, node_id: str, node: DbtNode) -> bool:
         self.visited_nodes.add(node_id)
 
         if node.resource_type == DbtResourceType.TEST:
-            node.tags = getattr(self.nodes.get(node.depends_on[0]), "tags", [])
+            dependency_ids = [node_id for node_id in node.depends_on if node_id.startswith("model.")]
```
tatiana, line 298: Why do we only care about model dependencies? Why don't we care about potential parent seeds/sources tags?

cosmos/dbt/selector.py
```diff
@@ -295,7 +295,14 @@ def _should_include_node(self, node_id: str, node: DbtNode) -> bool:
...
+            parent_id = dependency_ids[-1]
```
tatiana, line 299: Why do we have to pick only one parent? Wouldn't it be better if we took into account all the parents?

In your PR description, you mention:
> Almost all tests will have only one 'depends_on', but when they have two, ie. relationship, the true parent is the last one.

What is the definition of a true parent? If a test depends on two models, A and B, and they are run as independent Airflow tasks, why should the relationship with B be more relevant than the one with A?

Do we have any dbt docs on this?

cosmos/dbt/selector.py
```diff
@@ -295,7 +295,14 @@ def _should_include_node(self, node_id: str, node: DbtNode) -> bool:
...
+            if len(dependency_ids) > 1:
+                logger.warning(
+                    f"Test node {node.name} has more than one model dependency {dependency_ids}, selected tags from parent assumed to be {parent_id}."
```
tatiana, line 302: I liked that you added this warning. Before users were left out about what was happening.

tests/dbt/test_graph.py
```diff
@@ -665,11 +665,35 @@ def test_tag_selected_node_test_exist():
     assert len(dbt_graph.filtered_nodes) > 0
 
     for _, node in dbt_graph.filtered_nodes.items():
-        assert node.tags == ["test_tag"]
+        assert "test_tag" in node.tags
```
tatiana, line 668: It would be great if we could improve the way we're doing this selection without introducing a breaking change.

tatiana (commented):
Thanks for the contribution, @adammarples! I'm glad you're improving this part of the code.

I'll loop in @edgarasnavickas (@navedgaras), who previously worked on this, so we can get his thoughts as well.

I left some questions/comments inline

cosmos/dbt/selector.py
adammarples: This is a good point and should be taken out

tests/dbt/test_graph.py
adammarples: I think this line change is the right test to make, your test should only be asserting that the tag exists, because that is a sufficient condition to get the node included in the graph, it shouldn't be making the too-broad assertion that it knows what all the tags are, which just happens to be right as an implementation detail in how you initially set up your test manifest.

cosmos/dbt/selector.py
adammarples: We have to pick one parent because we're trying to make a run task followed by a test task, and our concern is which tests test which model. Tests can only be declared inside models/sources/seeds etc config blocks, they do explicitly belong to only one thing. For example in the "model.jaffle_shop.orders" model, there is a test on the column customer_id which tests that it has a relationship with the `ref('customers')` model. This is explicitly a test on the orders model, defined under the `- name: orders` section of the model config, the test shouldn't run after the customers table is built too, it isn't a test on that table.

There was a bit of discussion about this here https://github.com/dbt-labs/dbt-core/issues/6746

tatiana: @adammarples what about tests like the dbt_utils [equality](https://github.com/dbt-labs/dbt-utils?tab=readme-ov-file#equality-source), and others, such as:
* equal_rowcount
* cardinality_equality
* fewer_rows_than
* relationships_where

Example:
```
version: 2

models:
  - name: model_name
    tests:
      - dbt_utils.equality:
          compare_model: ref('other_table_name')
          compare_columns:
            - first_column
            - second_column
```

Even though they are declared part of the `model_name` block, they also work on the `other_table_name`.

I didn't check how dbt handles these, but I'd expect it to depend on both models if users use `ref`. Could you confirm the current behaviour and let me know your thoughts on these cases?

adammarples: In this example

```yaml
version: 2

models:
  - name: a

  - name: b
    tests:
      - dbt_utils.equality:
          compare_model: ref('a')
          compare_columns:
            - id
```

The test does depend on both tables a and b. In fact running both of these commands will trigger the test, which I didn't realise.

```zsh
 0  ~/projects/demo/demo> dbt test --models a
11:02:13  Running with dbt=1.7.4
11:02:13  Registered adapter: duckdb=1.7.0
11:02:13  Found 2 models, 1 test, 0 sources, 0 exposures, 0 metrics, 505 macros, 0 groups, 0 semantic models
11:02:13  
11:02:13  Concurrency: 1 threads (target='dev')
11:02:13  
11:02:13  1 of 1 START test dbt_utils_equality_b_id__ref_a_ .............................. [RUN]
11:02:13  1 of 1 PASS dbt_utils_equality_b_id__ref_a_ .................................... [PASS in 0.03s]
11:02:13  
11:02:13  Finished running 1 test in 0 hours 0 minutes and 0.06 seconds (0.06s).
11:02:13  
11:02:13  Completed successfully
11:02:13  
11:02:13  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1

 0  ~/projects/demo/demo> dbt test --models b
11:02:17  Running with dbt=1.7.4
11:02:17  Registered adapter: duckdb=1.7.0
11:02:17  Found 2 models, 1 test, 0 sources, 0 exposures, 0 metrics, 505 macros, 0 groups, 0 semantic models
11:02:17  
11:02:17  Concurrency: 1 threads (target='dev')
11:02:17  
11:02:17  1 of 1 START test dbt_utils_equality_b_id__ref_a_ .............................. [RUN]
11:02:17  1 of 1 PASS dbt_utils_equality_b_id__ref_a_ .................................... [PASS in 0.03s]
11:02:17  
11:02:17  Finished running 1 test in 0 hours 0 minutes and 0.06 seconds (0.06s).
11:02:17  
11:02:17  Completed successfully
11:02:17  
11:02:17  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
```

At the moment, at least when using the manifest load function combined with model tagging as a selection criteria, astronomer cosmos selects only one node to be the 'parent' of the test, which tells it which task to run after, the problem is that it's selecting the wrong one, imo, if we have to select one here it should be `b` but we're currently selecting `a` .

```python
depends_on = manifest['nodes']['test.demo.dbt_utils_equality_b_id__ref_a_.54b4729a08']['depends_on']['nodes']
print(depends_on)
```
```
['model.demo.a', 'model.demo.b']
```
Where we select [0] instead of [-1]

What would be interesting would be then to include all of the dependent model tags, not just one of them, so that the test could appear after model a is run and after model b is run, but that seems like a disaster for the way astronomer cosmos works at the moment, where the graph is ordered by model dependencies, not tests dependencies (I think). That might be well outside of the scope of a change such as this, which really is just about picking the last from the list over the first in the list as being the more important model to test.

tatiana: Thank you for the thorough analysis, @adammarples, and detailed explanation. I appreciate it.

Given the observed behaviour of dbt, wouldn't it be more accurate to keep all the test node parents in `depends_on` - not only the `-1`?

The downside would be that if the user doesn't apply filters, the test would be duplicated in multiple model task groups (if using `TestBehavior.AFTER_EACH`).

However, the benefit is we would not miss a test if the user selected `'model.demo.a'`, as you illustrated.

jbandoro: I agree with @tatiana that it would be better to keep all of the tests assigned, and during execution when running the tests, the user can use the `ExecutionConfig.test_indirect_selection` if they want to configure the dbt test behavior when using indirect selection.

@adammarples in the example you provided above if `ExecutionConfig.test_indirect_selection = TestIndirectSelection.CAUTIOUS` then it would only run the tests on model b. If model b depended on model a then a user could use `TestIndirectSelection.BUILDABLE` which would also run the test only for model b.

ms32035:
@adammarples as you're working on this, I'd like to bring to your attention a case you're missing, and which we're running into recently:
```
  File "/home/airflow/.local/lib/python3.11/site-packages/cosmos/dbt/selector.py", line 298, in _should_include_node
    node.tags = getattr(self.nodes.get(node.depends_on[0]), "tags", [])
                                       ~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```
Although not a dbt best practice, it's entirely possible to have a test with 0 dependencies, or model dependencies (test on sources)

cosmos/dbt/selector.py
adammarples: @jbandoro it looks like this issue for me is actually solved with a piece of TestIndirectSelection config that I wasn't aware of then? Thank you, I guess there is no need for this PR?

tatiana:
Hi @adammarples , this PR - and the discussion - was beneficial. Thank you. And thanks, @jbandoro, for giving an alternative solution to the problem.

Cosmos can improve its implementation towards tests with multiple parents - we should make it consistent with dbt. @adammarples, you make a great start - would you be interested in further contributing to this - or would you rather close this PR and let someone else take over the work?

@ms32035 I didn't realise that was possible! Although it is related to the original problem, I logged a separate ticket:
https://github.com/astronomer/astronomer-cosmos/issues/782

tatiana:
Closing this PR in favour of the ticket: https://github.com/astronomer/astronomer-cosmos/issues/978

PR #755, opened 2023-12-08
Add support for dbt `selector` arg for DAG parsing

### LINKED ISSUE #718: Add support to dbt selector argument, in addition to select and exclude
**Context**

As of Cosmos 1.2.5, users can specify dbt `select` and `exclude` causes via `RenderConfig` (https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html).

That said, dbt supports an additional argument called `selector`, as described in the official docs:
https://docs.getdbt.com/reference/commands/list

As explained in the doc, there is a different semantic to `--select` and `--selector`:
```
[--select SELECTION_ARG [SELECTION_ARG ...]]
[--selector YML_SELECTOR_NAME]
```

This was raised during a discussion with Cindy in the #airflow-dbt slack channel:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1701120337223689?thread_ts=1695419298.963199&cid=C059CC42E9W

**Acceptance criteria**

**Disclaimer:** To implement this feature for `LoadMode.DBT_LS` is straightforward. We probably want to look into supporting this feature for `LoadMode.DBT_MANIFEST` and `LoadMode.CUSTOM`, but we can have these in separate PRs/tickets.

* [ ] Add a `selector` argument in the `RenderConfig` class to support complex dbt selectors via `YML_SELECTOR_NAME`, in this [part of the code](https://github.com/astronomer/astronomer-cosmos/blob/aea4ee7f79a149f9aabeb0f70e3db844985d4c32/cosmos/dbt/graph.py#L190-L191)
* [ ] Add tests covering this use case
* [ ] Update docs [here](https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html) and [here](https://astronomer.github.io/astronomer-cosmos/configuration/selecting-excluding.html)

jbandoro:
## Description

This PR adds support for a new `selector` arg in `RenderConfig` so users can reference[ dbt yaml selectors](https://docs.getdbt.com/reference/node-selection/yaml-selectors) when rendering a project using dbt ls by adding the `--selector` arg to the cli command. This was pretty straightforward for dbt ls, though in order to support custom and manifest filtering we would need to add logic for translating the yaml selector definitions in follow up PR(s).

Added a coverage test and integration test here, and a small update in `tests/dbt/test_graph.py` to use a fixture for a ProfileConfig that is repeated 9 times.

An open question I have is whether we want to log a warning in `RenderConfig.__post_init__` if a user uses both `select`/`exclude` with `selector`.  dbt docs [here](https://docs.getdbt.com/reference/node-selection/syntax#examples) state that when using `--selector` with `--exclude` and/or `--select` the select/exclude flags are ignored (no error or warnings are logged by dbt if both are used). 

## Related Issue(s)

Closes #718  

## Breaking Change?

None

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -195,6 +195,9 @@ class DbtGraph:
         if self.project.dbt_vars:
             ls_command.extend(["--vars", yaml.dump(self.project.dbt_vars)])
 
+        if self.render_config.selector:
+            ls_command.extend(["--selector", self.render_config.selector])
+
         ls_command.extend(self.local_flags)
 
         stdout = run_command(ls_command, tmp_dir, env_vars)
--- a/docs/configuration/selecting-excluding.rst
+++ b/docs/configuration/selecting-excluding.rst
@@ -3,7 +3,11 @@
 Selecting & Excluding
 =======================
 
-Cosmos allows you to filter to a subset of your dbt project in each ``DbtDag`` / ``DbtTaskGroup`` using the ``select`` and ``exclude`` parameters in the ``RenderConfig`` class.
+Cosmos allows you to filter to a subset of your dbt project in each ``DbtDag`` / ``DbtTaskGroup`` using either the ``select`` and ``exclude`` parameters or the ``selector`` parameter in the ``RenderConfig`` class.
+
+
+Using ``select`` and ``exclude``
+--------------------------------
 
 The ``select`` and ``exclude`` parameters are lists, with values like the following:
 
@@ -82,5 +86,28 @@ Examples:
     jaffle_shop = DbtDag(
         render_config=RenderConfig(
             exclude=["node_name+"],  # node_name and its children
+        )
+    )
+
+Using ``selector``
+--------------------------------
+
+The ``selector`` parameter is a string that references a `dbt YAML selector <https://docs.getdbt.com/reference/node-selection/yaml-selectors>`_ already defined in a dbt project.
+
+.. note::
+
+    Only currently supported using the ``dbt_ls`` parsing method where the selector is passed directly to the dbt CLI command. If  ``select`` and/or ``exclude`` \
+    are used with ``selector``, dbt will ignore the ``select`` and ``exclude`` parameters.
+
+Examples:
+
+.. code-block:: python
+
+    from cosmos import DbtDag, RenderConfig, LoadMode
+
+    jaffle_shop = DbtDag(
+        render_config=RenderConfig(
+            selector="my_selector",  # this selector must be defined in your dbt project
+            load_method=LoadMode.DBT_LS,
         )
     )
--- a/tests/dbt/test_graph.py
+++ b/tests/dbt/test_graph.py
@@ -1,7 +1,7 @@
 import shutil
 import tempfile
 from pathlib import Path
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
 import yaml
 
 import pytest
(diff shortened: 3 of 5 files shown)


docs/configuration/selecting-excluding.rst
```diff
@@ -3,7 +3,11 @@
 Selecting & Excluding
 =======================
 
-Cosmos allows you to filter to a subset of your dbt project in each ``DbtDag`` / ``DbtTaskGroup`` using the ``select`` and ``exclude`` parameters in the ``RenderConfig`` class.
+Cosmos allows you to filter to a subset of your dbt project in each ``DbtDag`` / ``DbtTaskGroup`` using either the ``select`` and ``exclude`` parameters or the ``selector`` parameter in the ``RenderConfig`` class.
```
tatiana, line 6:
``` suggestion
Cosmos allows you to filter to a subset of your dbt project in each ``DbtDag`` / ``DbtTaskGroup`` using the ``select `` and ``exclude`` parameters in the ``RenderConfig`` class.

 Since Cosmos 1.3, the ``selector`` parameter is also available in ``RenderConfig`` when using the ``LoadMode.DBT_LS`` to parse the dbt project into Airflow.
```

docs/configuration/selecting-excluding.rst
```diff
@@ -84,3 +88,26 @@ Examples:
...
         )
     )
+
+Using ``selector``
+--------------------------------
+
+The ``selector`` parameter is a string that references a `dbt YAML selector <https://docs.getdbt.com/reference/node-selection/yaml-selectors>`_ already defined in a dbt project.
+
+.. note::
+
```
tatiana, line 98: I think it is worth moving the `:note` to the top, just before the `Using selector` heading - and to also state that this is available from 1.3 onwards.

tests/dbt/test_graph.py
```diff
@@ -44,6 +44,18 @@ def tmp_dbt_project_dir():
     shutil.rmtree(tmp_dir, ignore_errors=True)  # delete directory
 
 
+@pytest.fixture
```
tatiana, line 47: Great improvement, @jbandoro , reducing the redundancy!

cosmos/dbt/graph.py
```diff
@@ -195,6 +195,9 @@ def run_dbt_ls(
         if self.project.dbt_vars:
             ls_command.extend(["--vars", yaml.dump(self.project.dbt_vars)])
 
+        if self.render_config.selector:
```
tatiana, line 198: We should raise an exception - somewhere - if a user tries to use this argument with a load method other than `dbt ls`.

tatiana (commented):
Hi @jbandoro, as usual, great work!

> An open question is whether we want to log a warning in RenderConfig.__post_init__ if a user uses both select/exclude with selector. dbt docs [here](https://docs.getdbt.com/reference/node-selection/syntax#examples) state that when using --selector with --exclude and --select, the select/exclude flags are ignored (no error or warnings are logged by dbt if both are used).

I like the way you developed the feature. While checking for this in Cosmos would make users aware of the dbt limitation, if dbt decides to support both types of selectors simultaneously - we would have to change Cosmos to keep this.

The one check I think we need to add is to explicitly raise an exception if users try to use `selector` with a load mode that does not support it.

Do you think we should log a follow-up ticket to support this in the other load modes?

jbandoro:
> The one check I think we need to add is to explicitly raise an exception if users try to use selector with a load mode that does not support it.

Good point, I think that should be done in `DbtGraph` incase `RenderConfig.load_method = LoadMode.AUTOMATIC`. Will update.

> Do you think we should log a follow-up ticket to support this in the other load modes?

Yes definitely, I can follow-up and create one after #718 is closed.

docs/configuration/selecting-excluding.rst
jbandoro: I moved the note block up so it now looks like this: 
image

cosmos/dbt/graph.py
jbandoro: Added this to `DbtGraph` for manifest and custom load methods along with a test to check that an exception is raised.

tatiana (approved):
Looks great, thanks, @jbandoro for adding this feature. Once the checks pass, we can merge this PR.

jbandoro:
@tatiana added a follow-up issue for support manifest and custom load modes with the selector arg: https://github.com/astronomer/astronomer-cosmos/issues/767

Issue #520, opened 2023-09-07
Evaluate alternatives to rendering dbt projects other than during DAG parsing time


tatiana:
At the moment, Cosmos does several things during DAG parsing time:
- It may convert [connection -> profile](https://astronomer.github.io/astronomer-cosmos/profiles/index.html)
- It [parses the dbt project](https://astronomer.github.io/astronomer-cosmos/configuration/parsing-methods.html), to convert it into an Airflow DAG (dynamically, during parse-time)
- It may apply filters based on user-defined selectors if using `LoadMode.DBT_MANIFEST` or `LoadMode.CUSTOM`

We have yet to do a performance analysis (#519). However, this could bring issues to the scheduler in larger dbt projects.

Ideally, we should avoid this logic during scheduling / DAG parsing. Some alternatives:
1. Allow users to pre-compute an expanded Airflow DAG
2. Use [Airflow dynamic task mapping](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/dynamic-task-mapping.html)
3. Before running Airflow, extend manifest definitions to include filtered nodes (materialising into a JSON) before DAG-parsing
4. See if we could leverage [DAG setup and teardown](https://airflow.apache.org/docs/apache-airflow/stable/howto/setup-and-teardown.html) in any way
5. Others..?

Thoughts and ideas are very welcome!

DanMawdsleyBA:
I would welcome this! One thing that might be useful is if I wanted to run the airflow dag with a different select statement. For example I only want to rerun part of the dag not the entire dag. If my process loads from 10 sources in parallel but only 1 source file has come through it can just run the downstream of that one file. Could cutdown airflow dag times and unnecessary queries on the database.

tatiana:
#airflow-dbt users reported having to increase `AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT` to be able to parse DAGs with Cosmos:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1696435273519979

benjamin-awd:
FWIW switching from the default parsing mode to `LoadMode.DBT_MANIFEST` seems to have had a dramatic impact on DAG file processing durations -- I think this may be tangentially related to the DagBag import timeout issue.
![Screenshot 2023-10-06 at 11 33 13 AM](https://github.com/astronomer/astronomer-cosmos/assets/62495124/4f434c09-fdc6-4590-b439-acd5756daebb)

[...]

PR #757, opened 2023-12-09
Additions to docs regarding the `DBT_MANIFEST` load mode

dwreeves:
## Description

I thought that there were a few aspects regarding execution modes that could require more clarification in the docs.

- The parsing methods docs mentions that only `LOCAL` execution mode is supported for `DBT_LS`, but the reverse was not true (i.e. execution mode docs made no mention of parsing methods), so I added notes about that.
- GCC docs suggest using `VIRTUALENV` execution mode, but makes no mention of the fact that the `DBT_LS` parsing method is not supported in this execution mode. Naturally, in this case, users should be utilizing the `DBT_MANIFEST` load mode, but that means that the docs are incomplete since they don't include a `manifest_path=?` in the `ProjectConfig`.
    - Note that there are also discussions in the Airflow Slack regarding issues users have had parsing the `DbtDag` in GCC that are fixable via using a pre-compiled `manifest,json`, e.g. https://apache-airflow.slack.com/archives/C059CC42E9W/p1696435273519979
    - Also see #520 for more discussion.
- Generally speaking when doing the `DBT_MANIFEST` load method, the pattern is that you run `dbt deps && dbt compile` as part of your deployment, and upload your full dbt project including these artifacts. This deployment approach may be obvious to veteran users of Airflow and/or dbt, but it may not be obvious to everyone, so I think adding a couple sentences in `parsing-methods.rst` is beneficial.

## Related Issue(s)

Not explicitly related, but #520 discusses some issues encountered using the default parsing method. (Specifically, running `dbt deps` from a blank slate tends to slow everything down a lot.)

Part of my motivation for adding to the docs is to better advertise + better document this alternate method of parsing the dbt DAG.

## Breaking Change?

n/a

## Checklist

n/a


docs/getting_started/execution-modes.rst
```diff
@@ -72,7 +72,10 @@ In this case, users are responsible for declaring which version of ``dbt`` they
 
 Similar to the ``local`` execution mode, Cosmos converts Airflow Connections into a way ``dbt`` understands them by creating a ``dbt`` profile file (``profiles.yml``).
 
-A drawback with this approach is that it is slower than ``local`` because it creates a new Python virtual environment for each Cosmos dbt task run.
+Some drawbacks of this approach:
+
+- It is slower than ``local`` because it creates a new Python virtual environment for each Cosmos dbt task run.
+- The ``dbt_ls`` load method is not supported with this execution mode. See `Parsing Methods <parsing-methods.html>`_ for more information.
```
tatiana, line 78: The ``dbt_ls`` load method is indifferent to the execution mode - its only expectation is to have a dbt binary accessible from the instance running the Airflow scheduler. Therefore, I don't think it's accurate.

```suggestion
- If dbt is unavailable in the Airflow scheduler, the default ``LoadMode.DBT_LS`` will not work. In this scenario, users must use a `parsing method <parsing-methods.html>`_  that does not rely on dbt, such as ``LoadMode.MANIFEST``.
```

docs/getting_started/execution-modes.rst
```diff
@@ -72,7 +72,10 @@ In this case, users are responsible for declaring which version of ``dbt`` they
 
 Similar to the ``local`` execution mode, Cosmos converts Airflow Connections into a way ``dbt`` understands them by creating a ``dbt`` profile file (``profiles.yml``).
 
-A drawback with this approach is that it is slower than ``local`` because it creates a new Python virtual environment for each Cosmos dbt task run.
+Some drawbacks of this approach:
+
+- It is slower than ``local`` because it creates a new Python virtual environment for each Cosmos dbt task run.
```
tatiana, line 77: This will change once we merge #611 🙌

docs/getting_started/execution-modes.rst
```diff
@@ -91,6 +94,7 @@ The user has better environment isolation than when using ``local`` or ``virtual
 The other challenge with the ``docker`` approach is if the Airflow worker is already running in Docker, which sometimes can lead to challenges running `Docker in Docker <https://devops.stackexchange.com/questions/676/why-is-docker-in-docker-considered-bad>`__.
 
 This approach can be significantly slower than ``virtualenv`` since it may have to build the ``Docker`` container, which is slower than creating a Virtualenv with ``dbt-core``.
+Additionally, this execution mode does not support the ``dbt_ls`` load method (see `Parsing Methods <parsing-methods.html>`_ for more information).
```
tatiana, line 97:
```suggestion
If dbt is unavailable in the Airflow scheduler, the default ``LoadMode.DBT_LS`` will not work. In this scenario, users must use a `parsing method <parsing-methods.html>`_  that does not rely on dbt, such as ``LoadMode.MANIFEST``.
```

tatiana (commented):
Thanks for improving the docs, @dwreeves - some minor comments inline

docs/getting_started/execution-modes.rst
dwreeves: It depends on whether the venv is saved somewhere though, right? So this language change will still apply in some cases, i.e. if the flag is not set.

(Perhaps a separate discussion, but at the moment, my current setup is that I already run dbt out of a separate virtualenv, I just point the binary to `dbt_venv/bin/dbt`. I'm curious what the pros/cons are to saving the venv compared to that approach.)

dwreeves:
@tatiana Some context: I was basing this and all other documentation changes I've proposed on two things:

(1) the code that currently exists in `cosmos/dbt/graph.py` [here](https://github.com/astronomer/astronomer-cosmos/blob/43699870961b1c8d8b830fc8503c7404d077d793/cosmos/dbt/graph.py#L147) is that it will not automatically execute `load_via_dbt_ls()` if the execution mode is `LOCAL`.

image

(2) The docs already mention my wording change [somewhere else](https://astronomer.github.io/astronomer-cosmos/configuration/parsing-methods.html#dbt-ls):

image

I think I see what your changes are attempting to do. But before I continue with that train of thought, I am wondering: Is it possible that `execution_mode == ExecutionMode.LOCAL` should not be part of the predicate that checks whether `DBT_LS` mode is run? I think our divergence of views here makes more sense in a world where that `execution_mode == ExecutionMode.LOCAL` check is not part of `cosmos.dbt.graph.DbtGraph.load`.

dwreeves:
To add to that, if that does check out, then I think my confusion would be that you are viewing the AUTOMATIC and DBT_LS as separate modes, whereas I'm treating AUTOMATIC as a way to access DBT_LS. In that case where AUTOMATIC, I still think:

- the docs should make some note that AUTOMATIC will not use `dbt ls` to parse if the execution mode is local. (Although, you are right that "it is not supported" is too strong a statement; the more accurate statement would be to describe this in terms of how the AUTOMATIC load mode works.)
- Without a `manifest.json` specified, the GCC setup will default to the `LoadMode.CUSTOM` logic. Perhaps that is fine and my suggestion to specify a manifest is off-base, but I also believe that's an accurate description of the current behavior of the existing GCC suggested setup.

docs/getting_started/execution-modes.rst
tatiana: > It depends on whether the venv is saved somewhere, though, right? So, this language change will still apply in some cases, i.e. if the flag is not set.

Yes, you're right! It will also depend on how Airflow was configured. If the Airflow worker nodes run using `KubernetesExecutor`, the optimisation would only work if the dbt binary was in a shared volume - which could lead to concurrency issues.

> (Perhaps a separate discussion, but at the moment, my current setup is that I already run dbt out of a separate virtualenv, I point the binary to dbt_venv/bin/dbt. I'm curious what the pros/cons are to saving the venv and using the virtualenvoperator, compared to that approach of using local but pointing to a separate venv's bin/dbt.)

I'd encourage users to use `LoadMode.LOCAL`. It's the most tested approach, and - since the latest dbt and Airflow dependencies don't conflict (https://astronomer.github.io/astronomer-cosmos/getting_started/execution-modes-local-conflicts.html#execution-modes-local-conflicts) - this is the most straightforward setup.

If there are dependency conflicts, the second recommendation would be to use `ExecutionMode.LOCAL` customising the dbt binary path, as you mentioned.

I'm curious about your use case. Are you using Google Cloud Composer? How are you setting up the dbt binary? Are you adding to the shared volume/GCS bucket?

The `ExecutionMode.VIRTUALENV` is an alternative for users who may not be able to create a separate venv for dbt in the Airflow worker nodes and who are facing dependency conflicts but do not want the overhead of using `ExecutionMode.KUBERNETES`.

Overall, each ExecutionMode has its trade-offs.

tatiana:
Hey @dwreeves, first of all, thank you very much for working on this. We've been struggling to keep docs up-to-date, and they are as important as the code - directly affecting the adoption of Cosmos.

This particular discussion is also relevant for us to review and potentially improve Cosmos' behaviour.

> (1) the code that currently exists in cosmos/dbt/graph.py [here](https://github.com/astronomer/astronomer-cosmos/blob/43699870961b1c8d8b830fc8503c7404d077d793/cosmos/dbt/graph.py#L147) is that it will not automatically execute load_via_dbt_ls() if the execution mode is LOCAL.

If the manifest file is not set and the user-defined a profile configuration, yes, `LoadMode.AUTOMATIC` will call `load_via_dbt_ls()` by default when users use `ExecutionMode.LOCAL`. From the users I've supported so far, many end up using this.

> (2) The docs already mention my wording change [somewhere else](https://astronomer.github.io/astronomer-cosmos/configuration/parsing-methods.html#dbt-ls)

This is incorrect; users can use `LoadMode.DBT_LS` regardless of the `ExecutionMode` choice.

Could you remove this note as part of this PR?

During the last few weeks, we have been detaching the configuration of rendering the DAG from how Cosmos tasks execute dbt, where possible (an example issue: #568). This is particularly relevant when users use `ExecutionMode.Kubernetes` since the "environment" where the dbt commands run diverges significantly from where the DAG is parsed.

> I think I see what your changes are attempting to do. But before I continue with that train of thought, I am wondering: Is it possible that execution_mode == ExecutionMode.LOCAL should not be part of the predicate that checks whether DBT_LS mode is run?

I don't think so.

`ExecutionMode.LOCAL` can be used with any of the "four" DAG parsing modes (
`LoadMode.AUTOMATIC` , `LoadMode.MANIFEST` , `LoadMode.DBT_LS` and `LoadMode.CUSTOM`). We're expecting more parsing methods to be introduced, such as #733, which are also independent of how the DAG is rendered.

Additionally, `LoadMode.DBT_LS` can also be used with any of the supported execution modes (`ExecutionMode.LOCAL`, `ExecutionMode.VIRTUALENV`, `ExecutionMode.DOCKER` and `ExecutionMode.KUBERNETES`). The choice of how to render the DAG is  independent of how the tasks execute the dbt command. Again, there are a few ways we could extend the current execution modes - and they should still work with  `LoadMode.DBT_LS`. One approach, for instance, could be for us to introduce native Airflow execution modes that, for example, rely on dbt to compile the dbt project SQL but run the actual transformations using Airflow async operators.

I agree that usually users who opt to run the dbt commands using `ExecutionMode.KUBERNETES` will have a more complex CI/CD process and will likely want to "isolate" the dbt command from being run in their Airflow environment - very probably relying on `LoadMode.MANIFEST` (and soon `LoadMode.DBT_LS_FILE`) to parse their DAG - avoiding `LoadMode.DBT_LS`. But this is a choice, not a restriction imposed by Cosmos.

> I think our divergence of views here makes more sense in a world where execution_mode == ExecutionMode.LOCAL check is not part of cosmos.dbt.graph.DbtGraph.load. If that checks out, then I suppose I just may be confused about something.

I addressed this before; please let me know if not.

> To add to that, if that does check out, I think my confusion would be that you view the AUTOMATIC and DBT_LS as separate modes, whereas I'm treating AUTOMATIC as a way to access DBT_LS. In that case where AUTOMATIC is regarded as its own thing, I still think:

Yes, `LoadMode.AUTOMATIC` and `LoadMode.DBT_LS` are different. `LoadMode.AUTOMATIC` is the best effort to figure out how to parse the dbt project based on the user configuration; it can lead to any of the other three current load methods to be used (`LoadMode.MANIFEST`, `LoadMode.DBT_LS` and `LoadMode.CUSTOM`). If the user gives a manifest, that's what is used. If the user does not, and Cosmos has reasons to believe the `dbt` command is available locally, it will attempt to use dbt ls. If that doesn't work, it then uses the custom parsing method.

> the docs should note that AUTOMATIC will not use dbt ls to parse if the execution mode is local. (Although you are right that "it is not supported" is too strong a statement, the more accurate statement would be to describe how the AUTOMATIC load mode works.)

I've thought about this, and perhaps to simplify; we should change the following:
https://github.com/astronomer/astronomer-cosmos/blob/43699870961b1c8d8b830fc8503c7404d077d793/cosmos/dbt/graph.py#L173

To be:
```
                If self.profile_config:
```

This would simplify the docs and the explanation of `LoadMode.AUTOMATIC`. Since we're catching the exception, it should also be backwards compatible. WDYT? I can log a follow-up ticket for this - I don't think we should change the code as part of this docs update.

> Without a manifest.json specified, the GCC setup will default to the LoadMode.CUSTOM logic. That may be fine, and my suggestion to establish a manifest is off-base. Still, I also believe that's an accurate description of the current behaviour of the existing GCC suggested setup. So I would then ask: should LoadMode.DBT_MANIFEST be preferred over LoadMode.CUSTOM, and should the GCC docs mention anything about the default use of LoadMode?CUSTOM?

I agree with your proposal. For GCC, as of Cosmos 1.2.x, I believe we should encourage users to set the manifest using `ProjectConfig(manifest_path=...)`. It is a more accurate way of parsing the dbt project than fall backing to `LoadMode.CUSTOM`. From Cosmos 1.3.x, I'd recommend using the `LoadMode.DBT_LS_FILE`: #733, but since 1.3 is not out, we can also do this change as part of the 1.3 release, once this PR is merged.

Yes, I also agree we should highlight to GCC users that if they don't give a manifest, `LoadMode.AUTOMATIC` will fall back to using Cosmos' `LoadMode.CUSTOM` logic.

> The statement inside parsing-methods. rst should be relaxed from saying: "This only works for the local execution mode", and should instead say: "This only works if a dbt command is available to the scheduler."

Yes, I like your suggestion! Better than just removing, as a previously mentioned.

docs/getting_started/execution-modes.rst
dwreeves: >I'm curious about your use case. Are you using Google Cloud Composer? How are you setting up the dbt binary? Are you adding to the shared volume/GCS bucket?

I am not using GCC at the moment. I am using Astronomer and I pip install into a new venv as part of the Docker build step. Locally I have a command that aliases `dbt` to `rich-click venv_dbt/bin/dbt` and passes in a default project dir / profile dir and default port for docs.

dwreeves:
I've made changes to the PR as per the discussion.

docs/configuration/parsing-methods.rst
```diff
@@ -16,7 +16,7 @@ There are benefits and drawbacks to each method:
...
 - ``dbt_ls``: Cosmos will generate the manifest file for you. This method uses dbt's metadata AND dbt's selecting/excluding logic. This is the most robust method. However, this requires the dbt executable to be installed on your machine (either on the host directly or in a virtual environment).
 - ``dbt_ls_file`` (new in 1.3): Path to a file containing the ``dbt ls`` output. To use this method, run ``dbt ls`` using ``--output json`` and store the output in a file. ``RenderConfig.select`` and ``RenderConfig.exclude`` will not work using this method.
-- ``custom``: Cosmos will parse your project and model files for you. This means that Cosmos will not have access to dbt's metadata. However, this method does not require the dbt executable to be installed on your machine.
+- ``custom``: Cosmos will parse your project and model files for you. This means that Cosmos will not have access to dbt's metadata. However, this method does not require the dbt executable to be installed on your machine, and does not require the user to provide any dbt artifacts.
```
tatiana, line 19:
```suggestion
- ``custom``: Cosmos will parse your project and model files. This means that Cosmos will not have access to dbt's metadata. However, this method does not require the dbt executable to be installed on your machine, and does not require the user to provide any dbt artifacts.
```

tatiana (approved):
Thanks for addressing all the feedback, @dwreeves , we'll release this as part of 1.3!

PR #758, opened 2023-12-11
Athena - Get temporary credentials from the conn_id

octiva:
## Description

Passes the `conn_id` to the `AwsGenericHook` and uses `get_credentials()`, which handles the creation of a session, credentials, freezing of credentials & also masking. 

[See get_credentials() docs here](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/base_aws/index.html#airflow.providers.amazon.aws.hooks.base_aws.AwsGenericHook.get_credentials)

## Related Issue(s)

#691 

## Checklist

### DIFF
--- a/cosmos/profiles/athena/access_key.py
+++ b/cosmos/profiles/athena/access_key.py
@@ -3,20 +3,33 @@ from __future__ import annotations
 
 from typing import Any
 
+from cosmos.exceptions import CosmosValueError
+
 from ..base import BaseProfileMapping
 
 
 class AthenaAccessKeyProfileMapping(BaseProfileMapping):
     """
-    Maps Airflow AWS connections to a dbt Athena profile using an access key id and secret access key.
+    Uses the Airflow AWS Connection provided to get_credentials() to generate the profile for dbt.
 
-    https://docs.getdbt.com/docs/core/connect-data-platform/athena-setup
     https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/connections/aws.html
+
+
+    This behaves similarly to other provider operators such as the AWS Athena Operator.
+    Where you pass the aws_conn_id and the operator will generate the credentials for you.
+
+    https://registry.astronomer.io/providers/amazon/versions/latest/modules/athenaoperator
+
+    Information about the dbt Athena profile that is generated can be found here:
+
+    https://github.com/dbt-athena/dbt-athena?tab=readme-ov-file#configuring-your-profile
+    https://docs.getdbt.com/docs/core/connect-data-platform/athena-setup
     """
 
     airflow_connection_type: str = "aws"
     dbt_profile_type: str = "athena"
     is_community: bool = True
+    temporary_credentials = None
 
     required_fields = [
         "aws_access_key_id",
@@ -26,11 +39,7 @@ class AthenaAccessKeyProfileMapping(BaseProfileMapping):
         "s3_staging_dir",
         "schema",
     ]
-    secret_fields = ["aws_secret_access_key", "aws_session_token"]
     airflow_param_mapping = {
-        "aws_access_key_id": "login",
-        "aws_secret_access_key": "password",
-        "aws_session_token": "extra.aws_session_token",
         "aws_profile_name": "extra.aws_profile_name",
         "database": "extra.database",
         "debug_query_state": "extra.debug_query_state",
@@ -49,11 +58,43 @@ class AthenaAccessKeyProfileMapping(BaseProfileMapping):
     @property
     def profile(self) -> dict[str, Any | None]:
         "Gets profile. The password is stored in an environment variable."
+
+        self.temporary_credentials = self._get_temporary_credentials()  # type: ignore
+
         profile = {
             **self.mapped_params,
             **self.profile_args,
-            # aws_secret_access_key and aws_session_token should always get set as env var
+            "aws_access_key_id": self.temporary_credentials.access_key,
             "aws_secret_access_key": self.get_env_var_format("aws_secret_access_key"),
             "aws_session_token": self.get_env_var_format("aws_session_token"),
         }
+
         return self.filter_null(profile)
+
+    @property
+    def env_vars(self) -> dict[str, str]:
+        "Overwrites the env_vars for athena, Returns a dictionary of environment variables that should be set based on the self.temporary_credentials."
+
+        if self.temporary_credentials is None:
+            raise CosmosValueError(f"Could not find the athena credentials.")
+
+        env_vars = {}
+
+        env_secret_key_name = self.get_env_var_name("aws_secret_access_key")
+        env_session_token_name = self.get_env_var_name("aws_session_token")
+
+        env_vars[env_secret_key_name] = str(self.temporary_credentials.secret_key)
+        env_vars[env_session_token_name] = str(self.temporary_credentials.token)
+
+        return env_vars
+
+    def _get_temporary_credentials(self):  # type: ignore
+        """
+        Helper function to retrieve temporary short lived credentials
+        Returns an object including access_key, secret_key and token
+        """
+        from airflow.providers.amazon.aws.hooks.base_aws import AwsGenericHook
+
+        hook = AwsGenericHook(self.conn_id)  # type: ignore
+        credentials = hook.get_credentials()
+        return credentials
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ dbt-all = [
 ]
 dbt-athena = [
     "dbt-athena-community",
+    "apache-airflow-providers-amazon>=8.0.0",
 ]
 dbt-bigquery = [
     "dbt-bigquery",
@@ -110,7 +111,6 @@ tests = [
     "mypy",
     "sqlalchemy-stubs", # Change when sqlalchemy is upgraded https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html
 ]
-
 docker = [
     "apache-airflow-providers-docker>=3.5.0",
 ]
@@ -120,7 +120,6 @@ kubernetes = [
 pydantic = [
         "pydantic>=1.10.0,<2.0.0",
 ]
-
 
 [project.entry-points.cosmos]
 provider_info = "cosmos:get_provider_info"
--- a/tests/profiles/athena/test_athena_access_key.py
+++ b/tests/profiles/athena/test_athena_access_key.py
@@ -1,13 +1,41 @@
 "Tests for the Athena profile."
 
 import json
-from unittest.mock import patch
-
+from collections import namedtuple
+import sys
+from unittest.mock import MagicMock, patch
 import pytest
 from airflow.models.connection import Connection
 
 from cosmos.profiles import get_automatic_profile_mapping
 from cosmos.profiles.athena.access_key import AthenaAccessKeyProfileMapping
+
+Credentials = namedtuple("Credentials", ["access_key", "secret_key", "token"])
+
+mock_assumed_credentials = Credentials(
+    secret_key="my_aws_assumed_secret_key",
+    access_key="my_aws_assumed_access_key",
+    token="my_aws_assumed_token",
+)
+
+mock_missing_credentials = Credentials(access_key=None, secret_key=None, token=None)
+
+
+@pytest.fixture(autouse=True)
+def mock_aws_module():
+    mock_aws_hook = MagicMock()
+
+    class MockAwsGenericHook:
+        def __init__(self, conn_id: str) -> None:
+            pass
+
+        def get_credentials(self) -> Credentials:
+            return mock_assumed_credentials
+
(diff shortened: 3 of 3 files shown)


pixie79:
Do you mean dbt-Athena or dbt-athena-community. Dbt-Athena is the original module but the community version is much improved and optimised. https://github.com/dbt-athena/dbt-athena

dev/dags/example_athena_profile.py
jbandoro, line 1: Cosmos doesn't have example DAGs for all of the profile mappings it supports, so I would suggest removing this file. The DAGs in `dev/dags` are run for integration testing and this one I think would fail.

Docs for all of the profile mappings are [here](https://astronomer.github.io/astronomer-cosmos/profiles/index.html) which are sourced from the profile mapping classes. I'd suggest updating the sphinx docstring of the the `AthenaAccessKeyProfileMapping` class with any useful details that would be helpful for end users in using the profile.

jbandoro (commented):
Thanks @octiva for the contribution!

I'm not familiar with Athena access keys, so if anyone from #691 wants to chime in here it would be helpful. My main comment below is if we should keep the current profile mapping of the Airflow connection, are there cases where users would still want to use the Airflow connection params, or is it broken and this PR fixes it?

dev/dags/example_athena_profile.py
octiva: Thanks that is a great point. I will remove, and also update the docs.

octiva:
@jbandoro Thanks for the feedback. If the user has an Airflow connection with no `extra.role_arn` provided, their credentials  (secret + key id) used in the DBT profile will be unchanged, except there will now always be a `aws_session_token`.

cosmos/profiles/athena/access_key.py
```diff
@@ -49,11 +49,27 @@ class AthenaAccessKeyProfileMapping(BaseProfileMapping):
...
+
         profile = {
             **self.mapped_params,
             **self.profile_args,
-            # aws_secret_access_key and aws_session_token should always get set as env var
-            "aws_secret_access_key": self.get_env_var_format("aws_secret_access_key"),
-            "aws_session_token": self.get_env_var_format("aws_session_token"),
+            "aws_secret_access_key": credentials.secret_key,
+            "aws_session_token": credentials.token,
+            "aws_access_key_id": credentials.access_key,
```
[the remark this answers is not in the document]
octiva: Good call out, we can safely remove those. 

The AwsGenericHook handles the secret masking, so there is no need to store them in the envs, we can also remove that logic.  [AWS docs](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/base_aws/index.html#airflow.providers.amazon.aws.hooks.base_aws.AwsGenericHook.get_credentials)

I also want to wait for @benjamin-awd's feedback before remove too much of the existing logic.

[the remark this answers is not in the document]
octiva: @jbandoro removed `aws_secret_access_key`, `aws_session_token` & `aws_access_key_id` from `airflow_param_mapping` as these always come from the hook. 

I dont think we can remove them from the `required_field`, as they are a check on the connection that we will be able to create a profile, mainly when creating via the automatic_mapping function. We could use the `test_connection()` to ensure the credentials are usable, however that doesnt really fit the current pattern for profile from Airflow connections.

dev/dags/example_athena_profile.py
octiva: @jbandoro Removed all updates to `/dev`. Also updated the docs in the sphinx docstring to update what is happening and where we can find information.

cosmos/profiles/athena/access_key.py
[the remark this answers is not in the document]
jbandoro: Thanks!

You're right about the `required_fields` that they should be kept. Though they might need to be removed from the `secret_fields` or else trying to claim the connection will fail:

https://github.com/astronomer/astronomer-cosmos/blob/7b3e8ab93eda1dfd233ad201a231fb5033615bd7/cosmos/profiles/base.py#L90-L98

The unit tests should confirm this though once the GH actions are fixed.

cosmos/profiles/athena/access_key.py
```diff
@@ -49,11 +55,27 @@ class AthenaAccessKeyProfileMapping(BaseProfileMapping):
...
         }
+
         return self.filter_null(profile)
+
+    def get_temporary_credentials(self):  # type: ignore
+        """
+        Helper function to retrieve temporary short lived credentials
+        Returns an object including access_key, secret_key and token
+        """
+        from airflow.providers.amazon.aws.hooks.base_aws import AwsGenericHook
```
benjamin-awd, line 76: Should the AWS Airflow provider (`apache-airflow-providers-amazon`) be listed somewhere as an optional dependency in the pyproject.toml?

benjamin-awd:
LGTM, I tried this out today and it seems to work for my use-case -- unfortunately I don't have time to look into the memory issue and verify that this works at scale, but that's hopefully going to be resolved with the use of the AWS hook. Ran a couple of models concurrently and didn't run into any issues.

cosmos/profiles/athena/access_key.py
octiva: Thanks @benjamin-awd, ive added aws as a optional-dependency in the pyproject.toml

pyproject.toml
```diff
@@ -120,6 +119,9 @@ kubernetes = [
 pydantic = [
         "pydantic>=1.10.0,<2.0.0",
 ]
+aws = [
+    "apache-airflow-providers-amazon>=8.0.0",
+]
```
tatiana, line 124: Since, ATM, this dependency is specific to Athena, I suggest we add it to the previously existing:
```
dbt-athena = [
```

tests/profiles/athena/test_athena_access_key.py
```diff
@@ -136,8 +197,8 @@ def test_athena_profile_args_overrides(
     assert profile_mapping.profile == {
         "type": "athena",
         "aws_access_key_id": mock_athena_conn.login,
-        "aws_secret_access_key": "{{ env_var('COSMOS_CONN_AWS_AWS_SECRET_ACCESS_KEY') }}",
-        "aws_session_token": "{{ env_var('COSMOS_CONN_AWS_AWS_SESSION_TOKEN') }}",
+        "aws_secret_access_key": mock_assumed_credentials.secret_key,
+        "aws_session_token": mock_assumed_credentials.token,
```
tatiana, line 201: Before, we were obfuscating the secrets and setting them in environment variables. This behaviour should continue for temporary credentials as well.

tests/profiles/athena/test_athena_access_key.py
```diff
@@ -122,9 +173,19 @@ def test_athena_profile_args_overrides(
...
+        yield
+
     profile_mapping = get_automatic_profile_mapping(
         mock_athena_conn.conn_id,
-        profile_args={"schema": "my_custom_schema", "database": "my_custom_db", "aws_session_token": "override_token"},
+        profile_args={
+            "schema": "my_custom_schema",
+            "database": "my_custom_db",
+            "aws_session_token": "override_token",
+        },
```
tatiana, line 188: Before, we were obfuscating the secrets and setting them in environment variables. This behaviour should continue for temporary credentials as well.

tests/profiles/athena/test_athena_access_key.py
```diff
@@ -100,15 +146,20 @@ def test_athena_profile_args(
...
     )
 
     assert profile_mapping.profile == {
         "type": "athena",
-        "aws_access_key_id": mock_athena_conn.login,
-        "aws_secret_access_key": "{{ env_var('COSMOS_CONN_AWS_AWS_SECRET_ACCESS_KEY') }}",
-        "aws_session_token": "{{ env_var('COSMOS_CONN_AWS_AWS_SESSION_TOKEN') }}",
+        "aws_access_key_id": mock_assumed_credentials.access_key,
+        "aws_secret_access_key": mock_assumed_credentials.secret_key,
+        "aws_session_token": mock_assumed_credentials.token,
```
tatiana, line 162: Before, we were obfuscating the secrets and setting them in environment variables. This behaviour should continue for temporary credentials as well.

tatiana (commented):
Thanks for your contribution, @octiva! It is an excellent improvement to leverage the hook method instead of directly retrieving the values from the environment variable.

Before, we were obfuscating the secrets and setting them as environment variables. I suggest we keep this behaviour, even if the credentials are temporary. The description of your PR states that the Airflow `get_credentials` method masks those secrets. From what I understood from the docs:
>  By use this method also secret_key and token will mask in tasks logs.
https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/base_aws/index.html#airflow.providers.amazon.aws.hooks.base_aws.AwsGenericHook.get_credentials

In other words, it seems they are masked only in the logs - not in the `profiles.yml` file that Cosmos creates.

Assuming the value we wrote in the `profiles.yml` file was masked, the dbt commands using them would fail to run - since they would try to use invalid credentials.

tests/profiles/athena/test_athena_access_key.py
octiva: Great call out, sorry I missed it. Ive fixed it up now. Let me know what you think of the solution.

octiva:
@tatiana I am unable to write a test for the `_get_temporary_credentials`method, as I am unable to import `airflow.providers.amazon` via the test dependencies as we get some dependency conflicts when using Python 3.8/3.9 + Airflow 2.3. 

The amazon provider at 8.0.0 is compatible with 2.3 and python down to 3.7.

https://pypi.org/project/apache-airflow-providers-amazon/8.0.0/

I tried many ways to fix this on the weekend, but could not. I ended up resolving this by removing the dependency and mocking the class function. 

This has the unfortunately side-affect of reducing our test coverage

jbandoro:
Thanks @octiva addressing all the feedback! For your question below on test coverage:

> I tried many ways to fix this on the weekend, but could not. I ended up resolving this by removing the dependency and mocking the class function.
> 
> This has the unfortunately side-affect of reducing our test coverage

You can get test coverage for the `_get_temporary_credential` method by instead patching the provider module that can't be imported, like in the examples [here](https://github.com/astronomer/astronomer-cosmos/blob/9fcee8e785ea3bcb0d19e8d36f3f5b94bedafc98/tests/operators/test_local.py#L439) and [here](https://github.com/astronomer/astronomer-cosmos/blob/9fcee8e785ea3bcb0d19e8d36f3f5b94bedafc98/tests/test_export.py#L76-L83).

It might be easier to do this patching in a fixture and reuse it in all of tests that require it.

spencer-horton-wcq:
@jbandoro Thanks for the hint. I had attempted this, but was doing it at the wrong level, and now ive got it working quite nicely. Let me know what you think

octiva:
reposting from correct account @jbandoro Thanks for the hint. I had attempted this, but was doing it at the wrong level, and now ive got it working quite nicely. Let me know what you think 🚀

tests/profiles/athena/test_athena_access_key.py
```diff
@@ -1,20 +1,75 @@
...
+
+    with patch("airflow.hooks.base.BaseHook.get_connection", return_value=conn):
+        yield conn
+
 
 @pytest.fixture()
 def mock_athena_conn():  # type: ignore
     """
     Sets the connection as an environment variable.
     """
```
jbandoro, line 71: Do you think we need both the `mock_athena_assume_role_conn` and the `mock_athena_conn` fixture herd, could we keep only 1. 

It looks like the only difference is that one has `"role_arn"` field in its extra which isn't tested or used in the tests.

jbandoro (approved):
Thanks @octiva! Mocking the hook looks much better. Approving with some minor comments below to address before merging.

tests/profiles/athena/test_athena_access_key.py
spencer-horton-wcq: Agreed, removed

cosmos/profiles/athena/access_key.py
```diff
@@ -49,11 +58,44 @@ class AthenaAccessKeyProfileMapping(BaseProfileMapping):
...
+        return env_vars
+
+    def _get_temporary_credentials(self):  # type: ignore
+        """
+        Helper function to retrieve temporary short lived credentials
+        Returns an object including access_key, secret_key and token
+        """
+        from airflow.providers.amazon.aws.hooks.base_aws import AwsGenericHook
+
+        # AwsGenericHook takes care of secret masking for us
```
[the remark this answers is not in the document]
spencer-horton-wcq: thanks !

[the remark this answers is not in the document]
octiva: thanks !

tests/profiles/athena/test_athena_access_key.py
octiva: Agreed, removed

PR #761, opened 2023-12-11
Add package location to work with hatchling 1.19.0

### LINKED ISSUE #760: Cosmos tests failing due to hatchling v1.19.0 that mandates packages setting
See [this example](https://github.com/astronomer/astronomer-cosmos/actions/runs/7165058515/job/19524989835?pr=759) of cosmos tests that have been failing since earlier today that is because of a new release of [hatchling v1.19.0](https://github.com/pypa/hatch/releases/tag/hatchling-v1.19.0), where it is mandatory to set the package files location:

```
An error will now be raised for the wheel target if no file selection options are defined
```

See https://github.com/pypa/hatch/issues/1113 for other users experiencing the same issue.

I think a fix for this would be to add the following to the pyproject.toml:

```toml
[tool.hatch.build.targets.wheel]
packages = ["cosmos"]
```

jbandoro:
## Description

Fixes issue described in #760 where Cosmos tests are failing from a recent `hatchling` update that now requires package path to be specified if it can't be inferred from the package's name.

## Related Issue(s)

Closes #760 

## Breaking Change?

None

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,6 +138,9 @@ include = [
     "/cosmos",
 ]
 
+[tool.hatch.build.targets.wheel]
+packages = ["cosmos"]
+
 ######################################
 # TESTING
 ######################################


octiva (approved):

tatiana (approved):
You rock, @jbandoro ! Thanks a lot for keeping on top of things and solving this so quickly

Issue #724, opened 2023-11-28
Support disabling dbt "Anonymous usage stats"


ykuc:
How can I set send_anonymous_usage_stats=False?
https://docs.getdbt.com/reference/global-configs/usage-stats

jbandoro:
This is a great issue you brought up @ykuc since any cosmos profiles that are mapped from Airflow connections won't have the following block for [dbt usage stats](https://docs.getdbt.com/reference/global-configs/usage-stats) in the profiles.yml:
```yaml
config:
  send_anonymous_usage_stats: False
```
which means it defaults to True when not used.

There are 2 workarounds you could have at the moment for preventing anonymous usage stats tracking:
1. Supply your own `profiles.yml` file to cosmos, see docs [here on using your own file](https://astronomer.github.io/astronomer-cosmos/profiles/index.html#using-your-own-profiles-yml-file) where you could add the config above. Though the limitation of this is that you won't be able to use Airflow connections.
2. As noted in dbt docs linked above you could also use the environment variables: `DO_NOT_TRACK=1` or `DBT_SEND_ANONYMOUS_USAGE_STATS=False`. With cosmos currently you would have to use [operator_args](https://astronomer.github.io/astronomer-cosmos/configuration/operator-args.html#dbt-related) with either `"append_env": True` if you have the env var already set, or use `"env": {"DO_NOT_TRACK": "1"}`.

In Cosmos to add support for not sending usage stats, I think this could be added to `BaseProfileMapping` args and then it could be added to profile contents in the method below:

https://github.com/astronomer/astronomer-cosmos/blob/e1f34ea424b8940f1c06660554298008876ea11b/cosmos/profiles/base.py#L175-L180

PR #769, opened 2023-12-15
add-configs-in-profile

ykuc:
## Description
https://github.com/astronomer/astronomer-cosmos/issues/724

## Checklist


tatiana:
@ykuc

cosmos/config.py
```diff
@@ -212,6 +212,8 @@ class ProfileConfig:
     :param target_name: The name of the dbt target to use.
     :param profiles_yml_filepath: The path to a profiles.yml file to use.
     :param profile_mapping: A mapping of Airflow connections to dbt profiles.
+    :param dbt_config_vars: Dictionary of dbt configs for the project. This argument overrides configs defined in your profiles.yml
+        file. The dictionary is dumped to a yaml string. Details https://docs.getdbt.com/docs/core/connect-data-platform/profiles.yml
```
jbandoro, line 216: The arg description here says it overrides configs in the profiles.yml, but if a user uses `profiles_yml_filepath` instead of `profile_mapping` it does not update the profile config in that file. It should be made clear that configs are added to the mapped profile file and an exception should be raised if `profiles_yml_filepath` is used with `dbt_config_vars`.

cosmos/config.py
```diff
@@ -236,6 +240,44 @@ def validate_profile(self) -> None:
...
+            "printer_width": {'var_type': int},
+            "write_json": {'var_type': bool},
+            "warn_error": {'var_type': str},
+            "log_format": {'var_type': str, 'accepted_values': {"text", "json", "default"}},
+            "debug": {'var_type': bool},
+            "version_check": {'var_type': bool},
+            "fail_fast": {'var_type': bool},
+            "use_experimental_parser": {'var_type': bool},
+            "static_parser": {'var_type': bool},
+        }
```
jbandoro, line 259: Some of these configuration options at the moment will not affect how Cosmos will execute so it may be worth considering if we want to expose these for configuration.

For example [fail_fast ](https://docs.getdbt.com/reference/global-configs/failing-fast) for `dbt run` will prevent other models from running, however since Cosmos executes each model as its own task, it will not terminate other tasks that are running in their own subprocess. This may cause confusion for a user if one task fails on an error but the other ones are still running if they set this configuration option. The same with `use_colors` will not have an effect on task logs.

jbandoro (commented):
@ykuc thanks for opening up this PR, being able to add configs to the mapped profile is really cool and would be great to support! There are some items that need to be addressed below:

1. Did you see that #768 was merged that adds an option for disabling dbt event tracking? Since `dbt_config_vars` also allows users to configure anonymous event tracking, even though #768 has not been released yet, we should discuss if we're going to support both or only use `dbt_config_vars` and then update the Cosmos docs.
2. As @tatiana commented, unit tests need to be added.
3. Cosmos docs also need to be updated. Specifically this one: https://github.com/astronomer/astronomer-cosmos/blob/main/docs/templates/index.rst.jinja2

ykuc:
Hi, @jbandoro 
1. Sorry but I did see it... Do you think will we need this pr?
2. As @tatiana commented, unit tests need to be added. Thanks see it, will fix it soon
3. will do it

jbandoro:
> 1. Sorry but I did see it... Do you think will we need this pr?

This is great, I just wanted to flag it for you since your branch had merge conflicts and wasn't sure if we were both working on it at the same time.

ykuc:
done

1. fix type check
2. add description and warn if profiles_yml_filepath is used with dbt_config_vars or we need exception?

to do

1. check configuration https://github.com/astronomer/astronomer-cosmos/pull/769#discussion_r1428258595
2. merge conflicts https://github.com/astronomer/astronomer-cosmos/pull/768 Since dbt_config_vars also allows users to configure anonymous event tracking, even though https://github.com/astronomer/astronomer-cosmos/pull/768 has not been released yet, we should discuss if we're going to support both or only use dbt_config_vars and then update the Cosmos docs.
3. add unit tests
4. update docs. Specifically this one: https://github.com/astronomer/astronomer-cosmos/blob/main/docs/templates/index.rst.jinja2

cosmos/config.py
ykuc: @jbandoro Hi,
I think we can support:
- send_anonymous_usage_stats
- partial_parse
- use_experimental_parser
- static_parser
- printer_width
- write_json
- warn_error
- warn_error_options
- log_format
- debug
- version_check

ykuc:
finish pr in https://github.com/astronomer/astronomer-cosmos/pull/794

PR #772, opened 2023-12-21
Remove upper limit to Pydantic

jlaneve:
## Description

[Airflow's upgraded to Pydantic 2.0 (!!)](https://github.com/apache/airflow/pull/35551), so this removes the restriction on Cosmos to ensure it's compatible.

## Breaking Change?

No

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,7 +118,7 @@ kubernetes = [
     "apache-airflow-providers-cncf-kubernetes>=5.1.1",
 ]
 pydantic = [
-        "pydantic>=1.10.0,<2.0.0",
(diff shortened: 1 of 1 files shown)


tatiana (approved):
Thanks @jlaneve for fixing this!

tatiana:
Thanks, @jlaneve !

Giving additional context, this change is particularly relevant from Airflow 2.8 (https://pypi.org/project/apache-airflow/2.8.0/) or, in the case of Astronomer runtime 10 users (https://docs.astronomer.io/astro/runtime-release-notes#astro-runtime-1000), both of which released on the 18th December 2023.

PR #779, opened 2024-01-03
Add Airflow 2.8 to test matrix and update dbt & Airflow conflicts doc

tatiana:
Apache Airflow 2.8 was released on 18 December 2023:
https://pypi.org/project/apache-airflow/2.8.0/

Add Airflow 2.8 to Cosmos integration tests matrix.

Update Airflow & dbt dependencies conflict document to include the new conflicts introduced by this release:

### DIFF
--- a/docs/getting_started/execution-modes-local-conflicts.rst
+++ b/docs/getting_started/execution-modes-local-conflicts.rst
@@ -1,10 +1,10 @@
 .. _execution-modes-local-conflicts:
 
-Airflow and DBT dependencies conflicts
+Airflow and dbt dependencies conflicts
 ======================================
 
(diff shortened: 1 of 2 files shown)


jlaneve (approved):
yikes, does this mean airflow 2.8 only works with dbt 1.5? so a user will essentially always use an alternative execution mode, right?

tatiana:
> yikes, does this mean airflow 2.8 only works with dbt 1.5? So a user will essentially always use an alternative execution mode, right?

Hey @jlaneve, yes, it's a shame. It seemed we were heading in a good direction, with "core" dbt and Airflow dependencies no longer conflicting - but the recent Pydantic change in Airflow 2.8 led to issues again

Issue #754, opened 2023-12-08
Support individual model runs (2 possible implementations)


dwreeves:
## Context

A common pattern is-- I will deploy a code change, and I will want to either do a full refresh on an existing model, or run a new model so I can see it immediately in production without waiting on the next scheduled DAG run. 

I've been running dbt inside Airflow since prior to Astronomer Cosmos even existing, and this usage pattern has come up a lot in my experience. Right now, I don't think right now that Astronomer Cosmos supports it very well. The recent addition in #623 is much needed for supporting the pattern I am describing, but it does not go far enough.

Right now, in our Astronomer Cosmos deployment, we have a parameterized DAG that allows for this single node invocation, but in order to implement it, I needed to subclass the dbt operator and make everything I wanted into templated fields. I think though that astronomer-cosmos should support this pattern more directly without needing to do that.

## Implementation idea #1 (simpler)

Allow the user to more easily parameterize a DAG with the following additions:

- Make `dbt_cmd` a template field, and allow it to support `str` types (and `if isinstance(dbt_cmd, str): dbt_cmd = [dbt_cmd]`).
- Make `select`, `selector`, etc. all template fields.

This will allow the user to set up a simple DAG that passes in `Param()`s that looks something like this:

```python
with DAG(
    ...,
    params={
        "full_refresh": Param(default=True, ...),
        "dbt_cmd": Param(default="build", ...)
        "select": Param(...)
    },
    render_template_as_native_obj=True
) as dag:
    DbtLocalOperator(
        task_id="dbt_task",
        full_refresh="{{ params.full_refresh }}",
        dbt_cmd=" {{ params.dbt_cmd }}"
        select="{{ params.select }}"
    )
```

The main benefit to this approach compared to the alternative is it does not expand the high level API surface, however it does add functionality to the existing base operator that would be dangerous to ever deprecate. However I do not think the added functionality is hard to support or is even a downside to have; templating these fields could be beneficial to many users.

## Implementation idea #2 (more opinionated, easier to set up)

Set up an "individual run" DAG class for the user, that is pre-parameterized.

```python
from cosmos import DbtParametrizedDag

parametrized_dag = DbtParametrizedDag(
    profile_config=...,
    dag_id="dbt_run_manually",
    schedule_interval=None,
    start_date=...
)
```

[...]

PR #786, opened 2024-01-07
Add more template fields to `DbtBaseOperator`

dwreeves:
This partially addresses #754 via allowing for built-in templating support for the `DbtBaseOperator`.

I also noticed `--full-refresh` was not documented so I added that in.

## Still missing

Manual run pattern is not documented; the fact that these fields are templated is not documented. I don't really know where in the docs to put this. The docs are very API-focused more than narrative-based or suggestive, and Cosmos's maintainers prefer this style of documentation so it's hard to find a spot for this. It's possible that that's fine and we just keep this as a feature for more advanced users who dig into source code to discover for themselves? 🤷


tatiana (commented):
Hey @dwreeves, thanks a lot for the contribution!

Great catch on the missing docstring and docs.

>The docs are more API-focused than narrative-based or suggestive, and Cosmos's maintainers prefer this documentation style, so it's hard to find a spot for this. 

We can and should certainly improve Cosmos docs. I agree they are more API-focused, but it isn't a maintainers' preference. We're open to having other styles of docs, and I'd be super happy if we could follow https://diataxis.fr/. It has not been a priority so far - but we'd welcome help.

Regarding template fields, given the existing docs, I think we could either:

(1) Create a new page inside `docs/configuration` with a list of existing template fields

(2) Highlight the Cosmos arguments that are template fields where we expose them. Eg. In the case of `select`, `exclude` and `selection`, this could be in:
https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html

In the case of env and vars, we could probably set it in `dbt_vars` and `env_vars` in:

We should probably add `full_refresh` together with `models`:
https://astronomer.github.io/astronomer-cosmos/configuration/operator-args.html

As of compiled SQL, it seems we have a dedicated section for it:
https://astronomer.github.io/astronomer-cosmos/configuration/compiled-sql.html
Which could probably be merged into `operator_args`.

joppevos:
@dwreeves upps I missed documenting it. Thanks for taking care of it 🙇

dwreeves:
I think the thing that would make the most sense for now is to just add a section under [/operator-args.html](https://astronomer.github.io/astronomer-cosmos/configuration/operator-args.html) that is just about the templated variables. With the exception of `--full-refresh`, all of these variables are either (1) potentially used for rendering if `DBT_LS` is true (`vars` and `env`), or (2) used for model selection (`select`, `selector`, `models`, `exclude`) and are thus typically out of scope for `operator_args` inside of the dbt->Airflow converter context. A note should be included that clarifies this, although it's a little confusing because `--full-refresh` is essentially the only variable where it would make sense to do something like: `operator_args={"full_refresh": "{{ params.full_refresh }}"}`.

> (2) Highlight the Cosmos arguments that are template fields where we expose them. Eg. In the case of select, exclude and selection, this could be in:
https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html

So this is a bit weird because templating does not work when _rendering_ a DAG, only when _executing_ the task. I will keep in mind that a note should be made about this.

---

This also does beg a question though about the recent treatment of the `dbt_vars` and `dbt_env`. Right now in `validate_initial_user_config()`, an error is raised when both `operator_args={"vars": ...}` and `project_config.dbt_vars` are set. Since both rendering and execution use `--vars` and environment variables, this means that rendering and execution are strongly coupled and it's not possible to have these differ.

This was something I was thinking about irrespective of this particular PR. I think it's possible that it should be possible to set both `operator_args={"vars": ...}` and `project_config.dbt_vars`, and that the behavior would be that `project_config.dbt_vars` is used for rendering and `operator_args={"vars": ...}` could either merge or overwrite. Alternatively, render config and execution config could each have their own notion of dbt vars that can get set to the project configs' vars if either is not defined at the execution or rendering level.

In general the render and execution configs share a lot of configuration in the case that `DBT_LS` needs to be run, and maybe a separate discussion but I think there should be some sort of way to make the configurations more decoupled at the user level, but more DRY in their under-the-hood implementations. If that makes sense.

---

The reason why this discussion is potentially relevant to this PR is because of my note above about how the only operator args you can seemingly reliably utilize templating for at Cosmos's "highest level" API (i.e. DbtDag + DbtTaskGroup) is `full_refresh`.

I mean, you can access `vars` and `env` technically, but these will either:

-  if set in `operator_args`, they will raise deprecation warnings as of 1.3
- if set in `project_config` _and_ `LoadMode.DBT_LS` is used, then it will not properly render at all during `LoadMode.DBT_LS` rendering, i.e. the string literal will be used for either the dbt var or env var, since it never gets passed through Jinja and the scheduler wouldn't have potentially relevant context variables anyway.

This all feels both somewhat suboptimal, and also extremely confusing, to users. It's strange that `ProjectConfig(dbt_vars={'logical_date': '{{ context["logical_date"] }}')` is valid code, and works perfectly fine and as-intended if the user is using `LoadMode.DBT_MANIFEST`, but does not behave so well when using `LoadMode.DBT_LS`. It's tricky to convey this in documentation as a general statement. "Templating does not work with `ProjectConfig(dbt_vars=...)`" is not a true statement, but it is not a false statement either. The truth is templating works only in execution, and may or may not fail in rendering. Here are 2 example cases when it fails in rendering:

- Since executing `dbt ls` does involve dbt parsing all the Jinja, and if your SQL query has something like `{{ modules.datetime.datetime.fromisoformat(var("logical_date")) }}` then the code will fail when _rendering_ even though it would succeed when _executing_. In this case I think the user would want to either not have `logical_date` set at all (and rely on a default instead), or alternatively they should be allowed to set `logical_date` differently in rendering and execution.

- Env vars can be a little more realistic a problem if, say, an env var is being used as part of a connection or to enable or disable models. For example, imagine the user sets a dbt profile with something like `role: {{ env_var("DBT_ROLE") }}`, and in their DAG they want to set `DAG(..., params={"dbt_role": Param(...)})` and inside a DbtTaskGroup they want to set `DbtTaskGroup(..., project_config=ProjectConfig(..., dbt_vars={"dbt_role": "{{ params.dbt_role }}"}))`. Here, again, it works fine in execution, but it does not work in the case where the load method is `LoadMode.DBT_LS` specifically.

That said, I am aware these are potentially extreme edge cases-- for example in most cases passing the string literal `{{ logical_date }}` into the dbt rendering will have zero impact on parsing with `dbt ls` and everything will render fine. So this _adds_ to the confusion! The fully qualified correct statement is: **Templating works in execution but does not work in rendering when the load method is `LoadMode.DBT_LS`; it may or may not matter depending on your dbt project.** And that's just way more confusing than saying it doesn't work with `dbt ls` fully unqualified, and instead pushing users toward a separate layer of abstraction in the event that they want access to templating.

---

OK, this is all very confusing and in the weeds so I hope what I am saying makes sense. TLDR is:

- Probably just stick this in `operator-args.html` as a separate section.
- I would not mention this at all in `ProjectConfig`; and if we do, we may want to revisit some of the 1.3 changes regarding deprecation of `vars=` and `env=` in the `operator_args`?
- The use of templating is actually really complicated and isn't fully addressed by the abstractions. It may be best to revisit this.
- For now, without any further code changes, the only real thing that can be templated via DbtDag or DbtTaskGroup is `full_refresh`. (Maybe also warn users that `render_template_as_native_obj=True` should be true for that to work, since right now it cannot handle strings; alternatively make it so `full_refresh` can handle strings??).

dwreeves:
@tatiana:

- I have updated the docs
- I have made another minor change to handle type-checking for templating of `full_refresh`.
- I added `full_refresh` to the `DbtBuildMixin`

Sorry about the above long post, but it is still relevant to the use of ``env`` and ``vars`` in templating and DAG execution. Specifically the case of passing `{{ logical_date }}` as an environment variable or dbt var is far from being a stretch and the current API is not super friendly to this.

To _partially_ address it I added the following note in the docs:

> ```rst
> .. note::
>     Using Jinja templating for ``env`` and ``vars`` may cause problems when using ``LoadMode.DBT_LS`` to render your DAG.
> ```

---

I would love it if this PR made it into Cosmos 1.4. I feel like this is a pretty straightforward PR and shouldn't be hard to merge or have any unintended side effects.

I discussed with @jlaneve that I have a massive blog post with multiple optimizations (including one I've never seen discussed anywhere that shaves an average of 3.3 seconds off each task execution; that really adds up). In that blog post, one of the "quality of life" improvements I wrote about assumes this PR is merged into Cosmos 1.4: 😄👉👈 

image

dwreeves:
It looks like the only reason this is failing is due to the strict coverage requirement. Hrm. I really really do want this available for 1.4.0, but I don't know when I'll have time to address that. Probably this weekend at the earliest.

dwreeves:
Cov requirements are still failing. There is no universe where I get this done before the weekend, just as an FYI.

tatiana:
Thanks, @dwreeves , I just managed to get some time to add the missing coverage bit - it will be released as part of the alpha today!

dwreeves:
Yes! Thank you! Can't wait for this 😁

docs/configuration/operator-args.rst
```diff
@@ -88,3 +89,34 @@ Sample usage
...
+
+Some of the operator args are `template fields <https://airflow.apache.org/docs/apache-airflow/stable/howto/custom-operator.html#templating>`_ for your convenience.
+
+These template fields can be useful for hooking into Airflow `Params <https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/params.html>`_, or for more advanced customization with `XComs <https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/xcoms.html>`_.
+
+The following operator args support templating, and are accessible both through the  ``DbtDag`` and ``DbtTaskGroup`` constructors in addition to being accessible standalone:
+
+- ``env``
+- ``vars``
+- ``full_refresh`` (for the ``build``, ``seed``, and ``run`` operators.)
```
tatiana, line 105:
```suggestion
- ``full_refresh`` (for the ``build``, ``seed``, and ``run`` operators since Cosmos 1.4.)
```

docs/configuration/operator-args.rst
```diff
@@ -88,3 +89,34 @@ Sample usage
...
+
+.. note::
+    Using Jinja templating for ``env`` and ``vars`` may cause problems when using ``LoadMode.DBT_LS`` to render your DAG.
+
+The following template fields are only selectable when using the operators in a standalone context:
```
tatiana, line 110:
```suggestion
The following template fields are only selectable when using the operators in a standalone context (starting in Cosmos 1.4):
```

docs/configuration/operator-args.rst
```diff
@@ -88,3 +89,34 @@ Sample usage
...
+
+The following template fields are only selectable when using the operators in a standalone context (starting in Cosmos 1.4):
+
+- ``select``
+- ``exclude``
+- ``selector``
+- ``models``
+
+The aforementioned args are not available to be templated via ``DbtDag`` and ``DbtTaskGroup`` because they need to select dbt nodes to render the DAG's tasks.
+Since template fields are rendered on each ``DagRun``,
```
tatiana, line 118:
```suggestion
```

docs/configuration/operator-args.rst
```diff
@@ -88,3 +89,34 @@ Sample usage
...
+    Using Jinja templating for ``env`` and ``vars`` may cause problems when using ``LoadMode.DBT_LS`` to render your DAG.
+
+The following template fields are only selectable when using the operators in a standalone context (starting in Cosmos 1.4):
+
+- ``select``
+- ``exclude``
+- ``selector``
+- ``models``
+
+The aforementioned args are not available to be templated via ``DbtDag`` and ``DbtTaskGroup`` because they need to select dbt nodes to render the DAG's tasks.
```
tatiana, line 117:
```suggestion
Since Airflow resolves template fields during Airflow DAG execution and not DAG parsing,  the args above cannot be templated via ``DbtDag`` and ``DbtTaskGroup`` because both need to select dbt nodes during DAG parsing. 
```

tatiana (approved):
Looks great, thank you very much, @dwreeves both for the contribution and persistence to get this into the finishing line!

PR #788, opened 2024-01-08
fix docs and path for ls file (#773)

### LINKED ISSUE #773: dbt_ls_path: wrong docs and argument should support both Path and str types
[According to the documentation](https://github.com/astronomer/astronomer-cosmos/blob/main/docs/configuration/parsing-methods.rst?plain=1#L96) we support `str` as an argument to `dbt_ls_path`:

```
DbtDag(
        render_config=RenderConfig(
            load_method=LoadMode.DBT_MANIFEST, dbt_ls_path="/path/to/dbt_ls_file.txt"
        )
        # ...,
    )
```

1. There is a typo in the docs: it should be `LoadMode.DBT_LS_FILE`
2. The code in the doc will not run, since `dbt_ls_path` is expected to be a `Path` according to `dbt_ls_path: Path | None = None` in https://github.com/astronomer/astronomer-cosmos/blob/82e8db97d6a31757fb2da019fc36df52c342eb70/cosmos/config.py#L62C5-L62C36 . This is not consistent with other `*_path` arguments, which allow `str | Path`, for example `dbt_project_path: InitVar[str | Path | None] = None` in the line just above.

I propose to fix 1. and change allow `str | Path` types for the argument `dbt_ls_path`

flinz:
## Description

fix docs and support str argument for dbt_ls_file

## Related Issue(s)

closes #773 

## Breaking Change?

No

## Checklist

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -45,7 +45,8 @@ class RenderConfig:
     :param node_converters: a dictionary mapping a ``DbtResourceType`` into a callable. Users can control how to render dbt nodes in Airflow. Only supported when using ``load_method=LoadMode.DBT_MANIFEST`` or ``LoadMode.DBT_LS``.
     :param dbt_executable_path: The path to the dbt executable for dag generation. Defaults to dbt if available on the path.
     :param env_vars: (Deprecated since Cosmos 1.3 use ProjectConfig.env_vars) A dictionary of environment variables for rendering. Only supported when using ``LoadMode.DBT_LS``.
-    :param dbt_project_path Configures the DBT project location accessible on the airflow controller for DAG rendering. Mutually Exclusive with ProjectConfig.dbt_project_path. Required when using ``load_method=LoadMode.DBT_LS`` or ``load_method=LoadMode.CUSTOM``.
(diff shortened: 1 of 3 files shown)


tatiana (approved):
Looks great, thanks for fixing the docs, @flinz !

cosmos/config.py
```diff
@@ -70,6 +71,8 @@ def __post_init__(self, dbt_project_path: str | Path | None) -> None:
                 DeprecationWarning,
             )
         self.project_path = Path(dbt_project_path) if dbt_project_path else None
+        # allows us to initiate this attribute from Path objects and str
+        self.dbt_ls_path = Path(self.dbt_ls_path) if self.dbt_ls_path else None
```
tatiana, line 75: @flinz, could we add a unit test to cover the support for `dbt_ls_path` to be a string so we don't revert this behaviour unintentionally in future?

flinz: @tatiana I added one simple test that invokes the constructor with a string argument.
https://github.com/flinz/astronomer-cosmos/commit/c69e0c7d6f65b38c20e87e11aa0592657d68fe8c

PR #784, opened 2024-01-05
Bugfix disable event tracking throwing error

jbandoro:
## Description

As reported in #783 the new disabling event tracking feature was throwing an error. I didn't test an update I made changing the value from boolean to a string 🤦‍♂️. 

This PR fixes the issue and also adds it to a cosmo dag for integration testing so the error would be caught.

## Related Issue(s)

closes #783

## Breaking Change?

None
## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #794, opened 2024-01-11
Add dbt profile config variables to mapped profile

ykuc:
## Checklist


ykuc:
@jbandoro  Hi! could you check it?)

ykuc:
[old pr version](https://github.com/astronomer/astronomer-cosmos/pull/769)

cosmos/profiles/base.py
```diff
@@ -24,6 +25,66 @@
 logger = get_logger(__name__)
 
 
+@dataclasses.dataclass
+class DbtConfigVars:
```
jbandoro, line 29: This is great! Can this class instead be named `DbtProfileConfigVars`, so there is no confusion with other dbt [configs](https://docs.getdbt.com/reference/dbt-jinja-functions/config).

cosmos/profiles/base.py
```diff
@@ -24,6 +25,66 @@
...
+                var_type = vars_check.get("var_type", Any)
+
+                if not isinstance(field_value, var_type):
+                    raise CosmosValueError(f"dbt config var {field_name}: {field_value} must be a {var_type}")
+
+                if accepted_values:
+                    if field_value not in accepted_values:
+                        raise CosmosValueError(
+                            f"dbt config var {field_name}: {field_value} must be one of {accepted_values}"
+                        )
```
jbandoro, line 72: You could instead use [pydantic dataclasses](https://docs.pydantic.dev/1.10/usage/dataclasses/) so you could have the type hints be used in the validation itself so you wouldn't need the dictionary that repeats the dataclass fields. e.g. something like the below with the accepted types you're currently using:
```python
from typing import Literal
from pydantic.dataclasses import dataclass

@dataclass
class DbtConfigVars:
    send_anonymous_usage_stats: bool | None = False
    warn_error_options: Literal["include", "exclude"] | None = None
    log_format: str | None = None
...
```
Then you wouldn't even need the post init and `_validate_data` methods.

cosmos/profiles/base.py
```diff
@@ -41,11 +102,13 @@ class BaseProfileMapping(ABC):
 
     _conn: Connection | None = None
 
-    def __init__(self, conn_id: str, profile_args: dict[str, Any] | None = None, disable_event_tracking: bool = False):
+    def __init__(
+        self, conn_id: str, profile_args: dict[str, Any] | None = None, dbt_config_vars: DbtConfigVars | None = None
+    ):
```
jbandoro, line 107: This is a breaking change since `disable_event_tracking` was [released in v1.3.0](https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.3.0) and cosmos follows [semver versioning](https://semver.org/), and we cannot remove `disable_event_tracking` until cosmos v2.0. So it can't be removed, but we could consider warning users in v1.4.0 (when this feature is released) that `disable_event_tracking` is deprecated and will be removed in 2.0.

cosmos/profiles/base.py
```diff
@@ -180,8 +243,9 @@ def get_profile_file_contents(
             }
         }
 
-        if self.disable_event_tracking:
-            profile_contents["config"] = {"send_anonymous_usage_stats": False}
+        config_vars = self.dbt_config_vars.as_dict()
+        if config_vars:
+            profile_contents["config"] = config_vars
```
jbandoro, line 248: Would suggest having a check with `if self.dbt_config_vars` and then calling the `as_dict()` method and adding it to the the profile contents.

dev/dags/basic_cosmos_dag.py
```diff
@@ -18,7 +18,7 @@
     profile_mapping=PostgresUserPasswordProfileMapping(
         conn_id="airflow_db",
         profile_args={"schema": "public"},
-        disable_event_tracking=True,
+        dbt_config_vars=DbtConfigVars(send_anonymous_usage_stats=True),
```
jbandoro, line 21: Should keep `disable_event_tracking` here.

docs/templates/index.rst.jinja2
```diff
@@ -83,19 +83,15 @@ but override the ``database`` and ``schema`` values:
...
-when dbt commands are invoked. Users have an option to opt out of event tracking by updating their ``profiles.yml`` file.
-
-If you'd like to disable this behavior in the Cosmos generated profile, you can pass ``disable_event_tracking=True`` to the profile mapping like in
-the example below:
+The parts of ``profiles.yml``, which aren't specific to a particular data platform `dbt docs <https://docs.getdbt.com/docs/core/connect-data-platform/profiles.yml>`_
 
 .. code-block:: python
 
-    from cosmos.profiles import SnowflakeUserPasswordProfileMapping
+    from cosmos.profiles import SnowflakeUserPasswordProfileMapping, DbtConfigVars
```
jbandoro, line 94: Need to keep disable event tracking docs, but you can add Dbt config vars for release 1.4.0. Would also suggest showing an example with all of the options that are available.

tests/profiles/test_base_profile.py
```diff
@@ -36,17 +36,58 @@ def test_validate_profile_args(profile_arg: str):
         )
 
 
-@pytest.mark.parametrize("disable_event_tracking", [True, False])
-def test_disable_event_tracking(disable_event_tracking: str):
+@pytest.mark.parametrize("dbt_config_var,dbt_config_value", [("debug", True), ("debug", False)])
+def test_validate_dbt_config_vars(dbt_config_var: str, dbt_config_value: Any):
```
jbandoro, line 40: this test should be kept to validate that `disable_event_tracking` still works as expected and you can add the new tests below.

jbandoro (commented):
Thanks @ykuc this is looking good! Main comment below related to the breaking change that was introduced since cosmos v1.3.0 was released.

cosmos/profiles/base.py
ykuc: fix

cosmos/profiles/base.py
ykuc: airflow have version of pydantic. It depend on airflow version. Will check diff

cosmos/profiles/base.py
ykuc: returned)

cosmos/profiles/base.py
ykuc: fix

dev/dags/basic_cosmos_dag.py
ykuc: fix

docs/templates/index.rst.jinja2
ykuc: add

docs/templates/index.rst.jinja2
```diff
@@ -85,6 +85,8 @@ you specify in ``ProfileConfig``.
 
 Disabling dbt event tracking
 --------------------------------
+
+    Note: deprecated and will be removed in 2.0.0. Use dbt_config_vars instead.
 .. versionadded:: 1.3
```
jbandoro, line 90: I'd suggest moving the note into a sphinx note block like below and  you should also mention how to use dbt config vars to have the same effect:
```
.. note:
   Deprecated in v.1.4 and will be removed in v2.0.0. Use dbt config vars instead with ....
.. versionadded:: 1.3
```

docs/templates/index.rst.jinja2
```diff
@@ -112,6 +114,43 @@ the example below:
 
     dag = DbtDag(profile_config=profile_config, ...)
 
+Dbt config vars
```
jbandoro, line 117: nit: can we rename this section to Dbt profile config variables

cosmos/profiles/base.py
```diff
@@ -180,8 +248,14 @@ def get_profile_file_contents(
...
         }
 
+        if self.dbt_config_vars:
+            profile_contents["config"] = self.dbt_config_vars.as_dict()
+
         if self.disable_event_tracking:
-            profile_contents["config"] = {"send_anonymous_usage_stats": False}
+            logger.warning(
+                "disable_event_tracking is deprecated and will be removed in 2.0. You can use instead dbt_config_vars"
+            )
```
jbandoro, line 257: This should be moved to the constructor and use the deprecation warning category, see an example here: https://github.com/astronomer/astronomer-cosmos/blob/1b7f4a06e06cb5af75bf411a28dca932ef37ab5d/cosmos/config.py#L69-L72
Would also suggest being explicit on how to get the same behavior by adding "send_anonymous_usage_stats=False"

cosmos/profiles/base.py
```diff
@@ -41,11 +102,18 @@ class BaseProfileMapping(ABC):
 
     _conn: Connection | None = None
 
-    def __init__(self, conn_id: str, profile_args: dict[str, Any] | None = None, disable_event_tracking: bool = False):
+    def __init__(
+        self,
+        conn_id: str,
+        profile_args: dict[str, Any] | None = None,
+        disable_event_tracking: bool | None = None,
+        dbt_config_vars: DbtProfileConfigVars | None = None,
```
jbandoro, line 110: I think in the meantime cosmos will need to properly handle the case where `disable_event_tracking=True` and `DbtProfileConfigVars(send_anonymous_usage_stats=True)` and should raise a [CosmosValueError](https://github.com/astronomer/astronomer-cosmos/blob/1b7f4a06e06cb5af75bf411a28dca932ef37ab5d/cosmos/exceptions.py#L4-L5)
This will need a unit test.

cosmos/profiles/base.py
```diff
@@ -41,11 +102,18 @@ class BaseProfileMapping(ABC):
...
+    ):
         self.conn_id = conn_id
         self.profile_args = profile_args or {}
         self._validate_profile_args()
         self.disable_event_tracking = disable_event_tracking
+        self.dbt_config_vars = dbt_config_vars or DbtProfileConfigVars()
```
jbandoro, line 116: Since the default for `dbt_config_vars = None` why do you need to instantiate an empty `DbtProfileConfigVars` here, I think leaving it as `None` will work below.

cosmos/profiles/base.py
ykuc: fix)

tests/profiles/test_base_profile.py
ykuc: return it)

cosmos/profiles/base.py
ykuc: fix and add test

cosmos/profiles/base.py
ykuc: fix

cosmos/profiles/base.py
ykuc: have problem in python 3.8

tests/profiles/test_base_profile.py
```diff
@@ -47,6 +48,75 @@ def test_disable_event_tracking(disable_event_tracking: str):
...
     if disable_event_tracking:
-        assert profile_contents["config"]["send_anonymous_usage_stats"] is False
+        assert profile_contents["config"]["send_anonymous_usage_stats"] == "False"
+
+
+def test_disable_event_tracking_and_send_anonymous_usage_stats():
+    expected_cosmos_error = (
+        "Cannot set both disable_event_tracking and "
+        "dbt_config_vars=DbtProfileConfigVars(send_anonymous_usage_stats ..."
+    )
```
jbandoro, line 59: This expected error message is not checked in the test assertion below. You can use `with pytest.raises(CosmosValueError, match='regex match here')` to check that the raised error message matches a regex pattern.

cosmos/profiles/base.py
jbandoro: What was the issue? Can you point to the test that was failing for Python 3.8. Sometimes with 3.8 the failures are from type hint annotations like `some_arg: str | None`  that can be fixed by added `from __future__ import annotations`.

ykuc: Fix, thanks for help)
the main problem is that I did it at night)

jbandoro: That's great it looks much cleaner using pydantic dataclasses for validation!

cosmos/profiles/base.py
```diff
@@ -180,8 +249,24 @@ def get_profile_file_contents(
...
+            )
+            if (
+                isinstance(self.dbt_config_vars, DbtProfileConfigVars)
+                and self.dbt_config_vars.send_anonymous_usage_stats is not None
+            ):
+                raise CosmosValueError(
+                    "Cannot set both disable_event_tracking and "
+                    "dbt_config_vars=DbtProfileConfigVars(send_anonymous_usage_stats ..."
+                )
+            profile_contents["config"] = {"send_anonymous_usage_stats": "False"}
```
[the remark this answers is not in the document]
ykuc: fix

tests/profiles/test_base_profile.py
```diff
@@ -47,6 +48,75 @@ def test_disable_event_tracking(disable_event_tracking: str):
     )
     profile_contents = yaml.safe_load(test_profile.get_profile_file_contents(profile_name="fake-profile-name"))
 
-    assert ("config" in profile_contents) == disable_event_tracking
     if disable_event_tracking:
-        assert profile_contents["config"]["send_anonymous_usage_stats"] is False
+        assert profile_contents["config"]["send_anonymous_usage_stats"] == "False"
```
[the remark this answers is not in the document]
ykuc: fix

tests/profiles/test_base_profile.py
ykuc: fix)

tests/profiles/test_base_profile.py
```diff
@@ -47,6 +49,75 @@ def test_disable_event_tracking(disable_event_tracking: str):
     )
     profile_contents = yaml.safe_load(test_profile.get_profile_file_contents(profile_name="fake-profile-name"))
 
-    assert ("config" in profile_contents) == disable_event_tracking
```
jbandoro, line 50: This line should still be kept here to assert that if only `disable_event_tracking` is used and is False that config isn't added to the profile contents.

dev/dags/basic_cosmos_dag.py
```diff
@@ -18,7 +18,7 @@
     profile_mapping=PostgresUserPasswordProfileMapping(
         conn_id="airflow_db",
         profile_args={"schema": "public"},
-        disable_event_tracking=True,
+        dbt_config_vars=DbtProfileConfigVars(send_anonymous_usage_stats=True),
```
jbandoro, line 21: `disable_event_tracking` should be kept here so that issues like that which was fixed in #784 are caught in the integration tests.

Maybe a different field can be set in `dbt_config_vars` here?

tests/profiles/test_base_profile.py
ykuc: fix

dev/dags/basic_cosmos_dag.py
ykuc: fix

cosmos/profiles/base.py
```diff
@@ -24,6 +24,31 @@
...
+
+    def as_dict(self) -> dict[str, Any] | None:
+        result = {
+            field.name: getattr(self, field.name)
+            for field in self.__dataclass_fields__.values()
+            if getattr(self, field.name) is not None
+        }
+        if result != {}:
+            return result
+        return None
```
jbandoro, line 49: Can you add a test for returning `None` here as identified by Codecov

jbandoro (approved):
Thanks @ykuc for your contribution and work to get this ready, it looks good!

One comment below on code coverage.

cosmos/profiles/base.py
ykuc: add)

ykuc:
@jbandoro I fix last comment. If it ok, I will merge)

tatiana (approved):
This looks great, thanks for the great work @ykuc and @jbandoro !

Issue #722, opened 2023-11-28
Tasks are running full parse each time 


DanMawdsleyBA:
I've noticed that within each job it is running a full parse each time. I upload the manifest as part of the cicd so not sure why it isn't recognizing the existing manifest file. This could significantly improve the job performance currently full parse step takes 7 seconds per task. 
![image](https://github.com/astronomer/astronomer-cosmos/assets/134394384/38fe30e0-4ac8-4577-90cd-c8274b23c67c)

DanMawdsleyBA:
I think this is due to the syslink again as it does not copy across the target path by default. We could add a step to check if the manifest file is present then syslink just the manifest file. This should mean each task does not need to be reparsed fully each time

lrcuwicRSG1:
Hi @all, I am new to astronomer cosmos and I have the same performance issues as @DanMawdsleyBA noticed. When will it be fixed? @tatiana  Milestone 1.4.0 was updated for the release date

In the meanwhile I want to fix it temporary with the codesnippet from dosubot. What is the easiest way to modiy the graph.py file and take it over to my local docker-installation based on astronomer-runtime and astro-cli? I am looking forward to your suggestions. Thanks Chris

tatiana:
@DanMawdsleyBA @lrcuwicRSG1 Around two weeks ago, we managed to have an alpha release with #800 by @dwreeves, which seems to solve the problem. Could you try it out using [1.4.0a1](https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.4.0a1)? We're planning to release the stable 1.4 with this feature next week, and it would be great to confirm if/how much it helped you

lrcuwicRSG1:
@tatiana Thanks for your reply. The issue #800 was on my watchlist ;) I will try it soon. Cheers Chris

DanMawdsleyBA:
I am using 1.4.0a1 and wasn't able to get it to work because DBT was saying the profile and env vars have changed. 

[2024-03-14, 17:04:56 GMT] {{subprocess.py:94}} INFO - [0m17:04:56  Running with dbt=1.7.9
[2024-03-14, 17:04:57 GMT] {{subprocess.py:94}} INFO - [0m17:04:57  Registered adapter: snowflake=1.7.2
[2024-03-14, 17:04:57 GMT] {{subprocess.py:94}} INFO - [0m17:04:57  Unable to do partial parsing because config vars, config profile, or config target have changed
[...]

Issue #362, opened 2023-07-11
Slimming down tmp_dir to just the --target-dir


dwreeves:
# Background on Astronomer-Cosmos code

The `DbtLocalBaseOperator` copies the entire project directory into a temporary directory before executing. The reason why it does this is because usually where the dbt project actually exists may be read-only, and dbt attempts to write to two different locations during its runtime: to the logs folder and to the target folder.

# Issue this can cause

The Airflow Celery executor will execute multiple tasks concurrently on a single worker instance.

For a large-ish dbt project, if multiple tasks are executed concurrently on a single worker, and the worker runs out of space from these copies, the executor will SIGKILL all tasks on the worker node. Sigkilling the tasks means that the temporary directories do not get cleaned up, and the worker is now FUBAR because it is out of space.

So, this isn't just a nice-to-have optimization. Copy+pasting the entire project directory can cause incredibly frustrating issues for end users.

This also isn't just theoretical-- I've actually run into this. Not via Astronomer-Cosmos, but via my own Airflow+dbt implementation I've had prior to the existence of Astronomer-Cosmos, where we did the same approach Astronomer-Cosmos does for a while (create tmpdir, copy whole project directory into it), until our dbt project directory got too big to keep doing this.

# Workaround

As of dbt 1.5.2 and dbt 1.6.0, it should be possible to work around this problem. Remember the two directories that dbt writes to are `logs/` and `target/`:

- The `--target-path` flag can be set for any subcommand that uses it (dbt 1.5.0 had issues here but I believe this should be fully fixed as of 1.5.2).
- The `--log-level-file` (option for the `dbt` base command) can be set to `none`, which prevents writing logs to the file system entirely.

So the steps here are, instead of copying the entire project directory to the tmp dir, instead:

- Create a new temporary directory that will act as the `--target-path`.
- Optionally, copy `target/` from the project directory into the new temporary directory.
- Run the command `dbt --log-level-file none [subcommand] --target-path {tmpdir/target/} [options/args]`

# Complications adding to Astronomer-Cosmos

- There is no process for version-checking or asserting versioning in the Cosmos runtime, as far as I can tell. This only works with `dbt>=1.5.2`, and Cosmos supports much earlier versions of dbt than that.
- The `callback()` parameter runs on the project directory, and it seems difficult to implement backwards compatibility with that.
   - I'm not 100% sure what people are doing with that, but I assume some folks may specifically be looking to extract and dump the debug logs from the project directory.

[...]

PR #233, opened 2023-04-13
Run each operation in a temp dir

jlaneve:
This PR modifies the execution logic to run each operation in a temp dir. While this may be slightly less efficient, it's more reliable than file locking.

closes #228
closes #226
closes #217

PR #800, opened 2024-01-15
Support partial parsing

### LINKED ISSUE #785: Option to copy `partial_parse.msgpack` or `target/` dir into tmpdir.
# The Problem

For users with really large dbt projects, the entire project needs to be recompiled on each run.

`dbt-core` uses `partial_parse.msgpack` to get around this normally, but in Cosmos's execution, dbt is run out of a tmpdir with symlinks to each read-only dbt directory, and then everything else is fully recreated, including the entirety of the `target/` dir.

# A Solution

It may be beneficial to users to allow for the target directory to be copied into the tmpdir. Alternatively, for advanced use cases such as passing DagRun-specific `--vars` into the `dbt` operators (e.g. `{{ logical_date }}`), it may be useful to have some sort of `Compile` operator (something similar to the `DbtDocsOperator`) that is able to place `dbt compile` artifacts into a cloud storage directory (this would need to be templated so that it could be DagRun-specific) that could then be used for each model in the DagRun.

# API

The API is a little unclear to me but I imagine it would look something like this as an example:

```python
dbt_compile_task = DbtCompileS3Operator(
    aws_conn_id="my_aws_conn_id",
    bucket_name="my-dbt-artifacts",
    folder_dir="dbt_compile_artifacts/{{ run_id }}"
)

dbt_group = DbtTaskGroup(
    ...,
        ...,
        dbt_target_path="s3://my-dbt-artifacts/dbt_compile_artifacts/{{ run_id }}",
        dbt_target_path_conn_id="my_aws_conn_id",
        partial_parse=True
    )
)

dbt_compile_task >> dbt_group
```

Also, naturally you could use Airflow 2.8's new `ObjectStoragePath` feature for this, although (1) It doesn't help for users on older versions of Airflow, and (2) I'm not aware of if this supports templating out of the box?

I'm not fully satisfied with this API proposal-- for example, it's atypical in Airflow to have a `{thing}_conn_id` where `{thing}` refers to the variable or object being operated upon rather than what is being connected to. Just having `conn_id` also isn't necessarily appropriate because the `ExecutionConfig` also contains other paths that could plausibly be connected to as well (actually, that would be a pretty cool feature?  👀). But I think this is close to a good implementation. I'd love to hear ideas on any improvements to the high level API.

^ Note that this is the _full_ API and is designed to support advanced use cases such as users using `--vars`, who still want the full benefit of partial parsing. A typical user may just do something like, `dbt compile` as part of their deployment and they'd just use a local path:

```
# Just use `dbt_target_path is not None`

dbt_group = DbtTaskGroup(
    ...,
        ...,
        dbt_target_path="/usr/local/airflow/dags/dbt/target",
    )
)

# Or require `partial_parse=True`
dbt_group = DbtTaskGroup(
    ...,
        ...,
        partial_parse=True
    )
)
```

It's unclear if the API should be, `execution_config.target_path is not None` determines using `partial_parse.msgpack` or if there needs to explicitly be a flag for it:

```
dbt_group = DbtTaskGroup(
    ...,
        ...,
        dbt_project_path="/usr/local/airflow/dags/dbt",
        partial_parse=True,
    )
)
```

I am leaning toward having `partial_parse`. The reason why is: in this case where we use `partial_parse: bool`, this has the benefit that we can infer where the `target_path` is if the user does not supply one. Meaning a user who runs `dbt compile` with a very standard configuration, and who does not have any DagRun-specific `--vars`, would not need to also specify the target path. And this more explicitly maps the behavior that is occurring to the user-written code.

One last thing. Should `partial_parse` by default be `True`? I think **yes**, it should be the default, but it's not costless. In normal dbt operating circumstances, partial parse by default is essentially costless. For cosmos, there are two costs.

- First: (minor) `partial_parse.msgpack` will get copied into the tmpdir by default, which expands the utilization of the tmpdir. And when using virtual storage, this eats up some additional memory utilization on the scheduler.
- Second: (very very minor) the operator cannot warn (or fail) for lack of existence of a msgpack, unless you make it `bool | None` and have `None` mean "attempt to partial parse". This means that users who _think_ they are using partial parsing, but actually are not because the msgpack is missing, could end up surprised.

That said: (1) I think the benefits are high and this cost is low (2) the recent changes to symlink the read-only directories really helps to free up tmpdir space utilized by Cosmos (3) I think if a `partial_parse.msgpack` exists that it is natural for Cosmos to attempt to use it; otherwise the user can simply exclude it. For those reasons a default behavior of attempting to use a msgpack if one exists is sensible here.

### LINKED ISSUE #791: DBT_MANIFEST mode does not use manifest for task execution
I am using DBT_MANIFEST parsing mode, and seeing quite a bit of time being spent in each **task** on parsing the whole project, spending around 1/3 of the task total time on a full parse:

```
[2024-01-09, 16:51:27 CET] {subprocess.py:94} INFO - [0m15:51:27  Unable to do partial parsing because saved manifest not found. Starting full parse.
[2024-01-09, 16:51:30 CET] {subprocess.py:94} INFO - [0m15:51:30  Found 38 models, 6 tests, 20 sources, 0 exposures, 0 metrics, 532 macros, 0 groups, 0 semantic models
```

I see the command being executed for each task is along the lines of 
```
[2024-01-09, 16:51:23 CET] {subprocess.py:73} INFO - Running command: ['/opt/***/dbt_venv/bin/dbt', 'run', '--models', 'mart_facts_samples', '--profiles-dir', '/tmp/tmpnhr57up0', '--profile', 'default', '--target', 'prod']
```

I believe that providing the `target-dir` with the `manifest.json` in the command execution would probably allow skipping the re-parsing of the project for each **task**.

Related to this:
- https://github.com/astronomer/astronomer-cosmos/issues/520: this is at DAG parsing time, but not at task execution time
- https://github.com/astronomer/astronomer-cosmos/issues/785: is probably a more generic API for this -- i think for DBT_MANIFEST mode, this could be solved in an easier way by just using the existing manifest?

I am fairly new to cosmos, so I would be grateful for some pointers in the right direction.

dwreeves:
## Description

dbt uses `partial_parse.msgpack` to make rendering things a lot faster. This PR adds support for `partial_parse.msgpack` in the following places:

- `ExecutionMode.LOCAL`
- `ExecutionMode.VIRTUALENV`
- `LoadMode.DBT_LS`

This PR also allows users to explicitly _turn off_ partial parsing. If this is done, then `--no-partial-parse` will be passed as an explicit flag in all dbt command invocations (i.e. all `ExecutionMode`s and `LoadMode.DBT_LS`, albeit not the `dbt deps` invocation.)

This should address some performance complaints that users have, e.g. this message from Slack: https://apache-airflow.slack.com/archives/C059CC42E9W/p1704483361206829 Albeit, this user will also need to provide a `partial_parse.msgpack`.

My experimentation and looking at dbt-core source code confirms that dbt does not use `manifest.json` when partial parsing. It appears that these are just output artifacts, but not input artifacts. Only `partial_parse.msgpack` is used. (There are a couple ways to confirm this other than just checking source code

Also, I added a minor refactor of a feature I added a year ago: I added `send_sigint()` to the custom subprocess hook, since this custom subprocess hook did not exist back when I added it (if you want me to split this refactor into a different PR then let me know).

### API change

I decided the best way to go about this would be to just add a `partial_parse: bool` to both the execution config and render config. For example:

```python
dbt_group = DbtTaskGroup(
    ...,
        ...,
        partial_parse=True
    ),
    render_config=RenderConfig(
        ...,
        partial_parse=False
    )
)
```

That said, in all honesty users will not need to set this at all, except unless they want to suppress the little warning message about not being able to find the `partial_parse.msgpack`. This is because by default this addition searches for a msgpack if it exists, which is already the existing behavior in a sense, except right now the msgpack file never exists (dbt does look for it though).

When inserting into the `AbstractDbtBaseOperator`, I did not use `global_boolean_flags`. See the subsection below for why.

### Other execution performance improvements

The main reason I am adding this feature is that it should dramatically improve performance for users. However, it is not the only way to improve 

It's possible that we should add a way to add the flag `--no-write-json` as an explicit kwarg to the dbt base operator. Right now users can support this via `dbt_cmd_global_flags=["--no-write-json"]`. Some users (e.g. those using Elementary, or those using the dbt local operator `callback` kwarg) will want to write the JSON, but I suspect the majority of users will not. Similarly, `--log-level-file` is not used at all, and at minimum dbt should work best the vast majority of time with `--no-cache-selected-only` set.

It's also possible there should be some sort of "performant" mode that automatically sets all these defaults for optimal performance:

- `--no-write-json`
- `--log-level-file=none`
- `--no-cache-selected-only`

Perhaps a "performant" config would be too cumbersome to implement (I would agree with that). In which case the docs could also have a section on performance tips.

### A note on `global_boolean_flags`

I did not add the partial parse support to `global_boolean_flags` because it doesn't quite fit into the existing paradigm for this. Right now the default for each of these `global_boolean_flags` is False, whereas the default for partial parse is actually True. This makes fitting it in awkward.

I think it's possible that just having a `tuple[str]` is insufficient here, as some flags you may want to add (not just `--no-partial-parse` but also `--no-write-json` are by default _True_ and must be explicitly turned off. Meaning that the parsing Cosmos does with flags of `'--{flag.replace("_", "-")}'` is ineffective for flags like this.

Right now, we have an example of putting _no_ in front with `no_version_check`. Meaning that the default behavior of version checking is True, but the flag itself starts as negated so the default is actually `False`.

My proposal is, instead of `global_boolean_flags: tuple[str]`, this should instead be `global_boolean_flags: tuple[str | tuple[str, str | None, str | None]]`. In the case that a global flag is a `tuple[str, str | None, str | None]`, then the first arg should be the flag, the second should be "if true," and the third should be "if false." `None` indicates, when true/false (respectively), then do nothing.

For example:

```python
class AbstractDbtBaseOperator(BaseOperator, metaclass=ABCMeta):
    ...
    global_boolean_flags = (
        ("no_version_check", "--no-version-check", None),
        ("cache_selected_only", "-cache-selected-only", None),
        ("partial_parse", None, "--no-partial-parse"),
    )

```

And Cosmos want to support `str` parsing for backwards compatibility. It's pretty straightforward to convert the data type:

```python
if isinstance(flag, str):
    flag = (flag, '--{flag.replace("_", "-")}', None)
```

## Related Issue(s)

- Resolves #791
- Partially resolves #785
  - #785 should probably be split up into two different stages: (1) support for partial parsing (2) (a) dbt project dir / manifest / `partial_parse.msgpack` is allowed to come from cloud storage. (b) `dbt compile` is able to dump into cloud storage.

## Breaking Change?

Should not break anything. This doesn't do anything when `partial_parse.msgpack` is missing, and the default behavior (`partial_parse=True`) does not alter the dbt cmd flags.

## Checklist


dwreeves:
I don't know how the coverage thing works and if it is checking for 100% coverage of new commits vs just making sure the coverage goes up, but I vastly increased the coverage of the code by adding tests for the subprocess hook.

jlaneve (commented):
This is awesome for two reasons: the change in and of itself, and the writeup on where we should look in the future. Apologies it's taken longer to review, but I'm going to check this out today.

I agree the performant mode is nice in theory but may be difficult to maintain. Maybe we should create a docs page specifically on performance with our recommendations? Inclusive of this flag, using the manifest parsing method, install [LibYAML](https://docs.getdbt.com/reference/parsing#pyyaml--libyaml), etc. That way, for users who care about performance, there's a single doc for all optimization recommendations.

Quick question re: how this partial parsing plays out in practice. My assumption is that a user would:
- run `dbt parse` locally / in CICD (pre-deploying), which generates the necessary artifacts
- push the entire project to the Airflow deployment
- set the partial_parse flag

Does that sound right?

dwreeves:
> 1. run `dbt parse` locally / in CICD (pre-deploying), which generates the necessary artifacts

Yes or `dbt compile`. Both of these commands will produce a `manifest.json` and `partial_parse.msgpack`. (A note: This code addition only uses `partial_parse.msgpack`; however a user making use of pre-compilation for performance reasons would also probably want to use `manifest.json` as a load method similarly for performance reasons. So two birds with one stone.)

> 2. push the entire project to the Airflow deployment
> 3. set the partial_parse flag

By default, `partial_parse` is `True`. They wouldn't need to explicitly set it unless they wanted to.

I made this API decision in part because the user already receives the following warning from dbt when they are not using partial parsing:

```
Unable to do partial parsing because saved manifest not found. Starting full parse.
```

Meaning that a user who intended on using partial parsing, and one who did not, is already being warned by dbt that they are not; and by default dbt is looking for it. So I think having the default behavior of the operator match the default behavior of dbt is sensible. (Also, the performance implications of checking for the existence of the file, and failing to find it, are negligible, so very low harm in doing so.)

---

> I agree the performant mode is nice in theory but may be difficult to maintain. Maybe we should create a docs page specifically on performance with our recommendations? Inclusive of this flag, using the manifest parsing method, install [LibYAML](https://docs.getdbt.com/reference/parsing#pyyaml--libyaml), etc. That way, for users who care about performance, there's a single doc for all optimization recommendations.

I do like this approach. I agree that this would be better than a "performant" mode.

dwreeves:
Also, I just got my act together and got to full test coverage for this PR. 😅

jlaneve:
holding this until https://github.com/astronomer/astronomer-cosmos/pull/827 is complete so we can measure the before/after!

ms32035:
My question would be why `target` is excluded here: https://github.com/astronomer/astronomer-cosmos/blob/72690d31ba6822ae6df56f4d56f26196234c80b5/cosmos/dbt/project.py#L15C39-L15C58
We pre-compile our projects on a shared volume, so and option to copy the whole `target` would probably solve all our issues

dwreeves:
@ms32035 The history of Cosmos runtime is:

1. Initially they just ran out of the dbt project dir directly.
2. Then users quickly noted this is a non-starter because Airflow runtimes will often have the dags folder mounted as a read-only volume, so they switched to creating a temporary directory and copying the contents (see #233).
3. But this causes issues because of the Celery executor: shared workers on a small compute instance can clog the tmpdir and, if virtual storage runs out, this will cause a very very nasty SIGKILL deadlock (see #362 for description of the problem). So they (I believe very briefly, and never fully released?) switched to setting the `--target-dir` as a (blank) temporary directory (see #414).
4. But this only works for more recent versions of dbt; older versions don't let you specify a target dir via CLI or env vars. So they stopped doing that and instead started using symlinks (see #488).
5. Now, users are complaining that parsing is too slow and partial parsing is not being used (see #722, #785, #791; all three of these are the same issue). Hence this PR.

So why not just remove that line which excludes the target dir? Because that would cause a symlink to be created to the target dir, which would mean you're writing to the target dir, which is a non-starter for users with read-only mounts to their dags dir. OK, why use symlinks, why not just copy instead? Because that causes those nasty virtual storage issues for large dbt projects running on small Celery workers[*]. Why not symlink the read only files and copy the target dir? Because that's unnecessary; only `partial_parse.msgpack` is read by the dbt runtime pre-execution.[**]

Speaking as someone who has dealt with these problems pre-Cosmos and figured out a lot of the same lessons that Cosmos did, I think the way Cosmos works right now is very close to optimal, but they also have the added handicap of needing to support tons of different users and all their wacky idiosyncratic needs. The main suboptimality I see is the lack of partial parsing, which is why I wrote this PR. But the part where you don't symlink the target dir is absolutely correct, if you ask me. I don't see anything else I'd change with how Cosmos does parsing during task execution.

[*] Fun fact: there was a point when the popular `dbt-expectations` library [shipped with a 3.03 MB `.gif` file](https://github.com/calogica/dbt-expectations/blob/364c3fa7885c653c5b5b68833c822140b27720fc/expectations.gif) after running `dbt deps`, kid you not. So imagine a Celery worker with max concurrency of 5 tasks, that's 15 MB devoted to a single `.gif` file. That may not sound like _that_ much but this did cause a problem for me once, insofar that updating the package version to one without the `.gif` actually fixed my problems. The SIGKILL deadlock thing is not theoretically, it literally happened to me!

[**] In very niche situations, some extensions may use non-msgpack dbt artifacts. I believe dbt elementary is the most popular one. However, I believe it also uses these artifacts in post-hooks which would be generated via the command execution inside the temporary directory.

ms32035:
Thanks for detailed explanation @dwreeves . Never suggested that would copy would be a complete solution for every usecase, and can't wait to see this merged

dwreeves:
Yeah I get it, I'm just giving a very thorough history of how we ended up here and what all the alternatives are. 😁

cosmos/config.py
```diff
@@ -47,6 +47,7 @@ class RenderConfig:
     :param env_vars: (Deprecated since Cosmos 1.3 use ProjectConfig.env_vars) A dictionary of environment variables for rendering. Only supported when using ``LoadMode.DBT_LS``.
     :param dbt_project_path: Configures the DBT project location accessible on the airflow controller for DAG rendering. Mutually Exclusive with ProjectConfig.dbt_project_path. Required when using ``load_method=LoadMode.DBT_LS`` or ``load_method=LoadMode.CUSTOM``.
     :param dbt_ls_path: Configures the location of an output of ``dbt ls``. Required when using ``load_method=LoadMode.DBT_LS_FILE``.
+    :param partial_parse: If True, then when ``load_method=LoadMode.DBT_LS``, attempt to use the ``partial_parse.msgpack`` file if it exists.
```
jlaneve, line 50:
```suggestion
    :param partial_parse: If True (default), then when ``load_method=LoadMode.DBT_LS``, attempt to use the ``partial_parse.msgpack`` file if it exists.
```

cosmos/config.py
```diff
@@ -301,6 +306,7 @@ class ExecutionConfig:
 
     dbt_project_path: InitVar[str | Path | None] = None
     project_path: Path | None = field(init=False)
+    partial_parse: bool = True
```
jlaneve, line 309: do we need this in ExecutionConfig?

cosmos/dbt/project.py
```diff
@@ -21,6 +19,16 @@ def create_symlinks(project_path: Path, tmp_dir: Path, ignore_dbt_packages: bool
...
 
 
+def copy_manifest_for_partial_parse(project_path: Path, tmp_dir: Path) -> None:
+    partial_parse_file = Path(project_path) / DBT_TARGET_DIR_NAME / DBT_PARTIAL_PARSE_FILE_NAME
+
+    if partial_parse_file.exists():
+        tmp_target_dir = tmp_dir / DBT_TARGET_DIR_NAME
+        tmp_target_dir.mkdir(exist_ok=True)
+
+        shutil.copy(str(partial_parse_file), str(tmp_target_dir / DBT_PARTIAL_PARSE_FILE_NAME))
```
jlaneve, line 29: do we want to copy or symlink here?

cosmos/operators/base.py
```diff
@@ -51,6 +51,9 @@ class AbstractDbtBaseOperator(BaseOperator, metaclass=ABCMeta):
     :param skip_exit_code: If task exits with this exit code, leave the task
         in ``skipped`` state (default: 99). If set to ``None``, any non-zero
         exit code will be treated as a failure.
+    :param partial_parse: If True, then the operator will use the
```
jlaneve, line 54:
```suggestion
    :param partial_parse: If True (default), then the operator will use the
```

cosmos/dbt/project.py
```diff
@@ -21,6 +19,16 @@ def create_symlinks(project_path: Path, tmp_dir: Path, ignore_dbt_packages: bool
             os.symlink(project_path / child_name, tmp_dir / child_name)
 
 
+def copy_manifest_for_partial_parse(project_path: Path, tmp_dir: Path) -> None:
```
jlaneve, line 22: I don't want to be too nitpick-y but can we call this something like this instead given we're not copying the manifest?

```suggestion
def copy_msgpack_for_partial_parse(project_path: Path, tmp_dir: Path) -> None:
```

jlaneve (commented):
overall looks great - appreciate the thorough description and test coverage! I had a few minor comments but once we work through those, I'll approve and we can merge/release 🥂

cosmos/dbt/project.py
dwreeves: I'm like 98% sure you want to copy, not symlink. The reason is because the `partial_parse.msgpack` can be mutated by a dbt command invocation, and the symlink is only compatible with users' runtimes for read-only files. Because the `partial_parse.msgpack` can be overwritten, this means we cannot symlink to it.

cosmos/dbt/project.py
dwreeves: Sure. I think I used manifest because while researching it I found that under the hood that's how it's referred to; the partial parse is just a serialized manifest. But I do see how that can be confusing! I will update when I get a moment.

cosmos/config.py
dwreeves: I think so, but I don't know what you have in mind as an alternative. It's possible there is a different way to set this up (I'm obviously not as familiar with the high level design the Astronomer folks have in mind for Cosmos as you are).

My reasoning for including this here: both the execution config and the render config both potentially use partial parsing. The `RenderConfig` uses it when `LoadMode.DBT_LS` is set; the `ExecutionConfig` uses it when `ExecutionMode.LOCAL` or `ExecutionMode.VIRTUALENV` is set. So both of these may use partial parsing separately.

jlaneve: ah yeah I see the issue now. I don't believe we've run into too many of these cases, but the nuance that I'm thinking through here is that these are really the same config (meaning you wouldn't set these values differently across the `RenderConfig` and `ExecutionConfig`), right?

I can think of two alternatives:
- consolidate the config into a single field in the `ProjectConfig`, where we already have things like env vars that are used in both rendering and execution
- add it as a top-level field to the `DbtDag` / `DbtTaskGroup` classes - we already have `on_warning_callback` here 


Thoughts? I think this could make sense in the `ProjectConfig` but would also be curious to get @jbandoro or @tatiana thoughts

dwreeves:
I made updates as per everyone's suggestions. Thanks.

> The docs will have to make it clear that it's only used for rendering when dbt_ls is used and execution for local or virtualenv.

A note on the documentation. I am not documenting the behavior for `LoadMode.DBT_LS`. I think, in most cases, it will be rare for a user to both be using `DBT_LS` and have a `partial_parse.msgpack`. This would not be completely unheard of and can be a reasonable deliberate decision if the dbt project mutates a lot, but the `partial_parse.msgpack` is not being updated frequently, so you either set the load mode deliberately or just forego the manifest.json but keep the partial parse file. But all told this is a pretty niche situation-- I think a lot of users on `DBT_LS` mode are likely foregoing the use of dbt artifacts altogether. And I wasn't sure where it would fit in without feeling like it was adding clutter and potential confusion to the docs for most users. Let me know if you want to document this regardless.

cosmos/config.py
```diff
@@ -47,6 +47,7 @@ class RenderConfig:
     :param env_vars: (Deprecated since Cosmos 1.3 use ProjectConfig.env_vars) A dictionary of environment variables for rendering. Only supported when using ``LoadMode.DBT_LS``.
     :param dbt_project_path: Configures the DBT project location accessible on the airflow controller for DAG rendering. Mutually Exclusive with ProjectConfig.dbt_project_path. Required when using ``load_method=LoadMode.DBT_LS`` or ``load_method=LoadMode.CUSTOM``.
     :param dbt_ls_path: Configures the location of an output of ``dbt ls``. Required when using ``load_method=LoadMode.DBT_LS_FILE``.
+    :param partial_parse: If True (default), then when ``load_method=LoadMode.DBT_LS``, attempt to use the ``partial_parse.msgpack`` file if it exists.
```
jbandoro, line 50: This should be removed from the RenderConfig docstring with the update.

jbandoro (commented):
Thanks, this is great @dwreeves!

Could you also add the `partial_parse` arg to the project config docs here: https://github.com/astronomer/astronomer-cosmos/blob/main/docs/configuration/project-config.rst 

After that and the minor comment below, it looks good to me to merge.

dwreeves:
@jbandoro Done

PR #808, opened 2024-01-23
feat: add pip_install_options as an operator arg

octiva:
## Description

- In order to provide pip with install options when using the virtual env
- When using a private server, packages need to be provided locally through plugins, adding the install options will allow us to specify where to find the links. eg `--no-index --find-links=${AIRFLOW_HOME}/plugins`

## Checklist


tatiana:
@octiva, this is a great feature; thanks for working on it! I'm adding to the milestone 1.4. We'll review once the PR is out of draft :)

spencer-horton-wcq:
@tatiana Hey, would you be able to run run that failed test? I believe it is not caused by my changes. as it references failed api calls to: `/sql/1.0/warehouses/***`.

Maybe review too? I don't believe any tests are needed, as it is tested to the same level as: `py_requirements` in the same function.

tatiana (approved):
This looks great, thank you very much, @octiva ! We'll release it as part of 1.4

PR #809, opened 2024-01-23
Add more logs to troubleshoot custom selector

tatiana:
A user (ZD case: 38982) reported being unable to see the test node for one specific model in a particular DAG when using RenderConfig(select=tags). They were using the custom Cosmos selector. It seems the selector worked as expected for other models and for the same model/test in a different DAG.

This CR adds more logs for troubleshooting when using.


jbandoro (approved):
Thanks, helpful to have more of these logs for troubleshooting!

Issue #796, opened 2024-01-12
No datasets emitted in DbtLocalBaseOperator after Cosmos upgrade to 1.3.0


jakob-hvitnov-telia:
After updating Cosmos from 1.1.3 to 1.3.0 we have run into an issue where we are no longer seeing emission of datasets when using theDbtRunLocal operator.
At the same update we went from Airflow 2.6.3 to 2.7.2.

Dag log output (from dbt_run task) before the update:
```
[2024-01-04 12:49:06] {{local.py:238}} INFO - Inlets: [Dataset(uri='correct_uri_to_inlet01_anonymized', extra=None), Dataset(uri='correct_uri_to_inlet02_anonymized', extra=None), Dataset(uri='correct_uri_to_inlet03_anonymized', extra=None), Dataset(correct_uri_to_inlet04_anonymized', extra=None)]
[2024-01-04, 23:10:00] CET {{local.py:239}} INFO - Outlets: [Dataset(uri='correct_uri_to_outlet01_anonymized', extra=None), Dataset(uri='correct_uri_to_outlet01_anonymized', extra=None), Dataset(uri='correct_uri_to_outlet01_anonymized', extra=None), Dataset(uri='correct_uri_to_outlet01_anonymized', extra=None)]
```
Dag log output after the update:
```
[2024-01-05 10:26:17] {{local.py:238}} INFO - Inlets: []
[2024-01-05 10:26:17] {{local.py:239}} INFO - Outlets: []
```

We are running on AWS MWAA. 

I suspect the change in has something to do with this [commit](https://github.com/astronomer/astronomer-cosmos/commit/033a4b33ffa8f5801c5401399a9c58ed787dc56e#diff-cdc7f3cedeca298d5c3bdd8d9be8fa15791b915fbe1b2810e9c86c271eb86446) as this introduces the get_datasets method, which in its docstring clarifies that the method relies on:

- profiles
- {project_dir}/target/manifest.json
- {project_dir}/target/run_results.json

It is not quite clear to me if these are generated by our call to DbtRunLocal at the moment.
There might be configuration options I am overlooking:

```
dbt_run = DbtRunLocalOperator(
        profile_config=profile_config,
        project_dir= os.path.join(PurePath(__file__).parent.parent.parent.parent, "dbt"),
        task_id="dbt_run",
        select="path:path/to/our/dbt/models",
        install_deps=False,
        emit_datasets=True
        )
```

## Docstring typo
[...]

PR #810, opened 2024-01-23
Fix OpenLineage integration documentation

tatiana:
[Cosmos docs](https://astronomer.github.io/astronomer-cosmos/configuration/lineage.html) stated that users didn't have to install any dependency to use OpenLineage with Cosmos.

However, for inlets and outlets to be emitted, Airflow 2.7 users must install `apache-airflow-providers-openlineage` or `astronomer-cosmos[openlineage]`.

Closes: #796

### DIFF
--- a/docs/configuration/lineage.rst
+++ b/docs/configuration/lineage.rst
@@ -9,7 +9,7 @@ and virtualenv execution methods (read `execution modes <../getting_started/exec
 
 To emit lineage events, Cosmos can use one of the following:
 
-1. Airflow 2.7 `built-in support to OpenLineage <https://airflow.apache.org/docs/apache-airflow-providers-openlineage/1.0.2/guides/user.html>`_, or
+1. Airflow official OpenLineage provider <https://airflow.apache.org/docs/apache-airflow-providers-openlineage/1.0.2/guides/user.html>`_, or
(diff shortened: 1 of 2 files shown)


docs/configuration/lineage.rst
```diff
@@ -9,7 +9,7 @@ and virtualenv execution methods (read `execution modes <../getting_started/exec
 
 To emit lineage events, Cosmos can use one of the following:
 
-1. Airflow 2.7 `built-in support to OpenLineage <https://airflow.apache.org/docs/apache-airflow-providers-openlineage/1.0.2/guides/user.html>`_, or
+1. Airflow official OpenLineage provider <https://airflow.apache.org/docs/apache-airflow-providers-openlineage/1.0.2/guides/user.html>`_, or
```
jbandoro, line 12: I think there is a backtick missing somewhere here for the external link since it looks like this:

 
image

jbandoro (approved):
Thanks! Good to merge after the formatting comment below is addressed

docs/configuration/lineage.rst
tatiana: Thanks, @jbandoro , great catch - just fixed!

PR #812, opened 2024-01-23
Use Airflow constraint file for test environment setup

### LINKED ISSUE #811: Improve Cosmos test environment Airflow installation
Opening up this issue after #806 was merged where more pinning specific to Airflow version dependencies had to be done in overrides and a workaround to uninstall a provider that is not available in earlier Airflow versions.

A possible better way to install Airflow is to use the [recommended constraint file](https://airflow.apache.org/docs/apache-airflow/stable/installation/installing-from-pypi.html), this way we wouldn't have to keep track of pinning different Airflow dependencies for different versions. Installing with a constraint file can be done like the below where it depends on a specific airflow version and python version:

```shell
pip install "apache-airflow==2.8.1" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-2.8.1/constraints-3.8.txt"
```

After the work @tatiana did in #806 by adding a hatch `post-install-commands`, I think we could use the [pre-install-commands](https://hatch.pypa.io/1.9/config/environment/overview/#pre-install) to install airflow with a constraint file. The pyproject.toml would like:
```toml
######################################
# TESTING
######################################

[tool.hatch.envs.tests]
dependencies = [
    "astronomer-cosmos[tests]",
    "types-PyYAML",
    "types-attrs",
    "types-requests",
    "types-python-dateutil",
    "Werkzeug<3.0.0",
    "apache-airflow-providers-docker>=3.5.0",
]
pre-install-commands = [
    "pip install 'apache-airflow=={matrix:airflow}' --constraint 'https://raw.githubusercontent.com/apache/airflow/constraints-{matrix:airflow}/constraints-{matrix:python}.txt'"
]

[[tool.hatch.envs.tests.matrix]]
python = ["3.8", "3.9", "3.10"]
airflow = ["2.3.4", "2.4.3", "2.5.3", "2.6.3", "2.7.3", "2.8.1"]


[tool.hatch.envs.tests.scripts]
...
```
The update above would remove the need to have the `[tool.hatch.envs.tests.overrides]` section. However the matrix does change a bit where the Airflow patch version is required.

The GH workflow would also need to be updated so the matrix matches. 

An advantage of this is that we would not need to worry about newer releases of Airflow breaking the CI tests, but we would need to increment the patch version above when a latest version of Airflow is released.

jbandoro:
## Description

This PR installs apache-airflow in the hatch `pre-install-commands` section so Airflow related dependency conflicts do not need to be managed in the override section that has been removed.

Installing from the Airflow constraint file for versions <=2.6 fails because PyYAML is pinned in the constraint file for those versions to 6.0.0 and [a workaround ](https://github.com/yaml/pyyaml/issues/736) is required becaused older versions of PyYAML can no longer be installed from unmodified source or sdist with the release of Cython3.

## Related Issue(s)

closes #811 

## Breaking Change?

None

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
           architecture: 'x64'
 
       - run: pip3 install hatch
-      - run: hatch run tests.py3.9-2.7:type-check
+      - run: hatch run tests.py3.9-2.7.3:type-check
 
   Run-Unit-Tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,14 +34,14 @@ jobs:
           architecture: 'x64'
 
       - run: pip3 install hatch
-      - run: hatch run tests.py3.9-2.7.3:type-check
+      - run: hatch run tests.py3.9-2.7:type-check
 
   Run-Unit-Tests:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ["3.8", "3.9", "3.10"]
-        airflow-version: ["2.3.4", "2.4.3", "2.5.3", "2.6.3", "2.7.3", "2.8.1"]
+        airflow-version: ["2.3", "2.4", "2.5", "2.6", "2.7", "2.8"]
     steps:
       - uses: actions/checkout@v3
         with:
@@ -80,7 +80,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.8", "3.9", "3.10"]
-        airflow-version: ["2.3.4", "2.4.3", "2.5.3", "2.6.3", "2.7.3", "2.8.1"]
+        airflow-version: ["2.3", "2.4", "2.5", "2.6", "2.7", "2.8"]
     services:
       postgres:
         image: postgres
@@ -148,7 +148,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.10"]
-        airflow-version: ["2.6.3"]
+        airflow-version: ["2.6"]
 
     services:
       postgres:
@@ -228,7 +228,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.10"]
-        airflow-version: ["2.7.3"]
+        airflow-version: ["2.7"]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


jbandoro:
@tatiana do you know how to get the Github actions to run with the updated matrix made to the `.github/workflows/test.yml` in this PR - it seems to be running the workflow file from the `main` branch.

.github/workflows/test.yml
```diff
@@ -34,14 +34,14 @@ jobs:
...
-      - run: hatch run tests.py3.9-2.7:type-check
+      - run: hatch run tests.py3.9-2.7.3:type-check
 
   Run-Unit-Tests:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ["3.8", "3.9", "3.10"]
-        airflow-version: ["2.3", "2.4", "2.5", "2.6", "2.7"]
+        airflow-version: ["2.3.4", "2.4.3", "2.5.3", "2.6.3", "2.7.3", "2.8.1"]
```
tatiana, line 44: If we pin micro versions, we need to be aware of every micro release from Airflow and update Cosmos tests. What do you think of sticking to minor releases?

pyproject.toml
```diff
@@ -155,30 +155,21 @@ dependencies = [
     "apache-airflow-providers-cncf-kubernetes>=5.1.1",
     "apache-airflow-providers-docker>=3.5.0",
 ]
-post-install-commands = [
-  "pip uninstall -y apache-airflow-providers-common-io"
+# Airflow install with constraint file, Airflow versions < 2.7 require a workaround for PyYAML
+pre-install-commands = ["""
+    AIRFLOW_MINOR_VERSION=$(echo "{matrix:airflow}" | cut -c-3)
+    AIRFLOW_LESS_THAN_2_7=$(echo "$AIRFLOW_MINOR_VERSION < 2.7" | bc)
```
tatiana, line 161: This method to compare versions may not be accurate:
```
$ echo "2.1 < 2.7" | bc    
1
$  echo "2.10 < 2.7" | bc
1
```

.github/workflows/test.yml
jbandoro: Good point, I reverted this and it also solves the issue of using the branch workflow file since it's the same as main.

pyproject.toml
jbandoro: Good catch! Updated the check of the minor version.

tatiana (approved):
Much more maintainable, thanks a lot, @jbandoro !

Issue #813, opened 2024-01-24
Fix custom selector when `test` node has no `depends_on values`


tatiana:
A Cosmos user (Astronomer customer) reported the following issue while using Cosmos 1.3.1.

```
    def _should_include_node(self, node_id: str, node: DbtNode) -> bool:
        "Checks if a single node should be included. Only runs once per node with caching."
        if node_id in self.visited_nodes:
            return node_id in self.selected_nodes
    
        self.visited_nodes.add(node_id)
    
        if node.resource_type == DbtResourceType.TEST:
>           node.tags = getattr(self.nodes.get(node.depends_on[0]), "tags", [])
E           IndexError: list index out of range

cosmos/dbt/selector.py:298: IndexError
```

Based on the error, it seems their dbt project has a test without `depends_on`. This CR adds support to this use-case.

PR #814, opened 2024-01-24
Fix custom selector when `test` node has no `depends_on` values

tatiana:
A user reported the following issue while using Cosmos 1.3.1:

```
    def _should_include_node(self, node_id: str, node: DbtNode) -> bool:
        "Checks if a single node should be included. Only runs once per node with caching."
        if node_id in self.visited_nodes:
            return node_id in self.selected_nodes

        self.visited_nodes.add(node_id)

        if node.resource_type == DbtResourceType.TEST:
>           node.tags = getattr(self.nodes.get(node.depends_on[0]), "tags", [])
E           IndexError: list index out of range

cosmos/dbt/selector.py:298: IndexError
```

In order to reproduce this issue, it was necessary to add a tag-based select statement.

Based on the error, it seems their dbt project has a test without `depends_on`. This CR adds support to this use case.

Closes: #813

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -294,7 +294,7 @@ class NodeSelector:
 
         self.visited_nodes.add(node_id)
 
-        if node.resource_type == DbtResourceType.TEST:
+        if node.resource_type == DbtResourceType.TEST and node.depends_on:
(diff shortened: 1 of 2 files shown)


jbandoro (approved):
Thanks for the fix!

Issue #643, opened 2023-11-02
After all test behaviour with select and exclude


ndrluis:
Hello, the AFTER_ALL tests behavior is not considering the select and exclude settings defined in the rendering configuration.

Test Behavior: After All
Load Method: DBT ls
Execuction mode: Kubernetes
Test indirect selection: Buildable

<img width="1267" alt="image" src="https://github.com/astronomer/astronomer-cosmos/assets/1859449/4b22cef8-81e3-4430-a272-ae96785ad0c0">

tatiana:
Thanks for reporting this issue, @ndrluis! Would you like to help us fix this?

ndrluis:
Hello @tatiana, I am interested, but I need help understanding where I should make this correction.

I am studying the code, and I'm not sure if I got it right. But I believe I would need to pass the select and exclude values to the `build_airflow_graph` function and then to `create_test_task_metadata` when `TestBehavior.AFTER_ALL` is used. 

Within `create_test_task_metadata`, I would check if the load mode is `DBT_LS` and pass the values to the select and exclude arguments.

Is it something along these lines?

tatiana:
> Hello @tatiana, I am interested, but I need help understanding where I should make this correction.
> 
> I am studying the code, and I'm not sure if I got it right. But I believe I would need to pass the select and exclude values to the `build_airflow_graph` function and then to `create_test_task_metadata` when `TestBehavior.AFTER_ALL` is used.

Yes, I believe we could pass `render_config` as an argument of `build_airflow_graph` and expose select/exclude in `create_test_task_metadata`, as you suggested.

> Within `create_test_task_metadata`, I would check if the load mode is `DBT_LS` and pass the values to the select and exclude arguments.

I don't think we'd need to check if it is `DBT_LS`. Why do you think this would be necessary? This issue happens for other `LoadMode`s as well, and it makes sense for the test task craeted to only run tests for the subset of tasks filtered using `RenderConfig`.

binhnq94:
I have the same issue.

PR #816, opened 2024-01-25
Fix forwarding selectors to test task when using `TestBehavior.AFTER_ALL`

tatiana:
Before, when using `TestBehavior.AFTER_ALL`, it did not consider the `select` and `exclude` settings defined in the rendering configuration. This was an error since it would run all the dbt project tests, even if they were outside of the scope of the Cosmos-defined DAG.

Now we take into account `select`, `selector` and `exclude` when using this test behaviour.

Closes: #643


jlaneve (approved):
Nice! Great catch and great fix

PR #819, opened 2024-01-27
Update PyPI project description to better describe Cosmos

jbandoro:
## Description

I noticed that the project description in PyPi could use an update:

image

 I think the original intention of Cosmos was to render and run more than only dbt DAGs but has since changed to be specific to dbt.

Would welcome any other suggestions from the team/community!


jlaneve (approved):
Ah great catch!

PR #825, opened 2024-01-31
Fix error when running unit test `test_created_pod` with Py3.11

tatiana:
When the CI ran:
`hatch run tests.py3.11-2.7:test-cov`

It was consistently raising the exception:
```
FAILED tests/operators/test_kubernetes.py::test_created_pod - kubernetes.config.config_exception.ConfigException: Invalid kube-config file. No configuration found.
```

Example:
https://github.com/astronomer/astronomer-cosmos/actions/runs/7726274202/job/21063656952

I could not reproduce this issue locally, even when using Python 3.11.8. An explanation I see is that the built-in `unittest.mock` library changed in the Python version the CI was using - and for some reason, the way the mock was set stopped working. I changed how the mock was defined, and things seem to work fine.

### DIFF
--- a/tests/operators/test_kubernetes.py
+++ b/tests/operators/test_kubernetes.py
@@ -226,10 +226,11 @@ def test_dbt_test_kubernetes_operator_handle_warnings_and_cleanup_pod():
     test_operator._handle_warnings(context)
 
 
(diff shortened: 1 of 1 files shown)


jbandoro (approved):
Thanks for debugging and fixing the failing test!

PR #827, opened 2024-01-31
Add performance integration tests

jlaneve:
## Description

This PR adds a step to our CI to measure how quickly Cosmos can run models. This is part of a larger initiative to make the project more performant now that it's reached a certain level of maturity.

How it works:
- We now have [a test that generates a dbt project with a certain number of sequential models](https://github.com/astronomer/astronomer-cosmos/blob/performance-int-tests/tests/perf/test_performance.py) (based on a parameter that gets passed in), runs a simple DAG, and measures task throughput (measured in terms of models run per second
- I've extended our CI to run this test for 1, 10, 50, and 100 models to start
- This CI reports out a GitHub Actions output that gets shown in the actions summary, [at the bottom](https://github.com/astronomer/astronomer-cosmos/actions/runs/7894490582)

While this isn't perfect, it's a step in the right direction - we now have some general visibility! Note that these numbers may not be indicative of a production Airflow environment running something like the Kubernetes Executor, because this runs a local executor on GH Actions runners. Still, it's meant as a benchmark to demonstrate whether we're moving in the right direction or not.

As part of this, I've also refactored our tests to call a script from the pyproject file instead of embedding the scripts directly in the file. This should make it easier to maintain and track changes.

## Related Issue(s)

https://github.com/astronomer/astronomer-cosmos/pull/800

## Checklist


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, performance-int-tests] # TODO: remove me before merging
```
jlaneve, line 5: leaving this a comment here so I don't forget!

jlaneve: will delete this once the PR is approved and ready to be merged

pyproject.toml
```diff
@@ -175,51 +140,20 @@ matrix.airflow.dependencies = [
...
--k 'example_cosmos_sources or sqlite'"""
+test = 'sh scripts/test/unit.sh'
+test-cov = 'sh scripts/test/unit-cov.sh'
+test-integration-setup = 'sh scripts/test/integration-setup.sh'
+test-integration = 'sh scripts/test/integration.sh'
+test-integration-expensive = 'sh scripts/test/integration-expensive.sh'
+test-integration-sqlite-setup = 'sh scripts/test/integration-sqlite-setup.sh'
+test-integration-sqlite = 'sh scripts/test/integration-sqlite.sh'
+test-performance-setup = 'sh scripts/test/performance-setup.sh'
+test-performance = 'sh scripts/test/performance.sh'
```
jbandoro, line 151: Thanks for moving all of these to scripts, a lot easier to maintain!

tests/test_example_dags.py
```diff
@@ -24,6 +24,7 @@
 
 MIN_VER_DAG_FILE: dict[str, list[str]] = {
     "2.4": ["cosmos_seed_dag.py"],
+    "3.0": ["performance_dag.py"],
```
jbandoro, line 27: Just so it's clear that the peformance dag doesn't require Airflow v3.0, what do you think about instead having another constant like `IGNORED_DAG_FILES = ["performance_dag.py"]` and writing those to the ignore file?

jbandoro (approved):
🚀 this is great! Excited to test out the performance dag in #836 see how much a difference it makes.

PR #835, opened 2024-02-05
Fix: `operator_args` modified in place in Airflow converter

### LINKED ISSUE #833: Bug: `operator_args` are modified in place in the Airflow converter
A user reported this in the Slack thread [here](https://apache-airflow.slack.com/archives/C059CC42E9W/p1707150619491869). Even though we deprecated `operator_args` `vars` and `env` in favor of `ProjectConfig.dbt_vars` and `ProjectConfig.env_vars`, there is now a problem where if a user supplies `operator_args` to multiple `DbtDag` or `DbtTaskGroup`'s with vars on env, they will only be used in the first dag/task group object.

Example problem (copied from the user report)
```python
OSMOS_COMMON_RENDER_CONF = dict(
    emit_datasets=False, test_behavior=TestBehavior.AFTER_EACH, load_method=LoadMode.CUSTOM, dbt_deps=False
)


with DAG(
        dag_id="my-amazing-dag",
        default_args=default_args,
        schedule="50 4 * * *",
) as dag:
    dbt_task_group_base_config = {
        "project_config": MY_WONDERFUL_PROJECT_COSMOS_PROJ_CONF,
        "profile_config": MY_WONDERFUL_PROJECT_COSMOS_PROFILE_CONF,
        "execution_config": BI_PLATFORM_COSMOS_EXEC_CONF,
        "operator_args": {
            "vars": {"logical_date": "{{ execution_date }}"},
        },
    }

    dbt_tasks_google_data = DbtTaskGroup(
        group_id="dbt_tasks_google_data",
        render_config=RenderConfig(
            **COSMOS_COMMON_RENDER_CONF,
            select=[
                "config.tags:gam",
                "config.tags:dfp_perf",
                "path:models/staging/google",
                "path:models/marts/google",
                "path:models/reporting",
            ]
        ),
        **dbt_task_group_base_config,
    )

    dbt_tasks_facebook_data = DbtTaskGroup(
        group_id="dbt_tasks_facebook_data",
        render_config=RenderConfig(
            **COSMOS_COMMON_RENDER_CONF,
            select=[
                "path:models/staging/facebook",
                "path:models/marts/facebook",
            ]
        ),
        **dbt_task_group_base_config,  # logical_date doesn't get passed
    )

    dbt_tasks_customer_portal = DbtTaskGroup(
        group_id="dbt_tasks_customer_portal",
        render_config=RenderConfig(
            **COSMOS_COMMON_RENDER_CONF,
            select=[
                "path:seeds/customer_portal",
                "path:models/staging/customer_portal",
                "path:models/marts/customer_portal",
            ]
        ),
        **dbt_task_group_base_config, # logical_date doesn't get passed
    )
```

The responsible lines for this bug are in the converter is below:

https://github.com/astronomer/astronomer-cosmos/blob/714ecbde7666d8592f4aa8561f07d45a40ef75b0/cosmos/converter.py#L224-L225

The fix is really straightforward, to use `get` instead of `pop` in the lines above, this is ok because Cosmos will already raise an exception if both `operator_args` and `ProjectConfig` with env vars or dbt vars are both set [here](https://github.com/astronomer/astronomer-cosmos/blob/714ecbde7666d8592f4aa8561f07d45a40ef75b0/cosmos/converter.py#L131-L148).

jbandoro:
## Description

This is a fix for #833 where a user in Slack reported a bug where if the same `operator_args` are used in multiple `DbtTaskGroup`'s the `vars` were only used in the first task group.

I added a test which would have caught the bug, so that it isn't reintroduced later.

## Related Issue(s)

closes #833 

## Breaking Change?

None

## Checklist

### DIFF
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -221,8 +221,8 @@ class DbtToAirflowConverter:
 
         validate_adapted_user_config(execution_config, project_config, render_config)
 
-        env_vars = project_config.env_vars or operator_args.pop("env", None)
-        dbt_vars = project_config.dbt_vars or operator_args.pop("vars", None)
+        env_vars = project_config.env_vars or operator_args.get("env")
+        dbt_vars = project_config.dbt_vars or operator_args.get("vars")
 
         # Previously, we were creating a cosmos.dbt.project.DbtProject
         # DbtProject has now been replaced with ProjectConfig directly
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -405,3 +405,36 @@ def test_converter_project_config_dbt_vars_with_custom_load_mode(
         )
     _, kwargs = mock_legacy_dbt_project.call_args
     assert kwargs["dbt_vars"] == {"key": "value"}
+
+
+@patch("cosmos.config.ProjectConfig.validate_project")
+@patch("cosmos.converter.build_airflow_graph")
+@patch("cosmos.converter.DbtGraph.load")
+def test_converter_multiple_calls_same_operator_args(
+    mock_dbt_graph_load, mock_validate_project, mock_build_airflow_graph
+):
+    """Tests if the DbttoAirflowConverter is called more than once with the same operator_args, the
+    operator_args are not modified.
+    """
+    project_config = ProjectConfig(project_name="fake-project", dbt_project_path="/some/project/path")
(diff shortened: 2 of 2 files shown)


jlaneve (approved):
Nice, great catch & thanks for adding the test so this doesn't happen again!

PR #836, opened 2024-02-06
Add support for `InvocationMode.DBT_RUNNER` for local execution mode

### LINKED ISSUE #717: Add support to DBT_RUNNER execution mode
**Context**

There was a great recommendation from @sanromeo in the`#airflow-dbt` slack channel:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1701098801633179

To use `DbtRunner` (https://docs.getdbt.com/reference/programmatic-invocation) instead of `subprocess` for running (executing) dbt commands.

Historically, we decided not to adopt `dbt-core` as a dependency of Cosmos to avoid the conflicts between Airflow and `dbt-core`:

However,  as pointed out by @sanromeo, there are no more conflicts between dbt 1.7.0+ and Airflow 2.7.0+. So we could offer this as an alternative Cosmos [ExecutionMode](https://astronomer.github.io/astronomer-cosmos/getting_started/execution-modes.html) to users who are confident their dbt-core and Airflow versions do not conflict. If this approach is successful, we can also look into allowing users to use the same strategy in the `LoadMode.DBT_LS`.

**Acceptance criteria**

* [ ] Allow users to use `ExecutionConfig(execution_mode= ExecutionMode.DBT_RUNNER)`, which will not rely on Python `subprocess`, but call dbt `DbtRunner`

jbandoro:
## Description

This PR adds `dbtRunner` programmatic invocation for `ExecutionMode.LOCAL`. I decided to not make a new execution mode for each (e.g. `ExecutionMode.LOCAL_DBT_RUNNER`) and all of the child operators but instead added an additional config `ExecutionConfig.invocation_mode` where `InvocationMode.DBT_RUNNER` could be specified. This is so that users who are already using local execution mode could use dbt runner and see performance improvements. 

With the `dbtRunnerResult` it makes it easy to know whether the dbt run was successful and logs do not need to be parsed but are still logged in the operator:

image

## Performance Testing

After #827 was added, I modified it slightly to use postgres adapter instead of sqlite because the latest dbt-core support for sqlite is 1.4 when programmatic invocation requires >=1.5.0. I got the following results comparing subprocess to dbt runner for 10 models:

1. `InvocationMode.SUBPROCESS`:
```shell
Ran 10 models in 23.77661895751953 seconds
NUM_MODELS=10
TIME=23.77661895751953
```
2. `InvocationMode.DBT_RUNNER`:
```shell
Ran 10 models in 8.390100002288818 seconds
NUM_MODELS=10
TIME=8.390100002288818
```

So using `InvocationMode.DBT_RUNNER` is almost 3x faster, and can speed up dag runs if there are a lot of models that execute relatively quickly since there seems to be a 1-2s speed up per task.

One thing I found while working on this is that a [manifest](https://docs.getdbt.com/reference/programmatic-invocations#reusing-objects) is stored in the result if you parse a project with the runner, and can be reused in subsequent commands to avoid reparsing. This could be a useful way for caching the manifest if we use dbt runner for dbt ls parsing and could speed up the initial render as well.

I thought at first it would be easy to have this also work for virtualenv execution, since I at first thought the entire `execute` method was run in the virtualenv, which is not the case since the virtualenv operator creates a virtualenv and then passes the executable path to a subprocess. It may be possible to have this work for virtualenv and would be better suited for a follow-up PR.

## Related Issue(s)

closes #717 

## Breaking Change?

None

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works - added unit tests and integration tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -293,6 +293,7 @@ jobs:
       PYTHONPATH: /home/runner/work/astronomer-cosmos/astronomer-cosmos/:$PYTHONPATH
 
   Run-Performance-Tests:
+    needs: Authorize
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -346,13 +346,14 @@ jobs:
           AIRFLOW_CONN_AIRFLOW_DB: postgres://postgres:postgres@0.0.0.0:5432/postgres
           AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT: 90.0
           PYTHONPATH: /home/runner/work/astronomer-cosmos/astronomer-cosmos/:$PYTHONPATH
-          MODEL_COUNT: ${{ matrix.num-models }}
+          COSMOS_CONN_POSTGRES_PASSWORD: ${{ secrets.COSMOS_CONN_POSTGRES_PASSWORD }}
           POSTGRES_HOST: localhost
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
           POSTGRES_DB: postgres
           POSTGRES_SCHEMA: public
           POSTGRES_PORT: 5432
+          MODEL_COUNT: ${{ matrix.num-models }}
     env:
       AIRFLOW_HOME: /home/runner/work/astronomer-cosmos/astronomer-cosmos/
       AIRFLOW_CONN_AIRFLOW_DB: postgres://postgres:postgres@0.0.0.0:5432/postgres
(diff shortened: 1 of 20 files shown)


jlaneve (commented):
This is looking great! Will try it out locally over the weekend.

One thing I haven't thought a ton about is what config we should pass the performance tests. IMO the perf tests should cover the "best case" scenario (appropriately tune Cosmos for performance) so that we're always pushing the boundary vs the default. Thoughts on including a change in this PR to the perf tests to use this new method?

One other thought: do you think it's worth doing any auto-discovery to infer which invocation method is used? i.e. if you don't explicitly specify one, should we:
- try to import the dbt runner, if it works, great - we can use the more performant method
- if it doesn't work, no problem, we default to subprocess

jbandoro:
@jlaneve I'm closing this PR and opening up #850 because I couldn't update the GH action and have it run with the updates here on my forked branch.

PR #842, opened 2024-02-09
fix inplace mutation

dwreeves:
`operator_args.pop()` causes an in-place mutation, which long story short caused an issue for me today. I think it is a lot safer to rely on a `copy()` if the `DbtToAirflowConverter` is going to mutate the user inputs.

### DIFF
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -210,7 +210,7 @@ class DbtToAirflowConverter:
 
         execution_config = execution_config or ExecutionConfig()
         render_config = render_config or RenderConfig()
-        operator_args = operator_args or {}
+        operator_args = operator_args.copy() if operator_args else {}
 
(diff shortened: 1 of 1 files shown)


jbandoro:
Thanks @dwreeves! This fixes the same issue that https://github.com/astronomer/astronomer-cosmos/pull/835 tries to fix from user reports in Slack. The inplace updates for `operator_args` were for "env" or "vars" which #835 fixes, and there shouldn't be any updates in the rest of the converter because they're shallow copied below to the `task_args`:
https://github.com/astronomer/astronomer-cosmos/blob/72690d31ba6822ae6df56f4d56f26196234c80b5/cosmos/converter.py#L246-L253

dwreeves:
Ah I should have looked at existing PRs. I can close this then.

PR #849, opened 2024-02-16
Fix: Docker and Kubernetes operators execute method resolution

### LINKED ISSUE #848: Bug: MRO order of `DbtKubernetesBaseOperator.execute` causing issues
This was reported by slack users in the [thread here](https://apache-airflow.slack.com/archives/C059CC42E9W/p1707949188134889). The issue is that the `execute` method for the dbt kubernetes operator does not resolve to the base operator [execute method](https://github.com/astronomer/astronomer-cosmos/blob/8a1fa3f07fd0d85fdc5620089f564d17bd911c53/cosmos/operators/base.py#L245-L246) but instead to the `KubernetesPodOperator` and results in no dbt args passed to the pod args.

The fix is to change the order of the base classes and to explicitly call the kubernetes pod operator execute with `result = KubernetesPodOperator.execute(self, context)`.

jbandoro:
## Description
This was a bug reported by users using the kuberentes execution method where the pods wouldn't run any commands. With how the base operators inherited from their parent classes, the `execute` method for the operator was either the `DockerOperator.execute` or `KubernetesPodOperator.execute` skipping the `build_and_run_cmd` required to setup before containers were run.

As part of this fix, I added tests to check that when the `execute` method is invoked that the docker/kube args are first set before the executor pod is executed.

## Related Issue(s)

closes #848

### DIFF
--- a/cosmos/operators/docker.py
+++ b/cosmos/operators/docker.py
@@ -28,7 +28,7 @@ except ImportError:
     )
 
 
-class DbtDockerBaseOperator(DockerOperator, AbstractDbtBaseOperator):  # type: ignore
+class DbtDockerBaseOperator(AbstractDbtBaseOperator, DockerOperator):  # type: ignore
     """
     Executes a dbt core cli command in a Docker container.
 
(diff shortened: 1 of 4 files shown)


jlaneve (approved):
Nice! Love the quick fix

PR #850, opened 2024-02-17
Add support for `InvocationMode.DBT_RUNNER` for local execution mode

### LINKED ISSUE #717: Add support to DBT_RUNNER execution mode

jbandoro:
## Description

This PR adds `dbtRunner` programmatic invocation for `ExecutionMode.LOCAL`. I decided to not make a new execution mode for each (e.g. `ExecutionMode.LOCAL_DBT_RUNNER`) and all of the child operators but instead added an additional config `ExecutionConfig.invocation_mode` where `InvocationMode.DBT_RUNNER` could be specified. This is so that users who are already using local execution mode could use dbt runner and see performance improvements. 

With the `dbtRunnerResult` it makes it easy to know whether the dbt run was successful and logs do not need to be parsed but are still logged in the operator:

image

## Performance Testing

After #827 was added, I modified it slightly to use postgres adapter instead of sqlite because the latest dbt-core support for sqlite is 1.4 when programmatic invocation requires >=1.5.0. I got the following results comparing subprocess to dbt runner for 10 models:

1. `InvocationMode.SUBPROCESS`:
```shell
Ran 10 models in 23.77661895751953 seconds
NUM_MODELS=10
TIME=23.77661895751953
```
2. `InvocationMode.DBT_RUNNER`:
```shell
Ran 10 models in 8.390100002288818 seconds
NUM_MODELS=10
TIME=8.390100002288818
```

So using `InvocationMode.DBT_RUNNER` is almost 3x faster, and can speed up dag runs if there are a lot of models that execute relatively quickly since there seems to be a 1-2s speed up per task.

One thing I found while working on this is that a [manifest](https://docs.getdbt.com/reference/programmatic-invocations#reusing-objects) is stored in the result if you parse a project with the runner, and can be reused in subsequent commands to avoid reparsing. This could be a useful way for caching the manifest if we use dbt runner for dbt ls parsing and could speed up the initial render as well.

I thought at first it would be easy to have this also work for virtualenv execution, since I at first thought the entire `execute` method was run in the virtualenv, which is not the case since the virtualenv operator creates a virtualenv and then passes the executable path to a subprocess. It may be possible to have this work for virtualenv and would be better suited for a follow-up PR.

## Related Issue(s)

closes #717 

## Breaking Change?

None

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works - added unit tests and integration tests.


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, 717-add-dbtrunner-local-executor-2] # # TODO:remove before merge
```
jbandoro, line 5: Commenting here to remind me to remove this before merging. The performance dag was updated to use postgres and wanted to test the workflow in this PR.

jlaneve:
Screenshot 2024-02-20 at 8 24 54 PM

If I didn't write these integration tests myself, I don't know if I'd believe them!! This is super cool

jbandoro:
@jlaneve those test results look too good (unfortunately), I think it's because the job summary you're looking at is for the [on: pull_request_target](https://github.com/astronomer/astronomer-cosmos/actions/runs/7978515212) which is running the action that is in the `main` branch which has tasks in the dag failing because of the sqlite to postgres update that's causing it to fail - see [example here](https://github.com/astronomer/astronomer-cosmos/actions/runs/7978515212/job/21783798078). The 'real' summary is in the [on: push](https://github.com/astronomer/astronomer-cosmos/actions/runs/7978514995).

I can add in a check to the performance dag test to make sure the dag run was successful and no tasks failed - so this would have been caught.

Below is a comparison of the job summary between the the current branch and the last time the peformance dag was run on a commit to main:

| dbtRunner (this branch)      | subprocess |
| :---        |    :----:   |
| image  | image      |

.github/workflows/test.yml
```diff
@@ -2,9 +2,9 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
-  pull_request_target: # Also run on pull requests originated from forks
-    branches: [main]
+    branches: [main, 717-add-dbtrunner-local-executor-2] # TODO:remove before merge
+  # pull_request_target: # Also run on pull requests originated from forks # TODO: uncomment before merge
+    # branches: [main] # TODO: uncomment before merge
```
jbandoro, line 7: Reminder to revert this before merging, stops workflow from running with the action file from main and only this branch.

jbandoro:
The performance tests for this branch (push) are succeeding, but the performance tests for (pull_request_target) are failing because I added in a check to the peformance dag test to ensure it was successful and the main GH action file does not have the necessary postgres service that was added in this PR.

docs/getting_started/execution-modes.rst
```diff
@@ -184,3 +184,33 @@ Each task will create a new container on Azure, giving full isolation. This, how
...
+================
+.. versionadded:: 1.4
+
+For ``ExecutionMode.LOCAL`` execution mode, Cosmos supports two invocation modes for running dbt:
+
+1. ``InvocationMode.SUBPROCESS``: In this mode, Cosmos runs dbt cli commands using the Python ``subprocess`` module and parses the output to capture logs and to raise exceptions.
+
+2. ``InvocationMode.DBT_RUNNER``: In this mode, Cosmos uses the ``dbtRunner`` available for `dbt programmatic invocations <https://docs.getdbt.com/reference/programmatic-invocations>`__ to run dbt commands. \
+   In order to use this mode, dbt must be installed in the same environment, either local or virtualenv for the worker. This mode does not have the overhead of spawning new subprocesses or parsing the output of dbt commands and is faster than ``InvocationMode.SUBPROCESS``. \
+   This mode requires dbt version 1.5.0 or higher.
```
tatiana, line 200: It is probably worth linking that it is up to the user to resolve any dependencies conflicts between Airflow an dbt, and add a link to: https://astronomer.github.io/astronomer-cosmos/getting_started/execution-modes-local-conflicts.html#execution-modes-local-conflicts

jbandoro:
The performance tests are now failing because I removed this branch from the on-push from the test workflow. The performance tests passed in the previous commit here: https://github.com/astronomer/astronomer-cosmos/actions/runs/8068823803

PR #856, opened 2024-02-23
fix `folder_dir` not showing on logs for `DbtDocsS3LocalOperator`

PrimOox:
## Description

DAG logs is not printing `folder_dir`
```python
generate_dbt_docs_aws = DbtDocsS3Operator(
    task_id="generate_dbt_docs_aws",
    project_dir=f"{AIRFLOW_HOME}/dags/dbt/dbt-project",
    env=env_vars,
    append_env=True,
    # docs-specific arguments
    connection_id="aws_default",
    bucket_name="airflow-data-xxxxxxxx-us-east-2",
    folder_dir="dags/dbt/dbt-project/target",
    dag=dbt_cosmos_dag,
)
```
image

## Related Issue(s)
None

## Breaking Change?
No

## Checklist

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -574,9 +574,9 @@ class DbtDocsS3LocalOperator(DbtDocsCloudLocalOperator):
         )
 
         for filename in self.required_files:
-            logger.info("Uploading %s to %s", filename, f"s3://{self.bucket_name}/{filename}")
-
             key = f"{self.folder_dir}/{filename}" if self.folder_dir else filename
+            s3_path = f"s3://{self.bucket_name}/{key}"
+            logger.info("Uploading %s to %s", filename, s3_path)
 
             hook.load_file(
                 filename=f"{target_dir}/{filename}",


jbandoro (commented):
Thanks @PrimOox for catching this and contributing a fix!

My comment below is to use the existing `key` instead for logging the s3 path

cosmos/operators/local.py
```diff
@@ -574,7 +574,9 @@ def upload_to_cloud_storage(self, project_dir: str) -> None:
         )
 
         for filename in self.required_files:
-            logger.info("Uploading %s to %s", filename, f"s3://{self.bucket_name}/{filename}")
+            key = f"{self.folder_dir}/{filename}" if self.folder_dir else filename
+            s3_path = f"s3://{self.bucket_name}/{key}"
+            logger.info("Uploading %s to %s", filename, s3_path)
 
             key = f"{self.folder_dir}/{filename}" if self.folder_dir else filename
```
jbandoro, line 581: hey sorry if my first comment wasn't clear, `key` is already defined below on line 581 here which can be used, since it's now defined twice:
```suggestion
            key = f"{self.folder_dir}/{filename}" if self.folder_dir else filename
            s3_path = f"s3://{self.bucket_name}/{key}"
            logger.info("Uploading %s to %s", filename, s3_path)
```

PrimOox: didn`t see it, my bad

jbandoro (approved):
Thanks again for the contribution @PrimOox, will get this into the next release.

Issue #838, opened 2024-02-06
Invalid selector values passed to DbtTestMixin


ms32035:
https://github.com/astronomer/astronomer-cosmos/blame/72690d31ba6822ae6df56f4d56f26196234c80b5/cosmos/airflow/graph.py#L107

In `DbtTestMixin` the types are
```
        exclude: str | None = None,
        select: str | None = None,
        selector: str | None = None,
```

whereas what comes from `RenderConfig` is
```
    select: list[str] = field(default_factory=list)
    exclude: list[str] = field(default_factory=list)
    selector: str | None = None
```

this results in an invalid test command generation
```
'--select', '[]', '--exclude', '[]',
```

ms32035:
This has fixed the issue for me
```
        task_args["select"] = render_config.select[0] if render_config.select else None
        task_args["selector"] = render_config.selector
        task_args["exclude"] = render_config.exclude[0] if render_config.exclude else None
```
though I don't exactly understand why as by empty list evaluate to `False` in Python

pankajkoti:
probably just an annotation issue after the PR https://github.com/astronomer/astronomer-cosmos/pull/863 was contributed, but would be nice to confirm once and close the issue after correcting the annotation

PR #863, opened 2024-02-27
Fix global flags for lists

ms32035:
## Description

Correctly deals with global flags when they are a list.
Note - in the module there's no distinguishing which are and which aren't.

## Breaking Change?

Fixes the resulting bug in #838, though the type discrepancy is still there.

## Checklist

### DIFF
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -340,18 +340,6 @@ class DbtTestMixin:
         self.selector = selector
         super().__init__(exclude=exclude, select=select, selector=selector, **kwargs)  # type: ignore
 
-    def add_cmd_flags(self) -> list[str]:
-        flags = []
-        if self.exclude:
-            flags.extend(["--exclude", *self.exclude])
-
-        if self.select:
-            flags.extend(["--select", *self.select])
-
-        if self.selector:
-            flags.extend(["--selector", self.selector])
-        return flags
-
 
 class DbtRunOperationMixin:
     """
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -198,7 +198,7 @@ class AbstractDbtBaseOperator(BaseOperator, metaclass=ABCMeta):
                     yaml_string = yaml.dump(global_flag_value)
                     flags.extend([dbt_name, yaml_string])
                 elif isinstance(global_flag_value, list):
-                    flags.extend([i for j in global_flag_value for i in [dbt_name, j]])
+                    flags.extend([dbt_name, " ".join(global_flag_value)])
                 else:
                     flags.extend([dbt_name, str(global_flag_value)])
         for global_boolean_flag in self.global_boolean_flags:
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -198,6 +198,8 @@ class AbstractDbtBaseOperator(BaseOperator, metaclass=ABCMeta):
                 if isinstance(global_flag_value, dict):
                     yaml_string = yaml.dump(global_flag_value)
                     flags.extend([dbt_name, yaml_string])
+                elif isinstance(global_flag_value, list):
+                    flags.extend([i for j in global_flag_value for i in [dbt_name, j]])
                 else:
                     flags.extend([dbt_name, str(global_flag_value)])
         for global_boolean_flag in self.global_boolean_flags:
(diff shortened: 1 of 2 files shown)


tatiana (commented):
Thanks for working to fix this, @ms32035 !

Could you add a test that represents the original issue you found?

Would it make sense to fix the types as part of this PR? As you noticed,  `select` and `exclude` are likely lists.

ms32035:
I can add a test, but fixing the whole issue is beyond the scope of the PR. The problem is 
you have a duplication of handling `select` and `exclude` flags
between
https://github.com/astronomer/astronomer-cosmos/blob/da50758dc667caf5a1673d299586573920a8f785/cosmos/operators/base.py#L66
and 
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/base.py#L341
which you need to address on a design level

ms32035:
@tatiana maybe even this whole method is not needed, as it's handled by the global flags?
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/base.py#L341

tatiana:
> @tatiana maybe even this whole method is not needed, as it's handled by the global flags? https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/base.py#L341

I think you're right, @ms32035 , could you delete this method and we can see if tests continue passing?

ms32035:
@tatiana this should now be addressed in my latest commit

cosmos/operators/base.py
```diff
@@ -197,6 +197,8 @@ def add_global_flags(self) -> list[str]:
                 if isinstance(global_flag_value, dict):
                     yaml_string = yaml.dump(global_flag_value)
                     flags.extend([dbt_name, yaml_string])
+                elif isinstance(global_flag_value, list):
+                    flags.extend([i for j in global_flag_value for i in [dbt_name, j]])
```
tatiana, line 201:
```suggestion
                elif isinstance(global_flag_value, list):
                    flags.extend(dbt_name, " ".join(global_flag_value)
```
@ms32035 I've been thinking about this, and I believe we need to change what I proposed.

From a node selection perspective (`select` and `exclude`), Cosmos has been dealing with items of the list as items of the same select/exclude statement, e.g.:
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/dbt/selector.py#L438

Currently, if users do:
```
select: ["a", "b"]
```

During DAG parsing, Cosmos resolves it as `--select "a b"`.

The way that was implemented in da50758dc667caf5a1673d299586573920a8f785  would result in `--select "a" --select "b"`.

This is not very useful since dbt ignores the first select statement and only takes into account `--select "b"`

tests/operators/test_local.py
```diff
@@ -566,27 +566,35 @@ def test_store_compiled_sql() -> None:
...
             {"args": {"days": 7, "dry_run": True}, "macro_name": "bla"},
-            {"context": {}, "cmd_flags": ["--args", "days: 7\ndry_run: true\n"]},
+            {"context": {}, "env": {}, "cmd_flags": ["run-operation", "bla", "--args", "days: 7\ndry_run: true\n"]},
         ),
     ],
 )
-@patch("cosmos.operators.local.DbtLocalBaseOperator.build_and_run_cmd")
-def test_operator_execute_with_flags(mock_build_and_run_cmd, operator_class, kwargs, expected_call_kwargs):
+@patch("cosmos.operators.local.DbtLocalBaseOperator.run_command")
+def test_operator_execute_with_flags(mock_run_cmd, operator_class, kwargs, expected_call_kwargs):
```
tatiana, line 597: Since we're changing the method `add_global_flags`, it may be more representative to change the test `test_dbt_base_operator_add_global_flags` to something like:

```
def test_dbt_base_operator_add_global_flags() -> None:
    dbt_base_operator = ConcreteDbtLocalBaseOperator(
        task_id="my-task",
        project_dir="my/dir",
        vars={
            "start_time": "{{ data_interval_start.strftime('%Y%m%d%H%M%S') }}",
            "end_time": "{{ data_interval_end.strftime('%Y%m%d%H%M%S') }}",
        },
        no_version_check=True,
        select=["my_first_model", "my_second_model"]
    )
    assert dbt_base_operator.add_global_flags() == [
        "--select",
        "my_first_model my_second_model",
        "--vars",
        "end_time: '{{ data_interval_end.strftime(''%Y%m%d%H%M%S'') }}'\n"
        "start_time: '{{ data_interval_start.strftime(''%Y%m%d%H%M%S'') }}'\n",
        "--no-version-check",
    ]
```

cosmos/operators/base.py
ms32035: indeed, my mistake

ms32035:
@tatiana this should be good for merge now

tatiana (approved):
Thanks for working on this, @ms32035 ! We'll release it as part of 1.4

PR #771, opened 2023-12-20
Add Azure Container Instance as Execution Mode

danielvdende:
## Description

This PR adds support for Azure Container Instance as an Execution Mode. In this mode, a container is run on Azure to execute the dbt commands.

## Related Issue(s)

This closes #635 

## Checklist

- [x] I have made corresponding changes to the documentation (if required)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #872, opened 2024-03-01
Fix ACI integration

danielvdende:
The inheritance order of the `DbtAzureContainerInstanceBaseOperator` was incorrect, as well as the wrong execute method being called. This led to no dbt command actually being built nor run. I noticed this when testing the 1.4.0a1 release.

Example log before:
```
[2024-03-01, 20:27:29 UTC] {container_instances.py:366} INFO - Usage: dbt [OPTIONS] COMMAND [ARGS]...
[2024-03-01, 20:27:29 UTC] {container_instances.py:366} INFO - 
```

After this fix:
```
[2024-03-01, 20:28:44 UTC] {container_instances.py:366} INFO - [0m20:28:44  Found 5 models, 3 seeds, 20 tests, 0 sources, 0 exposures, 0 metrics, 403 macros, 0 groups, 0 semantic models
[2024-03-01, 20:28:44 UTC] {container_instances.py:366} INFO - [0m20:28:44
```

## Checklist

### DIFF
--- a/cosmos/operators/azure_container_instance.py
+++ b/cosmos/operators/azure_container_instance.py
@@ -79,13 +79,13 @@ class DbtAzureContainerInstanceBaseOperator(AbstractDbtBaseOperator, AzureContai
         self.command: list[str] = dbt_cmd
 
 
-class DbtLSAzureContainerInstanceOperator(DbtLSMixin, DbtAzureContainerInstanceBaseOperator):
+class DbtLSAzureContainerInstanceOperator(DbtLSMixin, DbtAzureContainerInstanceBaseOperator):  # type: ignore
     """
     Executes a dbt core ls command.
     """
 
 
-class DbtSeedAzureContainerInstanceOperator(DbtSeedMixin, DbtAzureContainerInstanceBaseOperator):
+class DbtSeedAzureContainerInstanceOperator(DbtSeedMixin, DbtAzureContainerInstanceBaseOperator):  # type: ignore
     """
     Executes a dbt core seed command.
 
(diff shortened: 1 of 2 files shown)


danielvdende:
@tatiana @jbandoro Hi! This seems to have slipped through during the refactor done towards the end of #771

cosmos/operators/azure_container_instance.py
```diff
@@ -66,7 +66,7 @@ def __init__(
     def build_and_run_cmd(self, context: Context, cmd_flags: list[str] | None = None) -> None:
         self.build_command(context, cmd_flags)
         self.log.info(f"Running command: {self.command}")
-        result = super().execute(context)
+        result = AzureContainerInstancesOperator.execute(self, context)
```
tatiana, line 69: @danielvdende is there any chance you could get some test coverage over here?

tatiana (commented):
Thank you very much for fixing the issue, @danielvdende! I left a minor comment, we'll make sure this is released before the stable 1.4.0 release

cosmos/operators/azure_container_instance.py
danielvdende: I'm not sure what the scope of the test would be tbh. I could add tests for this function, but it would essentially be validating that a number of mocks were being called. Is that what you have in mind?

danielvdende: @tatiana gentle reminder about this :-). Would be great if we can merge this PR soon

tatiana: Hi @danielvdende it would be great if you could add a test that reproduces the error you were facing before making this fix:
```
The inheritance order of the DbtAzureContainerInstanceBaseOperator was incorrect, as well as the wrong execute method being called.
```
Ideally, this test would fail before your fix and pass after your fix. Unless we have a test, we're at risk of incurring the same error again in the future.

danielvdende: Hi @tatiana , I've added a test that covers this. It indeed fails if I run it on code as it was before this fix.

tatiana (approved):
Thanks for addressing feedback, @danielvdende !

PR #874, opened 2024-03-03
Update docs regarding docs plugin in Astronomer cloud

### LINKED ISSUE #871: [Bug] [1.4.0a] dbt docs page does not work in Astronomer
I will need to investigate on my side when I get time, but I was struggling to get the dbt docs working in Astronomer.

This is a bit vague, I know, and I should probably investigate further before posting this, but I'm also the one who designed the dbt docs feature so I know how it works, and this was not anticipated.

- Locally, everything works fine.
- I also tried the S3 method, i.e. pointing my docs to `s3://my-bucket/dbt_docs_folder`, and that works fine both locally and in Astronomer (after fixing the `DbtDocsS3Operator`; see #869).
- Running in Astronomer, however, the local file path method does not work. And that is weird because:
  - I run `dbt docs generate` in prod alongside the `dbt compile` command, so all the relevant files should be there and deployed.
  - I also ran a `ls .dags/dbt/target` inside my CICD env and confirmed that `index.html` is generated.
  - I get a 500 error and an error message saying that `/usr/local/airflow/dags/dbt/target/index.html` is not found when I read the logs.

Could just be a configuration error on my end. Or not, and maybe it is something where in the Astronomer runtime, plugins cannot access some folders. We'll see. I am going to investigate further and print my `target/` inside the webserver instance. But I'm making a thread here in case anyone else runs into a similar issue.

dwreeves:
Resolves #871

Turns out, there is a lot of nuance in regards to how Astronomer interacts with stuff under the hood.

### DIFF
--- a/docs/configuration/hosting-docs.rst
+++ b/docs/configuration/hosting-docs.rst
@@ -105,10 +105,28 @@ For example, if your dbt project directory is ``/usr/local/airflow/dags/my_dbt_p
 
     AIRFLOW__COSMOS__DBT_DOCS_DIR="/usr/local/airflow/dags/my_dbt_project/target"
 
-Using docs out of local storage has the downside that some values in the dbt docs can become stale unless the docs are periodically refreshed and redeployed:
+Using docs out of local storage has a couple downsides. First, some values in the dbt docs can become stale, unless the docs are periodically refreshed and redeployed:
 
 - Counts of the numbers of rows.
 - The compiled SQL for incremental models before and after the first run.
+
+Second, deployment from local storage may only be partially compatible with some managed Airflow systems.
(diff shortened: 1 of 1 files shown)


tatiana (approved):
Thanks, @dwreeves ! I've shared this with other Astronomer team members to see how we could improve this.

dwreeves:
@tatiana Spitballing from the outside without any knowledge of Astronomer's architecture but I think if the `dags/` folder, wherever it's deployed, were mounted as a read volume to the webserver instance, that solves both issues.

That's roughly how MWAA works, for example (albeit MWAA does that for both the scheduler and webserver instances, whereas here it's only useful for the webserver instance.)

PR #884, opened 2024-03-09
Enhance manifest structure validation and add debug logging

AlgirdasDubickas:
This commit introduces several improvements to the manifest file processing logic to ensure robustness and ease of debugging:

1. **Manifest Structure Validation**: Added checks to verify that the manifest file contains the expected top-level keys ("nodes", "sources", "exposures"). 

2. **Data Type Validation**: Implemented additional validation to confirm that the "nodes" key maps to a dictionary. This step helps catch issues where the manifest's format deviates from expected structures, particularly in how nodes are represented.

3. **Key Presence Validation**: For each node within the manifest, we now verify the presence of essential keys ("resource_type", "original_file_path", "tags", "config").

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -398,8 +398,26 @@ class DbtGraph:
         with open(self.project.manifest_path) as fp:  # type: ignore[arg-type]
             manifest = json.load(fp)
 
+            # Validate expected structure in manifest
+            if not all(key in manifest for key in ["nodes", "sources", "exposures"]):
+                logger.error("Manifest file structure is not as expected.")
+                raise CosmosLoadDbtException("Invalid structure in manifest file")
+            
+            # Further validation can be performed here, e.g., checking data types
+            if not isinstance(manifest["nodes"], dict):
+                logger.error("Expected 'nodes' to be a dictionary.")
+                raise CosmosLoadDbtException("Invalid 'nodes' structure in manifest file")
+
+            
             resources = {**manifest.get("nodes", {}), **manifest.get("sources", {}), **manifest.get("exposures", {})}
             for unique_id, node_dict in resources.items():
+
+                expected_keys = ["resource_type", "original_file_path", "tags", "config"]
+                if not all(key in node_dict for key in expected_keys):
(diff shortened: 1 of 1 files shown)


tatiana (commented):
@AlgirdasDubickas thanks for the contribution!

Some comments:

1. Did you run into some issue that motivated this PR..?

2. Is the overall goal to check if the manifest contains fields that Cosmos relies on, or to check the manifest is compliant to dbt schema in a broader sense? I was particularly curious of in which scenario `nodes` are not a dictionary.

3. If the answer for (2) is to validate the dbt schema, would it make sense to use dbt's artifact schemas to validate this? https://schemas.getdbt.com/, and validate the manifest based on the dbt version?

4. Should we allow users to opt-in validating the schema or not, via a configuration in `RenderConfig`? Since this may add additional processing to DAG parsing, it may be worth to have a configuration for enabling / disabling this type of check.

Last but not least, it would be great to have some tests for this feature, once we confirm the original problem / overall approach

cosmos/dbt/graph.py
```diff
@@ -399,8 +399,24 @@ def load_from_dbt_manifest(self) -> None:
         with open(self.project.manifest_path) as fp:  # type: ignore[arg-type]
             manifest = json.load(fp)
 
+            # Validate expected structure in manifest
+            if not all(key in manifest for key in ["nodes", "sources", "exposures"]):
+                logger.error("Manifest file structure is not as expected.")
+                raise CosmosLoadDbtException("Invalid structure in manifest file")
```
pankajastro, line 405: wondering what happens today if this validation isn't in place when you have an incorrect manifest?

pankajastro (commented):
Hello @AlgirdasDubickas, I wanted to share some thoughts regarding the proposed change. It seems like implementing this might introduce some additional latency during DAG parsing. Additionally, I noticed that there haven't been any updates on the PR for quite some time. I was just wondering if you are still planning to work on this or if there's anything I can assist with.

tatiana:
@AlgirdasDubickas, we're closing this PR since it's stale. Please feel free to reopen it or create a new one in the future if you have a chance to work on this again - we really appreciate your contribution and would love to see it merged in the future!

PR #886, opened 2024-03-10
Expose the dbt graph in the `DbtToAirflowConverter` class

### LINKED ISSUE #885: dbt_graph in DbtToAirflowConverter should be exposed for use-cases that require custom DAG logic based on the properties of the graph instance
`dbt_graph` local variable inside `DbtToAirflowConverter` should be exposed for use-cases that require custom DAG logic based on the properties of the graph object (both in the sense of dbt graph as well as airflow task group)

Example snippet of how graph instance can be used:
```python
def filter_leaf_nodes(node_item: BaseOperator, group_id: str) -> bool:
    """
    Filters for all leafs nodes, ie.
    DbtRun nodes that belong this task group, and have
    either no downstream nodes or only DbtTest nodes downstream
    """
    return (
        node_item[1].task_type == "DbtRunOperator"
        and node_item[1].task_group.group_id == group_id
        and (
            len(list(node_item[1].downstream_list)) == 0
            or list(
                filter(
                    lambda downstream_node: downstream_node.task_type
                    == "DbtTestOperator",
                    node_item[1].downstream_list,
                )
            )
            == node_item[1].downstream_list
        )
    )

def filter_leaf_models(dbt_task_group: DbtTaskGroup) -> List[Node]:
    """
    Get all model nodes that have no downstream dependencies. These are the models
    that will be generated by this graph

    """
    return list(
        map(
            lambda item: item[0],
            filter(
                lambda item: filter_leaf_nodes(item, dbt_task_group.group_id),
                dbt_task_group.graph.nodes.items(),
            ),
        )
    )
```
The filter is then used to generate other airflow task groups
Example:
Screenshot 2024-03-12 at 11 00 34 AM
All task groups after `computation_task_group` require custom filtering based on the above code snippets, which rely on properties of the `dbt_graph` object inside `DbtToAirflowConverter`

Note: using the compiled artifacts or manifest is not practical because the graph itself can also be arbitrarily filtered based on custom selectors, complicating the dag generation logic if it has no access to the generated graph's properties.

tommyjxl:
## Description

`dbt_graph` is now accessible from generated dbt task groups, through `DbtToAirflowConverter`

## Related Issue(s)

closes #885 

## Breaking Change?

non-breaking change

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Callable, Union, Dict
+from typing import Any, Callable, Dict, Union
 
 from airflow.models import BaseOperator
 from airflow.models.dag import DAG
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Callable, Dict, Union
+from typing import Any, Callable, Union, Dict
 
 from airflow.models import BaseOperator
 from airflow.models.dag import DAG
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Callable, Union
+from typing import Any, Callable, Union, Dict
 
 from airflow.models import BaseOperator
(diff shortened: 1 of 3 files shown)


tatiana (approved):
Thanks for sharing the use-case and implementing this improvement, @tommyjxl ! This is very useful.

PR #887, opened 2024-03-11
Update pip install instructions in docs

linchun3:
### Current Behaviour

The pip3 install step does not work
```
pip3 install -e apache-airflow[cncf.kubernetes,openlineage]
```

Screenshot 2024-03-11 at 21 33 28

### Proposed change

Update pip install with double quotation marks so that the command works for both unix and windows environment

## Checklist


tatiana (approved):

tatiana:
Thanks for improving the docs, @linchun3 !

PR #893, opened 2024-03-14
fix:Adjust dbt_packages handling based on packages.yml existence

AlgirdasDubickas:
his commit refines the logic for handling the `dbt_packages` directory during the symlink creation process in `create_symlinks` function. Previously, the decision to ignore `dbt_packages` was solely based on the `ignore_dbt_packages` flag. The updated logic now also checks for the presence of a `packages.yml` file within the project directory.

- If `ignore_dbt_packages` is True and `packages.yml` exists, the function will add `dbt_packages` to the list of directories to ignore. This ensures that the symlink creation process respects the project's dependency management setup.

## Checklist

### DIFF
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -14,7 +14,7 @@ DBT_TARGET_DIR_NAME = "target"
 DBT_PARTIAL_PARSE_FILE_NAME = "partial_parse.msgpack"
 DBT_LOG_FILENAME = "dbt.log"
 DBT_BINARY_NAME = "dbt"
-DBT_PACKAGES_YML_PATH = Path(os.path.expanduser("~")).joinpath(".dbt/packages.yml") 
+DBT_PACKAGES_YML_PATH = Path(os.path.expanduser("~")).joinpath(".dbt/packages.yml")
 
 DEFAULT_OPENLINEAGE_NAMESPACE = "cosmos"
 OPENLINEAGE_PRODUCER = "https://github.com/astronomer/astronomer-cosmos/"
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -14,7 +14,7 @@ DBT_TARGET_DIR_NAME = "target"
 DBT_PARTIAL_PARSE_FILE_NAME = "partial_parse.msgpack"
 DBT_LOG_FILENAME = "dbt.log"
 DBT_BINARY_NAME = "dbt"
-DBT_PACKAGES_YML_PATH = Path(os.path.expanduser("~")).joinpath(".dbt/packages.yml")
+DBT_PACKAGES_CONFIG_PATH = Path(os.path.expanduser("~")).joinpath(".dbt/packages.yml")
 
 DEFAULT_OPENLINEAGE_NAMESPACE = "cosmos"
 OPENLINEAGE_PRODUCER = "https://github.com/astronomer/astronomer-cosmos/"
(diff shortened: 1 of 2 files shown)


cosmos/dbt/project.py
```diff
@@ -12,9 +12,13 @@
 def create_symlinks(project_path: Path, tmp_dir: Path, ignore_dbt_packages: bool) -> None:
     """Helper function to create symlinks to the dbt project files."""
     ignore_paths = [DBT_LOG_DIR_NAME, DBT_TARGET_DIR_NAME, "profiles.yml"]
+
     if ignore_dbt_packages:
-        # this is linked to dbt deps so if dbt deps is true then ignore existing dbt_packages folder
-        ignore_paths.append("dbt_packages")
+        # Check for the presence of 'packages.yml' in the project directory:
+        if (project_path / "packages.yml").exists():
```
tatiana, line 18: Let's avoid adding strings representing constants in the code's middle. Could you add `packages.yml` to the `constants.py` file, similar to the other constants? E.g.:
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/constants.py#L14

Also, why not have both checks in the same line?
```
if ignore_dbt_packages and (project_path / "packages.yml").exists():
```

tatiana (commented):
Hi @AlgirdasDubickas , thanks for contributing!

Did you face an error or exception when using `ignore_dbt_packages`? Please, could you add a test to represent the problem this PR solves?

cosmos/constants.py
```diff
@@ -14,6 +14,7 @@
 DBT_PARTIAL_PARSE_FILE_NAME = "partial_parse.msgpack"
 DBT_LOG_FILENAME = "dbt.log"
 DBT_BINARY_NAME = "dbt"
+DBT_PACKAGES_CONFIG_PATH = Path(os.path.expanduser("~")).joinpath(".dbt/packages.yml")
```
tatiana, line 17: Isn't the `packages.yml` relative to the dbt project, which may be in any path? It's fine to have in constants only `packages.yml`, similar to `DBT_PARTIAL_PARSE_FILE_NAME` and build the full path where we need it

cosmos/dbt/project.py
```diff
@@ -6,15 +6,22 @@
...
+)
 
 
 def create_symlinks(project_path: Path, tmp_dir: Path, ignore_dbt_packages: bool) -> None:
     """Helper function to create symlinks to the dbt project files."""
     ignore_paths = [DBT_LOG_DIR_NAME, DBT_TARGET_DIR_NAME, "profiles.yml"]
-    if ignore_dbt_packages:
-        # this is linked to dbt deps so if dbt deps is true then ignore existing dbt_packages folder
+
+    if ignore_dbt_packages and DBT_PACKAGES_CONFIG_PATH.exists():
```
tatiana, line 21: To create a test to validate this change and make sure we don't miss this use-case again, we could have a fixture similar to:
https://github.com/astronomer/astronomer-cosmos/blob/ddb25cbbcb6de3bde9d784b5fa59786c5b3beae8/tests/dbt/test_graph.py#L31-L45

And a test similar to:
https://github.com/astronomer/astronomer-cosmos/blob/ddb25cbbcb6de3bde9d784b5fa59786c5b3beae8/tests/dbt/test_project.py#L12-L20

tatiana:
HI, @AlgirdasDubickas thank you very much for contributing; we'd love to get this to the finishing line. Please, if you have a chance, let us know more about original issue you faced that led to this PR.

tatiana:
Since this PR has become stale, but the problem remains relevant, I created a follow-up PR: #1030 
I've added @AlgirdasDubickas as a co-author since he spotted the issue and was the first to propose a solution for it.

Issue #595, opened 2023-10-12
Set `append_env ` to True by default


tatiana:
As a follow-up suggestion from the Slack thread:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1696942075155009

@jlaneve suggested we could set `append_env ` to True by default, which sounds like a good idea! This way Cosmos will be consistent (`LoadMethod.DBT_LS` and operator's exposing environment variables).

Relates to: #593

PR #899, opened 2024-03-21
Enable `append_env` in `operator_args` by default

tatiana:
The default behaviour when parsing Cosmos DAGs is to use the system environment variables, but to execute dbt commands from Airflow operators is not. This PR solves this, making the behaviour consistent in DAG parsing and dbt command execution.

Closes: #595

### DIFF
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -43,10 +43,11 @@ class AbstractDbtBaseOperator(BaseOperator, metaclass=ABCMeta):
         environment variables for the new process; these are used instead
         of inheriting the current process environment, which is the default
         behavior. (templated)
-    :param append_env: If False(default) uses the environment variables passed in env params
-        and does not inherit the current process environment. If True, inherits the environment variables
+    :param append_env: . If True (default), inherits the environment variables
         from current passes and then environment variable passed by the user will either update the existing
-        inherited environment variables or the new variables gets appended to it
+        inherited environment variables or the new variables gets appended to it.
+        If False, only uses the environment variables passed in env params
+        and does not inherit the current process environment.
     :param output_encoding: Output encoding of bash command
     :param skip_exit_code: If task exits with this exit code, leave the task
         in ``skipped`` state (default: 99). If set to ``None``, any non-zero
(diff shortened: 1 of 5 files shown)


pankajkoti (approved):
LGTM. Since we're changing the default value from False to True, wondering if it will have end user impact for existing setups and whether it could be a breaking change? If we agree that it could be a breaking change, we would need to release this in a major release?

pankajastro (approved):
LGTM. 
But as mentioned by @pankajkoti, I'm wondering if we need to be concerned about changing the default value. Now, it may inject some env secrets. IMHO this case looks rare here

tatiana:
> LGTM. Since we're changing the default value from False to True, wondering if it will have end user impact for existing setups and whether it could be a breaking change? If we agree that it could be a breaking change, we would need to release this in a major release?

Hi @pankajkoti ! I agree that changing defaults usually is a breaking change.

In this particular case, this has caused users much pain and felt more like a bug and inconsistency. This is why we had planned to have it as part of 1.4. We're far from 2.0 - otherwise, we could wait.

Cosmos DAG Rendering already has this as a default. I believe the benefits outweigh the risks introduced by exposing the environment variables by default at the worker level. That said, I'll clarify in the release notes that this "bug fix" can be perceived as a breaking change.

pankajkoti:
> > LGTM. Since we're changing the default value from False to True, wondering if it will have end user impact for existing setups and whether it could be a breaking change? If we agree that it could be a breaking change, we would need to release this in a major release?
> 
> Hi @pankajkoti and @pankajastro ! I agree that changing defaults usually is a breaking change.
> 
> In this particular case, this has caused users much pain and felt more like a bug and inconsistency. This is why we had planned to have it as part of 1.4. We're far from 2.0 - otherwise, we could wait.
> 
> Cosmos DAG Rendering already has this as a default. I believe the benefits outweigh the risks introduced by exposing the environment variables by default at the worker level. That said, we can clarify in the release notes that this "bug fix" can be perceived as a breaking change - what do you think?

makes sense to me, thanks for the detailed clarification!

PR #904, opened 2024-03-28
Improve performance by 22-35% or more by caching partial parse artefact

tatiana:
Improve the performance to run the benchmark DAG with 100 tasks by 34% and the benchmark DAG with 10 tasks by 22%, by persisting the dbt partial parse artifact in Airflow nodes. This performance can be even higher in the case of dbt projects that take more time to be parsed.

With the introduction of #800, Cosmos supports using dbt partial parsing files. This feature has led to a substantial performance improvement, particularly for large dbt projects, both during Airflow DAG parsing (using LoadMode.DBT_LS) and also Airflow task execution (when using `ExecutionMode.LOCAL` and `ExecutionMode.VIRTUALENV`).

There were two limitations with the initial support to partial parsing, which the current PR aims to address:

1. DAGs using Cosmos `ProfileMapping` classes could not leverage this feature. This is because the partial parsing relies on profile files not changing, and by default, Cosmos would mock the dbt profile in several parts of the code. The consequence is that users trying Cosmos 1.4.0a1 will see the following message:
```
13:33:16  Unable to do partial parsing because profile has changed
13:33:16  Unable to do partial parsing because env vars used in profiles.yml have changed
```

2. The user had to explicitly provide a `partial_parse.msgpack` file in the original project folder for their Airflow deployment - and if, for any reason, this became outdated, the user would not leverage the partial parsing feature. Since Cosmos runs dbt tasks from within a temporary directory, the partial parse would be stale for some users, it would be updated in the temporary directory, but the next time the task was run, Cosmos/dbt would not leverage the recently updated `partial_parse.msgpack`  file.

The current PR addresses these two issues respectfully by:

1. Allowing users that want to leverage Cosmos `ProfileMapping` and partial parsing to use `RenderConfig(enable_mock_profile=False)`

2. Introducing a Cosmos cache directory where we are persisting partial parsing files. This feature is enabled by default, but users can opt out by setting the Airflow configuration `[cosmos][enable_cache] = False` (exporting the environment variable `AIRFLOW__COSMOS__ENABLE_CACHE=0`).  Users can also define the temporary directory used to store these files using the `[cosmos][cache_dir]` Airflow configuration. By default, Cosmos will create and use a folder `cosmos` inside the system's temporary directory: https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir .

This PR affects both DAG parsing and task execution. Although it does not introduce an optimisation per se, it makes the partial parse feature implemented #800 available to more users.

Closes: #722

I updated the documentation in the PR: #898

Some future steps related to optimization associated to caching to be addressed in separate PRs:
i. Change how we create mocked profiles, to create the file itself in the same way, referencing an environment variable with the same name - and only changing the value of the environment variable (#924)
ii. Extend caching to the `profiles.yml` created by Cosmos in the newly introduced `tmp/cosmos` without the need to recreate it every time (#925).
iii. Extend caching to the Airflow DAG/Task group as a pickle file - this approach is more generic and would work for every type of DAG parsing and executor. (#926)
iv. Support persisting/fetching the cache from remote storage so we don't have to replicate it for every Airflow scheduler and worker node. (#927)
v. Cache dbt deps lock file/avoid installing dbt steps every time. We can leverage `package-lock.yml` introduced in dbt t 1.7 (https://docs.getdbt.com/reference/commands/deps#predictable-package-installs), but ideally, we'd have a strategy to support older versions of dbt as well. (#930)
vi. Support caching `partial_parse.msgpack` even when vars change: https://medium.com/@sebastian.daum89/how-to-speed-up-single-dbt-invocations-when-using-changing-dbt-variables-b9d91ce3fb0d
vii. Support partial parsing in Docker and Kubernetes Cosmos executors (#929)
viii. Centralise all the Airflow-based config into Cosmos settings.py & create a dedicated docs page containing information about these (#928)

**How to validate this change**

Run the performance benchmark against this and the `main` branch, checking the value of `/tmp/performance_results.txt`.

Example of commands run locally:

```
# Setup
AIRFLOW_HOME=`pwd` AIRFLOW_CONN_AIRFLOW_DB="postgres://postgres:postgres@0.0.0.0:5432/postgres" PYTHONPATH=`pwd` AIRFLOW_HOME=`pwd` AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT=20000 AIRFLOW__CORE__DAG_FILE_PROCESSOR_TIMEOUT=20000 hatch run tests.py3.11-2.7:test-performance-setup

# Run test for 100 dbt models per DAG:
MODEL_COUNT=100 AIRFLOW_HOME=`pwd` AIRFLOW_CONN_AIRFLOW_DB="postgres://postgres:postgres@0.0.0.0:5432/postgres" PYTHONPATH=`pwd` AIRFLOW_HOME=`pwd` AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT=20000 AIRFLOW__CORE__DAG_FILE_PROCESSOR_TIMEOUT=20000 hatch run tests.py3.11-2.7:test-performance
```

An example of output when running 100 with the main branch:
```
NUM_MODELS=100
TIME=114.18614888191223
MODELS_PER_SECOND=0.8757629623135543
DBT_VERSION=1.7.13
```

And with the current PR:
```
NUM_MODELS=100
TIME=75.17766404151917
MODELS_PER_SECOND=1.33018232576064
DBT_VERSION=1.7.13
```


cosmos/cache.py
```diff
@@ -0,0 +1,97 @@
...
+import shutil
+from enum import Enum
+from pathlib import Path
+
+from cosmos import settings
+from cosmos.constants import DBT_PARTIAL_PARSE_FILE_NAME, DBT_TARGET_DIR_NAME
+from cosmos.dbt.project import get_partial_parse_path
+
+
+def create_cache_identifier(dag_name: str, task_group_name: str = "") -> str:
```
ashb, line 12: Need on the usage this should be

```suggestion
def create_cache_identifier(dag_name: str, task_group_name: str | None = None) -> str:
```

jlaneve:
do we want to cache this locally or to s3/cloud store? it'd be really cool to do to blob store as well, especially for KE, etc

kaxil:
> do we want to cache this locally or to s3/cloud store? it'd be really cool to do to blob store as well, especially for KE, etc

Worth using [`ObjectStore`](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/objectstorage.html) for the newer Airflow version too, that way we can support more Cloud Stores.

And not great, but a "hacky" approach can be to see if users have a "XCom backend" configured, and if yes, you can store the cache in XCom. From 2.9, you can also use ObjectStore as XCom backend: https://airflow.apache.org/blog/airflow-2.9.0/#object-storage-as-xcom-backend

tatiana:
Thanks for the feedback, @jlaneve and @kaxil !

Since the current PR already improved significantly the performance, I've created a separate ticket for us to support remote caching: #927, adding your suggestions!

cosmos/dbt/project.py
```diff
@@ -20,8 +20,15 @@ def create_symlinks(project_path: Path, tmp_dir: Path, ignore_dbt_packages: bool
...
 
 
+def get_partial_parse_path(project_dir_path: Path) -> Path:
+    """
+    Return the partial parse (partial_parse.msgpack) path for a given dbt project directory.
+    """
+    return project_dir_path / DBT_TARGET_DIR_NAME / DBT_PARTIAL_PARSE_FILE_NAME
+
+
 def copy_msgpack_for_partial_parse(project_path: Path, tmp_dir: Path) -> None:
```
pankajkoti, line 30: Is this method getting used in our source code anymore? I could not find references apart from getting used in a test. If it's only for test maybe could move it there? Or do we want to keep it here since it has been a public method?

cosmos/cache.py
```diff
@@ -0,0 +1,124 @@
...
+    age_cosmos_cached_partial_parse_filepath = get_timestamp(cosmos_cached_partial_parse_filepath)
+
+    if age_project_partial_parse and age_cosmos_cached_partial_parse_filepath:
+        if age_project_partial_parse > age_cosmos_cached_partial_parse_filepath:
+            return project_partial_parse_path
+        else:
+            return cosmos_cached_partial_parse_filepath
+    elif age_project_partial_parse:
+        return project_partial_parse_path
+    elif age_cosmos_cached_partial_parse_filepath:
```
pankajkoti, line 89: I am willing to understand when it can happen that the original project does not have the partial parse file but our cache can have it? Is it that the original project will have a parse file if a user has given in their project and otherwise no. And that when dbt parses the directory(with partial parsing on), it creates this partial parse in the project tmp dir and from there we copy it to our cache?

pankajkoti (approved):
LGTM. Some minor suggestions and questions inline for my understanding.

cosmos/cache.py
tatiana: That's a good question, @pankajkoti !

Cosmos users are not guaranteed to make the `partial_parse.msgpack` available to the Airflow scheduler/worker nodes.

This file is just generated after a user runs a dbt command, parses the project, and it is stored in the `target` folder.

Since Cosmos never runs `dbt` commands from within the original project folder, unless the user explicitly had made this file part of their deployment, the original project will not contain it.

cosmos/dbt/project.py
tatiana: Well spotted, its removed in 6401096e916c53f208a965f7dd7700a9d3c45bcc

dev/dags/basic_cosmos_task_group.py
```diff
@@ -25,7 +25,10 @@
     ),
 )
 
-shared_execution_config = ExecutionConfig(invocation_mode=InvocationMode.DBT_RUNNER)
+shared_execution_config = ExecutionConfig(
+    invocation_mode=InvocationMode.SUBPROCESS,
+    # invocation_mode=InvocationMode.DBT_RUNNER
```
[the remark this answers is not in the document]
tatiana: Thanks for pointing this out - I realised no other example DAG using `InvocationMode.DBT_RUNNER`, so I just removed this comment and it's now being used in another example DAG, as part of 6401096e916c53f208a965f7dd7700a9d3c45bcc

cosmos/settings.py
```diff
@@ -0,0 +1,11 @@
...
+from pathlib import Path
+
+from airflow.configuration import conf
+
+from cosmos.constants import DEFAULT_COSMOS_CACHE_DIR_NAME
+
+# In MacOS users may want to set the envvar `TMPDIR` if they do not want the value of the temp directory to change
+DEFAULT_CACHE_DIR = Path(tempfile.gettempdir(), DEFAULT_COSMOS_CACHE_DIR_NAME)
+cache_dir = Path(conf.get("cosmos", "cache_dir", fallback=DEFAULT_CACHE_DIR) or DEFAULT_CACHE_DIR)
+enable_cache = conf.get("cosmos", "enable_cache", fallback=True) or True
```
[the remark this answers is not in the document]
tatiana: Changed in 7424a7f

cosmos/settings.py
```diff
@@ -0,0 +1,11 @@
+import tempfile
+from pathlib import Path
+
+from airflow.configuration import conf
+
+from cosmos.constants import DEFAULT_COSMOS_CACHE_DIR_NAME
+
+# In MacOS users may want to set the envvar `TMPDIR` if they do not want the value of the temp directory to change
+DEFAULT_CACHE_DIR = Path(tempfile.gettempdir(), DEFAULT_COSMOS_CACHE_DIR_NAME)
+cache_dir = Path(conf.get("cosmos", "cache_dir", fallback=DEFAULT_CACHE_DIR) or DEFAULT_CACHE_DIR)
```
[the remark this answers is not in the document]
tatiana: That's how I initially implemented it, but mypy raises:
```
- hook id: mypy
- exit code: 1

cosmos/settings.py:10: error: Argument 1 to "Path" has incompatible type "str | None"; expected "str | PathLike[str]"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

Do you have any suggestions?

cosmos/cache.py
```diff
@@ -0,0 +1,124 @@
...
+    :param dag_name: Name of the Cosmos DbtDag being cached
+    :param task_group_name: (optional) Name of the Cosmos DbtTaskGroup being cached
+    :return: Unique identifier representing the cache
+    """
+    if task_group:
+        if task_group.dag_id is not None:
+            cache_identifiers_list = [task_group.dag_id]
+        if task_group.group_id is not None:
+            cache_identifiers_list.extend([task_group.group_id.replace(".", "_")])
+        cache_identifier = "_".join(cache_identifiers_list)
```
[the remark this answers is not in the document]
tatiana: Addressed in 2a3f00742fa3b228a95b4293fde047e63336f31f

tatiana:
Thanks a lot for the feedback, @pankajkoti ! I addressed all of them except for the minor https://github.com/astronomer/astronomer-cosmos/pull/904#discussion_r1586358258, which can be improved in a follow-up ticket once you reply. Since you approved the PR, I'll mark this comment as resolved and merging .

maslick:
Enabling caching can quickly fill up the 30GB ephemeral storage on MWAA workers if there are 500 or more DAGs.

Issue #909, opened 2024-04-09
Testing behavior docs outdated regarding supported modes for on_warning_callback feature


mc51:
The testing behavior docs are outdated. [They say ](https://github.com/astronomer/astronomer-cosmos/blob/7c4d7d6e0c680a4847414ed4eb525b7baa27c8c1/docs/configuration/testing-behavior.rst?plain=1#L40 )that only the `local` executor mode is supporting the `on_warning_callback` feature: However, it also works with the `virtualenv` mode. The reason for this seems to bet that `virtualenv` mode didn't exist when the warning feature was implemented (here's the original feature [PR ](https://github.com/astronomer/astronomer-cosmos/pull/225)).

It would be helpful to fix this in the docs: We wanted to implement this functionality and gave up after checking the docs. But after looking at the source ourselves, we figured out that it actually also works with `virtualenv` execution mode. We then implemented and tested the functionality successfully.

pankajastro:
Hey @mc51 would you like to submit a PR for this?

pankajastro:
Looks like docs has been updated here https://astronomer.github.io/astronomer-cosmos/configuration/testing-behavior.html#warning-behavior

PR #910, opened 2024-04-09
update testing behavior docs

mc51:
## Description

Update testing behavior docs to reflect support of warning feature in virtualenv execution mode. Currently, the docs say that only `local` executor mode is supporting `on_warning_callback`. However, this also works with `virtualenv` mode.

## Related Issue(s)

#909 

## Breaking Change?

None

## Checklist

- [x ] I have made corresponding changes to the documentation (if required)

### DIFF
--- a/docs/configuration/testing-behavior.rst
+++ b/docs/configuration/testing-behavior.rst
@@ -37,7 +37,7 @@ Warning Behavior
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


tatiana (approved):

tatiana:
Thanks for the docs improvement, @mc51 !

Issue #915, opened 2024-04-18
on DbtTaskGroup creation dag parameter is ignored.


tboutaour:
Currently, there is an issue when attempting to create a `DbtTaskGroup` using the dag parameter directly. However, this limitation can be bypassed using the with statement, where there is no issue.

Functional example:

```
with DAG(...):
    DbtTaskGroup(...) 
```

Problematic example:
```
dag = DAG(...):
DbtTaskGroup(dag=dag...) 
```

This problem stems from the `airflow_kwargs` function within `converter.py`. Although the `dag` argument is assumed to be present in `DbtToAirflowConverter`, this parameter is not being passed to the creation of `TaskGroup`.

I suggest two potential solutions to the problem:

1. Adjusting the `DbtTaskGroup` operator to explicitly pass the `dag` parameter when creating the task group.
2. Adjusting the behavior of the `airflow_kwargs` function to ensure that the dag parameter is properly handled when creating the TaskGroup.

PR #916, opened 2024-04-18
Fix instantiating `DbtTaskGroup` with argument `dag` (outside of `DAG` context)

tboutaour:
## Description

This code fixes the problem that dag parameter cannot be passed as argument at `DbtTaskGroup`.
This change has been made in order to not affecting DbtToAirflowConverter class at all.

## Related Issue(s)

This PR fixes bug at #915 

## Breaking Change?
There is not breaking changes at this PR.

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works (There is any test for task_group file.)


tatiana (commented):
This looks great, @tboutaour , thank you very much for the contribution!

Could you add a test to illustrate the original problem, so we avoid regressions in future?

tatiana (approved):
Thanks for adding the test, @tboutaour !

PR #917, opened 2024-04-20
Fix: Typo

SiddiqueAhmad:
it is `has_test` property instead of `has_text`


jlaneve (approved):
thank you @SiddiqueAhmad!

PR #921, opened 2024-04-24
fix: missed required parameter for non method authentication

AlexandrKhabarov:
## Description

Usage of TrinoBaseProfileMapping leads to the issue related with missed user field in profile with authentication method equals to None
```
Runtime Error
  Credentials in profile "test_profile", target "dev" invalid: 'user' is a required property
```

## Breaking Change?

No breaking changes

## Checklist


tatiana (approved):
Thanks for fixing this, @AlexandrKhabarov ! We'll release this as part of Cosmos 1.4

PR #923, opened 2024-04-26
Fixed getting values from extra__ keys in airflow con

glebkrapivin:
## Checklist


tatiana (commented):
Thanks for fixing this, @glebkrapivin !

It seems one of the pre-commit checks is failing, could you fix it, please? We'll aim to release this as part of 1.4
```
ruff..........................................................................Failed
- hook id: ruff
- exit code: 1

cosmos/profiles/base.py:244:9: C901 `get_dbt_value` is too complex (9 > 8)
Found 1 error.
```

glebkrapivin:
> Thanks for fixing this, @glebkrapivin !
> 
> It seems one of the pre-commit checks is failing, could you fix it, please? We'll aim to release this as part of 1.4
> 
> ```
> ruff..........................................................................Failed
> - hook id: ruff
> - exit code: 1
> 
> cosmos/profiles/base.py:244:9: C901 `get_dbt_value` is too complex (9 > 8)
> Found 1 error.
> ```

@tatiana 
Personally i would ignore that as i think branching for different `extras` increases readability in this case, but i will think of a way to refactor for it to pass

glebkrapivin:
> > Personally i would ignore that as i think branching for different extras increases readability in this case, but i will think of a way to refactor for it to pass
> 
> Thanks a lot, @glebkrapivin !

@tatiana  I've changed a couple of lines, now all the pre-commit checks do pass :)

tatiana (approved):
Thank you very much, @glebkrapivin !

PR #933, opened 2024-05-01
Ensure tags don't run into index errors when there are no upstream nodes

jlaneve:
## Description

This PR ensures we only try to access `node.depends_on[0]` if there's at least one element in the list.

## Breaking Change?

No

## Checklist

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -296,7 +296,7 @@ class NodeSelector:
 
         self.visited_nodes.add(node_id)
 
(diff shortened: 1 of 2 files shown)


tatiana (commented):
Hi @jlaneve , thanks a lot for the fix!

Would you have bandwidth for adding a test, or should someone else from the team wrap this up?

tatiana (approved):

tatiana:
I just added the missing test - and we'll be releasing this as part of 1.4

Issue #936, opened 2024-05-02
Bug: `ModuleNotFoundError: No module named 'pydantic'`


tatiana:
**Steps to reproduce**

From an Airflow project folder that contains at least one DAG inheriting from `DbtDag`:

```
pip install astronomer-cosmos==1.4.0a2
airflow db init
```

We see the error:
```
ERROR [airflow.models.dagbag.DagBag] Failed to import: /Users/pankajkoti/astronomer/astronomer-cosmos/dags/example_cosmos_python_models.py
Traceback (most recent call last):
  File "/Users/pankajkoti/astronomer/astronomer-cosmos/cosmos/lib/python3.8/site-packages/airflow/models/dagbag.py", line 346, in parse
    loader.exec_module(new_module)
  File "<frozen importlib._bootstrap_external>", line 843, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/pankajkoti/astronomer/astronomer-cosmos/dags/example_cosmos_python_models.py", line 20, in <module>
    from cosmos import DbtDag, ProfileConfig, ProjectConfig
  File "/Users/pankajkoti/astronomer/astronomer-cosmos/cosmos/lib/python3.8/site-packages/cosmos/__init__.py", line 11, in <module>
    from cosmos.airflow.dag import DbtDag
  File "/Users/pankajkoti/astronomer/astronomer-cosmos/cosmos/lib/python3.8/site-packages/cosmos/airflow/dag.py", line 11, in <module>
    from cosmos.converter import DbtToAirflowConverter, airflow_kwargs, specific_kwargs
  File "/Users/pankajkoti/astronomer/astronomer-cosmos/cosmos/lib/python3.8/site-packages/cosmos/converter.py", line 15, in <module>
    from cosmos.airflow.graph import build_airflow_graph
  File "/Users/pankajkoti/astronomer/astronomer-cosmos/cosmos/lib/python3.8/site-packages/cosmos/airflow/graph.py", line 9, in <module>
    from cosmos.config import RenderConfig
  File "/Users/pankajkoti/astronomer/astronomer-cosmos/cosmos/lib/python3.8/site-packages/cosmos/config.py", line 24, in <module>
    from cosmos.profiles import BaseProfileMapping
  File "/Users/pankajkoti/astronomer/astronomer-cosmos/cosmos/lib/python3.8/site-packages/cosmos/profiles/__init__.py", line 7, in <module>
[...]

PR #736, opened 2023-12-02
Optional pydantic

pixie79:
## Description

Converts pydantic to being optional, as it is required for Airflow 2.6 but not others.

## Related Issue(s)

Closes #725 
Closes #654 

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #939, opened 2024-05-03
Make Pydantic a required dependency

pankajkoti:
The pull request #736 initially designated Pydantic as an 
optional dependency. However, a subsequent pull request, 
#794 as an [import](https://github.com/astronomer/astronomer-cosmos/pull/794/files#diff-bd3fa47d7a9b96d7bb365f3ba3b60eaf0b20e06a48b28814cf6f6e6fb64d4da6R12), introduced an implementation that requires Pydantic 
to be a mandatory dependency for astronomer-cosmos. 
This pull request addresses the bug encountered while 
running `airflow db init`, as described in #936, by enforcing 
Pydantic as a required dependency.

closes: #936

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,11 @@ dependencies = [
     "apache-airflow>=2.3.0",
     "importlib-metadata; python_version < '3.8'",
     "Jinja2>=3.0.0",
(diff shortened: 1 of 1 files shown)


tatiana (approved):

PR #944, opened 2024-05-07
Support for running dbt tasks in AWS EKS

VolkerSchiewe:
## Description
We are using MWAA in combination with EKS so that all our dags in airflow are running in our EKS. We would like to use the same setup with cosmos.

### What changes?
- New AwsEksOperator classes (inheriting from KubernetesOperators) - Based on the original [EksOperator](https://github.com/apache/airflow/blob/main/airflow/providers/amazon/aws/operators/eks.py#L995)
- Tests
- Adjusted documentation

## Related Issue(s)

-

## Breaking Change?
No - only an additional feature

## Checklist

### DIFF
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -3,6 +3,7 @@ from enum import Enum
 from pathlib import Path
 
 import aenum
+from packaging.version import Version
 
 DBT_PROFILE_PATH = Path(os.path.expanduser("~")).joinpath(".dbt/profiles.yml")
 DEFAULT_DBT_PROFILE_NAME = "cosmos_profile"
@@ -19,6 +20,10 @@ DBT_BINARY_NAME = "dbt"
 
 DEFAULT_OPENLINEAGE_NAMESPACE = "cosmos"
 OPENLINEAGE_PRODUCER = "https://github.com/astronomer/astronomer-cosmos/"
+
+# Cosmos will not emit datasets for the following Airflow versions, due to a breaking change that's fixed in later Airflow 2.x versions
+# https://github.com/apache/airflow/issues/39486
+PARTIALLY_SUPPORTED_AIRFLOW_VERSIONS = [Version("2.9.0"), Version("2.9.1")]
 
 
 class LoadMode(Enum):
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -51,6 +51,7 @@ class ExecutionMode(Enum):
     LOCAL = "local"
     DOCKER = "docker"
     KUBERNETES = "kubernetes"
+    EKS = "eks"
     VIRTUALENV = "virtualenv"
     AZURE_CONTAINER_INSTANCE = "azure_container_instance"
 
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -56,7 +56,7 @@ class ExecutionMode(Enum):
     LOCAL = "local"
     DOCKER = "docker"
     KUBERNETES = "kubernetes"
-    EKS = "eks"
+    AWS_EKS = "aws_eks"
     VIRTUALENV = "virtualenv"
     AZURE_CONTAINER_INSTANCE = "azure_container_instance"
 
--- a/cosmos/operators/eks.py
+++ b/cosmos/operators/eks.py
@@ -20,7 +20,6 @@ from cosmos.operators.kubernetes import (
 try:
     from airflow.providers.amazon.aws.hooks.eks import EksHook
 except ImportError as e:
-    breakpoint()
     raise ImportError("EksOperator is not compatible with airflow < 2.4") from e
 
 DEFAULT_CONN_ID = "aws_default"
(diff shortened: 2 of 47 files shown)


VolkerSchiewe:
> Thanks for the contribution, @VolkerSchiewe ! Please, could you address the tests currently failing?

Hi @tatiana thanks for getting back to me! I was already working on it, but I ran into this issue: https://github.com/apache/airflow/issues/39103

Seems to be affecting the latest version of the amazon airflow provider. I already tried with pinning an older version of `xmlsec1` with no success :/

tatiana:
Hi @VolkerSchiewe , just a head's up: we're wrapping up the Cosmos 1.4 release and either myself or @pankajkoti will support you on this next week!

pankajkoti:
It appears that earlier versions of Flask is [not compatible](https://stackoverflow.com/questions/71718167/importerror-cannot-import-name-escape-from-jinja2/71735103#71735103) with newer versions of Jinja2. 

[These](https://raw.githubusercontent.com/apache/airflow/constraints-2.3.0/constraints-3.8.txt) are the constraints for Airflow 2.3 which suggests Flask 1.1.2 and Jinja2 3.0.3, however dbt-core is upgrading it to jinja2-3.1.4 (Jinja2<4,>=3.1.3 (from dbt-core)) which is not supported on Flask 1.1.2 that comes with Airflow.

VolkerSchiewe:
Hey there, the only failing parts are the tests with Airflow 2.3 and python 3.9. For some reason the install step fails with 
```
pip._vendor.resolvelib.resolvers.ResolutionTooDeep: 200000
```
when installing the test requirements . I don't really understand why and how to solve this. Is this problem you ran into earlier already?

VolkerSchiewe:
Fixing boto3 in the runs with airflow 2.3 solved the problem 🚀

cosmos/constants.py
```diff
@@ -56,6 +56,7 @@ class ExecutionMode(Enum):
     LOCAL = "local"
     DOCKER = "docker"
     KUBERNETES = "kubernetes"
+    EKS = "eks"
```
tatiana, line 59:
```suggestion
    AWS_EKS = "aws_eks"
```

cosmos/operators/eks.py
```diff
@@ -0,0 +1,135 @@
...
+    DbtRunOperationKubernetesOperator,
+    DbtSeedKubernetesOperator,
+    DbtSnapshotKubernetesOperator,
+    DbtTestKubernetesOperator,
+)
+
+try:
+    from airflow.providers.amazon.aws.hooks.eks import EksHook
+except ImportError as e:
+    raise ImportError("EksOperator is not compatible with airflow < 2.4") from e
```
pankajkoti, line 22:
```suggestion
from airflow.providers.amazon.aws.hooks.eks import EksHook
```

pankajkoti: hi @VolkerSchiewe , we're planning to drop support for Airflow 2.3 in the upcoming Cosmos 1.5 release. The PR is already merged to main. So if you rebase your PR and get the latest changes, the CI would no longer run tests on Airflow 2.3 and then we don't need this Import handling.

pyproject.toml
```diff
@@ -138,6 +142,7 @@ airflow = ["2.3", "2.4", "2.5", "2.6", "2.7", "2.8", "2.9"]
 [tool.hatch.envs.tests.overrides]
 matrix.airflow.dependencies = [
     { value = "typing_extensions<4.6", if = ["2.6"] },
+    { value = "boto3==1.22.0", if = ["2.3"] },
```
pankajkoti, line 145:
```suggestion
```

Same as above as we're dropping support for Airflow 2.3

pankajkoti (commented):
This is a great contribution @VolkerSchiewe . I think we're close and soon can merge this. 

Would it be possible to also paste a snapshot of a successful DAG run using AWS EKS operator?

cosmos/operators/eks.py
VolkerSchiewe: Very nice, love to remove the 2.3 related code 👍

cosmos/constants.py
VolkerSchiewe: Renamed everything 👍

pankajkoti (approved):
LGTM. Appreciate your contribution @VolkerSchiewe 👏🏽

tatiana (approved):
This looks great, thanks a lot for iterating and improving it so quickly, @VolkerSchiewe !

docs/getting_started/execution-modes.rst
```diff
@@ -159,6 +164,38 @@ Example DAG:
             "secrets": [postgres_password_secret],
         },
     )
+AWS_EKS
+----------
+
+The ``aws_eks`` approach is very similar to the ``kubernetes`` approach, but it is specifically designed to run on AWS EKS clusters.
+It uses the `EKSPodOperator <https://airflow.apache.org/docs/apache-airflow-providers-amazon/2.2.0/operators/eks.html#perform-a-task-on-an-amazon-eks-cluster>`_
```
pankajastro, line 171:
```suggestion
It uses the `EKSPodOperator <https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/operators/eks.html#perform-a-task-on-an-amazon-eks-cluster>`_
```

pankajastro (approved):
looks great 👍

docs/getting_started/execution-modes.rst
pankajkoti, line 171:
```suggestion
It uses the `EKSPodOperator <https://airflow.apache.org/docs/apache-airflow-providers-amazon/8.19.0/operators/eks.html#perform-a-task-on-an-amazon-eks-cluster>`_
```

pankajkoti: Stable would point to latest and it may change. I think better to tag it to a specific version and also would denote that it works fine according to that version. Since we don't allow latest versions yet, I think 8.19.0 is a good candidate

pankajkoti: I think and let the CI run and complete since it takes time and then we can push this change.

Issue #937, opened 2024-05-02
Bug when using `enable_mock_profile=False` in `simple_dag.py` with 1.4.0a2


tatiana:
**Steps to reproduce**

Change the [simple_dag.py](https://github.com/astronomer/cosmos-demo/blob/main/dags/basic/simple_dag.py) to use `render_config=RenderConfig(enable_mock_profile=False)` and try to run it using `astronomer-cosmos==1.4.0a2`.

This raised the exception:
```
[2024-05-02, 14:32:47 UTC] {taskinstance.py:1824} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/cosmos/operators/base.py", line 266, in execute
    self.build_and_run_cmd(context=context, cmd_flags=self.add_cmd_flags())
  File "/usr/local/lib/python3.10/site-packages/cosmos/operators/local.py", line 469, in build_and_run_cmd
    result = self.run_command(cmd=dbt_cmd, env=env, context=context)
  File "/usr/local/lib/python3.10/site-packages/cosmos/operators/local.py", line 354, in run_command
    self.handle_exception(result)
  File "/usr/local/lib/python3.10/site-packages/cosmos/operators/local.py", line 179, in handle_exception_subprocess
    raise AirflowException(
[...]

PR #950, opened 2024-05-10
Correct stale `root_path` in partial parse file

pankajkoti:
With the introduction of enabling partial parse in PR #904, 
upon testing the implementation, it is observed that the seeds 
files were not been able to be located as the partial parse file 
contained a stale `root_path` from previous command runs. 
This issue is observed on specific earlier versions of dbt-core like
`1.5.4` and `1.6.5`, but not on recent versions of dbt-core `1.5.8`, `1.6.6`
and `1.7.0`. I am suspecting that PR https://github.com/dbt-labs/dbt-core/pull/8762
is likely the fix and the fix appears to be backported to later version 
releases of `1.5.x` and `1.6.x`.

However, irrespective of the dbt-core version, this PR attempts to 
correct the `root_path` in the partial parse file by replacing it with 
the needed project directory where the project files are located. 
And thus ensures that the feature runs correctly for older and 
newer versions of dbt-core.

closes: #937

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,278 +19,278 @@ jobs:
     steps:
       - run: true
 
-  Type-Check:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          ref: ${{ github.event.pull_request.head.sha || github.ref }}
-
-      - uses: actions/setup-python@v3
-        with:
-          python-version: "3.9"
-          architecture: "x64"
-
-      - run: pip3 install hatch
-      - run: hatch run tests.py3.9-2.7:type-check
-
-  Run-Unit-Tests:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
-        airflow-version: ["2.3", "2.4", "2.5", "2.6", "2.7", "2.8", "2.9"]
-        exclude:
-          - python-version: "3.11"
-            airflow-version: "2.3"
-          - python-version: "3.11"
-            airflow-version: "2.4"
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          ref: ${{ github.event.pull_request.head.sha || github.ref }}
-
-      - uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cache/pip
-            .nox
-          key: unit-${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.airflow-version }}-${{ hashFiles('pyproject.toml') }}-${{ hashFiles('cosmos/__init__.py') }}
-
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
-        with:
-          python-version: ${{ matrix.python-version }}
-
-      - name: Install packages and dependencies
-        run: |
-          python -m pip install hatch
-          hatch -e tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }} run pip freeze
-
-      - name: Test Cosmos against Airflow ${{ matrix.airflow-version }} and Python ${{ matrix.python-version }}
-        run: |
-          hatch run tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }}:test-cov
-
-      - name: Upload coverage to Github
-        uses: actions/upload-artifact@v2
-        with:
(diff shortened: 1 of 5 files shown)


cosmos/cache.py
```diff
@@ -121,4 +122,18 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
     source_manifest_filepath = partial_parse_filepath.parent / DBT_MANIFEST_FILE_NAME
     target_manifest_filepath = target_partial_parse_file.parent / DBT_MANIFEST_FILE_NAME
     shutil.copy(str(partial_parse_filepath), str(target_partial_parse_file))
+
+    # Update root_path in partial parse file to point to the target project directory. This is necessary because in some
```
pankajkoti, line 126:
```suggestion
    # Update root_path in partial parse file to point to the needed project directory. This is necessary because in some
```

cosmos/cache.py
```diff
@@ -121,4 +122,18 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
     source_manifest_filepath = partial_parse_filepath.parent / DBT_MANIFEST_FILE_NAME
     target_manifest_filepath = target_partial_parse_file.parent / DBT_MANIFEST_FILE_NAME
     shutil.copy(str(partial_parse_filepath), str(target_partial_parse_file))
+
+    # Update root_path in partial parse file to point to the needed project directory. This is necessary because in some
+    # earlier versions of dbt (e.g. 1.5.4), the root_path was hardcoded to a stale directory and is not updated to the
+    # needed target directory. This seems to have been resolved in later versions of dbt, but we still need to handle
```
pankajkoti, line 128:
```suggestion
    # needed project directory. This seems to have been resolved in later versions of dbt, but we still need to handle
```

cosmos/cache.py
```diff
@@ -121,4 +122,18 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
     source_manifest_filepath = partial_parse_filepath.parent / DBT_MANIFEST_FILE_NAME
     target_manifest_filepath = target_partial_parse_file.parent / DBT_MANIFEST_FILE_NAME
     shutil.copy(str(partial_parse_filepath), str(target_partial_parse_file))
+
+    # Update root_path in partial parse file to point to the needed project directory. This is necessary because in some
+    # earlier versions of dbt (e.g. 1.5.4), the root_path was hardcoded to a stale directory and is not updated to the
+    # needed project directory. This seems to have been resolved in later versions of dbt, but we still need to handle
```
tatiana, line 128: It's worth saying explicitly this was not needed in dbt 1.7.4 and 1.80

tatiana (commented):
This is lookign great, @pankajkoti ! Probably worth to add a comment in the PR description that we observed this issue only for specific versions of dbt (1.5.4) and that they were not observed in dbt 1.7.4 onwards.

pyproject.toml
```diff
@@ -164,13 +166,14 @@ markers = ["integration", "sqlite", "perf"]
 [tool.hatch.envs.docs]
 dependencies = [
     "aenum",
-    "sphinx",
-    "pydata-sphinx-theme",
-    "sphinx-autobuild",
-    "sphinx-autoapi",
-    "openlineage-airflow",
     "apache-airflow-providers-cncf-kubernetes>=5.1.1",
+    "msgpack",
```
tatiana, line 170: We'll also need to add this to `docs/requirements.txt `

cosmos/cache.py
pankajkoti: yes, added some more information.

pankajkoti:
> This is lookign great, @pankajkoti ! Probably worth to add a comment in the PR description that we observed this issue only for specific versions of dbt (1.5.4) and that they were not observed in dbt 1.7.4 onwards.

added some more description. please check.

pyproject.toml
pankajkoti: Thanks a lot! That helped!

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, correct-project-dir-partial-parse-file]
```
```suggestion
    branches: [main]
```

pankajkoti: Changing this back to contain only main before merging the PR

tatiana (approved):
Thanks for fixing this, @pankajkoti ! We're close to the 1.4 release!

PR #954, opened 2024-05-13
Set default value for `append_env` based on execution modes

pankajkoti:
This PR sets the default value of dbt argument `append-env` that 
exposes the operating system environment variables to the 
`dbt` command based on execution modes.
For most execution modes, the default is set to `False`, 
however, for execution modes `ExecuteMode.LOCAL` and 
`ExecuteMode.VIRTUALENV`, the default is set to True. 
This behavior is consistent with the `LoadMode.DBT_LS` 
command in forwarding the environment variables to the 
subprocess by default.

### DIFF
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -46,7 +46,7 @@ class AbstractDbtBaseOperator(BaseOperator, metaclass=ABCMeta):
         environment variables for the new process; these are used instead
         of inheriting the current process environment, which is the default
         behavior. (templated)
-    :param append_env: . If True, inherits the environment variables
+    :param append_env: If True, inherits the environment variables
         from current process and then environment variable passed by the user will either update the existing
         inherited environment variables or the new variables gets appended to it.
         If False (default), only uses the environment variables passed in env params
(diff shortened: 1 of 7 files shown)


cosmos/operators/base.py
```diff
@@ -46,10 +46,10 @@ class AbstractDbtBaseOperator(BaseOperator, metaclass=ABCMeta):
         environment variables for the new process; these are used instead
         of inheriting the current process environment, which is the default
         behavior. (templated)
-    :param append_env: . If True (default), inherits the environment variables
-        from current passes and then environment variable passed by the user will either update the existing
+    :param append_env: . If True, inherits the environment variables
+        from current process and then environment variable passed by the user will either update the existing
```
pankajastro, line 50:
```suggestion
    :param append_env: If True, inherits the environment variables
        from current process and then environment variable passed by the user will either update the existing
```

cosmos/operators/local.py
```diff
@@ -113,6 +113,11 @@ class DbtLocalBaseOperator(AbstractDbtBaseOperator):
     :param target_name: A name to use for the dbt target. If not provided, and no target is found
         in your project's dbt_project.yml, "cosmos_target" is used.
     :param should_store_compiled_sql: If true, store the compiled SQL in the compiled_sql rendered template.
+    :param append_env: . If True(default), inherits the environment variables
```
pankajastro, line 116:
```suggestion
    :param append_env: If True(default), inherits the environment variables
```

pankajastro (approved):

tatiana (approved):
This looks great, thanks for the quick fix, @pankajkoti

Issue #952, opened 2024-05-10
Fix failing pre-commit hook `mypy-python`


tatiana:
**Context**

If we run 
```
$ pre-commit run --all-files
```

We currently get:
```
Check if .env file is empty...................................................Passed
check for added large files...................................................Passed
check for merge conflicts.....................................................Passed
check toml....................................................................Passed
check yaml....................................................................Passed
debug statements (python).....................................................Passed
fix end of files..............................................................Passed
mixed line ending.............................................................Passed
pretty format json............................................................Passed
trim trailing whitespace......................................................Passed
Run codespell to check for common misspellings in files.......................Passed
rst ``code`` is two backticks.................................................Passed
check for not-real mock methods...............................................Passed
CRLF end-lines remover........................................................Passed
Tabs remover..................................................................Passed
pyupgrade.....................................................................Passed
ruff..........................................................................Passed
black.........................................................................Passed
blacken-docs..................................................................Passed
mypy-python...................................................................Failed
- hook id: mypy
- exit code: 1

cosmos/profiles/base.py:45: error: "DbtProfileConfigVars" has no attribute "__dataclass_fields__"  [attr-defined]
cosmos/operators/local.py:228: error: Item "None" of "BaseOperator | MappedOperator | None" has no attribute "template_fields"  [union-attr]
Found 2 errors in 2 files (checked 65 source files)
```

**Acceptance criteria**

* [ ] Pre commit checks should pass
* [ ] Understand/fix why this check is being skipped in the CI

PR #957, opened 2024-05-14
Enable pre-commit run and fix type-check job

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/952

This PR addresses the following issues:

**1. Pre-commit Setup:**
- Pre-commit was not running in the CI pipeline because it was neither configured for this repository nor were we using the pre-commit command in CI to run it manually. To fix this:
Installed and added pre-commit to this repository. This allows pre-commit to automatically run the pre-commit configuration on every push.

**2. Type Check Job:**
- Previously, the type-check job was running with mypy directly, and the pre-commit type check command had different configurations. This PR aligns the configurations.

**3. Type Check Fixes:**
- Ignored the type check for __dataclass_fields__ since mypy seems to not recognize it.
- Added a check for None for ti.task

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -238,6 +238,8 @@ class DbtLocalBaseOperator(AbstractDbtBaseOperator):
         ti = context["ti"]
 
         if isinstance(ti, TaskInstance):  # verifies ti is a TaskInstance in order to access and use the "task" field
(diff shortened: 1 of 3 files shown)


tatiana (approved):
This is looking great, @pankajastro , thanks for solving this problem! 🎉

PR #968, opened 2024-05-15
Update CODEOWNERS

tatiana:
Update CODEOWNERS based on Astronomer organizational changes and new committer.


tatiana:
> wondering why we have 2 codeowners files?

I think the original motivation was for changes to the CI, that could incur a cost to Astronomer, to have to be reviewed by someone from Astronomer

pankajkoti (approved):

PR #969, opened 2024-05-15
Clean databricks credentials in test/CI

tatiana:
We're changing the credentials for the Databricks cluster used to validate dbt Python models, and I noticed there were a few unnecessary credentials set in the CI. This PR aims to clean them up, making it easier for us to manage them.

All the tests passed in: https://github.com/astronomer/astronomer-cosmos/actions/runs/9109622858

The push tests failed because they relied on workflow configuration from the `main` branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -129,6 +129,9 @@ jobs:
         env:
           AIRFLOW_HOME: /home/runner/work/astronomer-cosmos/astronomer-cosmos/
           AIRFLOW_CONN_AIRFLOW_DB: postgres://postgres:postgres@0.0.0.0:5432/postgres
+          DATABRICKS_HOST: mock
+          DATABRICKS_WAREHOUSE_ID: mock
+          DATABRICKS_TOKEN: mock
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, clean-databricks-credentials]
```
tatiana, line 5: Before merging, we should rollback this change:
```suggestion
    branches: [main]
```

.github/workflows/test.yml
```diff
@@ -2,9 +2,9 @@ name: test
...
   pull_request_target: # Also run on pull requests originated from forks
-    branches: [main]
+    branches: [main, clean-databricks-credentials]
```
tatiana, line 7:
```suggestion
    branches: [main]
```

tatiana, line 7:
```suggestion
    branches: [main]
```

tatiana:
All the tests passed in the CI: https://github.com/astronomer/astronomer-cosmos/actions/runs/9097739441

The push tests failed because they relied on workflow configuration from the `main` branch.

PR #989, opened 2024-05-20
Update Dockerfile -- use latest Astro Runtime

RNHTTR:
## Checklist


pankajastro (approved):
We should automate it

Issue #963, opened 2024-05-15
Review usage of Cosmos with Airflow 2.3 & older version of dbt


tatiana:
**Context**

We are aware there are conflicts between [Airflow/dbt dependencies](https://astronomer.github.io/astronomer-cosmos/getting_started/execution-modes-local-conflicts.html#execution-modes-local-conflicts). And there is a cost involved with supporting older versions of Airflow, as seen in https://github.com/astronomer/astronomer-cosmos/pull/944.

**Acceptance criteria**

* [x] Check which versions of Airflow are being used by Astronomer customers who use Cosmos
* [x] Make a survey with the community to check which versions of dbt & versions of Airflow they are using
* [x] Share the survey with the community
* [ ] ~Make a decision about what is the oldest version of Airflow we want to support~ (the survey will be open for a while, we should close this ticket for now, and have a follow up task to check on this)

tatiana:
Initially I thought about doing a "/poll" in the #airflow-dbt channel. However, since it will disappear from Slack in 90 days, I decided to have this in a Google Form.

We already planned to have two questions, and this could be an opportunity to add more. I drafted a survey:
* https://docs.google.com/forms/d/e/1FAIpQLSdlFdh3UkZNlTNX2_7V0Wo0tZdVHMzGZsY3Q4j_0_e6JCzQhw/viewform
* https://docs.google.com/forms/d/1xYKsClnedr-e_d5BHEvVc0xsTOJgL7bfW6GlZM66gKs/edit

So far, I have shared with the following people for initial feedback:
* [ ] @cmarteepants
* [ ] @jlaneve 
* [ ] @kaxil 
* [ ] @pankajkoti 
* [ ] @pankajastro

tatiana:
I queried Cosmos task runs since 1 January 2024 by Astronomer Customers, and identified which versions of Airflow and Runtime (that maps to Python versions), they used:
* Airflow versions: 2.4.3 - 2.9.1
* Runtime versions: 6.2.1 - 11.3.0 (Python 3.8 - Python 3.12, [based on the docs](https://docs.astronomer.io/astro/runtime-image-architecture#python-versioning))

Based exclusively on this, we could drop Airflow 2.3 from Cosmos test matrix. I also checked the usage since 1 May 2024, and unfortunately Airflow 2.4.3 was still there.

tatiana:
Feedback from @cmarteepants:
- Open question: How relevant is to know job titles?
[...]

PR #994, opened 2024-05-22
Drop support for Airflow 2.3

pankajkoti:
Airflow 2.3 was released more than 2 years ago on Apr 30, 2022
and few of the apache-airflow providers are known to no longer
support it. We've observed cosmos users are not on Airflow 2.3 
based on the analysis in https://github.com/astronomer/astronomer-cosmos/issues/963#issuecomment-2121165722
and hence, to avoid maintenance & support efforts for an older version,
this PR drops Cosmos support for Airflow 2.3


tatiana (approved):
Looks good, @pankajkoti , thank you!

Issue #924, opened 2024-04-29
Change how the `profiles.yml` is created when using mocked profiles 


tatiana:
**Context**

As of Cosmos 1.3, when users use Cosmos `ProfileMapping` we create the `profiles.yml` file in two different ways:
- **functional**: mapping keys to environment variables (necessary to run dbt commands that need to access the database). Example:
  https://github.com/astronomer/astronomer-cosmos/blob/d1dbc4125154920021b6440899b10c657a8b6840/cosmos/profiles/postgres/user_pass.py#L39
- **mocked**: mapping keys to hard-coded fake values (to parse the dbt project when using `dbt ls` without the need to run dbt commands that reach the database). Example:
  https://github.com/astronomer/astronomer-cosmos/blob/d1dbc4125154920021b6440899b10c657a8b6840/cosmos/profiles/postgres/user_pass.py#L55
  https://github.com/astronomer/astronomer-cosmos/blob/d1dbc4125154920021b6440899b10c657a8b6840/cosmos/profiles/base.py#L174

The mocked approach was introduced so users could validate their DAGs in their CI without creating Airflow connections and without exposing database credentials.

The challenge with this approach is that, when using partial parsing, dbt expects the `profiles.yml` to remain unchanged, as observed during #904. On the positive side, dbt is happy to use partial parse if only the content of the environment variable changes (validated with dbt 1.7.10).

This PR  #904 worked around the issue by allowing users to leverage partial parsing to avoid creating mocked profiles (using `RenderConfig(enable_mock_profile=False)`). However, users must ensure the CI has the desired connection (potentially with fake values). The goal of this ticket is to overcome this issue.

**Acceptance criteria**

* [ ] Cosmos should create the dbt `profiles.yml` in the same way, referencing environment variables
* [ ] When using mock, Cosmos should use mocked values in the environment variables
* [ ] Users should still be able to parse the Cosmos DAG that uses ProfileMapping by running `airflow dags list` when no connection was created
* [ ] Cosmos users should be able to use partial parsing during DAG building with `dbt ls` even when not using`RenderConfig(enable_mock_profile=False)`.

To validate the last criteria, the developer can change the example DAG [dev/dags/basic_cosmos_task_group.py](https://github.com/astronomer/astronomer-cosmos/blob/47aa75dda06afd4ab0d20b2d5290c62b6225f012/dev/dags/basic_cosmos_task_group.py), by removing `enable_mock_profile=False`, run it twice with a command similar to:
[...]

PR #1006, opened 2024-05-29
Read mock value from env for mock profile

pankajastro:
## Description

Read the mock value for mock_profile from the environment. This change allows partial parsing without enable_mock_profile=False, but it may require setting the necessary environment variables.

## Related Issue(s)
closes: https://github.com/astronomer/astronomer-cosmos/issues/924

## Checklist


pankajastro:
See https://github.com/astronomer/astronomer-cosmos/issues/924#issuecomment-2138193746

Issue #977, opened 2024-05-17
Create issue templates for Cosmos


tatiana:
**Context**

At the moment, we don't have issue templates for Cosmos.

This leads to many people logging very unstructured issues, which leads to us having to make lots of follow-up questions (versions, for example, DAG, error, how to reproduce, etc). This can be improved by having issue templates.

**Acceptance critearia**

* [ ] Issues are created using templates :D

pankajkoti:
Tested this with creating test issues using the templates
1. Bug report: https://github.com/astronomer/astronomer-cosmos/issues/1010
2. Feature request: https://github.com/astronomer/astronomer-cosmos/issues/1011

PR #1009, opened 2024-05-31
Add GitHub issue templates for bug reports and feature requests

pankajkoti:
We currently do not have GitHub issue templates when 
the community creates issues. Having issue templates
would aid us in catching relevant information that would
help in faster resolution of those issues reported.

With this PR, we're are adding issue templates for bug 
reports and feature requests.

closes: #977

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-bug.yml
+++ b/.github/ISSUE_TEMPLATE/01-bug.yml
@@ -1,0 +1,145 @@
+---
+name: Bug Report
+description: File a bug report.
+title: "[Bug]: "
+labels: ["bug", "triage-needed"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report!
+  - type: dropdown
+    id: cosmos-version
+    attributes:
+      label: Astronomer Cosmos Version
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/01-bug.yml
```diff
@@ -0,0 +1,145 @@
...
+      placeholder: ex. 2.9.0
+    validations:
+      required: true
+  - type: input
+    attributes:
+      label: Operating System
+      description: What Operating System are you using?
+      placeholder: "You can get it via `cat /etc/os-release` for example"
+    validations:
+      required: true
```
tatiana, line 58: This is probably not relevant - I don't remember us having an issue OS-specific so far..!

tatiana (approved):
This looks great, @pankajkoti , it will probably save us a lot of time on triaging Github issues. A minor comment, and I'm happy for the change to be merged.

.github/ISSUE_TEMPLATE/01-bug.yml
tatiana: Fair enough, we can keep it - we can always revisit this in the future!

pankajkoti: yes, thank you for reviewing and highlighting this.

PR #1013, opened 2024-06-03
Use uv in CI

dwreeves:
## Description

The CI takes a really long time to run sometimes:

image

I'm hoping that switching to `uv` will make it run significantly faster.

## Related Issue(s)

n/a

## Breaking Change?

nope

## Checklist

n/a


dwreeves:
A note: The workflows running for this PR are the old one, so if this passes it doesn't mean the actual changes to the Github Action workflow succeeded.

dwreeves:
Seems to run a a little bit faster (shaves about 30 seconds off each run), but Hatch (not uv) is still getting stuck on resolving the Airflow 2.7 dependencies.

.github/workflows/test.yml
```diff
@@ -2,9 +2,9 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, switch-ci-to-uv]
```
```suggestion
    branches: [main]
```

.github/workflows/test.yml
```diff
@@ -2,9 +2,9 @@ name: test
...
   pull_request_target: # Also run on pull requests originated from forks
-    branches: [main]
+    branches: [main, switch-ci-to-uv]
```
tatiana, line 7:
```suggestion
    branches: [main]
```

tatiana (approved):
Every second counts, thanks a lot, @dwreeves for checking the impact of these changes.

I've reverted the branch reference and we're good to merge this 🎉

Issue #990, opened 2024-05-21
Reduce task queueing latency when using Cosmos


tatiana:
**Context**

This issue happened before the last release (1.4) - and can also be reproduced with Cosmos 1.4.1.

Users have observed long task queueing times for Cosmos tasks:
![Screenshot 2024-05-21 at 11 33 12](https://github.com/astronomer/astronomer-cosmos/assets/272048/7af46cba-cbc1-43c1-9e6e-3771b29c9ece)

This is not observed when using, for instance, `BashOperator` task instances:
![Screenshot 2024-05-21 at 11 33 33](https://github.com/astronomer/astronomer-cosmos/assets/272048/59943d80-93d4-4511-bfb7-6775a0414868)

The task queueing time for the Cosmos DAG is consistently 5s, while it is close to 0s for the `BashOperator` one.

**How to reproduce**

Example `Cosmos` DAG:

```python
from datetime import datetime

from cosmos import DbtDag, ProjectConfig, ProfileConfig, RenderConfig
from cosmos.constants import TestBehavior

from include.constants import jaffle_shop_path, venv_execution_config

dbt_profile_example = DbtDag(
    # dbt/cosmos-specific parameters
    project_config=ProjectConfig(jaffle_shop_path),
    profile_config=ProfileConfig(
        # these map to dbt/jaffle_shop/profiles.yml
        profile_name="airflow_db",
        target_name="bq",
        profiles_yml_filepath=jaffle_shop_path / "profiles.yml",
    ),
    render_config=RenderConfig(test_behavior=TestBehavior.NONE),
    execution_config=venv_execution_config,
    # normal dag parameters
    schedule_interval="@daily",
    start_date=datetime(2023, 1, 1),
    catchup=False,
    dag_id="bq_profile_example",
    tags=["profiles"],
)
```

Example `BashOperator` DAG:

```
from pendulum import datetime

from airflow import DAG
from airflow.operators.bash_operator import BashOperator

from include.constants import jaffle_shop_path as DBT_PROJECT_DIR
from include.constants import dbt_executable

DBT_BIN_PATH = dbt_executable

BASE_DBT_CMD = f"{DBT_BIN_PATH} %s --profiles-dir {DBT_PROJECT_DIR} --project-dir {DBT_PROJECT_DIR} --profile airflow_db --target bq"

DBT_DEPS =  BASE_DBT_CMD % " deps"
DBT_LS = BASE_DBT_CMD % " ls"
[...]

PR #992, opened 2024-05-21
Add support for DAG & TaskGroup level caching (performance improvement)

tatiana:
Introduce DAG level cache. Based on initial tests, this can improve the performance of running a Cosmos-powered DAG by 59% (validated with the repo's `basic_cosmos_dag`).

To enable this feature, users should set the environment variable `AIRFLOW__COSMOS__EXPERIMENTAL_CACHE=1`.

What is missing (why this is draft PR):
* Allow users to customise their own `calculate_current_version`
* Refactor so there isn't so much code duplicate between `DbtDag` and `TaskGroup`
* Allow users to purge the cache (via env var or param passed in the Airflow UI - second option is probably preferable)
* Unify `_create_cache_identifier` and `create_cache_identifier_v2`
* Write tests
* The cache should be invalidated if there is an error unpickling

This will be done in a separate PR:
* have a way of storing cache remotely (to be addressed in a separate ticket: https://github.com/astronomer/astronomer-cosmos/issues/927). There is a concern about how secure is pickle, so we may need to cache representation.
* button in the UI to purge cache for individual DAG/TaskGroup
* button in the UI to purge all Cosmos cache
* enable feature by default before 1.5 release (TODO: log ticket)

Closes: #926
Closes: #990

**How to reproduce**

Run using the new cache feature:
```
time AIRFLOW__COSMOS__ENABLE_CACHE=1 AIRFLOW__COSMOS__EXPERIMENTAL_CACHE=1 airflow dags test basic_cosmos_dag  `date -Iseconds`
```
Ran in `7.03s user 2.59s system 81% cpu 11.808 total`
Screenshot 2024-05-21 at 17 28 27

Run without the new cache feature:
```
time  AIRFLOW__COSMOS__ENABLE_CACHE=1 AIRFLOW__COSMOS__EXPERIMENTAL_CACHE=0 airflow dags test basic_cosmos_dag  `date -Iseconds`
```
Ran in `18.75s user 2.62s system 73% cpu 28.898 total`
Screenshot 2024-05-21 at 17 28 39

Issue #1001, opened 2024-05-27
Decouple `LoadMode.AUTOMATIC` from `load()` method in `DbtGraph`


dwreeves:
This PR is an offshoot of #895. I'm introducing it here as it is something that can be implemented reasonably well in Cosmos 1.x.

## Overview

TLDR-- `DbtGraph.load()` actually is two things:

- It loads the dbt graph by dispatching the load implementation
- It implements the `LoadMode.AUTOMATIC` implementation

My proposal is that it should actually only do one of these things, i.e. loads the dbt graph. This means that `automatic` should be its own separate implementation.

See the proposed code for `DbtGraph` inside #895 for a full implementation of the 2.x proposal. Note that some of the suggestions for `DbtGraph` in #895 are proposed deprecations that are out of scope for this issue, but we'd certainly want to move in this direction. Here it is a little slimmed down:

```python
class DbtGraph:

    load_method_mapping: dict[LoadMode, Callable[[], None]] = {}

    def __init__(
        ...
    ):
        default_load_method_mapping = {
            LoadMode.AUTOMATIC: self.load_with_automatic_method,
            LoadMode.CUSTOM: self.load_via_custom_parser,
            LoadMode.DBT_LS: self.load_via_dbt_ls,
            LoadMode.DBT_LS_FILE: self.load_via_dbt_ls_file,
            LoadMode.DBT_MANIFEST: self.load_from_dbt_manifest,
        }
        # It looks like this in case load methods are set at the class level,
        # i.e. `default_load_method_mapping` sets defaults, but does not override
        # anything a user might set at the class level.
        default_load_method_mapping.update(self.load_method_mapping)
        self.load_method_mapping = default_load_method_mapping

        self.is_loaded: bool = False

        self.nodes: dict[str, DbtNode] = {}
        self.filtered_nodes: dict[str, DbtNode] = {

    def load(
            self,
            method: LoadMode | None = None,
            reload: bool = False
    ) -> dict[str, DbtNode]:
        if not reload and self.is_loaded:
            return self.filtered_nodes
        # Load only dispatches;
        # does not implicitly have the automatic load mode inside.
        if method is None:
            method = self.render_config.load_method
        callback = self.load_method_mapping[method]
        # Before the load() method only mutated DbtGraph's state.
        # It's possible that DbtGraph should not rely on mutated state.
        return callback()

    def load_with_automatic_method(self) -> None: ...
    def load_via_custom_parser(self) -> None: ...
[...]

Issue #895, opened 2024-03-18
Introducing composability in the middle layer of Cosmos's API


dwreeves:
# TLDR

Sorry, this is long!!!! But there is a lot to cover. This is a massive proposal to reorganize massive chunks of Cosmos.

Here's a TLDR:

- The "middle layer" of Cosmos's API (things right below the highest-level API) are very hard to customize for end-users.
- With some smart changes of this middle layer of the API, the floodgates can open for users to inject a lot of their own custom logic into Cosmos.
- These changes should mostly revolve around support for subclassing with method overrides.
- There are some other more minor changes to the API that also make sense, while we are at it.

# Introduction

At my last company, I built a Dbt Airflow integration prior to the existence of Cosmos. At my current company, we are using Cosmos.

There are a handful of features I really miss from that custom integration. The big one is, when I have a `ref()` to a task in a different DAG, we'd automatically create an external task sensor to the dbt node in that separate DAG. I don't think Cosmos should necessarily implement that feature, but I do think right now it would be extremely complicated to do so as a custom feature in an Airflow deployment.

# Background: Issues with custom implementations today

Right now, the only real way to dive deep into Cosmos's internals is to use the `node_converters` API. This is typed as a `dict[DbtResourceType, Callable[..., Any]]`, but a more accurate typing would be `dict[DbtResourceType, NodeConverterProtocol]` with the following definition for `NodeConverterProtocol`:

```python
class NodeConverterProtocol(Protocol):
    __call__(
            self,
            dag: DAG,
            task_group: TaskGroup | None,
            node: DbtNode,
            execution_mode: ExecutionMode,
            task_args: dict[str, Any],
            test_behavior: TestBehavior,
            test_indirect_selection: TestIndirectSelection,
            on_warning_callback: Callable[..., Any] | None,
            **kwargs: Any
    ) -> BaseOperator | TaskGroup | None:
        ...
```

Where the `kwargs` don't do anything.

This API is hard to work with, to be honest. Imagine for example you want to use a variable in the DbtNode's `config` to implement custom functionality. You'd have to do something like this:

```python
from cosmos.airflow.graph import generate_task_or_group

def custom_callback(
    dag: DAG,
    task_group: TaskGroup | None,
    node: DbtNode,
    execution_mode: ExecutionMode,
    task_args: dict[str, Any],
    test_behavior: TestBehavior,
    test_indirect_selection: TestIndirectSelection,
[...]

PR #1014, opened 2024-06-03
Speed up `LoadMode.DBT_LS` by caching dbt ls output in Airflow Variable

tatiana:
Improve significantly the `LoadMode.DBT_LS` performance. The example DAGs tested reduced the task queueing time significantly (from ~30s to ~0.5s) and the total DAG run time for Jaffle Shop from 1 min 25s to 40s (by more than 50%). Some users[ reported improvements of 84%](https://github.com/astronomer/astronomer-cosmos/pull/1014#issuecomment-2168185343) in the DAG run time when trying out these changes.  This difference can be even more significant on larger dbt projects.

The improvement was accomplished by caching the dbt ls output as an Airflow Variable. This is an alternative to #992, when we cached the pickled DAG/TaskGroup into a local file in the Airflow node. Unlike #992, this approach works well for distributed deployments of Airflow.

As with any caching solution, this strategy does not guarantee optimal performance on every run—whenever the cache is regenerated, the scheduler or DAG processor will experience a delay. It was also observed that the key value could change across platforms (e.g., `Darwin` and `Linux`). Therefore, if using a deployment with heterogeneous OS, the key may be regenerated often.

Closes: #990
Closes: #1061

**Enabling/disabling this feature**

This feature is enabled by default.
Users can disable it by setting the environment variable `AIRFLOW__COSMOS__ENABLE_CACHE_DBT_LS=0`.

**How the cache is refreshed**

Users can purge or delete the cache via Airflow UI by identifying and deleting the cache key.

The cache will be automatically refreshed in case any files of the dbt project change. Changes are calculated using the SHA256 of all the files in the directory. Initially, this feature was implemented using the files' modified timestamp, but this did not work well for some Airflow deployments (e.g., `astro --dags` since the timestamp was changed during deployments).

Additionally, if any of the following DAG configurations are changed, we'll automatically purge the cache of the DAGs that use that specific configuration:
* `ProjectConfig.dbt_vars`
* `ProjectConfig.env_vars`
* `ProjectConfig.partial_parse`
* `RenderConfig.env_vars`
* `RenderConfig.exclude`
* `RenderConfig.select`
* `RenderConfig.selector`

The following argument was introduced in case users would like to define Airflow variables that could be used to refresh the cache (it expects a list with Airflow variable names):
* `RenderConfig.airflow_vars_to_purge_cache`

Example:
```
RenderConfig(
    airflow_vars_to_purge_cache==["refresh_cache"]
)
```

**Cache key**

The Airflow variables that represent the dbt ls cache are prefixed by `cosmos_cache`. When using `DbtDag`, the keys use the DAG name. When using `DbtTaskGroup`, they consider the TaskGroup and parent task groups and DAG.

Examples:
1. The `DbtDag` "cosmos_dag" will have the cache represented by `"cosmos_cache__basic_cosmos_dag"`.
2. The `DbtTaskGroup` "customers"  declared inside teh DAG "basic_cosmos_task_group" will have the cache key `"cosmos_cache__basic_cosmos_task_group__customers"`.

**Cache value**

The cache values contain a few properties:
- `last_modified` timestamp, represented using the ISO 8601 format.
- `version` is a hash that represents the version of the dbt project and arguments used to run dbt ls by the time the cache was created
- `dbt_ls_compressed` represents the dbt ls output compressed using zlib and encoded to base64 to be recorded as a string to the Airflow metadata database.

Steps used to compress:
```
        compressed_data = zlib.compress(dbt_ls_output.encode("utf-8"))
        encoded_data = base64.b64encode(compressed_data)
        dbt_ls_compressed = encoded_data.decode("utf-8")
```

We are compressing this value because it will be significant for larger dbt projects, depending on the selectors used, and we wanted this approach to be safe and not clutter the Airflow metadata database.

Some numbers on the compression
* A dbt project with 100 models can lead to a dbt ls output of 257k characters when using JSON. Zlib could compress it by 20x.
* Another [real-life dbt project](https://gitlab.com/gitlab-data/analytics/-/tree/master/transform/snowflake-dbt?ref_type=heads) with 9,285 models led to a dbt ls output of 8.4 MB, uncompressed. It reduces to 489 KB after being compressed using `zlib` and encoded using `base64` - to 6% of the original size.
* Maximum cell size in Postgres: 20MB

The latency used to compress is in the order of milliseconds, not interfering in the performance of this solution.

**Future work**

* How this will affect the Airflow db in the long term
* How does this performance compare to `ObjectStorage`?

**Example of results before and after this change**

Task queue times in Astro before the change:
Screenshot 2024-06-03 at 11 15 26

Task queue times in Astro after the change on the second run of the DAG:
Screenshot 2024-06-03 at 11 15 44

This feature will be available in `astronomer-cosmos==1.5.0a8`.


cosmos/converter.py
```diff
@@ -225,19 +228,30 @@ def __init__(
         dbt_vars = project_config.dbt_vars or operator_args.get("vars")
 
         cache_dir = None
+        cache_identifier = None
         if settings.enable_cache:
-            cache_dir = cache._obtain_cache_dir_path(cache_identifier=cache._create_cache_identifier(dag, task_group))
+            cache_identifier = cache._create_cache_identifier(dag, task_group)
+            cache_dir = cache._obtain_cache_dir_path(cache_identifier=cache_identifier)
 
+        previous_time = time.process_time()
```
jedcunningham, line 236: Probably want `perf_counter()` instead?

cosmos/dbt/graph.py
```diff
@@ -190,6 +196,25 @@ def load(
         logger.info("Total nodes: %i", len(self.nodes))
         logger.info("Total filtered nodes: %i", len(self.nodes))
 
+    def load_via_dbt_ls_cache(self) -> bool:
+        """
+        Load dbt ls cache
+        """
+        logger.info(f"Trying to parse the dbt project using dbt ls cache {self.cache_identifier}...")
+        if settings.enable_cache and settings.experimental_cache:
+            dbt_ls_cache = Variable.get(self.cache_identifier, "")
```
jedcunningham, line 205: It might be better to query the db directly, so you bypass any configured secrets backend.

cosmos/converter.py
tatiana: Thanks, @jedcunningham , I'll change it!

kzajaczkowski:
Looks promising!

tatiana:
@kzajaczkowski, thanks a lot! The feature is currently available in the 1.5.0a7 release. Would you be interested in testing it out and giving early feedback?

kzajaczkowski:
> @kzajaczkowski, thanks a lot! The feature is currently available in the 1.5.0a7 release. Would you be interested in testing it out and giving early feedback?

@tatiana, we did some preliminary testing using 1.5.0a4 release and observed one of a simple dags go from 00:13:45 to 00:02:11. Queuing times went almost to zero. The  Experienced also `pytest` dag tests go from a few minutes to seconds.

cosmos/dbt/graph.py
```diff
@@ -223,10 +333,56 @@ def run_dbt_ls(
...
+            if dbt_ls_cache and not cache.was_project_modified(cache_version, current_version):
+                logger.info(
+                    f"Cosmos performance [{platform.node()}|{os.getpid()}]: The cache size for {self.dbt_ls_cache_key} is {len(dbt_ls_cache)}"
+                )
+                self.load_method = LoadMode.DBT_LS_CACHE
+
+                nodes = parse_dbt_ls_output(project_path=project_path, ls_stdout=dbt_ls_cache)
+                self.nodes = nodes
+                self.filtered_nodes = nodes
+                logger.warning(f"Cosmos performance: Cache hit for {self.dbt_ls_cache_key} - {current_version}")
```
pankajkoti, line 380:
```suggestion
                logger.info(f"Cosmos performance: Cache hit for {self.dbt_ls_cache_key} - {current_version}")
```
when cache is hit, do we still want it to be a warning?

dev/dags/.airflowignore.old
```diff
@@ -0,0 +1,12 @@
+basic_cosmos_dag.py
+#basic_cosmos_task_group.py
```
pankajkoti, line 2: Should we uncomment this?

dev/dags/.airflowignore.old
```diff
@@ -0,0 +1,12 @@
+basic_cosmos_dag.py
```
pankajkoti, line 1: Do we need this file? If yes, would it make sense to add a comment at the top of the file mentioning the purpose of this file.

pankajkoti (approved):
LGTM. Some minor comments inline if we would like to consider.

cosmos/cache.py
```diff
@@ -171,3 +179,59 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
...
+
+    # Combined value for when the dbt project directory files were last modified
+    # This is fast (e.g. 0.01s for jaffle shop, 0.135s for a 5k models dbt folder)
+    dbt_project_hash = _create_folder_version_hash(project_dir)
+
+    # The performance for the following will depend on the user's configuration
+    hash_args = hashlib.md5("".join(cmd_args).encode()).hexdigest()
+
+    elapsed_time = time.process_time() - start_time
+    logger.info(f"Cosmos performance: time to calculate {cache_identifier} current version: {elapsed_time}")
```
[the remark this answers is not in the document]
tatiana: Thanks a lot, @pankajkoti , addressed in fc9386f - yes, cmd_args are the arguments passed to the `dbt ls` command, that could invalidate the cache.

cosmos/converter.py
```diff
@@ -225,19 +228,30 @@ def __init__(
...
             cache_dir=cache_dir,
+            cache_identifier=cache_identifier,
             dbt_vars=dbt_vars,
         )
         self.dbt_graph.load(method=render_config.load_method, execution_mode=execution_config.execution_mode)
 
+        current_time = time.process_time()
+        elapsed_time = current_time - previous_time
+        logger.info(
+            f"Cosmos performance ({cache_identifier}) -  [{platform.node()}|{os.getpid()}]: It took {elapsed_time:.3}s to parse the dbt project for DAG {cache_identifier} using {self.dbt_graph.load_method}"
```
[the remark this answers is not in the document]
tatiana: Addressed in 76601aa9443731c8be7399c859684f4228598975

cosmos/dbt/graph.py
```diff
@@ -138,16 +146,131 @@ def __init__(
...
+        """
+        User-defined environment variables, relevant to running dbt ls.
+        """
+        return self.render_config.env_vars or self.project.env_vars or {}
+
+    @cached_property
+    def project_path(self) -> Path:
+        """
+        User-defined path to their dbt project. Tries to retrieve the configuration from render_config and (legacy support)
+        ExecutionConfig, where it was originally defined. Returns the
```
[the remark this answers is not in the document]
tatiana: Solved in b22c920a66de73530f2d858707d7f76e12dc07fe

cosmos/cache.py
```diff
@@ -171,3 +179,59 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
...
+
+    for filepath in sorted(filepaths):
+        with open(str(filepath), "rb") as fp:
+            buf = fp.read()
+            hasher.update(buf)
+
+    return hasher.hexdigest()
+
+
+def calculate_dbt_ls_cache_current_version(cache_identifier: str, project_dir: Path, cmd_args: list[str]) -> str:
```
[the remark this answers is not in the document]
tatiana: Done in d0c323d2de6029f0dfd34c6e15d0d66b43a02f28

cosmos/dbt/graph.py
```diff
@@ -138,16 +146,131 @@ def __init__(
...
+        env_vars = self.env_vars
+        if env_vars:
+            envvars_str = json.dumps(env_vars, sort_keys=True)
+            cache_args.append(envvars_str)
+        if self.render_config.airflow_vars_to_purge_dbt_ls_cache:
+            for var_name in self.render_config.airflow_vars_to_purge_dbt_ls_cache:
+                airflow_vars = [var_name, Variable.get(var_name, "")]
+                cache_args.extend(airflow_vars)
+
+        return cache_args
```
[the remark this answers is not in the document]
tatiana: Done in ea452f72634bed497767dcde98604cfde358300b

cosmos/dbt/graph.py
```diff
@@ -138,16 +146,131 @@ def __init__(
...
+        # This compression reduces the dbt ls output to 10% of the original size
+        compressed_data = zlib.compress(dbt_ls_output.encode("utf-8"))
+        encoded_data = base64.b64encode(compressed_data)
+        dbt_ls_compressed = encoded_data.decode("utf-8")
+        cache_dict = {
+            "version": cache.calculate_dbt_ls_cache_current_version(
+                self.dbt_ls_cache_key, self.project_path, self.dbt_ls_cache_key_args
+            ),
+            "dbt_ls_compressed": dbt_ls_compressed,
+            "last_modified": datetime.datetime.now().isoformat(),
```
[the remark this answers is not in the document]
tatiana: Done in 7ba3b63e57b25a2451a72436cdd063f436e450dc

cosmos/dbt/graph.py
tatiana: Nice catch, @pankajkoti ! I was just using that for troubleshooting, just reverted to info in: 97dddba00ac8cdb0e7a94a80d7c8230af66c990d

dev/dags/.airflowignore.old
tatiana: No, just deleted: 3cb6bedefd79c315e7308f2befdbb6343b4d81ca

dev/dags/.airflowignore.old
tatiana: I remove the file, so this no longer exists :)

cosmos/dbt/graph.py
tatiana: @jedcunningham how do you advise us to do this?
Wouldn't there be a risk that with this, we'd create a larger coupling of Cosmos to Airflow that could be more sensitive to different versions of Airflow?

tatiana: Jed, since this has been long standing, I'll be merging as it is - and I can make a follow up PR to address after your feedback.

docs/configuration/cosmos-conf.rst
```diff
@@ -30,6 +30,22 @@ This page lists all available Airflow configurations that affect ``astronomer-co
     - Default: ``True``
     - Environment Variable: ``AIRFLOW__COSMOS__ENABLE_CACHE``
 
+.. enable_cache_dbt_ls:
```
pankajkoti, line 33:
```suggestion
.. _enable_cache_dbt_ls:
```

pankajkoti (approved):
Looks good to me. There are some minor cosmetic suggestions for the documentation that we can address iteratively in a subsequent PR.

Great feature support! Thank you! 👏🏽

dwreeves:
@tatiana A little late to this PR, but not late to 1.5.0:

I'm thinking the `cache_identifier` and `cache_dir` should go in the `render_config`, since the render config is supposed to handle all the options for things like this.

I think, going into something like Cosmos 2.0, it makes sense to consolidate the `DbtGraph()` init function's signature to something simple like:

```python
class DbtGraph:

    load_method_mapping: dict[LoadMode, Callable[[], None]] = {}

    def __init__(
            self,
            project_config: ProjectConfig | None = None,
            profile_config: ProfileConfig | None = None,
            execution_config: ExecutionConfig | None = None,
            **kwargs
    ):
```

as per the code example in #895.

Adding more kwargs means either we cannot do this, or we need to add more deprecations.

I think doing this will also help make the API cleaner, more consistent, and easier for users to reason about. Right now there are just a lot of things for users to tweak and it can be overwhelming. Keeping it consistent and locked inside the configs can reduce confusion.

WDYT?

tatiana:
> @tatiana A little late to this PR, but not late to 1.5.0:

Hi @dwreeves, Thanks a lot for all the very relevant feedback! I'm sorry I missed it, as it was made after the PR was merged.

> I'm thinking the cache_identifier and cache_dir should go in the render_config, since the render config is supposed to handle all the options for things like this.

We can expose them there as part of a follow-up PR / future version of Cosmos. Just added: #1110

I'm also happy with the proposal to refactor the `DbtGraph` interface and #895. Would you be interested in working on this?

I created a PR to address the feedback on the dead code: #1111.

Your proposal to refactor the load implementation in #1001 is very good as well. Would you like to work on it?

dwreeves:
> Your proposal to refactor the load implementation in https://github.com/astronomer/astronomer-cosmos/issues/1001 is very good as well. Would you like to work on it?

I would, but I've been busy... still trying to find time to contribute to this project. ☹️ The time will eventually come.

PR #839, opened 2024-02-07
Make GoogleCloudServiceAccountDictProfileMapping dataset profile argument optional

oliverrmaa:
## Description

This PR follows the methodology in https://github.com/astronomer/astronomer-cosmos/pull/683/ by modifying the GCP `GoogleCloudServiceAccountDictProfileMapping()` `profile_args` not require dataset as a required argument. 

## Related Issue(s)

Closes #837 

## Breaking Change?

Not to my knowledge.

## Checklist

- [x] I have made corresponding changes to the documentation (if required)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #683, opened 2023-11-16
Make PostgresUserPasswordProfileMapping schema profile argument schem…

FouziaTariq:
## Description

Made PostgresUserPasswordProfileMapping schema profile argument optional

## Related Issue(s)

Closes: #675

PR #1017, opened 2024-06-03
Make GoogleCloudServiceAccountDictProfileMapping dataset profile argu…

### LINKED ISSUE #837: Make GoogleCloudServiceAccountDictProfileMapping dataset profile argument optional
**Context**

Given the issue [here](https://github.com/astronomer/astronomer-cosmos/issues/675) and the discussion around it [in slack](https://apache-airflow.slack.com/archives/C059CC42E9W/p1700049843620749), it follows that it would not be too far off for other profile mapping classes (like GCP) to have their schema (dataset in GCP terms) arguments made optional. 

The configurations in dbt (profiles.yml, dbt_project.yml, model level configs etc.) are meant to provide the dataset. As such, me and my team's initial thoughts were to not provide any arguments here since usually we would rely on this being handled inside of dbt instead of it being hard coded at the DAG (orchestration) level.

**Solution**

I believe its possible modifying this line [here ](https://github.com/astronomer/astronomer-cosmos/blob/72690d31ba6822ae6df56f4d56f26196234c80b5/cosmos/profiles/bigquery/service_account_keyfile_dict.py#L25) and the associated code similarly to this issue [here](https://github.com/astronomer/astronomer-cosmos/issues/675) would be the right path.

Me and/or my colleague are happy to to attempt this modification if there are no objections to this proposed change. Please let us know, thanks

pankajastro:
## Description

This PR follows the methodology in https://github.com/astronomer/astronomer-cosmos/pull/683/ by modifying the GCP `GoogleCloudServiceAccountDictProfileMapping()` `profile_args` not require dataset as a required argument. 

DAG RUN
Screenshot 2024-06-05 at 6 50 04 PM

Co-authored-by: "Ollie Ma" <oliver.zheyi.ma@gmail.com> 
Original PR by @oliverrmaa: https://github.com/astronomer/astronomer-cosmos/pull/839

## Related Issue(s)

Closes #837 

## Breaking Change?

Not to my knowledge.

## Checklist

### DIFF
--- a/tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
+++ b/tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
@@ -65,7 +65,7 @@ def test_connection_claiming_succeeds(mock_bigquery_conn_with_dict: Connection):
 
 def test_connection_claiming_fails(mock_bigquery_conn_with_dict: Connection):
     # Remove the `project` key, which is mandatory
-    mock_bigquery_conn_with_dict.extra = json.dumps({"keyfile_dict": sample_keyfile_dict})
+    mock_bigquery_conn_with_dict.extra = json.dumps({"dataset": "my_dataset", "keyfile_dict": sample_keyfile_dict})
     profile_mapping = GoogleCloudServiceAccountDictProfileMapping(mock_bigquery_conn_with_dict, {})
     assert not profile_mapping.can_claim_connection()
 
--- a/tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
+++ b/tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
@@ -64,8 +64,8 @@ def test_connection_claiming_succeeds(mock_bigquery_conn_with_dict: Connection):
 
 
 def test_connection_claiming_fails(mock_bigquery_conn_with_dict: Connection):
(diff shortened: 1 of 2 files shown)


tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
```diff
@@ -64,8 +64,8 @@ def test_connection_claiming_succeeds(mock_bigquery_conn_with_dict: Connection):
 
 
 def test_connection_claiming_fails(mock_bigquery_conn_with_dict: Connection):
-    # Remove the `dataset` key, which is mandatory
```
tatiana, line 67: Do we still have unit tests covering that if users were defining the Connection `extras.dataset`, it would be used?

Do we have a unittest covering that if users do not set `extras.dataset`, that we're not raising an exception?

pankajastro: Test where we set dataset https://github.com/astronomer/astronomer-cosmos/blob/main/tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py#L73

I have added a test where we set dataset and it does not raise error

tatiana: Thank you very much, @pankajastro !

Issue #962, opened 2024-05-14
[Bug]  Invalid aws session token


ssulca:
Cosmos in AthenaAccessKeyProfileMapping always tries to use a session token, even when it is not provided by the Airflow connection

**Context**

* astronomer-cosmos= "1.3.2"
* dbt-core = "1.7.14"
* dbt-athena-community = "1.7.2"

When setting up a new aws connection using only `aws_access_key_id` and `aws_secret_access_key`, it
generates a profile like this:

```yaml
profile_name:
    outputs:
        athena:
            aws_access_key_id: KEY_ID
            aws_secret_access_key: '{{ env_var(''COSMOS_CONN_AWS_AWS_SECRET_ACCESS_KEY'')}}'
            aws_session_token: '{{ env_var(''COSMOS_CONN_AWS_AWS_SESSION_TOKEN'')}}'
            database: db_name
            ...
    target: athena
```
But  `AthenaAccessKeyProfileMapping` sets the `aws_session_token` as an empty string, and dbt takes this value,
resulting in. 

```sh
'botocore.exceptions.ClientError: An error occurred (UnrecognizedClientException)
 when calling the GetDatabases operation: 
The security token included in the request is invalid', ''
```

**Propose solutions**
In [profile](https://github.com/astronomer/astronomer-cosmos/blob/e4aa3c48d8f9206c416dd94aba73ad3d4e6b391f/cosmos/profiles/athena/access_key.py#L59C1-L60C1) function

Add a condition that check if there is a token.
```python
profile = {
    **self.mapped_params,
    **self.profile_args,
    "aws_access_key_id": self.temporary_credentials.access_key,
    "aws_secret_access_key": self.get_env_var_format("aws_secret_access_key"),
    "aws_session_token": self.get_env_var_format("aws_session_token") if self.temporary_credentials.token else None,
}
```

tatiana:
HI @ssulca , thanks for reporting the issue and proposing a solution!
Would you like to contribute a PR with this fix? It would be highly appreciated.

PR #1022, opened 2024-06-04
Athena profile mapping set aws_session_token in profile only if it exist

pankajastro:
## Description
Set aws_session_token in the Athena profile only if it exists to avoid passing an empty string as a token to the AWS API.

## Related Issue(s)
closes: https://github.com/astronomer/astronomer-cosmos/issues/962

## Checklist


tatiana (commented):
Thanks, @pankajastro , it looks we're just missing a test!

pankajastro:
> Thanks, @pankajastro , it looks we're just missing a test!

Added in 89dd0eedcb5939ac8bb3793cb96bb309ecbb1a84

tatiana (approved):
Thanks, @pankajastro ! Super speedy fix

PR #975, opened 2024-05-16
Move airflow conf fetch call to setting.py

pankajastro:
## Description

- Centralizing environment or configuration fetching by moving the Airflow configuration call to the Cosmos settings.py file.
- Add documentation for cosmos config sections

Sample HTML page
Screenshot 2024-05-18 at 1 04 13 AM

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/928

## Breaking Change?

No

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

Issue #927, opened 2024-04-29
Support caching remotely


tatiana:
**Context**

Since #904, Cosmos introduced caching, contributing to the latest performance improvements in 1.4.

However, one of the limitations of this approach is that the cache is stored locally, on disk. This means that:
- if the Airflow scheduler/worker node that's running AIrflow is recreated, the cache will be have to be recreated
- each of the Airflow worker nodes/schedulers will have to create their cache. This can be remarkably inefficient when using Airflow `KubernetesExecutor`

During the code review of the PR mentioned above, one of the feedbacks was that it would be great if we supported caching this in S3/GCS/Blob storage: https://github.com/astronomer/astronomer-cosmos/pull/904#issuecomment-2032602233 (from @jlaneve).

Another feedback was to leverage Airflow 2.8 [ObjectStore](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/objectstorage.html): https://github.com/astronomer/astronomer-cosmos/pull/904#issuecomment-2049512157 or/and using an XCom backend to store the cache. (from @kaxil)

**Acceptance Criteria**

* [ ] Decide on an approach to store the remote cache
* [ ] Allow users to update/fetching cache from a remote location for all Airflow versions supported by Cosmos

dwreeves:
This is a duplicate of #870, although I prefer we use this issue as yours is newer and more general. (E.g. I don't mention the use of xcoms as the cache.) Just tagging that issue to relate these discussions.

dwreeves:
Remote filesystem stuff keeps coming up in multiple contexts. And we already have support for this in, of all places, `cosmos/plugin/__init__.py` with the `open_file()` function.

I think this function should be moved to some sort of utils file for interacting with remote filesystems, and the logic for getting the conn_id from the cosmos config should be decoupled from `open_file()` so it can be used more generically.

That said, we also want to make sure we are doing things idiomatically, as well. For Airflow 2.8+, `ObjectStoragePath` was essentially designed to do this. This means we have 3 options:

- **Only use `ObjectStoragePath`**: We can say "oh well" to Airflow users prior to 2.8 for remote filesystems and _exclusively_ work with `ObjectStoragePath` in all contexts where remote filesystems are required.
- **Never use `ObjectStoragePath`**: We can just ignore this new Airflow feature entirely.
- **Support both:** Let the user choose.

[...]

PR #1025, opened 2024-06-05
Fix Cosmos enable_cache setting

tatiana:
As of Cosmos 1.4.1, users are not able to disable the cache in Cosmos using `AIRFLOW__COSMOS__ENABLE_CACHE=0`.

During a recent refactoring #975, the `enable_cache` was changed to a non-boolean config. This PR fixes this issue.

### DIFF
--- a/cosmos/settings.py
+++ b/cosmos/settings.py
@@ -10,7 +10,7 @@ from cosmos.constants import DEFAULT_COSMOS_CACHE_DIR_NAME, DEFAULT_OPENLINEAGE_
 # In MacOS users may want to set the envvar `TMPDIR` if they do not want the value of the temp directory to change
 DEFAULT_CACHE_DIR = Path(tempfile.gettempdir(), DEFAULT_COSMOS_CACHE_DIR_NAME)
 cache_dir = Path(conf.get("cosmos", "cache_dir", fallback=DEFAULT_CACHE_DIR) or DEFAULT_CACHE_DIR)
-enable_cache = conf.get("cosmos", "enable_cache", fallback=True)
+enable_cache = conf.getboolean("cosmos", "enable_cache", fallback=True)
 propagate_logs = conf.getboolean("cosmos", "propagate_logs", fallback=True)
 dbt_docs_dir = conf.get("cosmos", "dbt_docs_dir", fallback=None)
 dbt_docs_conn_id = conf.get("cosmos", "dbt_docs_conn_id", fallback=None)
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -1,0 +1,11 @@
+import os
+from importlib import reload
+from unittest.mock import patch
+
+from cosmos import settings
+
+
+@patch.dict(os.environ, {"AIRFLOW__COSMOS__ENABLE_CACHE": "False"}, clear=True)
+def test_enable_cache_env_var():
+    reload(settings)
+    assert settings.enable_cache is False


pankajastro (approved):
Thanks for fixing it!

pankajkoti (approved):

maslick:
@tatiana Does setting `AIRFLOW__COSMOS__ENABLE_CACHE=False` work in 1.4.1?

tatiana:
@maslick There was an issue in the `1.4.1` release with this configuration, this was solved and is available on `1.4.2`

maslick:
@tatiana I understand. The reason I am asking is that we're using `AIRFLOW__COSMOS__ENABLE_CACHE=False` (and not `AIRFLOW__COSMOS__ENABLE_CACHE=0` as in the 1.4.2 release notes) with 1.4.1, but we can't easily migrate to 1.4.3. So my question is, would no caching even work with 1.4.3?

tatiana:
@maslick In Cosmos 1.4.1, if you want to disable the cache, you can do this via

```
export AIRFLOW__COSMOS__ENABLE_CACHE=''
```

If you check this PR, we fixed how we retrieve the envvar to use Airflow `conf.getboolean`:
```
enable_cache = conf.getboolean("cosmos", "enable_cache", fallback=True)
```

In Cosmos 1.4.1, this was implemented `conf.get` by mistake.

From Cosmos 1.4.2, both `AIRFLOW__COSMOS__ENABLE_CACHE=False` as well as `AIRFLOW__COSMOS__ENABLE_CACHE=0` will work.

Did you have any issues with cache that made you want to switch off this feature?

maslick:
@tatiana out issue with cache enabled was that it quickly fills up the 30 GB storage volume on MWAA workers. We are running a large number of DAGs (>500) with lots of dbt models/transformations... We can't increase the storage volume on MWAA either. That's why we decided to disable cache. Of course, this would increase the parsing times...

josefbits:
@tatiana With cache disabled it's no longer possible to do partial parsing from a supplied manifest. Is this intended behavior?

tatiana:
@maslick this issue will hopefully be resolved once #927 is implemented. I'm sorry that the cache had such a massive impact on your deployment - we weren't expecting for that to take 30 GB volume storage.

@josefbits I believe this is unintended behavior. Please open an issue using https://github.com/astronomer/astronomer-cosmos/issues, and we'll follow up on this ASAP. Contributions are also welcome; it should be a small fix.

josefbits:
@tatiana Thanks, I've opened https://github.com/astronomer/astronomer-cosmos/issues/1041 for this.

josefbits:
@tatiana Not sure if this is the right place to continue this discussion, but with regards to

> @maslick this issue will hopefully be resolved once #927 is implemented. I'm sorry that the cache had such a massive impact on your deployment - we weren't expecting for that to take 30 GB volume storage. It's a shame MWAA doesn't handle better temp files.

Is there some specific reason that you need to cache multiple copies of the manifests? Right now Cosmos is caching once per DbtTaskGroup and for projects with a large number of models, dags, task groups etc it can become problematic on managed Airflow solutions.

Speaking of dbt parsing specifically, I've done a bunch of comparisons on the manifest.json before and after run, with partial parse enabled/disabled, and different methods to change the dbt state. From what I've seen the only thing that really changes is metadata like created_at, raw_code, and compile/build paths. This is, of course, assuming parity between CI build environment and run time, and that one is not doing something strange that might change the model code or dependencies.

Just looking for a bit more insight :)

tatiana:
@josefbits, thanks for your insights!
This is a better place to discuss this: https://github.com/astronomer/astronomer-cosmos/issues/1042

The issue we're discussing was not introduced by this PR, which was already merged last week - so new discussions may get lost.

Issue #1032, opened 2024-06-07
[Docs]: Update outdated Data-Aware Scheduling docs


pankajastro:
### Astronomer Cosmos Version

main (development)

### dbt-core version

1.8.0

### LoadMode

AUTOMATIC

### ExecutionMode

AWS_EKS

### InvocationMode

None

### airflow version

2.9.0

### Operating System

Linux

### Deployment

Official Apache Airflow Helm Chart

### What happened?

The data-aware-scheduling docs look likes outdated

https://astronomer.github.io/astronomer-cosmos/configuration/scheduling.html#data-aware-scheduling

See: https://apache-airflow.slack.com/archives/C059CC42E9W/p1717713106253499

### How to reproduce

NA

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #1034, opened 2024-06-07
Fix docs so it does not reference non-existing `get_dbt_dataset`

tatiana:
[The documentation](https://astronomer.github.io/astronomer-cosmos/configuration/scheduling.html) was outdated. 
The method `get_dbt_dataset` no longer exists. It used to exist in older versions of Cosmos (before 1.1) when the URIs respected the format:
Dataset(f"DBT://{connection_id.upper()}/{project_name.upper()}/{model_name.upper()}")

More information on why we changed this: https://github.com/astronomer/astronomer-cosmos/issues/305

Closes: #1032

### DIFF
--- a/docs/configuration/scheduling.rst
+++ b/docs/configuration/scheduling.rst
@@ -24,23 +24,29 @@ To schedule a dbt project on a time-based schedule, you can use Airflow's schedu
 Data-Aware Scheduling
 ---------------------
 
-By default, Cosmos emits `Airflow Datasets <https://airflow.apache.org/docs/apache-airflow/stable/concepts/datasets.html>`_ when running dbt projects. This allows you to use Airflow's data-aware scheduling capabilities to schedule your dbt projects. Cosmos emits datasets in the following format:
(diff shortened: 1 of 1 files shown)


docs/configuration/scheduling.rst
```diff
@@ -24,23 +24,29 @@ To schedule a dbt project on a time-based schedule, you can use Airflow's schedu
...
 - A second dbt project (``project_two``) with a model called ``my_other_model`` that you want to run immediately after ``my_model``
 
+We are assuming that the Database used is Postgres, the host is ``host``, the database is ``database`` and the schema is ``schema``.
+
 Then, you can use Airflow's data-aware scheduling capabilities to schedule ``my_other_model`` to run after ``my_model``. For example, you can use the following DAGs:
 
 .. code-block:: python
 
-    from cosmos import DbtDag, get_dbt_dataset
+    from cosmos import DbtDa
```
pankajkoti, line 49:
```suggestion
    from cosmos import DbtDag
```

docs/configuration/scheduling.rst
```diff
@@ -50,9 +56,9 @@ Then, you can use Airflow's data-aware scheduling capabilities to schedule ``my_
 
     project_two = DbtDag(
         # for airflow <=2.3
-        # schedule=[get_dbt_dataset("my_conn", "project_one", "my_model")],
+        # schedule_interval=[Dataset("postgres://host:5432/database.schema.my_model")],,
```
pankajkoti, line 59: appears that Data-aware scheduling using Datasets was only introduced starting Airflow 2.4. So Airflow <=2.3 might not support this(?)

Ref: https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/datasets.html

tatiana: Well spotted, Pankaj, just fixed!

pankajkoti (approved):

PR #1038, opened 2024-06-09
Add deep linking of dbt docs

### LINKED ISSUE #1029: Support deep linking in the DBT docs plugin
### Description

Since the [DBT docs plugin](https://astronomer.github.io/astronomer-cosmos/configuration/hosting-docs.html) uses an iframe, it doesn't support deep linking.

For example, when browsing the docs with `dbt docs serve`, you can get links like `https://<hostname>/view#!/model/model.foo.bar`. This does not happen in this plugin.

### Use case/motivation

It's useful to be able to share links to a particular model in the docs. Without support for deep linking this isn't possible.

### Related issues

_No response_

dwreeves:
## Description

Ties the URL in the browser to the URL of the dbt docs

## Related Issue(s)

Resolves #1029

## Breaking Change?

n/a

## Checklist

- [x] (n/a) I have made corresponding changes to the documentation (if required)
- [x] (n/a) I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/cosmos/plugin/__init__.py
+++ b/cosmos/plugin/__init__.py
@@ -139,8 +139,25 @@ iframe_script = """
       )
     }
   }
+
+  // Prevent parent hash changes from sending a message back to the parent.
+  // This is necessary for making sure the browser back button works properly.
+  let hashChangeLock = true;
+
   window.addEventListener('hashchange', function () {
-    window.parent.postMessage(window.location.hash, '*');
+    if (!hashChangeLock) {
+      window.parent.postMessage(window.location.hash);
+    }
+    hashChangeLock = false;
+  });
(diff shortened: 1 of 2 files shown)


dwreeves:
(one sec, I'm trying to do something to make the back button work.)

tatiana (approved):
This looks great, @dwreeves ! I'm slightly worried that we currently don't have tests covering this part. I don't think we need to solve this as part of this PR, but as a follow up. I'll try to loop in people from Astronomer who may have more experience testing client side.

dwreeves:
Yeah I am just not a front-end person; I have no idea how to test this. (Or even if I'm doing things in the best way.) 😅

PR #1040, opened 2024-06-11
set task retries to 0 if TestBehavior.AFTER_ALL

davidsteinar:
## Description

if TestBehavior.AFTER_ALL we just want to run the `dbt_test` task a single time, no retries right?

## Related Issue(s)

No

## Breaking Change?

No

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -106,6 +106,7 @@ def create_test_task_metadata(
         task_args["select"] = render_config.select
         task_args["selector"] = render_config.selector
         task_args["exclude"] = render_config.exclude
+        task_args["retries"] = 0
 
     return TaskMetadata(
         id=test_task_name,


tatiana:
Hey @davidsteinar , thanks for opening this!

I'm actually curious about this. Why do you think we should override the DAG default behavior and force only test tasks that happen at the end of the `DbtDag`/`DbtTaskGroup` to not retry?

cosmos/airflow/graph.py
```diff
@@ -106,6 +106,7 @@ def create_test_task_metadata(
         task_args["select"] = render_config.select
         task_args["selector"] = render_config.selector
         task_args["exclude"] = render_config.exclude
+        task_args["retries"] = 0
```
tatiana, line 109: HI @davidsteinar ! Do you see a way of allowing users who want this behaviour to be able to configure it, but also allowing users who wish to test tasks to respect the retries they set at a DAG or task level to be respected?

davidsteinar: Hi @tatiana, I just find it strange that when the dbt tests are run and one of the tests fails, that we then rerun the task perhaps several times given the default_args retries. 
I guess I am betting on there not being issues with db connections since all other upstream tasks have to be successful for this one to be triggered. There is ofcourse a way to make this setting configurable and send it along with task_args, I am just wondering what is most useful?

tatiana: Hi @davidsteinar ! I agree with you that in many cases if a test fails, running the task once would be enough. Given the nature of Airflow being distributed (a K8s executor pod could be killed for some other reason) and potential issues with the communication in the database, I think it is very important we make this configurable, and don't enforce it on all Cosmos users.

I also believe we should not make this the default since it's non-intuitive that only tests would have a different retry than the other operators.

If you want to add some configuration such as a boolean or int `RenderConfig.tests_retry` where users could override the default retry behaviour only for tests, I think it is a promising approach.

I believe we'd like to make this work not only for `TestBehavior.AFTER_ALL`, but also for `TestBehavior.AFTER_EACH`

Issue #698, opened 2023-11-20
Expose "meta" model config dictionary on `TaskInstance` 


wseaton:
I'm working on integrating a PagerDuty setup with a Cosmos repo:

```python
def on_failure_callback(context: Context):
    summary = f"Airflow: DBT DAG {context.get('task_instance').dag_id} task failure"
    severity = "error"
    source = context.get("task_instance").task_id

    dedup_key = f"{context.get('task_instance').dag_id}-{context.get('task_instance').task_id}-{context.get('execution_date')}"

    pd_hook = PagerdutyEventsHook(pagerduty_events_conn_id="default_pagerduty")
    pd_hook.send_event(
        summary=summary,
        severity=severity,
        source=source,
        group=context.get("task_instance").dag_id,
        dedup_key=dedup_key,
        links=[{"href": context.get("task_instance").log_url, "text": "Airflow Log"}],
    )

```

This works well, but I'd love to be able to have metadata from a model like the `.meta.owner` field or even the tags attached to the `Context` object to be able to more intelligently route alerts within my callbacks. Is this possible to implement?

PR #700, opened 2023-11-20
Draft: Add config to `TaskInstance` context

wseaton:
## Description

forwards context from the parsed model config object to the Airflow `TaskInstance` `Context` so it is available in callback functions

## Related Issue(s)

closes #698 

## Breaking Change?

not that I'm aware

## Checklist

- [x] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

Would like guidance on where you'd like the tests for this, eg. mocked in the operators section or integration.

PR #1044, opened 2024-06-14
Add node config to TaskInstance Context

linchun3:
## Description

Add the node's attributes (config, tags, etc, ...) into a TaskInstance context for retrieval by callback functions in Airflow through the use of `pre_execute` to store these attributes into a task's context.

As [this PR](https://github.com/astronomer/astronomer-cosmos/pull/700/files) seems to be closed, and I have a use case for this feature, I attempt to recreate the needed feature. 

We leverage the `context_merge` utility function from Airflow to merge the extra context into the `Context` object of a `TaskInstance`.

## Related Issue(s)
closes https://github.com/astronomer/astronomer-cosmos/issues/698

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -105,7 +105,7 @@ def create_test_task_metadata(
         else:  # tested with node.resource_type == DbtResourceType.SEED or DbtResourceType.SNAPSHOT
             task_args["select"] = node.resource_name
 
-        extra_context = {f"{node.resource_type.value}_config": node.config}
+        extra_context = {"dbt_node_config": node.context_dict}
 
     elif render_config is not None:  # TestBehavior.AFTER_ALL
         task_args["select"] = render_config.select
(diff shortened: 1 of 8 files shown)


cosmos/airflow/graph.py
```diff
@@ -102,6 +104,9 @@ def create_test_task_metadata(
             task_args["select"] = f"source:{node.resource_name}"
         else:  # tested with node.resource_type == DbtResourceType.SEED or DbtResourceType.SNAPSHOT
             task_args["select"] = node.resource_name
+
+        extra_context = {f"{node.resource_type.value}_config": node.config}
```
tatiana, line 108: I'm pretty curious here, why did you add the dbt node type in the key of the config, as opposed to something like `cosmos_node_config`?

tatiana (commented):
Hi @linchun3 I left a minor question - we're close to merging this and I think we will be able to ship it in 1.5.

cosmos/airflow/graph.py
linchun3: > Since this is per task/node, wouldn't it be simpler always to have the same context identifier - and if the user is interested in the node type, they can check on the node.config

@tatiana oh yeah this makes sense, there would be less context switching required for the user using the field in the callback too. I'll make the change :) 

> Also, wouldn't it be more useful if we saved other properties of the dbt node to the context, including tags and resource type..? Perhaps we could have a property in DbtNode called context_dict to expose all the information of interest.

It'll be useful to expose all the attributes of a node, should I add that too in this PR?

linchun3: I have added the entire `DbtNode` object as a dictionary into `dbt_node_config` key as `extra_context`

tatiana (approved):
Thanks a lot for implementing this and addressing the feedback, @linchun3 ! It looks great!

Issue #925, opened 2024-04-29
Cache `profiles.yml` file when using Cosmos `ProfileMapping`


tatiana:
**Context**

In Cosmos 1.3, when users opt to use `ProfileMapping` Cosmos creates a `profiles.yml` file in potentially two places: 
* When using dbt ls to render the dbt project into an Airflow DAG
* When running dbt commands

Once we implement #924, which makes sure the profile file is created in the same way, we can extend the caching mechanism introduced in #904 also to cache the `profiles.yml` file.

Depends on: #924

**Acceptance criteria**

* [ ] Create a cache for the (if used, mocked and) non-mocked `profiles.yml` per DAG/TaskGroup when using `ProfileMapping`
* [ ] Decide on a criterion to decide when to use or not the cached `profiles.yml` (perhaps we could create a hash based on the keys in the connection and the profile args passed by the user?)
* [ ] Reuse the existing `profiles.yml` whenever possible without recreating it.
* [ ] The logs should contain less of the following:
```
Creating temporary profiles.yml at /var/folders/td/522y78v91d1f5wgh67mj3p0m0000gn/T/tmprpp66fdh/profiles.yml with the following contents:
default:
    outputs:
        dev:
            dbname: postgres
            host: 0.0.0.0
            password: '{{ env_var(''COSMOS_CONN_POSTGRES_PASSWORD'') }}'
            port: 5432
            schema: public
            type: postgres
            user: postgres
    target: dev
```

tatiana:
A key part to remember in this ticket is that we'll likely need two caches per ProfileMapping instance:
- one when it's mocked
- another when it's not mocked

tatiana:
Hi @pankajastro , following what we discussed, this PRs may help out:

Issue #647, opened 2023-11-03
Reuse connections accross entire Dag


DanMawdsleyBA:
I have run into issues when running an airflow dag with multiple tasks running in parallel which has caused tasks to hang waiting on a connection. As each tasks opens up 3 connections each it might be worth seeing if those connections can be reused across multiple tasks perhaps some sort of setup and teardown task (Airflow 2.7). Could also be extended to deal with initial profile mapping setup also 

I have raised a separate issue on the dbt snowflake repo to see if there is anything possible within dbt itself to help manage these connections better.
https://github.com/dbt-labs/dbt-snowflake/issues/824

tatiana:
@DanMawdsleyBA I've been thinking about this issue. Is it only happening when running models?

We could consider using an Airflow Snowflake (potentially async) operator to run the compiled SQL file using Cosmos. We could experiment with this using:
https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html#customizing-how-nodes-are-rendered-experimental

We may need to add the compiled SQL to the `DbtNode` definition.

If we see this approach is successful, we could consider having a Cosmos.AIRFLOW execution mode, where we'd attempt to use native Airflow operators to communicate with the database.

tatiana:
This is likely associated to the original request: https://github.com/astronomer/astronomer-cosmos/issues/925

tatiana:
We'll be doing the following ticket, and hopefully, the original issue will be solved:
1. https://github.com/astronomer/astronomer-cosmos/issues/924
2. https://github.com/astronomer/astronomer-cosmos/issues/925

pankajkoti:
hi @DanMawdsleyBA wish to check quickly if you faced this issue again of tasks hanging on waiting to get the connection?

PR #1046, opened 2024-06-17
Support to cache profiles created via `ProfileMapping`

pankajastro:
## Description

Add DBT profile caching mechanism.

1. Introduced env `enable_cache_profile` to enable or disable profile caching. This will be enabled only if global `enable_cache` is enabled. 
2. Users can set the env `profile_cache_dir_name`. This will be the name of a sub-dir inside `cache_dir` where cached profiles will be stored. This is optional, and the default name is `profile` 
3. Example Path for versioned profile: `{cache_dir}/{profile_cache_dir}/592906f650558ce1dadb75fcce84a2ec09e444441e6af6069f19204d59fe428b/profiles.yml`
4. Implemented profile mapping hashing: first, the profile is serialized using pickle. Then, the profile_name and target_name are appended before hashing the data using the SHA-256 algorithm

**Perf test result:**
In local dev env with command
```
AIRFLOW_HOME=`pwd` AIRFLOW_CONN_EXAMPLE_CONN="postgres://postgres:postgres@0.0.0.0:5432/postgres"  AIRFLOW_HOME=`pwd` AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT=20000 AIRFLOW__CORE__DAG_FILE_PROCESSOR_TIMEOUT=20000 hatch run tests.py3.10-2.8:test-performance
```

NUM_MODELS=100
- TIME=167.45248413085938 (with profile cache enabled)
- TIME=173.94845390319824 (with profile cache disabled)

NUM_MODELS=200
- TIME=376.2585120201111 (with profile cache enabled)
- TIME=418.14210200309753 (with profile cache disabled)

## Related Issue(s)

Closes: #925
Closes: #647

## Breaking Change?

No

## Checklist

### DIFF
--- a/cosmos/cache.py
+++ b/cosmos/cache.py
@@ -19,10 +19,10 @@ from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from cosmos import settings
-from cosmos.constants import DBT_MANIFEST_FILE_NAME, DBT_TARGET_DIR_NAME
+from cosmos.constants import DBT_MANIFEST_FILE_NAME, DBT_TARGET_DIR_NAME, DEFAULT_PROFILES_FILE_NAME
 from cosmos.dbt.project import get_partial_parse_path
 from cosmos.log import get_logger
-from cosmos.settings import dbt_profile_cache_path
+from cosmos.settings import cache_dir, dbt_profile_cache_dir_name, enable_cache, enable_profile_cache
 
 logger = get_logger(__name__)
 VAR_KEY_CACHE_PREFIX = "cosmos_cache__"
@@ -349,62 +349,27 @@ def delete_unused_dbt_ls_cache(
     return deleted_cosmos_variables
 
 
-def get_cache_profile_version_path(current_version: str) -> Path | None:
-    """
-    Check if profile version and correspond dbt profile yml exit in metadata.json and return the yml path or None
-    """
-    if not dbt_profile_cache_path:
-        raise Exception("profile_cache_path must be set.")
+def is_profile_cache_enabled() -> bool:
+    return enable_cache and enable_profile_cache
 
-    cache_metadata_path = Path(dbt_profile_cache_path) / "metadata.json"
-    cache_metadata_path.touch(exist_ok=True)
 
-    if cache_metadata_path.stat().st_size == 0:
-        return None
+def _get_or_create_profile_cache_dir() -> Path:
+    profile_cache_dir = cache_dir / dbt_profile_cache_dir_name
+    if not profile_cache_dir.exists():
+        profile_cache_dir.mkdir(parents=True, exist_ok=True)
+    return profile_cache_dir
 
-    with open(cache_metadata_path) as metadata_f:
-        json_data = json.load(metadata_f)
-        if json_data.get(current_version):
-            return Path(json_data.get(current_version))
+
+def get_cached_profile(version: str) -> Path | None:
+    profile_yml_path = _get_or_create_profile_cache_dir() / version[-8:] / DEFAULT_PROFILES_FILE_NAME
+    if profile_yml_path.exists() and profile_yml_path.is_file():
+        return profile_yml_path
     return None
 
 
-def _invalidate_profile_cache_metadata() -> None:
-    """
-    Invalidate (clear) the metadata stored in the profile cache.
-
-    This function clears the metadata stored in 'metadata.json' located in the specified
-    dbt_profile_cache_path directory by writing an empty dictionary to the file.
-    """
-    cache_metadata_path = Path(dbt_profile_cache_path) / "metadata.json"
-    metadata: dict[str, str] = {}
-    with open(cache_metadata_path, "w", encoding="utf-8") as file:
-        json.dump(metadata, file, indent=2)
-
-
-def cache_profile_version_path(version: str, profile_path: str) -> None:
-    """
-    Store the profile version and corresponding dbt profile path in metadata.json
-    """
-    cache_metadata_path = Path(dbt_profile_cache_path) / "metadata.json"
-
-    metadata = {}
-
-    if cache_metadata_path.exists() and cache_metadata_path.stat().st_size > 0:
-        try:
-            with open(cache_metadata_path) as file:
-                metadata = json.load(file)
-        except json.JSONDecodeError as e:
-            logger.info(f"Error decoding JSON from '{cache_metadata_path}': {str(e)}")
-            _invalidate_profile_cache_metadata()
-            return
-
-    metadata[version] = profile_path
-
-    try:
-        with open(cache_metadata_path, "w", encoding="utf-8") as file:
-            json.dump(metadata, file, indent=2)
-        logger.info(f"Updated profile cache with version '{version}' and profile path '{profile_path}'")
-    except Exception as e:
-        logger.info(f"Error writing JSON to '{cache_metadata_path}': {str(e)}")
-        _invalidate_profile_cache_metadata()
+def create_cache_profile(version: str, profile_content: str) -> Path:
+    profile_yml_dir = _get_or_create_profile_cache_dir() / version[-8:]
+    profile_yml_dir.mkdir(parents=True, exist_ok=True)
+    profile_yml_path = profile_yml_dir / DEFAULT_PROFILES_FILE_NAME
+    profile_yml_path.write_text(profile_content)
+    return profile_yml_path
--- a/cosmos/cache.py
+++ b/cosmos/cache.py
@@ -2,8 +2,8 @@ from __future__ import annotations
 
 import functools
 import hashlib
-import json
 import os
+import json
 import shutil
 import time
 from collections import defaultdict
@@ -22,6 +22,7 @@ from cosmos import settings
 from cosmos.constants import DBT_MANIFEST_FILE_NAME, DBT_TARGET_DIR_NAME
 from cosmos.dbt.project import get_partial_parse_path
 from cosmos.log import get_logger
+from cosmos.settings import dbt_profile_cache_path
 
 logger = get_logger(__name__)
 VAR_KEY_CACHE_PREFIX = "cosmos_cache__"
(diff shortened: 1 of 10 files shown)


tatiana (commented):
Hi @pankajastro , thanks a lot for advancing on this!

Some questions/comments on the overall feature:
1. So we are consistent with other enabling cache mechanism (e.g. #1014), please, rename the config to `enable_cache_profile`
2. Why are we forcing users to defined the cache path, differently from #904, where we dynamically define a path - and we give the possibility for users to override it with the `cache_dir` config, if they want to. We could perhaps allow users to force a specific profile path if they want to, by having `cache_dir_profile` - but it would be great if this was optional.
3. What does the `metadata.json` relate to? A dbt profile, an Airflow connection, a `DbtDag`, or to a `DbtTaskGroup`, to the Airflow deployment..? The name is very generic and the scope is unclear. How are we handling scenarios where concurrent tasks/DAGs would be writing at the same time to this file?
4. How sensitive is the identifier of the cache (sha256) to changes in the Airflow connection being used?

cosmos/config.py
```diff
@@ -267,35 +299,40 @@ def ensure_profile(
...
-            if desired_profile_path:
-                logger.info(
-                    "Writing profile to %s with the following contents:\n%s",
-                    desired_profile_path,
-                    profile_contents,
-                )
-                # write profile_contents to desired_profile_path using yaml library
-                desired_profile_path.write_text(profile_contents)
-                yield desired_profile_path, env_vars
+            if enable_profile_cache:
```
tatiana, line 307: This should be an end before the global setting `enable_cache` and the specific profile cache

cosmos/config.py
```diff
@@ -257,6 +259,36 @@ def validate_profiles_yml(self) -> None:
...
+    def _handle_profile_caching(self, desired_profile_path: Path | None = None, use_mock_values: bool = False) -> Path:
+        """
+        Handle the profile caching mechanism.
+
+        Check if profile object version is store in metadata.json and return the profile path
+        If Version not found in metadata.json then store version in metadata.json and create
+        corresponding profile at given path.
+        """
+        # TODO: Fix type
+        profile_version = self.profile_mapping.version(use_mock_values)  # type: ignore[union-attr]
```
tatiana, line 271:
```suggestion
        current_profile_version = self.profile_mapping.version(use_mock_values)  # type: ignore[union-attr]
```

cosmos/cache.py
```diff
@@ -171,3 +173,64 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
 
     if source_manifest_filepath.exists():
         shutil.copy(str(source_manifest_filepath), str(target_manifest_filepath))
+
+
+def get_cache_profile_version_path(current_version: str) -> Path | None:
+    """
+    Check if profile version and correspond dbt profile yml exit in metadata.json and return the yml path or None
+    """
+    if not dbt_profile_cache_path:
```
tatiana, line 182: This should be optional. We should not expect users to have to set this path - we already have a default `cache_dir`, why not reuse it?

cosmos/config.py
```diff
@@ -257,6 +259,36 @@ def validate_profiles_yml(self) -> None:
...
+                profile_path = Path(dbt_profile_cache_path) / profile_version / DEFAULT_PROFILES_FILE_NAME
+                cache_profile_version_path(profile_version, profile_path.__str__())
+                # TODO: Fix type
+                profile_contents = self.profile_mapping.get_profile_file_contents(  # type: ignore[union-attr]
+                    profile_name=self.profile_name, target_name=self.target_name, use_mock_values=use_mock_values
+                )
+
+                Path(Path(dbt_profile_cache_path) / profile_version).mkdir(parents=True, exist_ok=True)
+                logger.info("Profile not found in cache storing and using profile: %s.", profile_path)
+                profile_path.write_text(profile_contents)
```
tatiana, line 289: With the current approach, we'd be storing all the caches for all the versions of profiles - regardless if they exist or not. Ideally, we would only keep the relevant cache to the latest definition of a specific profile, overriding it if the parameters changed.

cosmos/config.py
```diff
@@ -257,6 +259,36 @@ def validate_profiles_yml(self) -> None:
         if self.profiles_yml_filepath and not Path(self.profiles_yml_filepath).exists():
             raise CosmosValueError(f"The file {self.profiles_yml_filepath} does not exist.")
 
+    def _handle_profile_caching(self, desired_profile_path: Path | None = None, use_mock_values: bool = False) -> Path:
```
tatiana, line 262: Could we find a name that represents better what this method does? Ultimately, it returns the profile path.
And the docstring could also give a brief description on what it does, before detailing the how, something along the lines:
`Return previously cached profile path or calculate and store the profile content into a file and return its path`

pankajastro:
> How sensitive is the identifier of the cache (sha256) to changes in the Airflow connection being used?

hashing connection is still a TODO item and we need to pass the actual value of a connection in the hashing input

pankajastro:
> What does the metadata.json relate to? A dbt profile, an Airflow connection, a DbtDag, or to a DbtTaskGroup, to the Airflow deployment..?

A JSON file that stores key-value pairs, where each key is a hash for a Profile Mapping object and the value is the corresponding path to the DBT profile YAML file. I'm using this file to check if we already have profile YAML for this profile object hash 

> The name is very generic and the scope is unclear.

I'm. ok to change the name.

>  How are we handling scenarios where concurrent tasks/DAGs would be writing at the same time to this file?

I haven't considered the concurrent case, do we need locking here?

pankajastro:
> Why are we forcing users to defined the cache path, differently from https://github.com/astronomer/astronomer-cosmos/pull/904, where we dynamically define a path - and we give the possibility for users to override it with the cache_dir config, if they want to. We could perhaps allow users to force a specific profile path if they want to, by having cache_dir_profile - but it would be great if this was optional.

We can make it optional, I wanted to keep the profile caching-related files in one place. Shall we keep the profile related files in `cache_dir / profile` ?

cosmos/cache.py
pankajastro: We can make it optional, I wanted to keep the profile caching-related files in one place. Shall we keep the profile related files in cache_dir / profile ?

cosmos/config.py
pankajastro: do you mean we should also check if `enable_cache` is also enable or something more?

cosmos/config.py
pankajastro: If the cache exists, i.e hash for the requested profile object exists in the cache path, then we are not rewriting it. 

> Ideally, we would only keep the relevant cache to the latest definition of a specific profile, overriding it if the parameters changed.

Could you please elaborate a bit more? Are you suggesting to keep only one profile cached per profile type? 

 was thinking that it makes sense to store in cache when object parameters change. For example, caching two profiles based on the object below, because the conn_id parameter has changed.

```
        conn_id="airflow_db",
        disable_event_tracking=True,
    ),
```

```
        conn_id="airflow_db_1",
        disable_event_tracking=True,
    ),
```

pankajastro:
Improved the below items after our first discussion and review

- Added environment variable enable_profile_cache to toggle profile caching on or off.
- Introduced environment variable profile_cache_dir, which will be created within cache_dir. The default name is 'profile', and it will store the versioned DBT profile. 
- Example Path for versioned provifile: {cache_dir}/{profile_cache_dir}/version/profiles.yml
- Implemented profile mapping hashing: first, the profile is serialized using pickle. Then, the profile_name and target_name are appended before hashing the data using the SHA-256 algorithm
- Removed metadata.json because now I'm dealing with `desired_profile_path` variable
- Added a check to enable_profile_cache only if global caching is enabled

docs/configuration/cosmos-conf.rst
```diff
@@ -54,6 +54,23 @@ This page lists all available Airflow configurations that affect ``astronomer-co
     - Default: ``None``
     - Environment Variable: ``AIRFLOW__COSMOS__DBT_DOCS_CONN_ID``
 
+.. _enable_profile_cache:
```
tatiana, line 57: So we are consistent with naming (e.g. #1014), let's use: `enable_cache_profile`. This way, all configs related to enabling cache share the same prefix.

Also, let's set this by default - so we are consistent with pre-existing cache.

docs/configuration/cosmos-conf.rst
```diff
@@ -54,6 +54,23 @@ This page lists all available Airflow configurations that affect ``astronomer-co
...
+    - Default: ``False``
+    - Environment Variable: ``AIRFLOW__COSMOS__ENABLE_PROFILE_CACHING``
+
+.. _profile_cache_path:
+
+`profile_cache_dir`_:
+    Folder path to store the DBT cached profile and related metadata
+
+    - Default: ``profile``
+    - Environment Variable: ``AIRFLOW__COSMOS__PROFILE_CACHE_DIR``
```
tatiana, line 71: I believe we should not have this setting. Why don't we create a subdirectory in the existing `cache_dir` and avoid forcing users to have to manually add another configuration?

pankajastro: This is optional and we are creating a sub dir default name  `profile` in `cache_dir` see https://github.com/astronomer/astronomer-cosmos/pull/1046/files#diff-16ea4e2752796e174ffdf472c3bd2ca64e4e7c62ab586e169f2631fc9bb7e564R182

tests/dbt/test_graph.py
```diff
@@ -429,7 +429,7 @@ def test_load_via_dbt_ls_does_not_create_target_logs_in_original_folder(
 
     used_cwd = Path(mock_popen.call_args[0][0][-5])
     assert used_cwd != project_config.dbt_project_path
-    assert not used_cwd.exists()
+    # assert not used_cwd.exists()
```
pankajastro, line 432: This check is failing at the moment

docs/configuration/cosmos-conf.rst
```diff
@@ -70,6 +70,23 @@ This page lists all available Airflow configurations that affect ``astronomer-co
...
+
+`enable_cache_profile`_:
+    Enable caching for the DBT profile.
+
+    - Default: ``True``
+    - Environment Variable: ``AIRFLOW__COSMOS__ENABLE_PROFILE_CACHING``
+
+.. _profile_cache_dir:
+
+`profile_cache_dir`_:
```
pankajastro, line 83:
```suggestion
`profile_cache_dir_name`_:
```

tatiana (commented):
HI @pankajastro thanks for addressing all the feedback!

Once you are able to uncomment that assert in the test and rename the config, I believe we're good to go! 🎉 

All the comments/feedback were addressed. Following up on:
> I haven't considered the concurrent case, do we need locking here?

I believe once we removed the `metadata.json` file, we're in a much safer position, since we reduce significantly concurrent writes to the same file. I suggest we go with the current solution and we review if users report any issues.

cosmos/profiles/base.py
```diff
@@ -82,6 +84,23 @@ def __init__(
...
+        :param target_name: Name of the DBT target
+        :param mock_profile: If True, use a mock profile.
+        """
+        if mock_profile:
+            profile = self.mock_profile
+        else:
+            profile = self.profile
+        profile["profile_name"] = profile_name
+        profile["target_name"] = target_name
+        hash_object = hashlib.sha256(pickle.dumps(profile))
```
jlaneve, line 101: do we need to pickle here? or can we just json serialize/deserialize?

pankajastro: I can opt for another serializer like `json.dumps` based on the complexity of the profile object. However, since pickle is already installed and works well to serialize complex objects, I'm using it here. I'm curious if there are any potential drawbacks to using pickle in this context?

tests/dbt/test_graph.py
```diff
@@ -429,7 +432,10 @@ def test_load_via_dbt_ls_does_not_create_target_logs_in_original_folder(
 
     used_cwd = Path(mock_popen.call_args[0][0][-5])
     assert used_cwd != project_config.dbt_project_path
-    assert not used_cwd.exists()
+    # When the cache profile is enabled, the profile is created in {cache_dir}/profile/{version}/profiles.yml
+    # rather than in a temporary file. This ensures it is persisted for future use and remains uncleared.
+    if not enable_cache_profile:
+        assert not used_cwd.exists()
```
pankajastro, line 438: @tatiana, `used_cwd` pointing to the dir where profiles.yml exists. When cache profile is enabled, it is created inside `cache_dir/profile/version`, ensuring persistence. If profile cache is enabled, does not it make sense, or I'm still missing something?

CI log:
```
FAILED tests/dbt/test_graph.py::test_load_via_dbt_ls_does_not_create_target_logs_in_original_folder[True] - AssertionError: assert not True
 +  where True = <bound method Path.exists of PosixPath('/tmp/cosmos/profile/470f1452ea88562203ec89a49bab7e2c6043ef3b7f93e61ce30772e24aa38289')>()
 +    where <bound method Path.exists of PosixPath('/tmp/cosmos/profile/470f1452ea88562203ec89a49bab7e2c6043ef3b7f93e61ce30772e24aa38289')> = PosixPath('/tmp/cosmos/profile/470f1452ea88562203ec89a49bab7e2c6043ef3b7f93e61ce30772e24aa38289').exists
```
Ref CI: https://github.com/astronomer/astronomer-cosmos/actions/runs/9679063910/job/26704228626

tests/dbt/test_graph.py
```diff
@@ -429,7 +429,7 @@ def test_load_via_dbt_ls_does_not_create_target_logs_in_original_folder(
 
     used_cwd = Path(mock_popen.call_args[0][0][-5])
```
tatiana, line 433: @pankajastro could you confirm the value of `used_cwd`, please?

pankajastro: When the cache profile is enabled 
local mac used_cwd: `PosixPath('/var/folders/t4/sk99_rj509127895hqk1m9x00000gn/T/cosmos/profile/470f1452ea88562203ec89a49bab7e2c6043ef3b7f93e61ce30772e24aa38289')`
CI used_cwd: `PosixPath('/tmp/cosmos/profile/470f1452ea88562203ec89a49bab7e2c6043ef3b7f93e61ce30772e24aa38289'`

tatiana: As we discussed, we've changed to `Path(mock_popen.call_args[0][0][5])`

cosmos/profiles/base.py
tatiana: Following up, @pankajastro , what we agreed:
1. It is worth changing from pickle to `json.dumps(sort_keys=True)
 for two reasons: the value needs to be deterministic and it must be safe to store it in a remote object store, something that's planned for Cosmos 1.6.
2. Add a test to validate that a given profile always has the same cache version. You may need to set a different value for Linux and MacOS.

tatiana (approved):
Looks great, thanks for addressing all the feedback, @pankajastro !

Issue #1008, opened 2024-05-29
Permission issue with Cosmos cache in some restricted environments


ghjklw:
Cosmos caching mecanism can trigger permission errors when the user does not have permission to chmod files in the cache location.

This can happen, for example, when using an Azure Files volume with AKS/Azure container instances/Azure container apps. A known (unfortunate) limitation of this system is that you can't change a file permissions, permissions are defined at mount time: https://stackoverflow.com/questions/58301985/permissions-on-azure-file.

I get the following error when trying to use Cosmos cache with such a volume:

```
[2024-05-29, 16:57:16 CEST] {taskinstance.py:2905} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 465, in _execute_task
    result = _execute_callable(context=context, **execute_callable_kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 432, in _execute_callable
    return execute_callable(context=context, **execute_callable_kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/baseoperator.py", line 400, in wrapper
    return func(self, *args, **kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/cosmos/operators/base.py", line 266, in execute
    self.build_and_run_cmd(context=context, cmd_flags=self.add_cmd_flags())
  File "/home/airflow/.local/lib/python3.10/site-packages/cosmos/operators/local.py", line 499, in build_and_run_cmd
    result = self.run_command(cmd=dbt_cmd, env=env, context=context)
  File "/home/airflow/.local/lib/python3.10/site-packages/cosmos/operators/local.py", line 366, in run_command
    cache._update_partial_parse_cache(partial_parse_file, self.cache_dir)
  File "/home/airflow/.local/lib/python3.10/site-packages/cosmos/cache.py", line 110, in _update_partial_parse_cache
    shutil.copy(str(latest_partial_parse_filepath), str(cache_path))
  File "/usr/local/lib/python3.10/shutil.py", line 418, in copy
    copymode(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/local/lib/python3.10/shutil.py", line 307, in copymode
    chmod_func(dst, stat.S_IMODE(st.st_mode))
[...]

PR #1051, opened 2024-06-18
Fix AKS permission error in restricted env

pankajastro:
## Description
~shutil.copy includes permission copying via chmod.
If the user lacks permission to run chmod, a PermissionError occurs.
To avoid this, we split the operation into two steps:
first, copy the file contents; then, copy metadata if feasible without raising exceptions.
Step 1: Copy file contents (no metadata)
Step 2: Copy file metadata (permission bits and other metadata) without raising exception~

use shutil.copyfile(...) instead of shutil.copy(...) to avoid running chmod

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/1008

## Breaking Change?

No

## Checklist

### DIFF
--- a/cosmos/cache.py
+++ b/cosmos/cache.py
@@ -97,10 +97,8 @@ def _get_latest_partial_parse(dbt_project_path: Path, cache_dir: Path) -> Path |
     return None
 
 
-def _copy(src, dst, *, follow_symlinks=True):
-    """
-    Copy a file from `src` to `dst` and preserve metadata if possible.
-    """
+def _copy(src: str, dst: str, *, follow_symlinks: bool = True) -> None:
+    """Copy a file from `src` to `dst` and preserve metadata if possible."""
     # Handle destination as directory
     if os.path.isdir(dst):
         dst = os.path.join(dst, os.path.basename(src))
--- a/cosmos/cache.py
+++ b/cosmos/cache.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import os
 import shutil
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


pankajastro:
Hi @ghjklw, could you please test this one? Thank you!

ghjklw:
Thanks @pankajastro!

I can confirm it works as expected:

- I get the following logs "INFO - (astronomer-cosmos) - Failed to copy the partial parse file metadata"
- I can see that the cache files are created
- And I see that they're being used

👌

cosmos/cache.py
```diff
@@ -96,6 +97,25 @@ def _get_latest_partial_parse(dbt_project_path: Path, cache_dir: Path) -> Path |
...
+    if os.path.isdir(dst):
+        dst = os.path.join(dst, os.path.basename(src))
+    # shutil.copy includes permission copying via chmod.
+    # If the user lacks permission to run chmod, a PermissionError occurs.
+    # To avoid this, we split the operation into two steps:
+    # first, copy the file contents; then, copy metadata if feasible without raising exceptions.
+    # Step 1: Copy file contents (no metadata)
+    shutil.copyfile(src, dst, follow_symlinks=follow_symlinks)
+
+    try:
```
tatiana, line 112: @pankajastro, what do you think if we simplified and didn't try to copy the file metadata? Step 1 should be enough

pankajastro: Look like we are using modification time for partial parse and copyfile does not update that ?
https://github.com/astronomer/astronomer-cosmos/issues/1008#issuecomment-2173848972

tatiana: This should be fine. This method will be invoked when it copies the most up-to-date cache to the "persistent" cache. The fact that its timestamp is newer than the original file should not be an issue. Next time Cosmos runs, it will try to use this "persistent" cache. If, for any reason, it's outdated, it should be overridden.

pankajastro: thanks for the clarification @tatiana  I have applied the feedback in 226f65440e91f1c547d711564a2062c9531afe79

tatiana (commented):
Thanks a lot for fixing this, @pankajastro ! I suggest we add a test that covers the problem we solved, related to permission. To make sure we don't have a regression in the future.

tatiana (approved):
Thanks, @pankajastro , looking good!

PR #1054, opened 2024-06-19
Change example DAGs to use `example_conn` as opposed to `airflow_db`

tatiana:
Using the Airflow metadata database connection as an example connection is misleading.

The mismatch in the environment variable value used in the Cosmos integration tests, particularly with sqlite as the Airflow metadata database, is an issue that can hide other underlining issues.

This PR decouples the test connection used by Cosmos example DAGs from the Airflow metadata Database connection.

Since this change affects the Github action configuration, it will only work for the branch-triggered GH action runs, such as:
https://github.com/astronomer/astronomer-cosmos/actions/runs/9596066209

Since this is a breaking change to the CI script itself, all the tests `pull_request_target` are expected to fail during the PR - and will pass once this is merged to `main`.

This improvement was originally part of #1014

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, rename-test-conn]
+    branches: [main]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,9 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
-  pull_request_target: # Also run on pull requests originated from forks
     branches: [main, rename-test-conn]
+  pull_request_target: # Also run on pull requests originated from forks
+    branches: [main]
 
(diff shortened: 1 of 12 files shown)


.github/workflows/test.yml
```diff
@@ -4,7 +4,7 @@ on:
   push: # Run on pushes to the default branch
     branches: [main]
   pull_request_target: # Also run on pull requests originated from forks
-    branches: [main]
+    branches: [main, rename-test-conn]
```
tatiana, line 7: Before this PR is merged, we should revert this change:
```suggestion
    branches: [main]
```

tests/dbt/test_graph.py
```diff
@@ -265,7 +265,7 @@ def test_load_automatic_without_manifest_with_profile_mapping(mock_load_via_dbt_
         profile_name="test",
         target_name="test",
         profile_mapping=PostgresUserPasswordProfileMapping(
-            conn_id="airflow_db",
+            conn_id="example",
```
pankajkoti, line 268:
```suggestion
            conn_id="example_conn",
```

pankajkoti (approved):
LGTM. One small correction suggested inline

PR #1058, opened 2024-06-20
Add test for response header

dwreeves:
Follow up from: #1055


tatiana (approved):
Thanks a lot, @dwreeves !

Issue #1041, opened 2024-06-11
[Bug] Unable to do partial parsing after disabling Cosmos caching.


josefbits:
### Astronomer Cosmos Version

Other Astronomer Cosmos version (please specify below)

### If "Other Astronomer Cosmos version" selected, which one?

1.4.3

### dbt-core version

1.7.15

### Versions of dbt adapters

dbt-postgres 1.7.15

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

VIRTUALENV

### InvocationMode

DBT_RUNNER

### airflow version

2.8.2

### Operating System

Ubuntu 22.04.4 LTS

### Deployment

Docker-Compose

### Deployment details

- Docker version 25.0.3, build 4debf41
- Docker Compose version v2.24.6

### What happened?

After setting `AIRFLOW__COSMOS__ENABLE_CACHE=False` dbt tasks are unable to run partial parsing due to the error "Unable to do partial parsing because saved manifest not found. Starting full parse."

The dbt artifacts are pre-generated in CI and the entire `target/` directory synced to the dbt project path on the Airflow host.

### Relevant log output

```shell
**** With AIRFLOW__COSMOS__ENABLE_CACHE=False ****
08a120f7b403
*** Found local files:
***   * /opt/airflow/logs/dag_id=cosmos_dag/run_id=manual__2024-06-11T13:37:41.235278+00:00/task_id=run_models.raw_payments_seed/attempt=1.log
[2024-06-11, 13:37:47 UTC] {taskinstance.py:1979} INFO - Dependencies all met for dep_context=non-requeueable deps ti=<TaskInstance: cosmos_dag.run_models.raw_payments_seed manual__2024-06-11T13:37:41.235278+00:00 [queued]>
[2024-06-11, 13:37:47 UTC] {taskinstance.py:1979} INFO - Dependencies all met for dep_context=requeueable deps ti=<TaskInstance: cosmos_dag.run_models.raw_payments_seed manual__2024-06-11T13:37:41.235278+00:00 [queued]>
[2024-06-11, 13:37:47 UTC] {taskinstance.py:2193} INFO - Starting attempt 1 of 3
[2024-06-11, 13:37:47 UTC] {taskinstance.py:2214} INFO - Executing <Task(DbtSeedVirtualenvOperator): run_models.raw_payments_seed> on 2024-06-11 13:37:41.235278+00:00
[...]

PR #1070, opened 2024-06-27
Support partial parsing when cache is disabled

tatiana:
Partial parsing support was introduced in #800 and improved in #904 (caching). However, as the caching layer was introduced, we removed support to use partial parsing if the cache was disabled.

This PR solves the issue.

Fix: #1041


cosmos/dbt/graph.py
```diff
@@ -424,18 +428,21 @@ def load_via_dbt_ls_without_cache(self) -> None:
...
+            logger.debug(f"Content of the dbt project dir {project_path}: `{os.listdir(project_path)}`")
             tmpdir_path = Path(tmpdir)
 
             create_symlinks(project_path, tmpdir_path, self.render_config.dbt_deps)
 
-            if self.project.partial_parse and self.cache_dir:
-                latest_partial_parse = cache._get_latest_partial_parse(project_path, self.cache_dir)
+            latest_partial_parse = None
+            if self.project.partial_parse:
+                if self.should_use_partial_parse_cache() and self.cache_dir:
```
pankajastro, line 438: Do we still need check `self.cache_dir` here, considering that `should_use_partial_parse_cache` already performs check `bool(self.cache_dir)`?

pankajastro (approved):
lgtm

cosmos/dbt/graph.py
```diff
@@ -424,18 +428,21 @@ def load_via_dbt_ls_without_cache(self) -> None:
...
+                    latest_partial_parse = cache._get_latest_partial_parse(project_path, self.cache_dir)
+                else:
+                    latest_partial_parse = get_partial_parse_path(project_path)
+
+            if latest_partial_parse is not None and latest_partial_parse.exists():
                 logger.info("Partial parse is enabled and the latest partial parse file is %s", latest_partial_parse)
-                if latest_partial_parse is not None:
-                    cache._copy_partial_parse_to_project(latest_partial_parse, tmpdir_path)
+                cache._copy_partial_parse_to_project(latest_partial_parse, tmpdir_path)
```
pankajkoti, line 445: I am wondering if we need another check here to see if cache is enabled and only then copy it?

cosmos/dbt/graph.py
```diff
@@ -470,9 +477,9 @@ def load_via_dbt_ls_without_cache(self) -> None:
                 self.nodes = nodes
                 self.filtered_nodes = nodes
 
-            if self.project.partial_parse and self.cache_dir:
+            if self.should_use_partial_parse_cache():
                 partial_parse_file = get_partial_parse_path(tmpdir_path)
-                if partial_parse_file.exists():
+                if partial_parse_file.exists() and self.cache_dir:
```
pankajkoti, line 482:
```suggestion
                if partial_parse_file.exists():
```
based on Pankaj's comment above that we already check this prior to getting here.

cosmos/dbt/graph.py
pankajkoti: +1

pankajkoti (approved):
LGTM. minor questions inline

cosmos/dbt/graph.py
tatiana: @pankajastro @pankajkoti MyPy didn't like it without it :( That's why I added it there. Do you have any suggestions

cosmos/dbt/graph.py
tatiana: This line needs to be run regardless of caching - this how we're fixing the original issue

cosmos/dbt/graph.py
pankajkoti: okay looks like MyPy is not able to figure out what's happening in the previous method. So I guess we can keep it for now until find a better resolution if any.

cosmos/dbt/graph.py
tatiana: Resolving this as https://github.com/astronomer/astronomer-cosmos/pull/1070#discussion_r1657159419

pankajastro (approved):

Issue #610, opened 2023-10-18
Support caching virtualenvs created when using ExecutionMode.VIRTUALENV


tatiana:
As of Cosmos 1.2, when using `ExecutionMode.VIRTUALENV`, each task will create a new Python virtual environment, in a temporary directory:
https://github.com/astronomer/astronomer-cosmos/blob/a433f15d6fd20f5fd713268040eba09727db4a2c/cosmos/operators/virtualenv.py#L68

This can cause delays, as discussed in the Slack thread:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1697614400289939

This could be improved, assuming the Airflow worker node is reused to run multiple tasks.

**Proposal**
Persist the virtual environment directory if users set the configuration: `ExecutionConfig.virtualenv_path`.

LennartKloppenburg:
As discussed with @tatiana I'd be happy to take a first stab at this :)

tatiana:
Thank you very much for taking the lead on this, @LennartKloppenburg ! Please, let us know if you'd like any support

LennartKloppenburg:
Thinking about whether it'd make sense to add a clean-up callback to delete the virtual env that should be executed after the whole DAG has completed?
Or should we defer this responsibility/flexibility to the end-user?

LennartKloppenburg:
More things to consider:
1. It could be useful to also centralize and cache the installation of `dbt deps`, as this is currently still happening in every task with this PR. Implementing this too would invalidate the `install_deps` flag on the operator level (on `DbtLocalBaseOperator` for instance) and make more sense on the ExecutionConfig instead, at least when using using `VirtualEnv` as ExecutionMode.

2. If we let an operator create a virtual env and cache it, it can cause concurrency issues with other operators simultaneously doing the same thing. My changes work when starting with 1 in isolation and then running the rest...
Perhaps the most elegant solution would be to add a "set-up" operator at the start of a VirtualEnv-based DAG that sets up the env, installs python requirements, installs dbt deps and everything, and then another one that tears it down at the end. That would take care of a lot of the complexity we'd otherwise end up implementing on the operator level. 

The current PR would also split the operator's responsibility into "do whatever DBT thing it needs to do" and "optionally set up a virtual env if necessary" and that feels like an anti-pattern.

tatiana:
@LennartKloppenburg, we discussed this last week, and I'll add some thoughts here as well:

[...]

PR #1079, opened 2024-07-05
Support persisting the `LoadMode.VIRTUALENV` directory

tatiana:
## Description

Added `virtualenv_dir` as an option to `ExecutionConfig` which is then propagated downstream to `DbtVirtualenvBaseOperator`.

The following now happens:
- If the flag is set, the operator will attempt to locate the `venv`'s `python` binary under the provided `virtualenv_dir`. 
    - If so, it will conclude that the `venv` exists and continues without creating a new one.
    - If not, it will create a new one at `virtualenv_dir`
- If the flag is not set, simply continue using the temporary directory solution that was already in place.

## Impact
A very basic test using a local `docker compose` set-up as per the contribution guide and the [example_virtualenv](https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/example_virtualenv.py) DAG saw the DAG's runtime go down from **2m31s** to just **32s**. I'd this improvement to be even more noticeable with more complex graphs and more python requirements.
 
## Related Issue(s)
Closes: #610 
Follow up ticket: #1157

## Breaking Change?
None, the flag is optional and is ignored (with a [warning](https://github.com/astronomer/astronomer-cosmos/compare/main...LennartKloppenburg:astronomer-cosmos:feature/cache-virtualenv?expand=1#diff-61b585fb903927b6868b9626c95e0ec47e3818eb477d795ebd13b0276d4fd76cR125))  when used outside of `VirtualEnv` execution mode.

## Important notice

Most of the changes in this PR were originally implemented in PR #611 by @LennartKloppenburg. It became stale over the last few months due to limited maintainer availability. Our sincere apologies to the original author.

What was accomplished since:
1. Rebased
2. Fixed conflicts
3. Fixed failing tests
4. Introduced new tests

Co-authored-by: Lennart Kloppenburg <lennartkloppenburg@live.nl>


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, feature/cache-virtualenv]
```
```suggestion
    branches: [main]
```

cosmos/operators/virtualenv.py
```diff
@@ -101,11 +113,113 @@ def run_subprocess(self, command: list[str], env: dict[str, str], cwd: str) -> F
...
+
+    # @depends_on_virtualenv_dir
+    def _is_lock_available(self) -> bool:
+        if self.__lock_file.is_file():
+            with open(self.__lock_file) as lf:
+                pid = int(lf.read())
+
+                self.log.info(f"Checking for running process with PID {pid}")
+                try:
+                    _process_running = psutil.Process(pid).is_running()
```
tatiana, line 192: Given that we're using subprocesses as well, we must check not only if the pid is running but also if its parent (`os. getppid()`) and potentially the children as well, before releasing the lock

tatiana: Since this is being created during the task execution, in the worker node, I don't see a scenario where the parent process would be creating the lock - I think we have a good starting point and we can introduce additional logic if needed in future.

cosmos/operators/virtualenv.py
```diff
@@ -44,18 +59,24 @@ class DbtVirtualenvBaseOperator(DbtLocalBaseOperator):
...
 
+    template_fields = DbtLocalBaseOperator.template_fields + ("virtualenv_dir", "is_virtualenv_dir_temporary")  # type: ignore[operator]
+
     def __init__(
         self,
         py_requirements: list[str] | None = None,
         pip_install_options: list[str] | None = None,
         py_system_site_packages: bool = False,
+        virtualenv_dir: Path | None = None,
+        is_virtualenv_dir_temporary: bool = False,
```
pankajastro, line 70: shall we add these param in docstring?

cosmos/operators/virtualenv.py
```diff
@@ -102,11 +117,110 @@ def run_subprocess(self, command: list[str], env: dict[str, str], cwd: str) -> F
...
+                py_bin = self.prepare_virtualenv()
+            return py_bin
+
+        # Use a reusable virtualenv
+        self.log.info(f"Checking if the virtualenv lock {str(self._lock_file)} exists")
+        while not self._is_lock_available() and self.max_retries_lock:
+            logger.info("Waiting for virtualenv lock to be released")
+            time.sleep(1)
+            self.max_retries_lock -= 1
+
```
pankajastro, line 166: just curious about how often we need to wait for the lock to be released. 
Do we have any data on how performance is affected if 100 parallel tasks try to use this virtual environment?

pankajastro (approved):
Looks good. Thanks @tatiana!

Just a few minor comments

cosmos/operators/virtualenv.py
```diff
@@ -102,11 +117,110 @@ def run_subprocess(self, command: list[str], env: dict[str, str], cwd: str) -> F
...
+                self.virtualenv_dir = Path(tempdir)
+                py_bin = self.prepare_virtualenv()
+            return py_bin
+
+        # Use a reusable virtualenv
+        self.log.info(f"Checking if the virtualenv lock {str(self._lock_file)} exists")
+        while not self._is_lock_available() and self.max_retries_lock:
+            logger.info("Waiting for virtualenv lock to be released")
+            time.sleep(1)
+            self.max_retries_lock -= 1
```
pankajkoti, line 165: what should happen when the max_retries_lock are exceeded? I am interpreting it as that when they're exhausted, we still proceed onto acquiring the lock.

cosmos/operators/virtualenv.py
```diff
@@ -102,11 +117,110 @@ def run_subprocess(self, command: list[str], env: dict[str, str], cwd: str) -> F
...
+
+        self.log.info(f"Acquiring the virtualenv lock")
+        self._acquire_venv_lock()
```
pankajkoti, line 168: Could this be a candidate for a potential race? I mean the lock is available now from the previous check, but 2 processes try to acquire the lock at the same time? 

Should we double-check in the acquire_venv_lock that the previous lock (if any) has been released?

cosmos/operators/virtualenv.py
```diff
@@ -102,11 +117,110 @@ def run_subprocess(self, command: list[str], env: dict[str, str], cwd: str) -> F
...
-        logger.info(output)
+        try:
+            super().execute(context)
+        finally:
+            self.clean_dir_if_temporary()
+
+    def on_kill(self) -> None:
+        self.clean_dir_if_temporary()
+
+    def prepare_virtualenv(self) -> str:
```
pankajkoti, line 137: can we have some identifier here to differentiate with the `prepare_virtualenv` method from `airflow.utils.python_virtual_env` below in `py_bin = prepare_virtualenv(....`? 

Maybe we could have a leading underscore for this method? or something else to avoid potential confusion while reading the code. My bad, but I first thought we are recursively calling the same method 🙃

cosmos/operators/virtualenv.py
pankajkoti: +1

cosmos/converter.py
```diff
@@ -220,12 +220,26 @@ def __init__(
...
+        dbt_vars = copy.deepcopy(project_config.dbt_vars or operator_args.get("vars"))
+
+        if execution_config.execution_mode != ExecutionMode.VIRTUALENV and execution_config.virtualenv_dir is not None:
+            logger.warning(
+                "`ExecutionConfig.virtualenv_dir` is only supported when \
+                ExecutionConfig.execution_mode is set to ExecutionMode.VIRTUALENV."
+            )
+
+        if not operator_args:
+            operator_args = {}
```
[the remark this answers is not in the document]
tatiana: @pankajastro, we should avoid mutable default arguments in Python because they are evaluated only once the function is defined, not every time the function is called. This behavior of Python can lead to several unforeseen issues that can be hard to troubleshoot.

There were recent discussions on this in the Python mailing list about changing the default behaviour:
https://discuss.python.org/t/revisit-mutable-default-arguments/37525/44

cosmos/operators/virtualenv.py
```diff
@@ -102,11 +117,110 @@ def run_subprocess(self, command: list[str], env: dict[str, str], cwd: str) -> F
...
+            python_bin=PY_INTERPRETER,
+            system_site_packages=self.py_system_site_packages,
+            requirements=self.py_requirements,
+            pip_install_options=self.pip_install_options,
+        )
+        return py_bin
+
+    def _get_or_create_venv_py_interpreter(self) -> str:
+        """Helper method that parses virtual env configuration
+        and returns a DBT binary within the resulting virtualenv"""
```
pankajkoti, line 150:
```suggestion
        and returns a Python binary within the resulting virtualenv"""
```

cosmos/operators/virtualenv.py
tatiana: We don't have this data right now, and I agree it would be great if we could get this. @pankajastro , could you help test this, please?

pankajkoti (commented):
Sorry for delayed review, missed reviewing this earlier

looks great to me mostly, some questions inline

cosmos/settings.py
```diff
@@ -22,6 +22,7 @@
 dbt_docs_index_file_name = conf.get("cosmos", "dbt_docs_index_file_name", fallback="index.html")
 enable_cache_profile = conf.getboolean("cosmos", "enable_cache_profile", fallback=True)
 dbt_profile_cache_dir_name = conf.get("cosmos", "profile_cache_dir_name", fallback="profile")
+virtualenv_max_retries_lock = conf.getint("cosmos", "virtualenv_max_retries_lock", fallback=120)
```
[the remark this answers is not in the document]
tatiana: Resolved in a712c274

cosmos/operators/virtualenv.py
tatiana: Yes, that's what's happening right now

cosmos/operators/virtualenv.py
tatiana: We try to wait for the lock to be released before. The consequence of two processes running pip concurrently wouldn't be disastrous if they happened to try to acquire the lock after x time, and it took a bit longer.

cosmos/operators/virtualenv.py
tatiana: Addressed in b465b2a21e7ee21a6017e379a16a607ea2259859

cosmos/operators/virtualenv.py
tatiana: This method was removed

pankajkoti (approved):

cosmos/operators/virtualenv.py
tatiana: Done

cosmos/operators/virtualenv.py
tatiana: We spoke to @pankajastro and we decided to release this feature to collect feedback from the community, and get support from others on testing this

cosmos/operators/virtualenv.py
tatiana: Spoke to @pankajkoti and we decided to wait for community feedback to see if this behaviour works well in practice or if we should error.

tests/dbt/test_graph.py
```diff
@@ -1518,7 +1518,7 @@ def test_save_dbt_ls_cache(mock_variable_set, mock_datetime, tmp_dbt_project_dir
     hash_dir, hash_args = version.split(",")
     assert hash_args == "d41d8cd98f00b204e9800998ecf8427e"
     if sys.platform == "darwin":
-        assert hash_dir == "a9879ec2ec503b0fe023d059caf50d41"
+        assert hash_dir == "6534a9963174a3adf58725f7f06a30a1"
     else:
         assert hash_dir == "9001bedf4aa8a329f7b669c89f337c24"
```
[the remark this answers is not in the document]
tatiana: We will check this in the future. We decided not to comment on this line intentionally since we care about the hash being deterministic. There may be something specific to MacOS - since the Linux hash seems stable. But let's see in a different PR/work.

PR #1081, opened 2024-07-05
Fix getting temporary AWS credentials with assume_role

piotrkubicki:
## Description

When Airflow is getting temporary AWS credentials by assuming role with `role_arn` as only `Connection` parameter, this cause task to fail due to missing credentials. This is due to the latest changes related to profile caching. The `env_vars` are accessed before `profile` which, in this case, means required values are not populated yet. 

## Related Issue(s)

None, but I can create one if needed?

## Breaking Change?

None

## Checklist

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -279,13 +279,6 @@ class ProfileConfig:
             logger.info("Profile not found in cache storing and using profile: %s.", profile_path)
             return profile_path
 
-    def _get_env_vars(self, use_mock_values: bool = False) -> dict[str, str]:
-        """
-        Return profile env_vars.
-        """
-        return {} if use_mock_values else self.profile_mapping.env_vars
-
-
     @contextlib.contextmanager
     def ensure_profile(
         self, desired_profile_path: Path | None = None, use_mock_values: bool = False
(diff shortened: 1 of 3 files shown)


tatiana:
Thank you very much, @piotrkubicki , this is looking great! Please, could you add some tests so we make sure we don't have regressions in - the future? We'll release this as part of 1.5.1, next week.

piotrkubicki:
Hi @tatiana, thanks. I wanted to add some test from the beginning as I also worried that the bug may be re-introduced in the future, but had to familiarise a bit more with the code first.
Just added few tests, hope that would do, if not any suggestions will be appreciated.

pankajastro:
Hmm, maybe the same issue has been reported here https://apache-airflow.slack.com/archives/C059CC42E9W/p1719956963750389

pankajastro (approved):
Nice!

tatiana (approved):
This is great, thank you very mich, @piotrkubicki !
We'll be releasing this in Cosmos 1.5.1, later this week.

piotrkubicki:
Great news, thank you 😄

Issue #1065, opened 2024-06-26
[Feature] add task owner to dbt operator


wornjs:
### Description

add owner information to DbtxxxOperator 

### Use case/motivation

Each model is created by multiple da.
When a failure occurs, I want to check the owner on the UI without checking the source.

### Are you willing to submit a PR?

- [X] Yes, I am willing to submit a PR!

tatiana:
@wornjs, this looks like a cool feature!

How do you currently represent in dbt a model owner? Do you use the following..?
https://docs.getdbt.com/reference/resource-configs/meta#designate-a-model-owner

It does not look like a big task to map this into the Airflow argument `owner`.

Do you know if dbt also supports setting the `owner` information at a pipeline level? Ideally we'd handle both cases.

PR #1082, opened 2024-07-07
Add task owner to dbt operator

wornjs:
Set dbt meta owner as task owner in all Cosmos `DbtxxxOperator`s.

Closes: #1065


wornjs:
here is  the result in airflow web ui
스크린샷 2024-07-07 오후 11 18 59

스크린샷 2024-07-07 오후 11 19 24

cosmos/dbt/graph.py
```diff
@@ -76,6 +76,10 @@ def name(self) -> str:
         """
         return self.resource_name.replace(".", "_")
 
+    @property
+    def owner(self) -> str:
+        return str(self.config.get("meta", {}).get("owner", "airflow"))
```
tatiana, line 81: An alternative to hard-code `airflow` could be to retrieve the default value from Airflow:
```
from airflow.configuration import conf

default_owner = conf.get("operators", "default_owner")
```

If you want to follow this path, please define `default_owner` in `settings.py`, together with other configurations.

tatiana (commented):
This is looking great, @wornjs !

I left  a minor comment (sorry for doing this while the PR is in draft) - we can finish reviewing once the checks are passing and the PR is ready for review.

cosmos/dbt/graph.py
wornjs: I'll check it out tonight

wornjs: if there is no owner in dbt config 
It would be better to specify the owner for the Airflow DAG.

wornjs:
@tatiana 
plz check this

cosmos/airflow/graph.py
```diff
@@ -106,14 +106,20 @@ def create_test_task_metadata(
...
+        task_owner = node.owner
 
     elif render_config is not None:  # TestBehavior.AFTER_ALL
         task_args["select"] = render_config.select
         task_args["selector"] = render_config.selector
         task_args["exclude"] = render_config.exclude
+        task_owner = ""
+
+    else:
+        task_owner = ""
```
tatiana, line 118: @wornjs thanks for addressing the feedback! Could you add a test to cover this use case as well? It is looking good and we'll release this as part of Cosmos 1.6.

tests/airflow/test_graph.py
```diff
@@ -42,7 +42,7 @@
     depends_on=[parent_seed.unique_id],
     file_path=SAMPLE_PROJ_PATH / "gen2/models/parent.sql",
     tags=["has_child"],
-    config={"materialized": "view"},
+    config={"materialized": "view", "meta": {"owner": "parent_node"}},
```
wornjs, line 45: @tatiana 
here is the test code to check task owner

cosmos/dbt/graph.py
tatiana: Thanks a lot, @wornjs !

cosmos/airflow/graph.py
tatiana: I just refactored the code not to need this conditional, I believe coverage will be happy 🤞

tatiana: It worked! I'll rebase and merge, thanks for the contribution, @wornjs !

tatiana (approved):

Issue #930, opened 2024-04-29
Cache dbt deps lock file


tatiana:
**Context**

dbt 1.7 creates/uses a dbt dependencies [lock file](https://docs.getdbt.com/reference/commands/deps#predictable-package-installs): `package-lock.yml`. 

As of Cosmos 1.3, we did not take advantage of this feature. This task aims to cache this file using the same caching mechanism introduced in #904, so running `dbt deps` is quicker.

**Acceptance criteria**

* [ ] When using `ExecutionMode.LOCAL` and `ExecutionMode.VIRTUALENV`, Cosmos should cache and try to reuse the `package-lock.yml` file in a similar way it does to `partial_parse.msgpack`

PR #1086, opened 2024-07-10
Cache package-lock.yml file

pankajastro:
## Description
This PR aims to cache the package-lock.yml in `cache_dir/dbt_project`

Since dbt version 1.7.0, executing the dbt deps command results in the generation of a package-lock.yml file. This file pins the dependencies and their versions for the dbt project. dbt uses this file to install packages, ensuring predictable and consistent package installations across environments.

- This feature is enabled only if the user checks in package-lock.yml in their dbt project. Also, I'm assuming if `package-lock.yml` their dbt-core version is >= 1.7.0 since this feature is available for only dbt >= 1.7.0
- package-lock.yml also contains the sha1_hash of the packages. This is used to check if the cached package-lock.yml is outdated or not in this PR
- The cached `package-lock.yml` is finally copied from from cached path to the tmp project and used
- To update dependencies or versions, it is expected that the user will manually update their package-lock.yml in the dbt project using the dbt deps command.

## Related Issue(s)
closes: https://github.com/astronomer/astronomer-cosmos/issues/930

## Checklist


cosmos/cache.py
```diff
@@ -397,3 +404,49 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
     profile_yml_path = profile_yml_dir / DEFAULT_PROFILES_FILE_NAME
     profile_yml_path.write_text(profile_content)
     return profile_yml_path
+
+
+def is_cache_lockfile_enabled() -> bool:
+    from dbt.version import __version__
```
tatiana, line 410: @pankajastro we may want to revisit this, since it won't work for a few use cases.

Although this will work for `ExecutionMode.LOCAL` usually, it will not work if users use `dbt_executable_path` to a `dbt` installation separate from Airflow (most Astro customers).

This will also not work for `ExecutionMode.VIRTUALENV`.

https://astronomer.github.io/astronomer-cosmos/getting_started/execution-modes.html

pankajastro: Thank you for the pointer. Yes, you are right; I have removed dbt as a dependency and now checking only for the presence of the lock file in the project to enable it.

cosmos/cache.py
```diff
@@ -397,3 +403,48 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
...
+            project_sha1_hash = _get_sha1_hash(project_lockfile)
+            cached_sha1_hash = _get_sha1_hash(cached_package_lockfile)
+            if project_sha1_hash == cached_sha1_hash:
+                return cached_package_lockfile
+        else:
+            cached_lockfile_dir = cache_dir / cache_identifier
+            cached_lockfile_dir.mkdir(parents=True, exist_ok=True)
+            shutil.copyfile(project_lockfile, cached_package_lockfile)
+            return cached_package_lockfile
+    except Exception as e:
```
tatiana, line 442: Can we be more specific about which exception we're raising?

cosmos/cache.py
```diff
@@ -397,3 +403,48 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
...
+        logger.warning(f"Error processing cached lockfile: {e}")
+    return None
+
+
+def _copy_cached_lockfile_to_project(cached_lockfile: Path, project_dir: Path) -> None:
+    """Copy the cached package-lock.yml to tmp project dir"""
+    lock_file = project_dir / PACKAGE_LOCK_YML
+    shutil.copyfile(cached_lockfile, lock_file)
```
tatiana, line 450: Lessons learned from #1051 🥇

cosmos/dbt/graph.py
```diff
@@ -467,6 +468,13 @@ def load_via_dbt_ls_without_cache(self) -> None:
                 env[DBT_TARGET_PATH_ENVVAR] = str(self.target_dir)
 
                 if self.render_config.dbt_deps and has_non_empty_dependencies_file(self.project_path):
+                    if self.execution_config.execution_mode in (
+                        ExecutionMode.LOCAL,
+                        ExecutionMode.VIRTUALENV,
+                    ) and is_cache_lockfile_enabled(project_path):
```
tatiana, line 474: Execution config is related to the task execution; we should avoid it wherever possible in the rendering phase (when we parse dbt).

Since we're already in `load_via_dbt_ls_without_cache`, I believe it would be safe to do:
```suggestion
                    if is_cache_lockfile_enabled(project_path):
```

tatiana (commented):
@pankajastro great work! I left some comments inline.

I believe my main concern with us caching this per dbt project is that there is a chance that multiple concurrent dbt deps commands running from diffferent task executions in the same Airflow Worker node could potentially face issues by having racing conditions - a process overriding or deleting another's package lock file.

That said, I believe we should try this approach. It would be great, however, if similar to other caches that we allow users to enable/disable, that we give support to users enabling/disabling this, similar to other `settings` we created. Could you add this, please?

pankajastro:
> I believe my main concern with us caching this per dbt project is that there is a chance that multiple concurrent dbt deps commands running from diffferent task executions in the same Airflow Worker node could potentially face issues by having racing conditions - a process overriding or deleting another's package lock file.

Thanks for the feedback. Yes, unfortunately, the copyfile operation here is not atomic and causes race conditions. This blog (https://alexwlchan.net/2019/atomic-cross-filesystem-moves-in-python/) suggests some workarounds, though I have not tried or tested them yet.

pankajastro:
> That said, I believe we should try this approach. It would be great, however, if similar to other caches that we allow users to enable/disable, that we give support to users enabling/disabling this, similar to other settings we created. Could you add this, please?

Make sense, I'll address this 👍

cosmos/cache.py
```diff
@@ -400,3 +407,60 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
...
+
+def _copy_cached_lockfile_to_project(cached_lockfile: Path, project_dir: Path) -> None:
+    """Copy the cached package-lock.yml to tmp project dir"""
+    lock_file = project_dir / PACKAGE_LOCK_YML
+    _safe_copy(cached_lockfile, lock_file)
+
+
+# TODO: Move this function at different location
+def _safe_copy(src: Path, dst: Path) -> None:
+    # Create a temporary file in the same directory as the destination
```
tatiana, line 459: @pankajastro could you please add a link to that article that discussed the issues with concurrency?

cosmos/cache.py
```diff
@@ -400,3 +407,60 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
     profile_yml_path = profile_yml_dir / DEFAULT_PROFILES_FILE_NAME
     profile_yml_path.write_text(profile_content)
     return profile_yml_path
+
+
+def is_cache_lockfile_enabled(project_dir: Path) -> bool:
```
tatiana, line 412: BTW: it may be worth to rename any occurrence of `lockfile` to `package_lockfile` - Cosmos has other lockfiles (we're introducing one in the #1079

tatiana (commented):
@pankajastro this is looking great, thanks for addressing all the feedback.
I added two minor comments:
1. Related to methods related to packages lockfile
2. Detail more why we're copying files the way we're doing, explaining the concurrency issues and if possible adding some reference

Once these are done, I'm happy to approve and have this merged!

cosmos/cache.py
pankajastro: Add docstring and also a reference. PTAL

cosmos/cache.py
pankajastro: Updated in commit d5b3e9b7bc9063a7a44c0e583bf4e79f8f51fe09 PTAL

cosmos/cache.py
```diff
@@ -400,3 +412,72 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
     profile_yml_path = profile_yml_dir / DEFAULT_PROFILES_FILE_NAME
     profile_yml_path.write_text(profile_content)
     return profile_yml_path
+
+
+def is_cache_package_lockfile_enabled(project_dir: Path) -> bool:
```
pankajkoti, line 417: Slightly confused with the name of this method. It appears that the implementation in this method checks whether the project has the PACKAGE_LOCKFILE_YML file, but the name of the method implies that we are checking if caching is enabled or not.

cosmos/cache.py
```diff
@@ -400,3 +412,72 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
...
+    """
+    cache_identifier = project_dir.name
+    package_lockfile = project_dir / PACKAGE_LOCKFILE_YML
+    cached_package_lockfile = cache_dir / cache_identifier / PACKAGE_LOCKFILE_YML
+
+    try:
+        if cached_package_lockfile.exists() and cached_package_lockfile.is_file():
+            project_sha1_hash = _get_sha1_hash(package_lockfile)
+            cached_sha1_hash = _get_sha1_hash(cached_package_lockfile)
+            if project_sha1_hash == cached_sha1_hash:
```
pankajkoti, line 444: what should happen when the hash does not match?

pankajkoti (commented):
LGTM mostly, have a couple of questions inline

cosmos/cache.py
```diff
@@ -400,3 +412,72 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
...
+    return None
+
+
+def _copy_cached_package_lockfile_to_project(cached_package_lockfile: Path, project_dir: Path) -> None:
+    """Copy the cached package-lock.yml to tmp project dir"""
+    package_lockfile = project_dir / PACKAGE_LOCKFILE_YML
+    _safe_copy(cached_package_lockfile, package_lockfile)
+
+
+# TODO: Move this function at different location
```
tatiana, line 462:
```suggestion
# TODO: Move this function to a different location
```

tatiana (approved):

cosmos/cache.py
```diff
@@ -400,3 +412,72 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
...
+def _get_latest_cached_package_lockfile(project_dir: Path) -> Path | None:
+    """
+    Retrieves the latest cached package-lock.yml for the specified project directory,
+    or creates and caches it if not already cached and hashes match.
+    """
+    cache_identifier = project_dir.name
+    package_lockfile = project_dir / PACKAGE_LOCKFILE_YML
+    cached_package_lockfile = cache_dir / cache_identifier / PACKAGE_LOCKFILE_YML
+
+    try:
```
tatiana, line 440: Usually, it would be better to have try/except covering one/or very few lines without many if/excepts inside. Is there any particular method we'd expect to raise `OSError`?

cosmos/cache.py
```diff
@@ -400,3 +412,72 @@ def create_cache_profile(version: str, profile_content: str) -> Path:
...
+def _get_sha1_hash(yaml_file: Path) -> str:
+    """Read package-lock.yml file and return sha1_hash"""
+    with open(yaml_file) as file:
+        yaml_content = file.read()
+    data = yaml.safe_load(yaml_content)
+    sha1_hash: str = data.get("sha1_hash", "")
+    return sha1_hash
+
+
+def _get_latest_cached_package_lockfile(project_dir: Path) -> Path | None:
```
tatiana, line 431: Would it make sense to use the method `_get_timestamp` that's being used to retrieve the newest partial parse file?
https://github.com/astronomer/astronomer-cosmos/blob/d5b3e9b7bc9063a7a44c0e583bf4e79f8f51fe09/cosmos/cache.py#L105

cosmos/cache.py
tatiana: We could follow the same pattern we're using for enabling/disabling other cache mechanisms:
https://github.com/astronomer/astronomer-cosmos/blob/60148ceb311e41c5c27a7be1b584bc5fcaafbcfe/cosmos/settings.py#L15-L17

And it would justify keeping this name

cosmos/cache.py
pankajastro: Thanks for the feedback @pankajkoti. I have updated to update the cache in case hash does not match

cosmos/cache.py
pankajastro: Added a config to enable and disable it with default value True to enable

cosmos/cache.py
pankajastro: Actually, it will not be required. I have cleaned it up a bitt

cosmos/cache.py
pankajastro: When we deploy a dbt project, does it update the stats for only the file that has changed, or does it update the stats for all files? Example, if we make a change to a dbt model and deploy, does the stats update apply only to this modified model file, or to all files? If stats change for all files, then the current approach looks more reliable. WDYT?

tatiana: That's a good question, we need to check. There is a chance that as part of the Docker image build process, we're creating new timestamps. Could you please confirm in an Astro deployment so we know?

pankajastro: I checked, and each time we deploy, the timestamp changes

cosmos/dbt/graph.py
```diff
@@ -487,6 +492,10 @@ def load_via_dbt_ls_without_cache(self) -> None:
                 env[DBT_TARGET_PATH_ENVVAR] = str(self.target_dir)
 
                 if self.render_config.dbt_deps and has_non_empty_dependencies_file(self.project_path):
+                    if is_cache_package_lockfile_enabled(project_path):
+                        latest_package_lockfile = _get_latest_cached_package_lockfile(project_path)
+                        if latest_package_lockfile:
+                            _copy_cached_package_lockfile_to_project(latest_package_lockfile, Path(tmpdir))
```
pankajkoti, line 498:
```suggestion
                            _copy_cached_package_lockfile_to_project(latest_package_lockfile, tmpdir_path))
```
We already have a `tmpdir_path` as a `Path` instance for the `tmpdir`

pankajkoti (approved):
LGTM, one minor suggestion inline

cosmos/cache.py
pankajastro: I'm resolving this conversion but please let me know if you have more input. thanks!

PR #1091, opened 2024-07-13
add DatabricksOauthProfileMapping profile

### LINKED ISSUE #1085: [Feature] Databricks - Allow each dbt node to run in its own configurable cluster
### Description

I am currently working in a company that is using an in-house solution to run dbt-like transformations on databricks and I am trying to switch to dbt & cosmos.  The issue arrives in the fact that the current solution has an extra level of control which is **the ability to spin up a cluster with custom resources, run the transformation in that cluster and, once concluded, the cluster stops**. The time wasted spinning up clusters for singular SQL tasks is not a big concern, since the money saved on the process is tremendous in comparison with deploying dbt models in an all-purpose cluster or in the databricks SQL warehouse, which is what dbt supports at the moment.

# General Idea

- For dbt-databricks DAGs, there should be an extra operator with the option to pass a json file as a parameter mapping cluster resources for individual nodes
- The dbt node and its corresponding tests should run in the same cluster

# Implementation ideas

## Functionality 

1. Allow cosmos to call databricks API for some custom Operator (for example, _DbtDatabricksClusterOperator_)

airflow [DatabricksSubmitRunOperator](https://airflow.apache.org/docs/apache-airflow-providers-databricks/stable/operators/submit_run.html) allows to submit **dbt_task** commands from it. The issue I saw is that the [PR](https://github.com/apache/airflow/pull/25623/files#diff-30731826b25d422cb6069898701e34f8ef23b2a32b7782444793a14bc52b0cb1R365) that added this feature hard-codes it to only work with git sync, which is the opposite direction of cosmos. I can submit a PR later on airflow repo to change that, but for now as a solution I can think of is directly calling databricks API which allows us to specify the [project path](https://docs.databricks.com/api/workspace/jobs/submit#tasks-dbt_task-project_directory). The drawback I can think about is that in the case the user is using the `manifest.json` to trigger jobs, this approach would not work by default.
An alternative would be to run a task's query as a databricks [sql_task](https://docs.databricks.com/api/workspace/jobs/submit#tasks-sql_task) API request, which would be a more universal approach since cosmos is essentially doing dbt's job by rendering the query in the first place.

2. add an extra field (`cluster_resources`) to this operator

The custom operator would have an extra field where we can map the dbt task name with the cluster configs it should have. one of the fields can be called `default` and if no task cluster config is given, the task should use the default resources.

`DbtDatabricksClusterOperator(...,
cluster_resources = {
"default" : { "new_cluster" : {...}},
"my_task_1" : {"new_cluster":{...}},
...
}
)`

## Observation

This is the most straightforward idea I could think of, but it has been a long time I do not touch the project and I am more than open for suggestions before I even start working on it. I thought about other ideas, such as maybe deploying the models to different warehouses based on tags (`low`, `medium`, `high`), spinning up the cluster with a custom task and shutting it down at the end of a successful task and so on, but all of them had implementation problems. What do you think @tatiana ? Any feedback before I start implementing? or recommendations? The project changed a lot since the last time I dealt with it, so maybe there is a better way I am not aware of.

## Other useful links
[specifying compute resources for nodes in the dbt_project.yaml](https://docs.getdbt.com/reference/resource-configs/databricks-configs#specifying-the-compute-for-models)

### Use case/motivation

- There is a huge price difference depending on how you deploy your dbt model on databricks, and unfortunately dbt does not support the most economical way possible : creating a cluster per task that is terminated once the task is done

- when you have an all-purpose data warehouse for dbt transformations, you will have to deal with dbt nodes competing for resources

- It is easier to track resource consumption per dbt node when you map one task to one cluster

### Related issues

_No response_

- [X] Yes, I am willing to submit a PR!

CorsettiS:
## Description

As of now, Token-based auth is the only possible way to access databricks from cosmos, even though dbt-databricks support OAuth-based access as can be seen [here ](https://docs.getdbt.com/docs/core/connect-data-platform/databricks-setup).

This PR aims to enhance the support offered by cosmos to databricks. The final goal is to close #1085

I have tested locally and got a positive result. 
## Breaking Change?
No

## Checklist


cosmos/profiles/databricks/client.py
pankajastro, line 1: Does Databricks refer to it as client authentication or OAuth authentication? If it's OAuth, should we rename the file to "oauth"? What do you think?

cosmos/profiles/databricks/client.py
```diff
@@ -0,0 +1,48 @@
...
+        "host": "host",
+        "schema": "schema",
+        "client_id": ["login", "extra.client_id"],
+        "client_secret": ["password", "extra.client_secret"],
+        "http_path": "extra.http_path",
+    }
+
+    @property
+    def profile(self) -> dict[str, Any | None]:
+        """Generates profile. The token is stored in an environment variable."""
```
pankajastro, line 41:
```suggestion
        """Generates profile. The client-id and client-secret is stored in an environment variable."""
```

tests/profiles/databricks/test_dbr_client.py
```diff
@@ -0,0 +1,73 @@
+"""Tests for the databricks profile."""
```
pankajastro, line 1:
```suggestion
"""Tests for the databricks OAuth profile."""
```

tests/profiles/databricks/test_dbr_client.py
```diff
@@ -0,0 +1,73 @@
...
+
+from unittest.mock import patch
+
+import pytest
+from airflow.models.connection import Connection
+
+from cosmos.profiles.databricks import (
+    DatabricksOauthProfileMapping,
+)
```
pankajastro, line 10:
```suggestion
from cosmos.profiles.databricks import DatabricksOauthProfileMapping
```

pankajastro (commented):
Thank you for your contribution! It looks great. I just posted a few minor comments

cosmos/profiles/databricks/client.py
CorsettiS: done

pankajastro (approved):

PR #1099, opened 2024-07-17
Install requirements.txt by default during dev env spin up

CorsettiS:
## Description

I bumped the astronomer image to 11.6 since it uses the newest airflow, incorporating [tatiana's change in airflow version](https://github.com/apache/airflow/pull/39670) for datasets and I am installing requirements.txt by default in the image, since if the user wants to add some custom package, he needs to change not only the requirements.txt file but also the Dockerfile. With this only adding dependencies to requirements.txt will already install them in the image for dev development. 

## Checklist


tatiana:
Thanks for fixing this, @CorsettiS !

tatiana (approved):

Issue #943, opened 2024-05-07
DbtTaskGroup - Kubernetes Execution with RenderConfig


harshag7dm:
When I use Kubernetes Execution in DbtTaskGroup and try to filter out the Models using tag, it doesn't work. There is no Kubernetes pod being run.
If I remove the RenderConfig, then all the models are run? 
Is this a bug or am I missing something?

tatiana:
To try to reproduce this:
1. Run https://astronomer.github.io/astronomer-cosmos/getting_started/kubernetes.html#kubernetes
2. Modify https://github.com/astronomer/cosmos-example/blob/main/dags/jaffle_shop_kubernetes.py to have RenderConfig with a selector using tag (e.g. https://github.com/astronomer/cosmos-demo/blob/main/dags/filtering/customers_tag.py)
3. Confirm we can reproduce the issue
4. Fix this, if we can

pankajastro:
This is a legitimate issue. I can replicate it using the steps outlined above. I’ll create a PR to address it.
<img width="1435" alt="Screenshot 2024-07-16 at 10 51 00 PM" src="https://github.com/user-attachments/assets/50e1effe-0556-4153-8ebd-5624cdd587c9">

pankajastro:
I investigated further and found that when the manifest is unavailable and the execution mode isn't set to ExecutionMode.LOCAL, we're using the custom parser to load the dbt project. However, this custom parser currently does not support selectors.

https://github.com/astronomer/astronomer-cosmos/blob/2994dba1059cc914343a477e71564300b3fa0896/cosmos/dbt/graph.py#L323-L329

The workaround I tried 
- Use load_via_dbt_ls for other execution modes as well, although I'm uncertain about any potential side effects.
- Extract the tag in the custom parser method and assign it to the node. This is currently set to empty. https://github.com/astronomer/astronomer-cosmos/blob/2994dba1059cc914343a477e71564300b3fa0896/cosmos/dbt/graph.py#L557

Issue #693, opened 2023-11-20
LoadMode.CUSTOM cannot select tags


adammarples:
https://github.com/astronomer/astronomer-cosmos/blob/d28d3df7eacac74d0fbb016be70434bcc30c51be/cosmos/dbt/graph.py#L314C17-L314C25

`load_via_custom_parser` hard codes DbtNode.tags attribute to be an empty list
 
 ```python
         for model_name, model in models:
            config = {item.split(":")[0]: item.split(":")[-1] for item in model.config.config_selectors}
            node = DbtNode(
                unique_id=f"{model.type.value}.{self.project.project_name}.{model_name}",
                resource_type=DbtResourceType(model.type.value),
                depends_on=list(model.config.upstream_models),
                file_path=Path(
                    model.path.as_posix().replace(
                        self.render_config.project_path.as_posix(), self.execution_config.project_path.as_posix()
                    )
                ),
                tags=[],
                config=config,
            )
  ```
  
  This causes `_is_tags_subset` to return False every time
  
  ```python
      def _is_tags_subset(self, node: DbtNode) -> bool:
        """Checks if the node's tags are a subset of the config's tags."""
        if not (set(self.config.tags) <= set(node.tags)):
            return False
        return True
 ```
 
 because set(node.tags) is always empty.

tatiana:
@adammarples I just checked on the main branch, and the latest versions of Cosmos, both 1.2.5 and 1.3.0a2 should have this fixed:
https://github.com/astronomer/astronomer-cosmos/blob/173436552c8362e229258386e63ceb395fa33a44/cosmos/dbt/graph.py#L112

Could you upgrade and confirm if it works as expected, please?

adammarples:
Hi @tatiana this is not a problem with DBT_LS but with load via custom parser https://github.com/astronomer/astronomer-cosmos/blob/2878d6accc2a725502158247dbd1af7ac37b4449/cosmos/dbt/graph.py#L314 which is still there in 1.2.5 etc

tatiana:
HI @adammarples , please, could you confirm if this is still an issue in Cosmos 1.4?
Would you like to contribute to the fix?

nathan-nguyen-webjet:
Can confirm dbt_ls fixed the bug for me, custom mode fail to create the graph

PR #1100, opened 2024-07-17
Fix empty tag in case of custom parser

pankajastro:
## Description
As of version 1.5.1, using a custom parser to parse a dbt project
hardcodes the tags value to an empty list [] when initializing 
the DbtNode object. This creates issues for users utilizing 
the select option in the DAG with render_config, 
resulting in distorted DAG rendering, as shown below.

Screenshot 2024-07-16 at 10 51 00 PM

This PR parses the tag values from the project and uses them to initialize the DbtNode object. Post this change rendering appears like
Screenshot 2024-07-17 at 7 01 53 PM

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/943 
closes: https://github.com/astronomer/astronomer-cosmos/issues/693

## Breaking Change?

No

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -545,6 +545,7 @@ class DbtGraph:
         )
         for model_name, model in models:
             config = {item.split(":")[0]: item.split(":")[-1] for item in model.config.config_selectors}
+            tags = [selector for selector in model.config.config_selectors if selector.startswith("tags:")]
             node = DbtNode(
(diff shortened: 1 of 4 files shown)


tatiana (approved):
@pankajastro this looks great, thank you!

Does it actually close #943 and #693? If yes, please, update the PR description to:
Close: #943
Close: #693

So they are automatically closed

PR #1101, opened 2024-07-17
Use `dbt ls` as the default parser when profile_config is provided

pankajastro:
## Description

The `load_via_custom_parser` method is not very effective 
when it comes to filtering. Currently, if the ExecutionMode 
is not Local, we always use load_via_custom_parser to parse 
the dbt project. However, I believe it's beneficial to utilize DBT LS 
whenever it's available. This PR remove the check if the 
execution mode is ExecutionMode.LOCAL before `load_via_dbt_ls`

## Related Issue(s)

Related: https://github.com/astronomer/astronomer-cosmos/issues/943

## Breaking Change?

I believe no :) 

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -323,7 +323,6 @@ class DbtGraph:
                 if self.profile_config and self.project_path:
                     try:
                         self.load_via_dbt_ls()
-                        logger.info("Using DBT ls")
                     except FileNotFoundError:
                         self.load_via_custom_parser()
                 else:
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -320,9 +320,10 @@ class DbtGraph:
             if self.project.is_manifest_available():
                 self.load_from_dbt_manifest()
             else:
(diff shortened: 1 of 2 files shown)


cosmos/dbt/graph.py
```diff
@@ -320,7 +320,7 @@ def load(
             if self.project.is_manifest_available():
                 self.load_from_dbt_manifest()
             else:
-                if execution_mode == ExecutionMode.LOCAL and self.profile_config:
+                if self.profile_config:
```
tatiana, line 323: Another condition we probably need to check is if the user set a path to the dbt project path (`self.project_dir`) - without this, we can't run dbt ls

pankajastro: yes, I have added this check.

pankajkoti (approved):
LGTM, let's wait for @tatiana's review.

tatiana (commented):
@pankajastro Please, let's add a test to cover the use-case that encouraged this change - so we make sure it continues working in future

tatiana (approved):

Issue #875, opened 2024-03-03
[bug] [1.3.2] tests on sources do not render


push2prod:
because sources are not natively rendered as models(only by using the experimental feature) i think tests on sources do not render as well.
I tried rendering the sources as dummy operators and trying the AFTER_ALL test behavior but nothing seemed to work.
it would be very convenient to test sources right before running to avoid any problems with dbt pipelines and would greatly appreciate a patch.

we're using local mode with manifest file without any selectors with airflow 2.5.2(if it matters).

tatiana:
@push2prod Please, could you share a reproducible example of DAG of what you tried to do, what you saw, and confirm if you'd like to see the tests bundled in task groups, together with the sources, or rendered by the end of the TaskGroup/DAG?

Issue #630, opened 2023-10-26
Add default behaviour for converting dbt Source nodes into Airflow tasks


tatiana:
As of Cosmos 1.2.1, it does not render [dbt Source nodes](https://docs.getdbt.com/docs/build/sources#snapshotting-source-data-freshness) as Airflow tasks by default.

Since [Cosmos 1.2.0](https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.2.0), we've introduced support for customizing how the library converts any dbt node into Airflow:
https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html#customizing-how-nodes-are-rendered-experimental

This means users can already customise the desired behaviour for Source nodes, using something like:
https://github.com/astronomer/astronomer-cosmos/blob/11ce2d718483f1d4eb1bfdac659417fb66a1492e/dev/dags/example_cosmos_sources.py#L63

The open questions are:
1. Do we want Cosmos to have a default behaviour for dbt Source nodes and automatically generate tasks for nodes of this type?
2. If the answer to (1) is yes, what do we want the behaviour for those nodes to be? Should we run:
```
dbt source freshness
```

pgoslatara:
In addition to freshness checks users can also (and sometimes do) add standard tests to their sources, i.e. `unique`, `not_null`, etc. I think a good default behaviour would be to display sources as a task in the Airflow UI as this provides an Airflow DAG to the end user familiar with dbt that is most similar to dbt's DAG, thereby enabling Cosmos adoption. These tasks can use EmptyOperator and therefore have no impact on compute usage or scheduler resources.

One small note, DummyOperator has been deprecated in favour of EmptyOperator in Airflow 2.4 (see [here](https://airflow.apache.org/docs/apache-airflow/2.3.4/_api/airflow/operators/dummy/index.html)).

linchun3:
+1 to this issue as this is something I'm interested in.

1. Do we want Cosmos to have a default behaviour for dbt Source nodes and automatically generate tasks for nodes of this type?

I'm in support for having each source render as an Airflow task, and running all tests attached to it. This improves the visibility of a source and whether a model update failure is due to an upstream failure.

2. If the answer to (1) is yes, what do we want the behaviour for those nodes to be?

`dbt test --select source:jaffle_shop.orders`

This command should run all the tests associated to a source. I'm not sure how we should handle `dbt source freshness` though.

Add on questions:
[...]

Issue #572, opened 2023-09-30
How to test sources in Astronomer Cosmos DAG


maver1ck:
Hi,
I'm experimenting with Astronomer Cosmos.
I read trough documentation and I didn't find an option to include sources tests in dag when using 'after_each' behaviour.
https://astronomer.github.io/astronomer-cosmos/configuration/testing-behavior.html

Is there an option for that? 

My dag
```
from datetime import datetime
import os
from airflow.datasets import Dataset

from cosmos import DbtDag, ProjectConfig, ProfileConfig, ExecutionConfig

profile_config = ProfileConfig(
    profile_name="test",
    target_name="dev",
    profiles_yml_filepath=os.environ['AIRFLOW_HOME'] +"/profiles.yml",
)

my_cosmos_dag = DbtDag(
    project_config=ProjectConfig(
        os.environ['AIRFLOW_HOME'] +"/dbt_project",
    ),
    profile_config=profile_config,
    execution_config=ExecutionConfig(
        dbt_executable_path="dbt",
    ),
    # normal dag parameters
    start_date=datetime.now(),
    catchup=False,
    dag_id="sample_dag",
    schedule=[Dataset("airbyte://example_source")],
)
```

tatiana:
Hi @maver1ck , thanks for reaching out!

At the moment, Cosmos doesn't expose this, but it would be a simple change, similar to #474 

Would you like to contribute?

The change would be around this part of the code:
https://github.com/astronomer/astronomer-cosmos/blob/435a6996cf404bc2a9b9a23c7957e72786fab8ec/cosmos/airflow/graph.py#L171

The critical part of this improvement would be adding tests to ensure we continue rendering the Airflow DAG as intended.

tatiana:
Hi @maver1ck , I wanted to follow up on this issue - since we made some progress.

As of Cosmos 1.2.1, it does not render [dbt Source nodes](https://docs.getdbt.com/docs/build/sources#snapshotting-source-data-freshness) as Airflow tasks by default.

Since [Cosmos 1.2.0](https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.2.0), we've introduced support for customizing how the library converts any dbt node into Airflow:
https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html#customizing-how-nodes-are-rendered-experimental

This means users can already customise the desired behaviour for Source nodes, using something like:
[...]

PR #661, opened 2023-11-09
Add default source nodes rendering

arojasb3:
## Description
I'm aiming to give a default behaviour to source nodes for checking their source freshness and their tests. 
One of the main limitations I found while using the custom_callback functions on source nodes to check freshness is that those node were being created to 100% of sources but not all of them required freshness checks, this made workers waste compute time.

I'm adding a new variable into the DbtNode class called `has_freshness` which would be True for sources with freshness checks and False for the ones that not and any other resource type.

All sources with the `has_freshness` == False will be rendered as Empty Operators, to keep the dbt's behavior of showing sources as suggested in issue #630.

## Related Issue(s)

#630 
## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1107, opened 2024-07-19
Add default source nodes rendering

arojasb3:
## Description
Re-Opening of PR #661 
This PR features a new way of rendering source nodes:
- Check freshness for sources with freshness checks
- Source tests
- Empty operators for nodes without tests or freshness.

One of the main limitations I found while using the `custom_callback` functions on source nodes to check freshness is that nodes were being created on 100% of sources but not all of them required freshness checks, this made workers waste compute time.

I'm adding a new variable into the DbtNode class called has_freshness which would be True for sources with freshness checks and False for any other resource type.

If this feature is enabled with the option `ALL`:
All sources with the has_freshness == False will be rendered as Empty Operators, to keep the dbt's behavior of showing sources as suggested in issue #630 

A new rendered template field is included too: `freshness` which is the sources.json generated by dbt when running `dbt source freshness`

This adds a new node type (source), which changes some tests behavior.
This PR also updates the dev dbt project jaffle_shop to include source nodes when enabled.
image

As seen in the image, source nodes with freshness checks are rendered with a blue color, while the ones rendered as EmptyOperator show a white/light green color

## Related Issue(s)
Closes: #630
Closes: #572 
Closes: https://github.com/astronomer/astronomer-cosmos/issues/875

## Breaking Change?
This won't be a breaking change since the default behavior will still be ignoring this new feature. That can be changed with the new RenderConfig variable called `source_rendering_behavior`.

## Checklist


arojasb3:
hey @pankajastro !
The errors from the tests are regarding the source node group naming ("source" instead of "run") which can be debated. Since we are checking freshness I'm not 100% sure we should still name those tasks "run".

On the other hand, there's a hashing test error I'm not 100% sure how to deal with 🤔 .

Please let me know any thoughts in this PR

pankajastro:
> On the other hand, there's a hashing test error I'm not 100% sure how to deal with 🤔 .

The test `test_save_dbt_ls_cache` checks the hash of the project dir since you are adding a new file `sources.yml` in the project the hash has  changed maybe you can update the hash value in test https://github.com/astronomer/astronomer-cosmos/blob/e61f3a365f6c3a0b550cda298516f64ad1321189/tests/dbt/test_graph.py#L1404

pankajastro:
> The errors from the tests are regarding the source node group naming ("source" instead of "run") which can be debated. Since we are checking freshness I'm not 100% sure we should still name those tasks "run".

If we decide to keep the source suffix then will have to update the test based on DbtResourceType https://github.com/astronomer/astronomer-cosmos/blob/e61f3a365f6c3a0b550cda298516f64ad1321189/tests/airflow/test_graph.py#L345

@tatiana /@pankajkoti any suggestions here?

cosmos/dbt/graph.py
```diff
@@ -138,6 +163,11 @@ def parse_dbt_ls_output(project_path: Path | None, ls_stdout: str) -> dict[str,
                 file_path=project_path / node_dict["original_file_path"],
                 tags=node_dict.get("tags", []),
                 config=node_dict.get("config", {}),
+                has_freshness=(
+                    is_freshness_effective(node_dict.get("freshness", False))
+                    if node_dict["resource_type"] == DbtResourceType.SOURCE
+                    else False
+                ),
```
tatiana, line 170: It seems the output of `dbt ls` does not include the `freshness`  property :/

I cloned the [stacktonic-dbt-example-project](https://github.com/stacktonic-com/stacktonic-dbt-example-project)  repo, which has a source:
```
$ cat models/staging/google_analytics/src_google_analytics.yml
version: 2

sources:
  - name: google_analytics
    database: stacktonic-cloud # gcp-project
    dataset: analytics_263482362 # gcp-dataset
    
    freshness:
      warn_after: { count: 24, period: hour } 
      error_after: { count: 30, period: hour } 
    
    loaded_at_field: "parse_timestamp(_table_suffix, '%Y%m%d')"

    tables:
      - name: events
        identifier: events_*
        description: "Default Google Analytics 4 (raw) data dumps. Tables are date sharded (events_YYYYMMDD)."%    
``` 

I set up the local environment using:
```
python3.11 -m venv venv   
source venv/bin/activate
pip install dbt-core dbt-bigquery
cp .dbt/profiles.yml . 
```

And I run the following command using `dbt 1.4` and `dbt 1.8.3`:
```
 dbt ls --output json
```

Both of them generated an output that contained:
```
{
	"name": "events",
	"resource_type": "source",
	"package_name": "stacktonic_example_project",
	"original_file_path": "models/staging/google_analytics/src_google_analytics.yml",
	"unique_id": "source.stacktonic_example_project.google_analytics.events",
	"source_name": "google_analytics",
	"tags": [],
	"config": {
		"enabled": true
	}
}
```

So far, we have not relied on the `manifest.json` when using the dbt ls parsing mode, but it seems the time has come. We'll probably have to use it here.

cosmos/dbt/graph.py
```diff
@@ -605,6 +642,11 @@ def load_from_dbt_manifest(self) -> None:
                     file_path=self.execution_config.project_path / Path(node_dict["original_file_path"]),
                     tags=node_dict["tags"],
                     config=node_dict["config"],
+                    has_freshness=(
+                        node_dict["freshness"] is not None
+                        if node_dict["resource_type"] == "source" and "freshness" in node_dict
+                        else False
+                    ),
```
tatiana, line 649:
```suggestion
                    has_freshness= node_dict["resource_type"] == "source" and node_dict.get("freshness")
```

cosmos/operators/local.py
```diff
@@ -816,3 +816,12 @@ def __init__(self, **kwargs: str) -> None:
         raise DeprecationWarning(
             "The DbtDepsOperator has been deprecated. " "Please use the `install_deps` flag in dbt_args instead."
         )
+
+
+class DbtSourceLocalOperator(DbtLocalBaseOperator):
```
tatiana, line 821: There was a recent refactor, which exposes the dbt command as generic Mixin classes, in the `cosmos/operators/base.py` module:
* https://github.com/astronomer/astronomer-cosmos/blob/e61f3a365f6c3a0b550cda298516f64ad1321189/cosmos/operators/base.py#L347
* https://github.com/astronomer/astronomer-cosmos/blob/e61f3a365f6c3a0b550cda298516f64ad1321189/cosmos/operators/base.py#L379

And then we're able to simply subclass them in all modules of Cosmos operators, such as:
* https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/local.py#L528
* https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/kubernetes.py#L128

This was an improvement by @jbandoro which made the code much cleaner and testable. It would be great if we could follow the same pattern here, by:
* creating a `DbtSourceMixin` class in `operators/base.py`
* and inheriting it in all supported execution modes within the modules in `cosmos/operators`

tatiana:
Thanks a lot for creating a new PR on this, @arojasb3 , 5he PR is looking great and I'm super excited we'll be merging this and releasing this in 1.6.

> Since this PR could affect some users with the custom rendering, I was thinking we could enable this with a flag?

I agree we should have a feature flag, and perhaps we can enable it by default - and users who want to opt-out can do, what do you think? We have a few other similar feature flags in `settings.py`.

> Since we are checking freshness I'm not 100% sure we should still name those tasks "run".

So far we were using run only for models, it feels worth respecting the naming we had for other node types
```
            task_id = f"{node.name}_{node.resource_type.value}"
```
https://github.com/astronomer/astronomer-cosmos/blob/e61f3a365f6c3a0b550cda298516f64ad1321189/cosmos/airflow/graph.py#L155

tatiana:
> > The errors from the tests are regarding the source node group naming ("source" instead of "run") which can be debated. Since we are checking freshness I'm not 100% sure we should still name those tasks "run".
> 
> If we decide to keep the source suffix then will have to update the test based on DbtResourceType
> 
> https://github.com/astronomer/astronomer-cosmos/blob/e61f3a365f6c3a0b550cda298516f64ad1321189/tests/airflow/test_graph.py#L345
> 
> @tatiana /@pankajkoti any suggestions here?

As mentioned in https://github.com/astronomer/astronomer-cosmos/pull/1107#issuecomment-2239280775, I believe we should be consistent with the rest of Cosmos naming, and have source nodes having task IDs using:
```
 task_id = f"{node.name}_{node.resource_type.value}"
```

cosmos/operators/local.py
```diff
@@ -816,3 +816,12 @@ def __init__(self, **kwargs: str) -> None:
...
+    """
+    Executes a dbt source freshness command.
+    """
+
+    ui_color = "#34CCEB"
+    base_cmd = ["source", "freshness"]
```
tatiana, line 827: What do you think if we added freshness to `template_fields`?

This way, users would be able to visualise the freshness values in the UI, similar to other ones:

cosmos_rendered_templates

Example of how to do:
https://github.com/astronomer/astronomer-cosmos/blob/e61f3a365f6c3a0b550cda298516f64ad1321189/cosmos/operators/local.py#L113

arojasb3: image
The latest PR added freshness like the picture.
It is adding the entire [sources.json file](https://docs.getdbt.com/reference/artifacts/sources-json#:~:text=This%20file%20contains%20information%20about,power%20its%20Source%20Freshness%20visualization.)

I couldn't find where dbt was compiling the freshness sql queries, it probably doesn't store it locally

cosmos/operators/local.py
arojasb3: Done ✔️

cosmos/dbt/graph.py
arojasb3: Ended up adding it like this
```
has_freshness=(
                        is_freshness_effective(node_dict.get("freshness"))
                        if DbtResourceType(node_dict["resource_type"]) == DbtResourceType.SOURCE
                        else False
                    ),
```

since both manifest and `ls` can return an empty freshness dictionary
```
{
                "warn_after": {
                    "count": null,
                    "period": null
                },
                "error_after": {
                    "count": null,
                    "period": null
                },
                "filter": null
            }
```

pankajastro:
@arojasb3, it looks like we're almost ready to merge. Could you please resolve the conflict and check for failing CI tests? A quick look suggests that we need to adjust a few parameters in the tests.

dwreeves (changes requested):
Hi! Sorry, been busy with other stuff, so haven't been able to respond to this, but I really want to give my 2 cents on the feature and high level API.

First of all, great job on this PR, this will be a very handy feature for Cosmos users.

Rendering all source nodes as `EmptyOperator`s is a good behavior if there are no has freshness tests, **and also so long as the user wants to render all source nodes.**

The issue is in practice, a dbt project can have a lot of source nodes. E.g. I use a third party data provider that gives me 150 tables, each one is a source node. But I only use a subset of them in my DAG. I render them using `dbt-codegen` and I keep them in the schema `yml`s because it makes adopting them as tables much easier, but I ultimately don't want to do anything with them.

It would significantly clutter my DAG to have all of these be empty operators. I suspect there are many other users in the same boat.

At the same time, **that doesn't mean I never want to render source nodes.** I would want to render them:

- On a case-by-case basis (i.e. within different task groups or Cosmos DAGs).
- Potentially _only_ render the source nodes with either freshness tests or dbt tests.

What I am ultimately getting at here is **I think the `AIRFLOW__COSMOS__RENDER_SOURCE_NODES` flag to enable this globally is not the best way to go about things.** Once you implement an API, it is really hard to un-implement it, so this is a detail I think we should get correct before we release it. This feature should absolutely be done via the `RenderConfig` API, and the user should have the option to do any of the following:

- Show all (selected) source nodes
- Show none
- Show only source nodes tests
- (Optional, this maybe too complicated for now) show only source nodes with non-source child nodes

pankajastro:
Thank you very much for your feedback, @dwreeves.

I understand your concerns and appreciate the idea of exposing this through the `RenderConfig` API. It also seems logical not to rely solely on the global configuration **AIRFLOW__COSMOS__RENDER_SOURCE_NODES**.

At the same time, given the community's interest in this feature, I believe it would be beneficial to release it sooner rather than later and gather their feedback.

Here’s what I propose:

1. Introduce an enum called [SourceBehaviour](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/constants.py) with the initial supported values
   1. **ALL**: Render all source nodes.
   2. **NONE**: Do not render source nodes.
2. Add a parameter `source_behavior` of type `SourceBehaviour` to the `RenderConfig` class.
3. Replace the usage of the global configuration with this new RenderConfig approach.

I believe these three suggestions can be easily implemented in the current PR.

This approach will address:
- Design Concerns: The user interface will remain consistent as you suggested.
- DAG Rendering: Users will now have control at the DAG level. This will address your concern at some level 

I will create a follow-up GitHub issue to address:
1. Source node rendering filters:
    1. Display only source nodes with tests (either freshness tests or data tests).
    2. (Optional, potentially complex for now) Display only source nodes with non-source child nodes.
2. Refactor the implementation and potentially relocate the filtering logic to DbtToAirflowConverter.
Does this plan align with your thoughts?

@dwreeves / @arojasb3, what are your thoughts on this approach?

dwreeves:
Yep, I understand this is a requested feature. I also think getting APIs correct to begin with is _really_ important; adding things to an API is easy, deleting or modifying things in APIs is very hard and causes severe headaches for users. Internals can always be moved around later but API decisions, in effect, cannot without causing a lot of problems. The highest-level API decisions for Cosmos must be made with a lot of care.

Another 2 cents. You say we shouldn't rely **solely** on the global config option. I would push back a little more: I am not sure the global configuration option should be there at all.

I am not aware of any Airflow configuration option that has the ability to mutate how a DAG actually renders. Airflow config options are things like database connection parameters, how the logging works, compute configuration of the scheduler, health checks for the webserver... but nothing impacts how a DAG renders.

Basically, it would be really weird for a global configuration option in Airflow to literally change whether or not a task shows up in a DAG. I would propose removing the global config option and solely relying on using the Cosmos `RenderConfig`.

arojasb3:
@pankajastro @dwreeves 
Hey! Thank you both for the feedback.
In the latest commit I moved config to RenderConfig with 3 options:
- ALL
- NONE
- WITH_TESTS_OR_FRESHNESS

I'll try to get the branch rebased and fix tests tomorrow =)
In the meantime please let me know any thoughts of the new config implementation.

pankajastro:
This feature has recently been requested in the #airflow-dbt Slack channel by a couple of users.
- https://apache-airflow.slack.com/archives/C059CC42E9W/p1721926644485259?thread_ts=1721618035.164499&cid=C059CC42E9W
- https://apache-airflow.slack.com/archives/C059CC42E9W/p1722119361007889

pankajastro (commented):
LGTM 👍 
@arojasb3, once CI is green, I'm happy for this to be merged. Really excited for this feature!

pankajastro:
We have updated the CI config. Could you please rebase again.

cosmos/dbt/graph.py
```diff
@@ -348,7 +378,14 @@ def run_dbt_ls(
...
     ) -> dict[str, DbtNode]:
         """Runs dbt ls command and returns the parsed nodes."""
-        ls_command = [dbt_cmd, "ls", "--output", "json"]
+        ls_command = [
+            dbt_cmd,
+            "ls",
+            "--output",
+            "json",
+            "--output-keys",
+            '"name alias unique_id resource_type depends_on original_file_path tags config freshness"',
```
pankajastro, line 387: Why is the `--output-keys` option necessary? I also suspect that the CI failure might be connected to this change.

arojasb3: yes, dbt ls doesn't output freshness by default.

I'm having a hard time debugging this, since the output in the tests is {}
but I'm testing locally with an env running dbt-core=1.5.4 and dbt-postgres=1.5.4
and the command does work from my terminal

pankajastro: Could you please test once with 
```
ls_command = [
 dbt_cmd,
 "ls",
 "--output",
 "json",
 "--output-keys",
 "name",
 "unique_id",
 "resource_type",
 "depends_on",
 "original_file_path",
 "tags",
 "config",
 "freshness"
 ]
```

arojasb3: That worked! thank you

pankajastro: yes, But look like now failing for dbt-1.4

cosmos/dbt/graph.py
```diff
@@ -348,7 +378,21 @@ def run_dbt_ls(
         self, dbt_cmd: str, project_path: Path, tmp_dir: Path, env_vars: dict[str, str]
     ) -> dict[str, DbtNode]:
         """Runs dbt ls command and returns the parsed nodes."""
-        ls_command = [dbt_cmd, "ls", "--output", "json"]
+        ls_command = [
```
pankajastro, line 381: Hey, I noticed that the way dbt ls --output-keys options are supplied has changed across different dbt versions. Here are some examples from various versions:

- [dbt v1.4.2](https://github.com/dbt-labs/dbt-core/blob/v1.4.2/tests/functional/list/test_list.py#L543)
- [dbt v1.3.7](https://github.com/dbt-labs/dbt-core/blob/v1.3.7/tests/functional/list/test_list.py#L543)
- [dbt v1.5.7](https://github.com/dbt-labs/dbt-core/blob/v1.5.7/tests/functional/list/test_list.py#L605)
- [dbt v1.8.3](https://github.com/dbt-labs/dbt-core/blob/v1.8.3/tests/functional/list/test_list.py#L741)

For version 1.5.x and the latest expected like below 
```bash
ls_command = [
  ....
  "--output-keys",
  "name",
  "unique_id"
]
```
And for 1.4.x and older it is like 

```
ls_command = [
  ...
  "--output-keys",
  "name, unique_id",
]
```

To handle both cases, I can see 2 option

1. Run the dbt version command, parse the version number, and then construct the appropriate dbt command based on that version. The downside is that this might be a bit slow and could introduce bugs if the output format of dbt -v changes.
2. Introduce a config variable, TEST_SOURCE_GT_V14, to determine the command construction. While this approach might seem a bit unconventional from a user experience side, it would be relatively bug-free and faster.

I'd appreciate any thoughts or suggestions you might have on this.

tatiana: @pankajastro, there is a third approach:

a) Try to run the following (which would work if `dbt-core>=1.5`):
```
ls_command = [ .... "--output-keys", "name", "unique_id" ]
```

b) If it fails (is there some specific message in the command output we could check, it would be great), we then try to run using the `dbt-core<=1.4` approach:
```
ls_command = [ ... "--output-keys", "name, unique_id", ]
```

The advantage of this approach is that it would not incur additional costs to end-users already using the latest versions of dbt. In contrast, it would incur one more dbt subprocess call to users who are using older versions of dbt-core, without the need for any additional configuration.

The cost for both of these is better than approach (1) because it would not be running subprocess twice every time.  Both of them will have performance improvements once we implement:
https://github.com/astronomer/astronomer-cosmos/issues/865

The challenge with approach (2) is to force users to set yet another configuration.

Approch (4), which may be cleaner on the Cosmos side, is to describe in the docs that Source nodes are only supported with dbt-core>=1.5.

pankajkoti: @tatiana & @pankajastro discussed this further and we feel with approach (4), we however also need to capture the specific error for invalid command and raise relevant warnings/logs

pankajastro: Yes, the 4th option seems reasonable to me. This is what I believe we should proceed with.

1. In the current implementation, the default behavior of source_rendering_behavior is set to SourceRenderingBehavior.ALL. I propose changing this default behavior to SourceRenderingBehavior.NONE. This adjustment ensures that the default behavior remains consistent with the old DAG rendering behavior, providing no change for users by default.
https://github.com/astronomer/astronomer-cosmos/blob/d788c040e87e9d11d06f3cf605c30683ca5ae4c3/cosmos/config.py#L80
```python
source_rendering_behavior: SourceRenderingBehavior = SourceRenderingBehavior.NONE 
```
2. If source node rendering is SourceRenderingBehavior.NONE then keep dbt ls command as it was before this PR otherwise change the command which work for dbt-core >= 1.5.0.
Something like
```
ls_command = []

if self.render_config.source_rendering_behavior != SourceRenderingBehavior.NONE:
    ls_command = [
        dbt_cmd,
        "ls",
        "--output",
        "json",
        "--output-keys",
        "name",
        "unique_id",
        "resource_type",
        "depends_on",
        "original_file_path",
        "tags",
        "config",
        "freshness"
    ]
else:
    ls_command = [dbt_cmd, "ls", "--output", "json"]

ls_args = self.dbt_ls_args
ls_command.extend(self.local_flags)
ls_command.extend(ls_args)

try:
    stdout = run_command(ls_command, tmp_dir, env_vars)
except CosmosLoadDbtException as e:
    if self.render_config.source_rendering_behavior != SourceRenderingBehavior.NONE:
        logger.warning(f"cosmos support source node rendering for dbt-core > 1.5.0 {e}")
    else:
        raise e
```
https://github.com/astronomer/astronomer-cosmos/blob/d788c040e87e9d11d06f3cf605c30683ca5ae4c3/cosmos/dbt/graph.py#L381
3. Add `source_rendering_behavior` into the integration tests that run for dbt-core >= 1.5.0
4. Document that this feature is available only dbt-core >=1.5.0

cosmos/airflow/graph.py
```diff
@@ -127,7 +128,11 @@ def create_test_task_metadata(
 
 
 def create_task_metadata(
-    node: DbtNode, execution_mode: ExecutionMode, args: dict[str, Any], use_task_group: bool = False
+    node: DbtNode,
+    execution_mode: ExecutionMode,
+    args: dict[str, Any],
+    use_task_group: bool = False,
+    source_rendering_behavior: SourceRenderingBehavior = SourceRenderingBehavior.ALL,
```
pankajastro, line 135:
```suggestion
    source_rendering_behavior: SourceRenderingBehavior = SourceRenderingBehavior.NONE,
```

cosmos/config.py
```diff
@@ -59,6 +60,7 @@ class RenderConfig:
     :param dbt_project_path: Configures the DBT project location accessible on the airflow controller for DAG rendering. Mutually Exclusive with ProjectConfig.dbt_project_path. Required when using ``load_method=LoadMode.DBT_LS`` or ``load_method=LoadMode.CUSTOM``.
     :param dbt_ls_path: Configures the location of an output of ``dbt ls``. Required when using ``load_method=LoadMode.DBT_LS_FILE``.
     :param enable_mock_profile: Allows to enable/disable mocking profile. Enabled by default. Mock profiles are useful for parsing Cosmos DAGs in the CI, but should be disabled to benefit from partial parsing (since Cosmos 1.4).
+    :param source_rendering_behavior: Determines how source nodes are rendered when using cosmos default source node rendering (ALL, NONE, WITH_TESTS_OR_FRESHNESS). Defaults to "ALL" (since Cosmos 1.6).
```
pankajastro, line 63:
```suggestion
    :param source_rendering_behavior: Determines how source nodes are rendered when using cosmos default source node rendering (ALL, NONE, WITH_TESTS_OR_FRESHNESS). Defaults to "NONE" (since Cosmos 1.6).
```

cosmos/operators/local.py
```diff
@@ -248,6 +252,29 @@ def store_compiled_sql(self, tmp_project_dir: str, context: Context, session: Se
         else:
             logger.info("Warning: ti is of type TaskInstancePydantic. Cannot update template_fields.")
 
+    @provide_session
+    def store_freshness_json(self, tmp_project_dir: str, context: Context, session: Session = NEW_SESSION) -> None:
```
pankajastro, line 256: shall we add some test for this method?

docs/configuration/source-nodes-rendering.rst
```diff
@@ -0,0 +1,36 @@
+.. _source-nodes-rendering:
+
+Source Nodes Rendering
+================
+
+.. note::
+    This feature is only available for dbt-core >= 1.5.
```
pankajastro, line 7:
```suggestion
    This feature is only available for dbt-core >= 1.5 and cosmos >= 1.6.0.
```

pankajastro (commented):
Looks great—just a few minor comments

scripts/test/integration.sh
```diff
@@ -3,6 +3,8 @@
 set -x
 set -e
 
+export SOURCE_RENDERING_BEHAVIOR=all
```
pankajastro, line 6: Do we need to export it to workflow config too https://github.com/astronomer/astronomer-cosmos/blob/main/.github/workflows/test.yml#L147

tests/airflow/test_graph.py
```diff
@@ -29,6 +31,7 @@
 from cosmos.profiles import PostgresUserPasswordProfileMapping
 
 SAMPLE_PROJ_PATH = Path("/home/user/path/dbt-proj/")
+SOURCE_RENDERING_BEHAVIOR_ENV = os.getenv("SOURCE_RENDERING_BEHAVIOR", "none")
```
pankajastro, line 34: May need to type cast?
```suggestion
SOURCE_RENDERING_BEHAVIOR_ENV = SourceRenderingBehavior(os.getenv("SOURCE_RENDERING_BEHAVIOR", "none"))
```

cosmos/operators/local.py
arojasb3: Do you have any suggestions for this?
I'm not 100% sure how to access the rendered template of a node after a run (if not using the UI)

scripts/test/integration.sh
arojasb3: Just added them to
- integration tests
- 1.5.4 tests
- integration expensive

Let's wait for the results of the CI...

cosmos/operators/local.py
pankajastro: I added some unit tests

cosmos/airflow/graph.py
```diff
@@ -154,6 +160,23 @@ def create_task_metadata(
...
+                and node.has_test is False
+            ):
+                return None
+            # TODO: https://github.com/astronomer/astronomer-cosmos
+            # pragma: no cover
+            task_id = f"{node.name}_source"
+            args["select"] = f"source:{node.resource_name}"
+            args.pop("models")
+            if use_task_group is True:
+                task_id = node.resource_type.value
```
tatiana, line 176: Should we keep only this and remove the lines 182-183, that contain the same?
```
            if use_task_group is True:
                task_id = node.resource_type.value
```

tatiana (approved):
@arojasb3 @pankajastro You can have done an outstanding work developing this feature, adjusting to all change requests, adapting the interfaces, making it backwards compatible, making compromises depending on the dbt version, documenting and testing.

This is an amazing piece of work and I can't wait to hear the community feedback once we release it in Cosmos 1.6.

As follow up tickets:
- We'll be improving the test coverage: https://github.com/astronomer/astronomer-cosmos/issues/1155
- In Cosmos 2.x, we'll change the default `SourceRenderingBehavior` so more people can benefit from this

I left a minor comment that can be addressed as part of the test coverage improvements, that are not a requirement for the release of this feature.

pankajastro:
Thanks, @arojasb3, for the contribution; we appreciate it! 🚀

fabiomx:
Hi!

Sorry for commenting here and now that it's already merged, but it's very small detail.

When implementing this new feature (thanks a lot for this, @arojasb3!), it felt natural to try importing `SourceRenderingBehavior` directly from `cosmos`, just like `LoadMode` or `TestBehavior`. However, I noticed that `SourceRenderingBehavior` isn't included in the `__init__` file and has to be imported from `cosmos.constants`.

I think it would be good if `SourceRenderingBehavior` could be imported in the same way as the others. What do you think about adding it?

P.S. The same might apply to the `InvocationMode` constant.

pankajastro:
Hi @fabiomx, Thanks for your feedback! Feel free to submit a PR anytime—I’d be happy to review and merge it.

Issue #906, opened 2024-04-04
Simplify logging


jlaneve:
We add colors and other formatting to the logs that make it very difficult to read. This came up in the #airflow-dbt channel on the Airflow Slack: https://apache-airflow.slack.com/archives/C059CC42E9W/p1711624722823329

We should look at how we do logging today and potentially think about an overhaul. We should be able to remove the colors, duplicate timestamps, etc.

dwreeves:
This is small but a simple starting point would be to move to the LoggingMixin instead of a global logger. This would still be backwards compatible I believe, wouldn't change any behaviors, but at least provides opportunities to users to override logging behaviors, and is overall more idiomatic in Airflow world.

pankajkoti:
This issue was reported again by the community:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1715739144301439

This was the Slack message body in case we lose the above thread data
```
Hi everyone,
Actually i'm using astronomer-cosmos==0.6.2 and the dbt models run correctly in astro airflow, the only problem is that scheduler logs are bigger and bigger, and almost all the logs contains the same info (check the sample below) but as i mentioned before the process executes as expected, the diagram is correctly. Anyone knows how to avoid this or fix it because the log size of scheduler is almost ~5 Gb per day?
[2024-05-14T16:56:57.780+0000] {logging_mixin.py:154} INFO - [2024-05-14T16:56:57.780+0000] {render.py:214} ERROR - Dependency stg_talent_companies not found for model DbtModel(name='stg_goal_processes', type='DbtModelType.DBT_MODEL', path='/usr/local/airflow/dbt/ch_analytics/models/staging/stg_talent/stg_goal_processes.sql', config=DbtModelConfig(config_selectors=set(), upstream_models={'stg_talent_companies', 'talent_innpulsa'}))
```

Temptively adding this to the 1.5 milestone

dwreeves:
Big post incoming

# Issue

Users have been seeing duplicate logs and other logging related issues (see #639, #747. #906) for a while.

In my experience, logging issues have gotten a little out of hand.  When running the dev container, I noticed that the logs were being tripled. That was new to me, and I hadn’t recalled that occurring the last time I worked with Cosmos. After running the `main` branch of Cosmos with a completely blank slate, here are what my logs look like:

```
[...]

PR #1108, opened 2024-07-21
Simplify logging

dwreeves:
## Description

This PR addresses #906 and fixes issues in the Cosmos logging once and for all.\*

> \* Actually, there is another issue with logs being polluting with warnings about "source" nodes in 1.5.0, but that is a separate matter!

I have a long explanation of how the `logging` module in Python works, and the sort of idioms it expects of end users of the module, here: https://github.com/astronomer/astronomer-cosmos/issues/906#issuecomment-2134150103

The choices I made, explained:

- Although I don't know that I entirely agree with adding `(astronomer-cosmos)` to all the logs, clearly at least one user, and possibly many more, want it, and I don't believe we should remove it. The objective of this PR was therefore to preserve the feature while future-proofing against future issues.
    - Why I can't say I'm a fan of it: It seems that adding `(astronomer-cosmos)` to logs is a symptom of other problems with the Cosmos library, specifically how it impacts performance when users do not set it up effectively. And the prefix was added as a way to assist people in diagnosing these issues. I think ultimately we want to move away from this. Other components of the Airflow ecosystem do not feel compelled to do things like this. Also, the module path is something that can be handled in the `log_format` if users really want it.
    - How I future-proofed: As per the long post I link above, basically the issue is that there should not be tons of StreamHandlers being created. The proper and typical use of the logging module, with few exceptions, is to allow for logs to propagate upwards to a root logger. The reason the Cosmos logs presented issues for so long was because it deviated a lot from this.
- I think default behavior being the "least astonishing" means making no modifications to the base logging behavior whatsoever. This is also less likely to morph into future issues if any further changes are made to the custom logging.
    - One thing I never mentioned: I found it odd that by default Cosmos did not "work out of the box" and that, despite using Astronomer's own Airflow platform (!), I had to set a config option that made Cosmos logging not be a nightmare (i.e. set `propagate_logs` = false). Previous logs referenced the Celery Executor as having issues, even though this is one of 2 of the most popular production ways to run Airflow. Something like this should just work out of the box for a majority of users!
- For task execution, Cosmos should make use of the more Airflow-idiomatic `LoggingMixin` class whenever appropriate. This can also be used in scheduler / webserver related logging contexts but I think it is less out-of-place there to use globally scoped loggers.
  - These will not use the `get_logger()` implementation. That is intentional and probably desirable. These logs do not need to be "enriched" because they are isolated in the task execution logs.

Oh also, I fixed an issue in the `project.entry_points` in the `pyproject.toml` while I was at it.

## Breaking Change?

- Removes `propagate_logging` conf option, although removing this will not break users' builds. There is now a `rich_logging` conf option instead, which by default is disabled.

## Checklist


dwreeves:
So what's going on with the status of this PR is: I swapped the `logger.info()` for `self.log.info()` (same for `error`, `warning` etc.) inside the operators.

It's failing due to coverage for `upload_to_cloud_storage()`, which was _previously not covered_, but due to the fact I change how the logging works, it is now being marked as a coverage miss.

@tatiana I am wondering if we can make an exception for the coverage on that? I don't really want to have to write a unit test for code I don't even use right now 😭

The PR otherwise works, is documented, and has unit tests. So I think it should be fine and I removed the "WIP" from it.

tatiana (approved):

Issue #448, opened 2023-08-08
Support loading manifest from S3/GCS/Azure BlobStorage


tatiana:
While talking to one of the community members, Edgaras Navickas, he mentioned it would be great if users could reference a manifest in an S3 bucket. This was a follow-up to issues reported in the [slack thread](https://apache-airflow.slack.com/archives/C059CC42E9W/p1691483255763189?thread_ts=1691398106.912769&cid=C059CC42E9W).

Example:
```
DbtDag(
    project_config=ProjectConfig(
        manifest_path="s3://path/to/manifest.json",
        manifest_conn_id="aws_conn",
    ),
    render_config=RenderConfig(
        load_mode=LoadMode.DBT_MANIFEST,
    )
    # ...,
)
```

We can have separate tickets to support loading manifests from other cloud providers.

MrBones757:
I'd like to add some additional ideas / comments here.

Rather than support the s3 uri, would it be worth creating a set of classes similar to the way we handle profiles.

I.e im thinking something like

ManifestSourceBase
 -> S3ManifestSource
 -> ArtifactoryManifestSource
 -> NexusManifestSource
 etc
 
This would make it really modular and allow us to source from numerous artifact stores, and support more than aws specific s3, think azure blob store, Google Storage, Cloudflare R2 as well as those above.

idealopamp:
> I'd like to add some additional ideas / comments here.
> 
> Rather than support the s3 uri, would it be worth creating a set of classes similar to the way we handle profiles.
> 
> I.e im thinking something like
> 
> ManifestSourceBase -> S3ManifestSource -> ArtifactoryManifestSource -> NexusManifestSource etc
> 
> This would make it really modular and allow us to source from numerous artifact stores, and support more than aws specific s3, think azure blob store, Google Storage, Cloudflare R2 as well as those above.
> 
> Thinking about integration, it would be fairly easy to just allow ManifestSourceBase as a possible type for the manifest path arg, and make ManifestSourceBase an abstract base.
> 
> This also relates in a way to #570, as they are somewhat related & competing ideas as they both deal with a remotely sourced manifest (and indeed, profiles.yml, for which the same logic could be used - Perhaps ManifestSourceBase -> CosmosFileSource or something, that would accept injects creds, or airflow conns)

This would be great. Our team would love to see this for Google Cloud Storage.

tatiana:
[...]

PR #1109, opened 2024-07-21
Add support for loading manifest from cloud store using Airflow Object Store

pankajkoti:
## Summary

This PR introduces the capability to load `manifest.json` files from various cloud storage services using Airflow's Object Store integration. The supported cloud storages include AWS S3, Google Cloud Storage (GCS), and Azure Blob Storage. The feature allows seamless integration with remote paths, providing enhanced flexibility and scalability for managing DBT projects.

### Key Changes

1. Parameters in DbtDag and DbtTaskGroup:
`manifest_path`: Accepts both local paths and remote URLs (e.g., S3, GCS, Azure Blob Storage).
`manifest_conn_id`: (Optional) An Airflow connection ID for accessing the remote path.
2. Automatic Detection of Storage Type:
The system automatically identifies the storage service based on the scheme of the URL provided (e.g., s3://, gs://, abfs://) by integrating with Airflow Object Store
3. If a `manifest_conn_id` is provided, it is used to fetch the necessary credentials.
4. If no `manifest_conn_id` is provided, the default connection ID for the identified scheme is used.

### Validation and Error Handling

1. Validates the existence of the `manifest.json` file when a path is specified.
2. Raises appropriate errors if a remote `manifest_path` is given but the required min Airflow version 2.8(Object Store feature) support is not available.

### Backward Compatibility
Ensures compatibility with existing workflows that use local paths for the manifest.json.

### How to Use

1. Local Path:
```
DbtDag(
    project_config=ProjectConfig(
        manifest_path="/path/to/local/manifest.json",
    ),
    ...
)
```
2. Remote Path (e.g., S3):
```
DbtDag(
    project_config=ProjectConfig(
        manifest_path="s3://bucket/path/to/manifest.json",
        manifest_conn_id="aws_s3_conn",
    ),
    ...
)
```
3.  Remote Path without Explicit Connection ID:
```
DbtDag(
    project_config=ProjectConfig(
        manifest_path="gs://bucket/path/to/manifest.json",
        # No manifest_conn_id provided, will use default Airflow GCS connection `google_cloud_default`
    ),
    ...
)
```

### Additional Notes

1. Ensure that the required Airflow version (2.8 or later) is used to take advantage of the Object Store features.
2. Review the updated documentation for detailed usage instructions and examples.

### Testing

1. Added unit tests to cover various scenarios including local paths, remote paths with and without manifest_conn_id.
2. Verified integration with different cloud storage services (AWS S3, GCS, Azure Blob Storage).
3. Ensured backward compatibility with existing local path workflows.

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/448

## Breaking Change?
No.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, 448-remote-manifest-support]
+    branches: [main]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, 448-remote-manifest-support]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, 448-remote-manifest-support]
+    branches: [main, "448-remote-manifest-support"]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -143,6 +143,9 @@ jobs:
           AIRFLOW__COSMOS__ENABLE_CACHE_DBT_LS: 0
           AIRFLOW_HOME: /home/runner/work/astronomer-cosmos/astronomer-cosmos/
           AIRFLOW_CONN_EXAMPLE_CONN: postgres://postgres:postgres@0.0.0.0:5432/postgres
+          AIRFLOW_CONN_AWS_S3_CONN: ${{ secrets.AIRFLOW_CONN_AWS_S3_CONN }}
+          AIRFLOW_CONN_GCP_GS_CONN: ${{ secrets.AIRFLOW_CONN_GCP_GS_CONN }}
+          AIRFLOW_CONN_AZURE_ABFS_CONN: ${{ secrets.AIRFLOW_CONN_AZURE_ABFS_CONN }}
           DATABRICKS_HOST: mock
           DATABRICKS_WAREHOUSE_ID: mock
           DATABRICKS_TOKEN: mock
@@ -213,6 +216,9 @@ jobs:
         env:
           AIRFLOW_HOME: /home/runner/work/astronomer-cosmos/astronomer-cosmos/
           AIRFLOW_CONN_EXAMPLE_CONN: postgres://postgres:postgres@0.0.0.0:5432/postgres
+          AIRFLOW_CONN_AWS_S3_CONN: ${{ secrets.AIRFLOW_CONN_AWS_S3_CONN }}
+          AIRFLOW_CONN_GCP_GS_CONN: ${{ secrets.AIRFLOW_CONN_GCP_GS_CONN }}
+          AIRFLOW_CONN_AZURE_ABFS_CONN: ${{ secrets.AIRFLOW_CONN_AZURE_ABFS_CONN }}
           PYTHONPATH: /home/runner/work/astronomer-cosmos/astronomer-cosmos/:$PYTHONPATH
           AIRFLOW_CONN_DATABRICKS_DEFAULT: ${{ secrets.AIRFLOW_CONN_DATABRICKS_DEFAULT }}
           DATABRICKS_CLUSTER_ID: ${{ secrets.DATABRICKS_CLUSTER_ID }}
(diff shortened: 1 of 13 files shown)


pyproject.toml
```diff
@@ -129,7 +129,7 @@ packages = ["/cosmos"]
 dependencies = [
     "astronomer-cosmos[tests]",
     "apache-airflow-providers-cncf-kubernetes>=5.1.1",
-    "apache-airflow-providers-amazon>=3.0.0,<8.20.0", # https://github.com/apache/airflow/issues/39103
+    "apache-airflow-providers-amazon[s3fs]>=3.0.0,<8.20.0", # https://github.com/apache/airflow/issues/39103
```
tatiana, line 132: Do we still need this upper bound limit? It seems the issue was merged in May

cosmos/config.py
```diff
@@ -196,24 +219,21 @@ def validate_project(self) -> None:
         if self.dbt_project_path:
             project_yml_path = self.dbt_project_path / "dbt_project.yml"
             mandatory_paths = {
```
tatiana, line 221: The first time I saw this, I was confused about why we're doing things this way. It may be worth to add some comment on why

cosmos/config.py
```diff
@@ -196,24 +219,21 @@ def validate_project(self) -> None:
...
-                "dbt_project.yml": project_yml_path,
-                "models directory ": self.models_path,
+                "dbt_project.yml": Path(project_yml_path) if project_yml_path else None,
```
tatiana, line 222: The first time I saw this, I was confused about why we're doing things this way. It may be worth to add some comment on why

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, 448-remote-manifest-support]
```
tatiana, line 5: Before we merge: 
```suggestion
    branches: [main]
```

tatiana (commented):
It looks great, @pankajkoti , great work!

I know this PR is still in draft, but some early feedback:
- docs: we need to update the docs (both on the `ProjectConfig` and also in the manifest parsing method, mentioning the minimum version of Cosmos/Airflow to support remote manifest locations
- dependencies: probably worth have extras specific per cloud provider, similar to Airflow, so we do not to introduce new extras if we add additional improvements per cloud provider

cosmos/config.py
pankajkoti: added a comment, please check.

pyproject.toml
pankajkoti: I have added extras for the cloud providers. Still need to check if we can remove the upper bound and also recap on why exactly it was added.

pankajkoti:
> It looks great, @pankajkoti , great work!
> 
> I know this PR is still in draft, but some early feedback:
> 
> * docs: we need to update the docs (both on the `ProjectConfig` and also in the manifest parsing method, mentioning the minimum version of Cosmos/Airflow to support remote manifest locations
> * dependencies: probably worth have extras specific per cloud provider, similar to Airflow, so we do not to introduce new extras if we add additional improvements per cloud provider

Thanks for the early feedback. I have added docs and also added cloud provider specific dependencies. I am taking the PR out of draft. I believe I have addressed the comments so far except for checking yet if we can remove the upper bound on the Amazon provider dependency.

.github/workflows/test.yml
pankajkoti, line 5: as suggested in previous comment https://github.com/astronomer/astronomer-cosmos/pull/1109#discussion_r1687979238, duplicating the comment so that it appears well before merging.
```suggestion
    branches: [main]
```

docs/configuration/parsing-methods.rst
```diff
@@ -34,26 +34,63 @@ When you don't supply an argument to the ``load_mode`` parameter (or you supply
...
+Examples of how to supply ``manifest.json`` using ``manifest_path`` argument:
+
+- Local path:
+
+.. literalinclude:: ../../dev/dags/cosmos_manifest_example.py
+    :language: python
+    :start-after: [START local_example]
+    :end-before: [END local_example]
+
+- AWS S3 URL:
```
tatiana, line 64:
```suggestion
- AWS S3 URL (available since Cosmos 1.6):
```

docs/configuration/parsing-methods.rst
```diff
@@ -34,26 +34,63 @@ When you don't supply an argument to the ``load_mode`` parameter (or you supply
...
+
+Ensure that you have the required dependencies installed to use the S3 URL. You can install the required dependencies
+using the following command: ``pip install "astronomer-cosmos[amazon]"``
+
+.. literalinclude:: ../../dev/dags/cosmos_manifest_example.py
+    :language: python
+    :start-after: [START aws_s3_example]
+    :end-before: [END aws_s3_example]
+
+- GCP GCS URL:
```
tatiana, line 74:
```suggestion
- GCP GCS URL (available since Cosmos 1.6):
```

docs/configuration/parsing-methods.rst
```diff
@@ -34,26 +34,63 @@ When you don't supply an argument to the ``load_mode`` parameter (or you supply
...
+
+Ensure that you have the required dependencies installed to use the GCS URL. You can install the required dependencies
+using the following command: ``pip install "astronomer-cosmos[google]"``
+
+.. literalinclude:: ../../dev/dags/cosmos_manifest_example.py
+    :language: python
+    :start-after: [START gcp_gs_example]
+    :end-before: [END gcp_gs_example]
+
+- Azure Blob Storage URL:
```
tatiana, line 84:
```suggestion
- Azure Blob Storage URL (available since Cosmos 1.6):
```

tatiana (approved):
This is a very waited feature, @pankajkoti , thanks a lot for addressing all the feedback! I left three minor comments, and we're good to merge.

dwreeves:
Hi, just noticed this one testing some other code and this caused my code to fail.

I don't think it is appropriate for Cosmos to create explicit dependencies on AWS, GCP, and Azure providers, which is what this code does by eagerly loading the providers into the level. This will be a disruption to a lot of users, and also adds additional bloat to their environments that they genuinely don't need.

dwreeves:
^ I've opened PR #1132 to address the issue raised in the above comment.

Issue #1096, opened 2024-07-17
Bug on Cosmos 1.5 `No such file or directory "/usr/local/aiflow/dags/dbt/dbt_venv/bin/python"`


tatiana:
Two Astro customers reported that when updating to Cosmos 1.5, they started seeing the error:
```
No such file or directory "/usr/local/aiflow/dags/dbt/dbt_venv/bin/python"
```

As illustrated in the two screenshots below:

![Image](https://github.com/user-attachments/assets/148fc806-c112-42f6-ace3-958e2386e7fa)
![Image](https://github.com/user-attachments/assets/859eb89e-8157-4d93-8492-052f4a7b1500)

tatiana:
After lots of time trying to reproduce and understand this issue, I narrowed it down:
- inside the customers dbt project folder, there was a symbolic link to a no longer existent file

Cosmos was raising an exception while trying to parse the file and create a version hash to represent the version of that dir.

I validated a possible solution with one of the customers earlier today, available in Cosmos 1.5.1a3 - and they confirmed it solved the problem. I'll make a quick PR with the solution and unittests today

tatiana:
We fixed the issue in #1097, and, in parallel to this, we would like to encourage users to have Python virtual environments outside the dbt project folder and outside of the Airflow DAGs folder.

Issue #1076, opened 2024-07-02
[Bug] RuntimeError: Detected recursive loop for /usr/local/airflow/dags/dbt/dbt_venv/lib


oliverrmaa:
### Astronomer Cosmos Version

Other Astronomer Cosmos version (please specify below)

### If "Other Astronomer Cosmos version" selected, which one?

1.4.3

### dbt-core version

1.7.17

### Versions of dbt adapters

dbt-bigquery==1.7.4
dbt-core==1.7.17
dbt-extractor==0.5.1
dbt-semantic-interfaces==0.4.4

### LoadMode

DBT_LS

### ExecutionMode

LOCAL

### InvocationMode

SUBPROCESS

### airflow version

apache-airflow==2.9.2+astro.1

### Operating System

Debian GNU/Linux 11 (bullseye)

### Deployment

Astronomer

### Deployment details

We have a main production deployment in Astro Cloud which we consider as production. We also do local development via `astro dev start`. We have continuous deployment set up through CircleCI which deploys merged PRs to our master branch to our production deployment via `astro deploy --dags`. For authentication to our data warehouse (Google BigQuery) in production, we use `GoogleCloudServiceAccountDictProfileMapping` and for local we use `ProfileConfig` where our dbt profiles.yml has a hardcoded path to a service account JSON file which is at the same path for each developer.

### What happened?

We noticed that we are having a RuntimeError constantly (I believe every second according to our Astro production deployment logs): `Detected recursive loop when walking DAG directory /usr/local/airflow/dags: /usr/local/airflow/dags/dbt/dbt_venv/lib has appeared more than once.`

We aren't sure whether this error is going to cause parts of our setup to work sub optimally or connected to any other issues we are seeing in our production environment, i.e. like in https://github.com/astronomer/astronomer-cosmos/issues/1075, or if it will cause further issues in the future. 

### Relevant log output

```shell
[...]

PR #1115, opened 2024-07-23
Add cleanup functions to fix recursive loop and no such file raised by broken symlinks

tatiana:
Introduce tools that help users facing two issues raised by problematic Airflow or dbt project directories:
1. File not found (as reported in #1096)
2. Recursive loop (as reported in #1076)

**Disclaimer**

Cosmos is not the cause of the two issues previously mentioned.  The reason for these problems is a misconfiguration of the Airflow and/or dbt directories created by users and set (by users) to be used in Cosmos DAGs.

This PR proposes methods to mitigate these problems, but users need to introduce this or similar tooling in their deployment processes to avoid the issue from happening again.

**Details on the problems addressed**

The first of these issues relates to the dbt project folder having a symbolic link to a file that no longer exists:
```
No such file or directory "/usr/local/aiflow/dags/dbt/dbt_venv/bin/python"
```

The second of these issues relates to the dbt project folder having a symbolic link to a parent folder, which leads Airflow not to be able to parse dbt DAGs at all:
```
RuntimeError: Detected recursive loop for /usr/local/airflow/dags/dbt/dbt_venv/lib
```

The second issue, in particular, can be very destructive since it can block the scheduler from parsing any DAGs in the Airflow project, leading to critical DAGs not running in the expected schedules.

**How to use this PR**

For problem (1), this PR has an example DAG, `example_cosmos_cleanup_dir_dag,` which can be scheduled as any other DAG in Airflow.

It was tested by using the following steps:

(a) Create symlinks and make them invalid

```
touch sample-file
ln -s `pwd`/sample-file dev/dags/dbt/bla2 
ln -s `pwd`/sample-file dev/dags/dbt/bla3 
rm sample-file
```

(b) Triggering this DAG
```
airflow dags test example_cosmos_cleanup_dir_dag  `date -Iseconds`
```

In its logs, we can see:
```
[2024-07-23T02:15:00.995+0100] {cleanup.py:25} INFO - Inspecting the directory /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt for broken symbolic links.
[2024-07-23T02:15:00.997+0100] {cleanup.py:38} WARNING - The folder /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt contains a symbolic link to a non-existent file: /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt/bla3
[2024-07-23T02:15:00.997+0100] {cleanup.py:40} INFO - Deleting the invalid symbolic link: /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt/bla3
[2024-07-23T02:15:00.997+0100] {cleanup.py:38} WARNING - The folder /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt contains a symbolic link to a non-existent file: /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt/bla2
[2024-07-23T02:15:00.997+0100] {cleanup.py:40} INFO - Deleting the invalid symbolic link: /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt/bla2
[2024-07-23T02:15:00.998+0100] {cleanup.py:44} INFO - After inspecting /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt, identified 2 broken links and deleted 2 of them.
```

When trying to run it again, it's possible to see no more broken symlinks were found:

```
After inspecting /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt, identified 0 broken links and deleted 0 of them.
```

For problem (2), since the problem happens at Airflow DAG parsing, this issue has to be solved outside of Airflow DAGs/tasks.

We are exposing a command line that allows users to visualize and delete recursive loops. This is an example of how to do both things:
```
python -m cosmos.cleanup -p dags/dbt -d 
```

The expectation is that end-users can run this command before deploying to Airflow, ensuring the deployed dbt project folder is valid and does not contain recursive links. This same command also checks for broken symbolic links.

To validate this feature, we did the same steps to reproduce the broken symlinks and also:
```
ln -s `pwd`/astronomer-cosmos `pwd`/astronomer-cosmos/dags/dbt/bla
```

Example of logs created by running this command line:
```
[2024-07-23T01:50:04.264+0100] {cleanup.py:49} INFO - Inspecting the directory dags/dbt for recursive loops.
[2024-07-23T01:50:04.267+0100] {cleanup.py:64} WARNING - Detected recursive loop from /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt/bla to /Users/tati/Code/cosmos-clean/astronomer-cosmos
[2024-07-23T01:50:04.267+0100] {cleanup.py:67} INFO - Deleting symbolic link: /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt/bla
[2024-07-23T01:50:04.267+0100] {cleanup.py:71} INFO - After inspecting /Users/tati/Code/cosmos-clean/astronomer-cosmos/dags/dbt, identified 1 recursive loops and deleted 1 of them.


[2024-07-23T01:50:04.267+0100] {cleanup.py:17} INFO - Inspecting the directory dags/dbt for broken symbolic links.
[2024-07-23T01:50:04.268+0100] {cleanup.py:30} WARNING - The folder dags/dbt contains a symbolic link to a non-existent file: dags/dbt/bla3
[2024-07-23T01:50:04.268+0100] {cleanup.py:32} INFO - Deleting the invalid symbolic link: dags/dbt/bla3
[2024-07-23T01:50:04.269+0100] {cleanup.py:30} WARNING - The folder dags/dbt contains a symbolic link to a non-existent file: dags/dbt/bla2
[2024-07-23T01:50:04.269+0100] {cleanup.py:32} INFO - Deleting the invalid symbolic link: dags/dbt/bla2
[2024-07-23T01:50:04.269+0100] {cleanup.py:36} INFO - After inspecting dags/dbt, identified 2 broken links and deleted 2 of them.
```

From an end-user perspective, the script that solves the problem (2) can be automatically run in a few places, including
- As part of the CI/CD pipelines
- As a pre-commit hook

Closes: #1096
Closes: #1076


cosmos/cleanup.py
```diff
@@ -0,0 +1,95 @@
...
+                logger.info(f"Deleting the invalid symbolic link: {filepath}")
+                os.unlink(filepath)
+                deleted_symlinks_count += 1
+
+    logger.info(
+        f"After inspecting {dir_path}, identified {broken_symlinks_count} broken links and deleted {deleted_symlinks_count} of them."
+    )
+
+
+# Airflow DAG parsing fails if recursive loops are found, so this method cannot be used from within an Airflow task
```
pankajkoti, line 49: Should we also add an example here on how to call this and where to call this like mentioned in the PR description.

Or should we also create a public docs page listing the steps that we can share with users?

cosmos/cleanup.py
pankajkoti, line 1: Should we add some tests for this module or exclude from codecov for the time being?

pankajkoti (commented):
LGTM mostly. Minor questions inline

cosmos/cleanup.py
pankajastro: I think a short docs will be great and also we could render example in the docs

cosmos/cleanup.py
pankajastro, line 1: It sounds like an operator or decorator. I'm ok with keeping it in the project's root path, but what are your thoughts on placing it in the operator directory instead?

dev/dags/example_cosmos_cleanup_dir_dag.py
```diff
@@ -0,0 +1,38 @@
+"""
+We've observed users who had dbt project directories containing symbolic links to files that no longer existed.
+
+Although this issue was not created by Cosmos itself, since this issue was already observed by two users, we thought it
+was useful to give an example DAG illustrating how to clean the problematic directories.
+
+Assuming the cause of the issue no longer exists, this DAG can be run only once.
+"""
+
+# [START dirty_dir_example]
```
pankajastro, line 10: It would be great if we could use tag `dirty_dir_example` to render this example in docs

cosmos/cleanup.py
pankajastro: +1

tatiana: The challenge is that while `identify_broken_symbolic_links`  could be exposed as an operator, ideally, users would check this before it is deployed to a remote instance of Airflow. Additionally, `identify_recursive_loops` cannot be exposed as an operator since it would be useless (the exception is raised at Airflow DAG parsing). Therefore, we'd need a non-operator place to expose the second use case, so it felt we could benefit from having both methods together - and it'd be up to the user on how to use them.

cosmos/cleanup.py
```diff
@@ -0,0 +1,95 @@
...
+
+    :param dir_path: Path to the directory to be analysed
+    :param should_delete: Users should set to True if they want the method to not only identify but also delete these links.
+    """
+    logger.info(f"Inspecting the directory {dir_path} for broken symbolic links.")
+    filepaths = []
+    broken_symlinks_count = 0
+    deleted_symlinks_count = 0
+    for root_dir, dirs, files in os.walk(dir_path):
+        paths = [os.path.join(root_dir, filepath) for filepath in files]
```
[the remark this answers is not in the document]
tatiana: It was being checked after, but it would be a good improvement!

tatiana:
Thanks a lot for the reviews, @pankajastro and @pankajkoti! I started addressing them, but we had a call with the team who initially raised this error, and we found a more straightforward workaround for the problem, as described in https://github.com/astronomer/astronomer-cosmos/issues/1076#issuecomment-2246631121

I believe we can close this PR for now, and if this feature becomes useful in the future, we can reopen/recreate it.

PR #1116, opened 2024-07-23
Remove upper bound for the amazon provider to test in CI

pankajkoti:
We previously had pinned the apache-airflow-amazon-provider version 
in PR #944 due to the issue https://github.com/apache/airflow/issues/39103 which has gotten [resolved](https://github.com/apache/airflow/issues/39103#issuecomment-2104729821). 
Hence, remove the upper bound.


pankajkoti:
Closing the PR. Had created for testing.

PR #1127, opened 2024-07-29
Run some example in Kubernetes execution mode in CI

pankajastro:
### Migrate example from [cosmos-example](https://github.com/astronomer/cosmos-example/) 

The [cosmos-example](https://github.com/astronomer/cosmos-example/) repository currently contains several examples, including those that run in Kubernetes execution mode. This setup has made testing local changes in Kubernetes execution mode challenging and keeping the documentation up-to-date is also not easy. Therefore, it makes sense to migrate the Kubernetes examples from [cosmos-example](https://github.com/astronomer/cosmos-example/) to this repository. This PR resolved the below issue in this regard
- Migrate the [jaffle_shop_kubernetes](https://github.com/astronomer/cosmos-example/blob/main/dags/jaffle_shop_kubernetes.py) example DAG to the this repository.
- Moved the Dockerfile from [cosmos-example](https://github.com/astronomer/cosmos-example/blob/main/Dockerfile.postgres_profile_docker_k8s) to this repository to build the image with the necessary DAGs and DBT projects
I also adjusted both the example DAG and Dockerfile to work within this repository.

### Automate running locally 
I introduce some scripts to make running Kubernetes DAG easy.

**postgres-deployment.yaml:** Kubernetes resource file for spinning up PostgreSQL and creating Kubernetes secrets.

**integration-kubernetes.sh:** Runs the Kubernetes DAG using pytest.

**kubernetes-setup.sh:**

- Builds the Docker image with the Jaffle Shop dbt project and DAG, and loads the Docker image into the local registry.
- Creates Kubernetes resources such as PostgreSQL deployment, service, and secret.

**Run DAG locally**
Prerequisites:

- Docker Desktop
- KinD (Kubernetes in Docker)
- kubectl

Steps:
1. Create cluster:  `kind create cluster`
2. Create Resource: `scripts/test/kubernetes-setup.sh` (This will set up PostgreSQL and load the DBT project into the local registry)
3. Run DAG: `cd dev && scripts/test/integration-kubernetes.sh` this will execute this DAG with a pytest you can also run directly with airflow command given that project is installed in your virtual env 
```
time AIRFLOW__COSMOS__PROPAGATE_LOGS=0 AIRFLOW__COSMOS__ENABLE_CACHE=1 AIRFLOW__COSMOS__CACHE_DIR=/tmp/ AIRFLOW_CONN_EXAMPLE_
CONN="postgres://postgres:postgres@0.0.0.0:5432/postgres" PYTHONPATH=`pwd` AIRFLOW_HOME=`pwd` AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT=20000 AIRFLOW__CORE__DAG_FILE_PROCESSOR_TIMEOUT=20000 airflow dags test jaffle_shop_kubernetes  `date -Iseconds`
```
### Run jaffle_shop_kubernetes in CI
To avoid regression we have automated running the jaffle_shop_kubernetes in CI

- Set up the GitHub Actions infrastructure to run DAGs using Kubernetes execution mode
- Use [container-tools/kind-action@v1](https://github.com/container-tools/kind-action) to create a KinD cluster.
- Used the bash script to streamline the creation of Kubernetes resources, build and load the image into a local registry, and execute tests.
- At the moment I'm running the pytest from virtual env

### Documentation changes
Given that the DAG [jaffle_shop_kubernetes](https://github.com/astronomer/cosmos-example/blob/main/dags/jaffle_shop_kubernetes.py) is now part of this repository, I have automated the example rendering for Kubernetes execution mode. This ensures that we avoid displaying outdated example code.

https://astronomer.github.io/astronomer-cosmos/getting_started/execution-modes.html#kubernetes
Screenshot 2024-08-15 at 8 03 59 PM

https://astronomer.github.io/astronomer-cosmos/getting_started/kubernetes.html#kubernetes

Screenshot 2024-08-15 at 8 04 22 PM

### Future work

- Use the hatch target to run the test. I have introduced the hatch target to run the Kubernetes example with hatch, but it's currently not working due to a mismatch between the local and container DBT project paths. This requires a bit more work.
- Remove the virtual environment step (Install packages and dependencies) in the CI configuration for Run-Kubernetes-Tests and use hatch instead.
- Update the profile YAML to use environment variables for the port, as it is currently hardcoded.
- Remove the host from the Kubernetes secret and replace it with the username and make corresponding change in DAG
- Currently, we need to export both POSTGRES_DATABASE and POSTGRES_DB in the Dockerfile because both are used in the project. To ensure consistency, avoid exporting both and instead make the environment variables consistent across the repository
- Not a big deal in this context, but we have some hardcoded values for secrets. It would be better to parameterize them

GH issue for future improvement: https://github.com/astronomer/astronomer-cosmos/issues/1160

### Example CI Run

- https://github.com/astronomer/astronomer-cosmos/actions/runs/10405590862

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/535

## Breaking Change?

No

## Checklist


.github/workflows/test.yml
```diff
@@ -2,9 +2,9 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, kube_mode_ci]
```
```suggestion
    branches: [main]
```

.github/workflows/test.yml
```diff
@@ -2,9 +2,9 @@ name: test
...
   pull_request_target: # Also run on pull requests originated from forks
-    branches: [main]
+    branches: [main, kube_mode_ci]
```
pankajastro, line 7:
```suggestion
    branches: [main]
```

tatiana (approved):
This looks great, @pankajastro , really happy this is automated, it will save us lots of time!

PR #1132, opened 2024-08-01
Fix import handling by lazy loading hooks introduced in PR #1109

dwreeves:
## Description

Making an update to #1109, which introduced module-level imports of optional dependencies. This is inappropriate as it will break if the user does not have them installed, and indeed the user really does not need them installed if they are not relying on them directly.

This PR lazy-loads the imports so that it does not impact users who do not need them.

Additionally, the scheme added for Azure only supported the Azure Data Lake Storage V2 protocol and not the (legacy, but also I believe more common?) `wasb://` protocol, which I added. Additionally, the `conn_id` was being pulled from the wrong hook for the `abfs://` scheme. This is not just nitpicking, as the default `conn_id` for each hook is actually different: the `conn_id` that corresponds with the `abfs://` scheme is `adls_default`, whereas for the `wasb://` scheme it is `wasb_default`.

**This PR should be merged before releasing `1.6` to prevent breaking anyone's environments. 😄** Sorry to sound the **bold** alarms, but this one is actually pretty important.

### DIFF
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -29,7 +29,6 @@ OPENLINEAGE_PRODUCER = "https://github.com/astronomer/astronomer-cosmos/"
 PARTIALLY_SUPPORTED_AIRFLOW_VERSIONS = [Version("2.9.0"), Version("2.9.1")]
 
 
-ABFS_FILE_SCHEME = "abfs"
 
 
 def _default_s3_conn() -> str:
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -29,8 +29,6 @@ OPENLINEAGE_PRODUCER = "https://github.com/astronomer/astronomer-cosmos/"
 PARTIALLY_SUPPORTED_AIRFLOW_VERSIONS = [Version("2.9.0"), Version("2.9.1")]
 
 
-
-
 def _default_s3_conn() -> str:
     from airflow.providers.amazon.aws.hooks.s3 import S3Hook
 
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -1,7 +1,7 @@
 import os
 from enum import Enum
 from pathlib import Path
-from typing import Dict, Callable
+from typing import Callable, Dict
 
 import aenum
 from packaging.version import Version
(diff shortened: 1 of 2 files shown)


cosmos/constants.py
```diff
@@ -31,14 +29,38 @@
...
+    return S3Hook.default_conn_name
+
+
+def _default_gcs_conn() -> str:
+    from airflow.providers.google.cloud.hooks.gcs import GCSHook
+
+    return GCSHook.default_conn_name
+
+
+def _default_wasb_conn() -> str:
```
pankajkoti, line 47: Did specifying a `wasb` URI work for you? Can you please share the format of the `wasb` URI you used? Perhaps we could include that URI format in our example DAG & docs. 
It didn't seem to work when I tried it, which is why I didn't include `wasb`—I considered it a limitation that we only support `abfs`. I tested various schemes like `abfs`, `wasb`, `azure`, and `az`, and was only able to successfully run our `cosmos_manifest_example.py` DAG with the `abfs` scheme, using the URI format specified in [this line](https://github.com/astronomer/astronomer-cosmos/blob/7889fbe7abf3ad79a14530e91ed90ef6352eba4e/dev/dags/cosmos_manifest_example.py#L93).

pankajkoti (commented):
Thanks @dwreeves for trying out from `main` and finding an issue. Many thanks for going ahead and creating a fix ❤️ 

LGTM, just have a question regarding the `wasb` URI format we could use and then I believe we are good to merge it.

cosmos/constants.py
dwreeves: Yes, I see what you mean, it seems that `wasb://` is not supported: https://github.com/fsspec/universal_pathlib/blob/92675f060775e7aeb3ffc92cb74abdb92458d231/README.md?plain=1#L86 It also seems that:

- In the `upath` library, `az:`, `adl:`, `abfs:` and `abfss:` are also all valid schemes, albeit Airflow only references the latter 3 in the code: https://github.com/apache/airflow/blob/e3824eaaba7eada9a807f7a2f9f89d977a210e15/airflow/providers/microsoft/azure/fs/adls.py#L29 So I think we should add `adl:`, `abfs:` and `abfss:`.
- Upon review, I'm actually really confused by the difference between the "data lake" hook and the "wasb" hook. 🤔 Azure is confusing. I am going to revert back to using the `WasbHook` default. I have a feeling that they both take in the same configuration but that the WasbHook is more commonly used.
- Seems like we may want to do a few updates to how the Cosmos dbt docs plugin works with Azure blob storage, too, but that is a separate PR.

Thanks for the review!

dwreeves:
@pankajkoti Thank you for the very speedy review. As per my comment above I looked into things a little bit more and incorporated your bit of feedback, plus learned a few more things (looks like `abfs`,  `abfss` and `adl` can all be supported?)

cosmos/constants.py
```diff
@@ -31,14 +29,33 @@
 PARTIALLY_SUPPORTED_AIRFLOW_VERSIONS = [Version("2.9.0"), Version("2.9.1")]
 
 
-S3_FILE_SCHEME = "s3"
-GS_FILE_SCHEME = "gs"
 ABFS_FILE_SCHEME = "abfs"
```
pankajkoti, line 32:
```suggestion
```

pankajkoti:
@dwreeves could we merge the PR? Or is there something more you'd like to add here?

dwreeves:
@pankajkoti It's fine to merge!

Issue #1072, opened 2024-06-27
[Feature] Allow storing dbt ls cache into Object Store


tatiana:
### Description

Following up on #1014 , we'd like to allow users to store the dbt ls cache into a remote object store, as an alternative to store it in Airflow Variables. This task relates to #927.

### Use case/motivation

Even though the dbt ls is being compressed without significantly affecting the Airflow Metadata database, some users prefer not to have cache information mixed up with their Airflow variables.

### Related issues

* #1014 
* #927 

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

pankajkoti:
Aiming to create a PR for this tonight. We discussed the approach and clarified doubts in our call today

PR #1147, opened 2024-08-09
Add support to store and fetch dbt ls cache in remote stores

pankajkoti:
## Summary
This PR introduces the functionality to store and retrieve the `dbt ls` output cache in remote storage systems. This enhancement improves the efficiency and scalability of cache management for Cosmos dbt projects that use the `dbt ls` cache option (enabled by default) introduced in PR #1014 

## Key Changes
1. **`dbt ls` Cache Storage in Remote Stores**:
Added support to store the dbt ls cache as a JSON file in remote storage paths configured in the Airflow settings under the `cosmos` section.
The cache is saved in the specified remote storage path & it includes the `cosmos_cache__` prefix.
2. **Cache Retrieval from Remote Stores**:
Implemented logic to check the existence of the cache in the remote storage path before falling back to the Variable cache.
If the `remote_cache_dir` is specified and it exists in the remote store, it is read and used; We try creating the specified path if it does not exist.
3. **Backward Compatibility**:
Maintained backward compatibility by allowing users to continue using local cache storage through Airflow Variables if a `remote_cache_dir` is not specified.

## Impact
1. **Scalability**: Enables the use of remote, scalable storage systems for dbt cache management.
2. **Performance**: Reduces the load on Airflow's metadata database by offloading cache storage to external systems.
3. **Flexibility**: Provides users with the option to choose between local (Airflow metadata using Variables) and remote cache storage based on their infrastructure needs.

## Configuration
To leverage this feature, users need to set the `remote_cache_dir` in their Airflow settings in the `cosmos` section. This path should point to a compatible remote storage location. You can also specify the `remote_cache_dir_conn_id` which is your Airflow connection that can connect to your remote store. If it's not specified, Cosmos will aim to identify the scheme for the specified path and use the default Airflow connection ID as per the scheme.

## Testing
1. Tested with various remote storage backends (AWS S3 and GCP GS) to ensure compatibility and reliability
2. Verified that cache retrieval falls back to Variable based caching approach if the `remote_cache_dir` is not configured.

## Documentation
Updated the documentation to include instructions on configuring `remote_cache_dir`.

## Limitations
1. Users must be on Airflow version 2.8 or higher because the underlying Airflow Object Store feature we utilise to access remote stores was introduced in this version. If users attempt to specify a `remote_cache_dir` on an older Airflow version, they will encounter an error indicating the version requirement.
2. Users would observe a slight delay for the tasks being in queued state (approx 1-2 seconds queued duration vs the 0-1 seconds previously in the Variable approach) due to remote storage calls to retrieve the cache from.

closes: #1072 

## Breaking Change?
No

## Checklist


cosmos/dbt/graph.py
```diff
@@ -296,7 +312,11 @@ def get_dbt_ls_cache(self) -> dict[str, str]:
         """
         cache_dict: dict[str, str] = {}
         try:
-            cache_dict = Variable.get(self.dbt_ls_cache_key, deserialize_json=True)
+            cache_dict = (
+                self._get_dbt_ls_remote_cache()
+                if settings.remote_cache_path
```
pankajkoti, line 317:
```suggestion
                if self.project.remote_cache_path
```

.github/workflows/test.yml
```diff
@@ -170,6 +170,81 @@ jobs:
           name: coverage-integration-test-${{ matrix.python-version }}-${{ matrix.airflow-version }}
           path: .coverage
 
+  Run-Integration-Tests-Remote-Cache-Path:
+    needs: Authorize
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ "3.8", "3.9", "3.10", "3.11" ]
+        airflow-version: [ "2.8", "2.9" ]
```
pankajkoti, line 179:
```suggestion
        airflow-version: [ "2.8", "2.9.3" ]
```

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,remote-db-ls-cache]
```
```suggestion
    branches: [main]
```

apply the suggestion to remove current branch before merging

pankajastro (approved):
looks great! 

just some minor comments

What are your thoughts on adding a mechanism for cleaning up the remote cache, similar to what we have for native?

https://github.com/astronomer/astronomer-cosmos/blob/4dd9549a76092fb39aca2deede1416b58d256926/cosmos/cache.py#L273

.github/workflows/test.yml
```diff
@@ -170,6 +170,81 @@ jobs:
           name: coverage-integration-test-${{ matrix.python-version }}-${{ matrix.airflow-version }}
           path: .coverage
 
+  Run-Integration-Tests-Remote-Cache-Path:
```
tatiana, line 173: @pankajkoti why did we create a dedicated job for these tests?

cosmos/__init__.py
```diff
@@ -5,7 +5,7 @@
 
 Contains dags, task groups, and operators.
 """
-__version__ = "1.5.1"
+__version__ = "1.6.0a6"
```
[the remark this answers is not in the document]
tatiana: @pankajastro we changed this to collect feedback on the feature from a customer!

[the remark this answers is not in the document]
pankajastro: does it mean we merge this change to main too?

cosmos/config.py
```diff
@@ -146,6 +147,7 @@ class ProjectConfig:
     seeds_path: Path | None = None
     snapshots_path: Path | None = None
     project_name: str
+    remote_cache_path: Path | None = None
```
tatiana, line 150: Previously, we attempted to have a single cache directory, at an Airflow deployment level:
Screenshot 2024-08-14 at 09 55 51

https://astronomer.github.io/astronomer-cosmos/configuration/partial-parsing.html#caching

Does it make sense to do the same here?

Perhaps we can also have the cache subpath in the `ProjectConfig` If we want to give users the granularity to define a subpath per project - but in that case, I believe it should be `cache_subpath` - , so it can be used both remotely and locally as well.

docs/configuration/cosmos-conf.rst
```diff
@@ -94,6 +94,27 @@ This page lists all available Airflow configurations that affect ``astronomer-co
     - Default: ``profile``
     - Environment Variable: ``AIRFLOW__COSMOS__PROFILE_CACHE_DIR_NAME``
 
+.. _remote_cache_path:
+
+`remote_cache_path`_:
+    The remote path to store the DBT cache. Starting with Cosmos 1.6.0, you can store the dbt ls output as cache in a
```
tatiana, line 100:
```suggestion
    The remote path to store the `dbt ls` cache. Starting with Cosmos 1.6.0, you can store the dbt ls output as cache in a
```

tatiana (commented):
Great work, @pankajkoti ! I left some feedback - we can discuss and iterate over it tomorrow in a call to speed up getting this merged

.github/workflows/test.yml
pankajkoti: The current integration tests have turned off dbt ls cache. I was unsure on why they are turned off and additionally kept a separate job for remote cache (in case we would have liked the existing job to follow the local/default/variable based cache)

.github/workflows/test.yml
```diff
@@ -170,6 +170,81 @@ jobs:
...
+    needs: Authorize
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ "3.8", "3.9", "3.10", "3.11" ]
+        airflow-version: [ "2.8" ]
```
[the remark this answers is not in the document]
pankajkoti: I had 2.8 and 2.9 previously, but we ignore runing example DAGs for 2.9, hence removed 2.9

cosmos/dbt/graph.py
```diff
@@ -281,7 +281,23 @@ def save_dbt_ls_cache(self, dbt_ls_output: str) -> None:
             "last_modified": datetime.datetime.now(datetime.timezone.utc).isoformat(),
             **self.airflow_metadata,
         }
-        Variable.set(self.dbt_ls_cache_key, cache_dict, serialize_json=True)
+        if self.project.remote_cache_path:
+            remote_cache_key_path = self.project.remote_cache_path / self.dbt_ls_cache_key / "dbt_ls_cache.json"
```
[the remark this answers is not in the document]
pankajkoti: End users would not have to specify a value per DAG. There is a single remote path that needs to be configured. I think yes, discussing this over a call can bridge up the gap quicker.

cosmos/config.py
pankajkoti: We still would have a single cache directory for the Airflow deployment. `ProjectConfig` is just a safe place I found it to retrieve and configure the `remote_cache_path` (we can find for a better place). 

Previously I was configuring it in the settings.py file itself https://github.com/astronomer/astronomer-cosmos/pull/1147/commits/ae0f45586912f1ccbfa7677c18720bdc781ea139#diff-2317d3342b8470278e95b3fcc555507acb062a595c518396ad5b94f1d3021e3eR41, but having it at the top level gave errors by the fsspec module complaining about it not being fork-safe.

We however are not exposing this param in the ProjectConfig but just using it for Cosmos internals. Can discuss more over our call.

docs/configuration/cosmos-conf.rst
pankajkoti: Since we would like to have a single cache path and not for `dbt ls` alone had mentioned only `dbt`. And added the second part because at the moment we would only support `dbt ls` in the cache.

tatiana (commented):
Hi @pankajkoti , summarising what we agreed:

1. Rename the configuration, so we minimize the amount of configuration and are consistent with how we are setting up variables when hosting docs:
   * drop `remote_cache_path` and reuse pre-existing `cache_dir`
   * rename `remote_cache_conn_id` to `cache_conn_id`
2. Avoid exposing the method in `ProjectConfig` - if we could keep cache aux functions in either `cache.py` or `graph.py`, it would be ideal. We should refactor this in the future, as we discussed. If for some reason this refactor lead to the error "this class is not fork safe", it's okay to expose it in the `ProjectConfig` as a property prefixed with `_`. If we manage to get this to work in `cache.py` or `graph.py`, we can consider decorating it with the lru cache method to avoid re-computing it twice.
3. Make sure `delete_unused_dbt_ls_cache` works for the remote cache
4. Do not have a dedicated CI job for this feature, but try to leverage using something like the following in the relevant tests, setting the necessary envvars: ```@patch.dict(os.environ, {"AIRFLOW__COSMOS__ENABLE_CACHE": "False"}, clear=True)```.  We could also consider running one of the integration DAGs with this enabled, but without the need to create a new CI job.

It's looking great, and it will be very exciting to have this feature out.

.github/workflows/test.yml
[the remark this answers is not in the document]
pankajkoti: will add the integration tests as part of a separate PR to follow-up on. Removed this for now.

.github/workflows/test.yml
pankajkoti: Removed the integration tests as part of this PR and will create a follow-up PR to have limited tests running as discussed.

cosmos/config.py
pankajkoti: yes, we have a single remote_cache_dir as discussed. Removed from ProjectConfig and added a method to the `cache.py` module as suggested, thanks!

pankajkoti:
Thanks @tatiana for the review. I have managed to address the review comments as per our recent discussion

> 1. Rename the configuration, so we minimize the amount of configuration and are consistent with how we are setting up variables when hosting docs:
>    
>    * drop `remote_cache_path` and reuse pre-existing `cache_dir`
>    * rename `remote_cache_conn_id` to `cache_conn_id`

At this time, we couldn't use the existing `cache_dir` due to other caching implementation implications that are challenging to address. I've renamed it to `remote_cache_dir` and noted in the documentation that this is an experimental feature, which will be merged into `cache_dir` in future releases.

> 2. Avoid exposing the method in `ProjectConfig` - if we could keep cache aux functions in either `cache.py` or `graph.py`, it would be ideal. We should refactor this in the future, as we discussed. If for some reason this refactor lead to the error "this class is not fork safe", it's okay to expose it in the `ProjectConfig` as a property prefixed with `_`. If we manage to get this to work in `cache.py` or `graph.py`, we can consider decorating it with the lru cache method to avoid re-computing it twice.

Yes, as per the pairing session, we moved it to `cache.py`. As of now, we're avoiding initiating the configuration in the module but deferring it while retrieving/updating the cache due to the `fsspec` constraints.

> 3. Make sure `delete_unused_dbt_ls_cache` works for the remote cache

I have created a new function `delete_unused_dbt_ls_remote_cache_files` and included this in the example DAG. I ensured that it is able to identify the stale cache files and delete those (verified that the file actually got deleted in the remote store).
I haven't yet added tests for these as we're missing to add remote storage config in the CI yet, I have marked it as nocover for now, but will add tests for it in the follow-up PR when we enable CI integration tests for remote storage. Hope that's okay.
Snapshot of deletion logs
Screenshot 2024-08-16 at 2 04 16 AM


> 4. Do not have a dedicated CI job for this feature, but try to leverage using something like the following in the relevant tests, setting the necessary envvars: `@patch.dict(os.environ, {"AIRFLOW__COSMOS__ENABLE_CACHE": "False"}, clear=True)`.  We could also consider running one of the integration DAGs with this enabled, but without the need to create a new CI job.

Yes, I have removed the separate job for now. And as discussed, I will follow-up in a subsequent PR to address this point.

Requesting a re-review please.

cc: @pankajastro

cosmos/__init__.py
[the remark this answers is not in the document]
tatiana: It's ok for main to have this as well :)

tatiana (approved):

PR #1080, opened 2024-07-05
Release 1.6.0

tatiana:
New Features

* Add support for loading manifest from cloud stores using Airflow Object Storage by @pankajkoti in #1109
* Cache ``package-lock.yml`` file by @pankajastro in #1086
* Support persisting the ``LoadMode.VIRTUALENV`` directory by @tatiana in #1079
* Add support to store and fetch ``dbt ls`` cache in remote stores by @pankajkoti in #1147
* Add default source nodes rendering by @arojasb3 in #1107
* Add Teradata ``ProfileMapping`` by @sc250072 in #1077

Enhancements

* Add ``DatabricksOauthProfileMapping`` profile by @CorsettiS in #1091
* Use ``dbt ls`` as the default parser when ``profile_config`` is provided by @pankajastro in #1101
* Add task owner to dbt operators by @wornjs in #1082
* Extend Cosmos custom selector to support + when using paths and tags by @mvictoria in #1150
* Simplify logging by @dwreeves in #1108

Bug fixes

* Fix Teradata ``ProfileMapping`` target invalid issue by @sc250072 in #1088
* Fix empty tag in case of custom parser by @pankajastro in #1100
* Fix ``dbt deps`` of ``LoadMode.DBT_LS`` should use ``ProjectConfig.dbt_vars`` by @tatiana in #1114
* Fix import handling by lazy loading hooks introduced in PR #1109 by @dwreeves in #1132
* Fix Airflow 2.10 regression and add Airflow 2.10 in test matrix by @pankajastro in #1162

Docs

* Fix typo in azure-container-instance docs by @pankajastro in #1106
* Use Airflow trademark as it has been registered by @pankajastro in #1105

Others

* Run some example DAGs in Kubernetes execution mode in CI by @pankajastro in #1127
* Install requirements.txt by default during dev env spin up by @@CorsettiS in #1099
* Remove ``DbtGraph.current_version`` dead code by @tatiana in #1111
* Disable test for Airflow-2.5 and Python-3.11 combination in CI by @pankajastro in #1124
* Pre-commit hook updates in #1074, #1113, #1125, #1144, #1154,  #1167

PR #1148, opened 2024-08-09
Add support for loading dbt project from cloud store using Airflow Object Store

CorsettiS:
## Description

This PR is based on #1109 and intends to allow users to load entire dbt projects that are stored at a cloud provider. That way, it is totally possible to decouple dbt and airflow with cosmos. 

As of now, the company I currently work at does not have dbt & airflow in the same repo, but by allowing cosmos to fetch entire dbt projects from a cloud provider we can work around that easily. dbt projects are not usually very heavy, so performance should not be drastically impacted (it was not in the local tests I have done).

Would be good if this could be included in the release 1.6.0 #1080

## Breaking Change?
No

## Checklist


cosmos/config.py
```diff
@@ -224,8 +245,8 @@ def validate_project(self) -> None:
         if self.dbt_project_path:
             project_yml_path = self.dbt_project_path / "dbt_project.yml"
             mandatory_paths = {
-                "dbt_project.yml": Path(project_yml_path) if project_yml_path else None,
-                "models directory ": Path(self.models_path) if self.models_path else None,
```
CorsettiS, line 228: honestly speaking I could not understand this even after reading the comment. both the self.dbt_project_path and self.models_path are already Path as specified in the __init__

tests/test_config.py
```diff
@@ -36,12 +36,8 @@ def test_init_with_manifest_path_and_project_path_succeeds():
     project_name in this case should be based on dbt_project_path
     """
     project_config = ProjectConfig(dbt_project_path="/tmp/some-path", manifest_path="target/manifest.json")
-    if AIRFLOW_IO_AVAILABLE:
```
CorsettiS, line 39: This test is actually wrong since AIRFLOW_IO_AVAILABLE just means we are using airflow >=2.8.0, and not that we are using object storage.

tatiana: @corsettigyg wasn't object storage introduced in Airflow 2.8?
https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/objectstorage.html

tatiana (commented):
@CorsettiS This is a very promising change, thank you!

Some questions:
1. How big are the dbt projects you've tested this feature with?
2. I'm slightly concerned about performance, since we'd be running this during DAG processing/parsing time very regularly usually in the scheduler - depending on the customer deployment. This may not scale for bigger dbt projects or for situations where users have many Cosmos `DbtDag`s. Did you run any performance tests? Could you share some numbers / how you validated this feature?
3. Following (2), by syncing the whole dbt project from object store, we would potentially not benefit from the caching introduced in Cosmos 1.5 in PR #1014. Do you have any proposals of how we could avoid performing this action every time?
4. Would using `LoadMethod.MANIFEST` + `ExecutionMode.KUBERNETES` be an option for your company? This should allow the decoupling you are aiming for. Or is there any particular reason why you'd need `LoadMethod.DBT_LS` and `LoadMethod.LOCAL`?

It seems unlikely we'll be able to release this in 1.6, given the number of things we're still trying to address before the release - but let's see!

tests/test_config.py
CorsettiS: @tatiana indeed, although using airflow 2.8 does not exactly mean we are using object storage, which was the assumption of this test, so I just fixed it.

CorsettiS:
I have decided to close the PR for now because I found a severe flaw when I was extending the functionality of this PR to the task execution, which is that each singular task will need to list and fetch all artifacts from the repo in s3, which is going to balloon costs in an unexpected way. I will work with it on the side and once I have a good alternative for the problem I will re-create the PR

tatiana:
Thank you very much, @CorsettiS , for the additional information and context! We're looking forward to seeing your follow up PRs

PR #1153, opened 2024-08-12
Add GCP_CLOUD_RUN_JOB execution mode

### LINKED ISSUE #1149: [Feature] New GCP Cloud Run Job Execution Mode
### Overview

Currently, using dbt with cosmos on GCP (Google Cloud Composer) is limited to `virtualenv` Execution Mode. The performance of this solution is tightly coupled with GCC's instance parameters and is, in general, rather slow.

### Proposed Solution
I would like cosmos to have ability to trigger [GCP Cloud Run Jobs](https://cloud.google.com/run/docs/overview/what-is-cloud-run) and run dbt models in there. 
Cloud Run Job is a lightweight, serverless container solution that allows more flexibility for each dbt model - we could adjust CR's instance parameters for either simple or heavy-load dbt models instead of relying solely on single GCC instance. Cloud Run Jobs also work well with Airflow's concurrency parameter, allowing concurrent task execution and, thus, scalability.

### Implementation Details
There is already Airflow CloudRunExecuteJobOperator in place that triggers Cloud Run Job with optional `container_overrides` parameter that overrides default code when a job is run. I intend to put a dbt command into `container_overrides` to run the model.

```python
    def build_and_run_cmd(self, context: Context, cmd_flags: list[str] | None = None) -> Any:
        self.build_command(context, cmd_flags)
        self.log.info(f"Running command: {self.command}")
        result = CloudRunExecuteJobOperator.execute(self, context)
        logger.info(result)

    def build_command(self, context: Context, cmd_flags: list[str] | None = None) -> None:
        # For the first round, we're going to assume that the command is dbt
        # This means that we don't have openlineage support, but we will create a ticket
        # to add that in the future
        self.dbt_executable_path = "dbt"
        dbt_cmd, env_vars = self.build_cmd(context=context, cmd_flags=cmd_flags)
        self.command: list[str] = dbt_cmd
        # Override Cloud Run Job default arguments with dbt command
        self.overrides = {
            'container_overrides': [
                {
                    "args": self.command,
                }
            ],
        }
```

### Use case/motivation

The solution would decouple Airflow instance and dbt models. It adds more flexibility and less risk of entire GCC failure when something goes wrong or the GCC resources are exceeded.

### Related issues

#958
Unfortunately I noticed that GCC worker's space disk is getting filled as well while running dbt models using `virtualenv` Execution Mode. 
As GCC is a managed instance with hidden resources, I am not able to investigate it further.

- [X] Yes, I am willing to submit a PR!

ags-de:
## Description

Added new `GCP_CLOUD_RUN_JOB` execution mode that triggers Google Cloud Platform's Cloud Run Job instance with dbt model in it.

It extends Airflow's `CloudRunExecuteJobOperator` and overrides Cloud Run Job's container with dbt command generated by cosmos.

Note: `CloudRunExecuteJobOperator` has `container_overrides` parameter implemented in `apache-airflow-providers-google==10.13.0` which is supported by `airflow >=2.6.0`.

## Related Issue(s)

resolves #1149 

## Breaking Change?

None, the new execution mode is optional.

## Checklist


pankajastro:
Hi @ags-de, Thank you so much for your contribution. Please let us know if you need any support.

pankajkoti:
hi @ags-de could you please rebase the PR. We just merged https://github.com/astronomer/astronomer-cosmos/pull/1182 that should fix the breaking CI jobs.

cosmos/__init__.py
```diff
@@ -40,6 +40,36 @@
 
 logger = get_logger(__name__)
 
+try:
+    from cosmos.operators.gcp_cloud_run_job import (
+        DbtLSGcpCloudRunJobOperator,
```
tatiana, line 45: I just noticed we are not importing the Build operator from other similar modules, apparently, except for `DbtBuildLocalOperator`. We can start correcting this by adding here:

```suggestion
        DbtBuildGcpCloudRunJobOperator,
        DbtLSGcpCloudRunJobOperator,
```

cosmos/__init__.py
```diff
@@ -142,6 +172,12 @@
     "DbtSnapshotLocalOperator",
     "DbtDag",
     "DbtTaskGroup",
+    "DbtLSGcpCloudRunJobOperator",
```
tatiana, line 175:
```suggestion
    "DbtBuildGcpCloudRunJobOperator",
    "DbtLSGcpCloudRunJobOperator",
```

docs/getting_started/execution-modes.rst
```diff
@@ -209,6 +214,27 @@ Each task will create a new container on Azure, giving full isolation. This, how
         },
     )
 
+GCP Cloud Run Job
```
tatiana, line 217: Two recommendations on the docs:
1. Add a link to Cloud Run documentation
2. Mention that for this execution mode to work, users should, previously:
* Create container
* Create Cloud Run job
3. (optional) If we could have something similar to this, but for GCP Cloud Run, it would be amazing:
https://github.com/astronomer/astronomer-cosmos/blob/main/docs/getting_started/azure-container-instance.rst

tatiana (commented):
Hi @ags-de, thank you for contributing this new execution mode to Cosmos!
It's great to see new execution modes that support isolation between the dbt project and the Airflow worker node. I've added some feedback inline.

Please, could you confirm if/how you validated this feature?

We'll likely be able to release this feature in Cosmos 1.7, planned to be released in a week time.

ags-de:
Hello @tatiana,

Thank you for your insights! I am working on code updates as your suggestions.

Regarding feature validation, I am in fact working on "getting started" guide for Cloud Run Job, based on [Azure Container Instance Execution Mode page](https://astronomer.github.io/astronomer-cosmos/getting_started/azure-container-instance.html). It will include all the information needed to use and validate the feature.

PR description updated.

cosmos/converter.py
```diff
@@ -120,6 +120,7 @@ def validate_initial_user_config(
         ExecutionMode.KUBERNETES,
         ExecutionMode.AWS_EKS,
         ExecutionMode.DOCKER,
+        ExecutionMode.GCP_CLOUD_RUN_JOB,
```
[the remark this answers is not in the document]
ags-de: Yes Tatiana, that makes a lot more sense. Updated as per your suggestion.

tatiana:
Thank you very much, @ags-de . Fantastic work, we're very grateful for your contribution! 

I created a PR to try to make some things consistent that were missing in Cosmos: #1215

And follow-ups for us to introduce breaking changes in Cosmos 2.0 when the time comes: #1214 (you already adopted `-`, so let's stick to this in this PR, #1212 #1213

docs/getting_started/execution-modes.rst
ags-de: 1 and 2 updated, working on 3.

tatiana:
@ags-de thank you very much for working so quickly on the changes requested. When you get a chance, please, rebase on the main branch.

docs/getting_started/execution-modes.rst
ags-de: Hello @tatiana,

I created step-by-step guide to setup GCP infrastructure and Airflow to validate the feature: docs/getting_started/gcp-cloud-run-job.rst

All of the files needed for tests are available in cosmos-example repo, [PR #8](https://github.com/astronomer/cosmos-example/pull/8).

By the way, I believe there is still another [PR #6](https://github.com/astronomer/cosmos-example/pull/6) pending on that repo.

tatiana: Thank you very much, @ags-de , this looks great!
Thanks for pointing out the lack of reviews in the cosmos-example repo. I created #1222 so we can address this in the mid-long term

ags-de:
Hello @tatiana,

The branch is rebased and I also fixed a few typos in docs. It should be ready to go.

pyproject.toml
```diff
@@ -69,7 +69,7 @@ openlineage = ["openlineage-integration-common!=1.15.0", "openlineage-airflow"]
 amazon = [
     "apache-airflow-providers-amazon[s3fs]>=3.0.0",
 ]
-google = ["apache-airflow-providers-google"]
+google = ["apache-airflow-providers-google>=10.11.0"]
```
tatiana, line 72: We should not add this lower bound since it would be a breaking change.

Cosmos still supports Airflow 2.4.0; in that case, users who currently install `astronomer-cosmos[google]` rely on `apache-airflow-providers-google<10.11.0`. They may only be interested in leveraging the BQ profile mapping and not Cloud Run, and they should still be allowed to install whichever version of the Google provider they want to.

tatiana, line 72:
```suggestion
google = ["apache-airflow-providers-google"]
```

tatiana (approved):
Hi @ags-de amazing work!
I left one last comment. I'll wait for your feedback/change before we merge into the main branch.

pyproject.toml
ags-de: Of course, change commited.

PR #1162, opened 2024-08-16
Fix Airflow 2.10 regression and Add Airflow 2.10 in test matrix

pankajastro:
We have a few classes that lack an `__init__` method, and their parent class is BaseOperator. These classes are failing to load due to changes introduced in the PR: https://github.com/apache/airflow/issues/41085.

I'm adding `__init__` at missing places in this PR and also adding Airflow 2.10 in our test matrix

CI: https://github.com/astronomer/astronomer-cosmos/actions/runs/10469548488

closes: https://github.com/astronomer/astronomer-cosmos/issues/1166
closes: https://github.com/astronomer/astronomer-cosmos/issues/1165

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, test-af-2.10]
+    branches: [main]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, test-af-2.10]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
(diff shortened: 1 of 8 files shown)


dwreeves (changes requested):
Can you check #1163 before merging? I think that #1163 has a fix that is a little simpler and more robust. (Note that my PR points to your branch, not main, so merging my PR merges to `test-af-2.10`. I based my PR off your PR especially with the changes to the CI + hatch env.) Let me know what you think.

(It also fixes an issue that's been annoying me for a while, which is that the Mixins needed lots of `type: ignore`s. 😆)

dwreeves:
@pankajastro I force pushed away my change, sorry for wasting time! ☹️ Let's just prioritize getting this out ASAP.

dwreeves (approved):

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, test-af-2.10]
```
```suggestion
    branches: [main]
```
Remove branch before merging

pankajkoti (approved):

pankajkoti:
Can we create a ticket for this in our repo/board for tracking the work done here and link it in the description?

cosmos/operators/aws_eks.py
```diff
@@ -79,12 +79,18 @@ class DbtBuildAwsEksOperator(DbtAwsEksBaseOperator, DbtBuildKubernetesOperator):
...
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+
 
 class DbtLSAwsEksOperator(DbtAwsEksBaseOperator, DbtLSKubernetesOperator):
     """
     Executes a dbt core ls command.
     """
 
+    def __init__(self, **kwargs: Any) -> None:
```
pankajkoti, line 91: should we also declare *args and pass it to super init

Issue #1161, opened 2024-08-16
[Bug] Airflow 2.10 compatibility


ms32035:
### Astronomer Cosmos Version

Other Astronomer Cosmos version (please specify below)

### If "Other Astronomer Cosmos version" selected, which one?

1.5.1

### dbt-core version

1.8.5

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

KUBERNETES

### InvocationMode

None

### airflow version

2.10.0

### Operating System

Debian GNU/Linux 12 (bookworm)

### Deployment

Official Apache Airflow Helm Chart

### What happened?

Dag parsing fails with error

`Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.11/site-packages/cosmos/operators/base.py", line 361, in __init__
    self.full_refresh = full_refresh
    ^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.11/site-packages/airflow/models/baseoperator.py", line 1198, in __setattr__
    if key in self.__init_kwargs:
              ^^^^^^^^^^^^^^^^^^
AttributeError: 'DbtRunLocalOperator' object has no attribute '_BaseOperator__init_kwargs'`

### How to reproduce

Install Airflow 2.10

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

pankajastro:
Feel free to submit a PR whenever you're ready. I'm happy to review and merge it.

pankajastro:
It looks like a side effect of the changes in PR https://github.com/apache/airflow/issues/41085. A quick workaround is to add the missing `__init__` method, which I have added in https://github.com/astronomer/astronomer-cosmos/pull/1162. Let's see how CI feel.

dwreeves:
@pankajastro I am wondering if that is the best fix, or if that even works at all (if it does work, it will only work due to opaque MRO trickery).

I think the best solution is just that `super().__init__(**kwargs)` needs to be at the top of the `AbstractDbtBaseOperator` init call instead of at the bottom.

The error occurs inside the `__setattr__` for the `BaseOperator`. If you look at the `BaseOperator` code in 3.10, the issue is that the first thing the `__init__()` does is set the `__init_kwargs` and every other `__setattr__` beyond that assumes the existence of `self.__init_kwargs`.

(I'm personally not convinced this is appropriate and that this is a small bug in Airflow 2.10. 🤔)

[...]

PR #1163, opened 2024-08-19
Proposal for different fix to 2.10 regression

dwreeves:
## Description

Fixes Airflow 2.10 regression / compat

## Related Issue(s)

This is a fix for #1161, and is designed to be a different proposal to the changes in #1162.

Ultimately the issue that popped up in 2.10 relates to how the MRO was impacted by the Mixin class.

The simplest, most robust fix to make the MRO work properly is to make each mixin a subclass of `BaseOperator`. This doesn't require the unintuitive hack of needing an `__init__()` call to shuffle the MRO around, and it makes things more easily maintainable going forward, I think.

A note: This PR is merging to `test-af-2.10`, not `main`.

Tested for 2.10.

I also removed an unnecessary `__init__` for the `DbtTestMixin` while I was at it.

## Checklist

### DIFF
--- a/cosmos/operators/aws_eks.py
+++ b/cosmos/operators/aws_eks.py
@@ -79,17 +79,11 @@ class DbtBuildAwsEksOperator(DbtAwsEksBaseOperator, DbtBuildKubernetesOperator):
         DbtAwsEksBaseOperator.template_fields + DbtBuildKubernetesOperator.template_fields  # type: ignore[operator]
     )
 
-    def __init__(self, **kwargs: Any) -> None:
-        super().__init__(**kwargs)
-
 
 class DbtLSAwsEksOperator(DbtAwsEksBaseOperator, DbtLSKubernetesOperator):
     """
     Executes a dbt core ls command.
     """
-
-    def __init__(self, **kwargs: Any) -> None:
-        super().__init__(**kwargs)
 
 
 class DbtSeedAwsEksOperator(DbtAwsEksBaseOperator, DbtSeedKubernetesOperator):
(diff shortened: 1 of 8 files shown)


dwreeves:
One more thing, after all of this gets resolved, we should either do this for each minor version from 1.3+, or alternatively put an upper bound on each previously released version as patch releases. I believe based on the Slack that there are a lot of users on 1.3 still.

pankajastro (approved):
Looks good. thanks @dwreeves for looking into.

pankajastro:
cc: @pankajkoti we should include this in 1.6.0

pankajastro:
> we should either do this for each minor version from 1.3+, or alternatively put an upper bound on each previously released version as patch releases. 

I'm not sure about this at the moment, but I'm open to hearing what others think. The user who has reported this issue using Cosmos 1.5.1.

> I believe based on the Slack that there are a lot of users on 1.3 still.

Which Slack are you referring to?

pankajastro:
Also, the CI is not running on this PR, possibly because the base branch is not main.

pankajastro:
I'm merging this to branch `test-af-2.10` to validate the CI check

pankajastro:
Look like CI is not happy

dwreeves:
@pankajastro That's weird. 😕 I did run tests locally for `2.10` and it worked, but I see now that 2.4 tests break. 

Let's rollback on the other PR and we can merge this if/when it is ready. There is a higher criticality to just releasing a fix than there is to fixing it in a "good way."

Sorry for making this so complicated.

dwreeves:
RE: upgrading other versions-- Fair enough, we won't backport then.

My one concern was that Airflow issues like this can be annoying because if a user upgrades their Airflow, _downgrading_ Airflow to rollback a change can be very annoying because of the Postgres db requiring a downgrade, so if they are stuck on an old Cosmos version for some reason and they upgrade Airflow, now they have 2 problems.

> Which Slack are you referring to?

Official Airflow slack, dbt channel.

Issue #1062, opened 2024-06-24
[Bug] Test task doesn't accept `full_refresh` kwarg


CamposContentful:
### Astronomer Cosmos Version

main (development)

### dbt-core version

1.8.2

### Versions of dbt adapters

dbt-redshift = "^1.8.1"

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

AWS_EKS

### InvocationMode

None

### airflow version

MWAA 2.7.2

### Operating System

Fedora

### Deployment

Amazon (AWS) MWAA

### What happened?

Hello everyone,
when passing the `full_refresh` parameter in `operator_args` for a `DbtTaskGroup` task. 
I get the following error:

```
airflow.exceptions.AirflowException: Invalid arguments were passed to DbtTestAwsEksOperator (task_id: test). Invalid arguments were:
**kwargs: {'full_refresh': True}
```

The task is defined as:

```
    example_calculation = DbtTaskGroup(
        group_id="example_calculation_models",
        project_config=ProjectConfig(
            manifest_path=f"{os.path.dirname(os.path.realpath(__file__))}/manifest.json",
            project_name="example_calculation",
        ),
        profile_config=ProfileConfig(
            profile_name="default_profile", target_name="airflow", profiles_yml_filepath="./dbt_profiles/profiles.yml"
        ),
        render_config=RenderConfig(
            select=["tag:example_calculation"],
            load_method=LoadMode.DBT_MANIFEST,
            node_converters={
                DbtResourceType("source"): convert_source,  # known dbt node type to Cosmos (part of DbtResourceType)
            },
            test_behavior=TestBehavior.AFTER_ALL,
        ),
        execution_config=ExecutionConfig(execution_mode=ExecutionMode.AWS_EKS, dbt_project_path="/usr/app/dbt/"),
        operator_args={
            "in_cluster": False,
            "image_pull_policy": "Always",
            "get_logs": True,
            "is_delete_operator_pod": True,
            "on_finish_action": "delete_pod",
            "full_refresh": True,
        },
    )
```
My understanding is that the `full_refresh` argument should be ignored by `dbt test`.
I couldn't find any similar issues in the repo, do you an idea if I'm just setting the argument wrong or is something else.

Thanks!

### How to reproduce

For `execution_mode=ExecutionMode.AWS_EKS` set up a dbt project with test and `full_refresh` option in `operator_args`.

### Are you willing to submit PR?

[...]

PR #590, opened 2023-10-10
fix: full_refresh unsused in test operators

EgorSemenov:
## Description

https://apache-airflow.slack.com/archives/C059CC42E9W/p1696857998959909

## Related Issue(s)

[issue 589](https://github.com/astronomer/astronomer-cosmos/issues/589)

## Breaking Change?

<details><summary>Details</summary>
<p>

airflow.exceptions.AirflowException: Invalid arguments were passed to DbtTestLocalOperator (task_id: test). Invalid arguments were:
**kwargs: {'full_refresh': True}

</p>
</details>

PR #1175, opened 2024-08-23
Fix invalid argument (`full_refresh`) passed to DbtTestAwsEksOperator (and others)

johnhoran:
https://github.com/astronomer/astronomer-cosmos/pull/590/ added a fix to consume the kwargs `full_refresh_ignore` if it wasn't consumed by a higher class as it was preventing the use of test in the DbtTaskGroup if `full_refresh_ignore` was set.  The previous patch fixed this by consuming the variable for the `DbtLocalBaseOperator`, leaving a bug in kubernetes and docker operator.  Since `AbstractDbtBaseOperator` has been added as a base of `DbtDockerBaseOperator`, `DbtKubernetesBaseOperator` and `DbtLocalBaseOperator`, moving the code there will fix all three.

Fixes https://github.com/astronomer/astronomer-cosmos/issues/1062

### DIFF
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -140,7 +140,6 @@ class AbstractDbtBaseOperator(BaseOperator, metaclass=ABCMeta):
         self.cache_dir = cache_dir
         self.extra_context = extra_context or {}
         kwargs.pop("full_refresh", None)  # usage of this param should be implemented in child classes
-
         super().__init__(**kwargs)
 
     def get_env(self, context: Context) -> dict[str, str | bytes | os.PathLike[Any]]:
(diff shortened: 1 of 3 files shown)


johnhoran:
https://github.com/astronomer/astronomer-cosmos/issues/1062

tatiana (commented):
@johnhoran thank you very much for making this contribution!

Could you add a unittest that covers that scenario that stopped working - so we make sure we don't have any regressions in this area in future? Perhaps it could be on an EKS or K8s operator, since that's where the issue was first spotted.

Also, could you please update the PR description to include "Fixes #1062" or "Closes #1062", so when we merge the PR the related issue is closed automagically?

johnhoran:
@tatiana sure, I've added a test.  My strategy there was to have an exception thrown after the kubernetes manifest is created and then examine the manifest to check the correct arguments are being passed to DBT.

pankajkoti:
hi @johnhoran could you please rebase the PR. We just merged #1182 that should fix the breaking CI jobs.

johnhoran:
Thanks @pankajkoti

tatiana (approved):
This looks great, @johnhoran , thank you very much for reporting and fixing the issue! We'll release this as part of Cosmos 1.7

Issue #1180, opened 2024-08-27
Fix broken CI


tatiana:
Recently our main branch checks stopped passing:
https://github.com/astronomer/astronomer-cosmos/actions/runs/10528925346/job/29255782679

We managed to reproduce the issue locally by running:
```
hatch -v run tests.py3.9-2.7:type-check
```

The stacktrace is:
```
Traceback (most recent call last):
  File "/Users/tati/Library/Application Support/hatch/env/virtual/astronomer-cosmos/4VBJdS-x/tests.py3.9-2.7/lib/python3.9/site-packages/pip/_internal/cli/base_command.py", line 180, in exc_logging_wrapper
    status = run_func(*args)
  File "/Users/tati/Library/Application Support/hatch/env/virtual/astronomer-cosmos/4VBJdS-x/tests.py3.9-2.7/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 245, in wrapper
    return func(self, options, args)
  File "/Users/tati/Library/Application Support/hatch/env/virtual/astronomer-cosmos/4VBJdS-x/tests.py3.9-2.7/lib/python3.9/site-packages/pip/_internal/commands/install.py", line 377, in run
    requirement_set = resolver.resolve(
  File "/Users/tati/Library/Application Support/hatch/env/virtual/astronomer-cosmos/4VBJdS-x/tests.py3.9-2.7/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 95, in resolve
    result = self._result = resolver.resolve(
  File "/Users/tati/Library/Application Support/hatch/env/virtual/astronomer-cosmos/4VBJdS-x/tests.py3.9-2.7/lib/python3.9/site-packages/pip/_vendor/resolvelib/resolvers.py", line 546, in resolve
    state = resolution.resolve(requirements, max_rounds=max_rounds)
  File "/Users/tati/Library/Application Support/hatch/env/virtual/astronomer-cosmos/4VBJdS-x/tests.py3.9-2.7/lib/python3.9/site-packages/pip/_vendor/resolvelib/resolvers.py", line 457, in resolve
    raise ResolutionTooDeep(max_rounds)
pip._vendor.resolvelib.resolvers.ResolutionTooDeep: 200000
```

The issue is happening during the dependency resolution. Locally these are the logs that happened beforehand:
```
  Using cached apache_airflow_providers_amazon-8.24.0-py3-none-any.whl.metadata (10 kB)
[...]

Issue #1103, opened 2024-07-18
Release Cosmos 1.6.0


phanikumv:
* [x] https://github.com/astronomer/astronomer-cosmos/issues/1090
* [x] https://github.com/astronomer/astronomer-cosmos/issues/1087
* [x] https://github.com/astronomer/astronomer-cosmos/issues/1085
* [x] https://github.com/astronomer/astronomer-cosmos/issues/1076
* [x] https://github.com/astronomer/astronomer-cosmos/issues/1072
* [x] https://github.com/astronomer/astronomer-cosmos/issues/1042)
* [x] https://github.com/astronomer/astronomer-cosmos/issues/953
* [x] https://github.com/astronomer/astronomer-cosmos/issues/943 
* [x] https://github.com/astronomer/astronomer-cosmos/issues/930 
* [x] https://github.com/astronomer/astronomer-cosmos/issues/906 
* [x] https://github.com/astronomer/astronomer-cosmos/issues/875 
* [x] https://github.com/astronomer/astronomer-cosmos/issues/823 
* [x] https://github.com/astronomer/astronomer-cosmos/issues/693 
* [x] https://github.com/astronomer/astronomer-cosmos/issues/630 
* [x] https://github.com/astronomer/astronomer-cosmos/issues/610
* [x] https://github.com/astronomer/astronomer-cosmos/issues/572
* [x] https://github.com/astronomer/astronomer-cosmos/issues/535
* [x] https://github.com/astronomer/astronomer-cosmos/issues/448

tatiana:
I updated the description of this release, to contain all the scope.

This is the list of tasks:
* https://github.com/astronomer/astronomer-cosmos/issues/958
* https://github.com/astronomer/astronomer-cosmos/issues/1042
* https://github.com/astronomer/astronomer-cosmos/issues/927
* https://github.com/astronomer/astronomer-cosmos/issues/894
* https://github.com/astronomer/astronomer-cosmos/issues/1078
* https://github.com/astronomer/astronomer-cosmos/issues/1075
* https://github.com/astronomer/astronomer-cosmos/issues/1072
* https://github.com/astronomer/astronomer-cosmos/issues/448
* https://github.com/astronomer/astronomer-cosmos/pull/661
* https://github.com/astronomer/astronomer-cosmos/issues/875
* https://github.com/astronomer/astronomer-cosmos/issues/572
* https://github.com/astronomer/astronomer-cosmos/issues/630
[...]

PR #1182, opened 2024-08-27
Resolve CI failures for Apache Airflow 2.7 jobs

pankajkoti:
## Description

This is an effort to fix the currently broken CI jobs. The CI pipeline was running smoothly until the Cosmos 1.6.0 release (#1103). However, after the recent release of Apache Airflow providers, we’ve been encountering deep resolution errors specifically with Apache Airflow 2.7 jobs.

To address this:

1. For Airflow 2.7 tests, we’re pre-installing the providers using Airflow 2.7 constraints.
2. We’re also upgrading certain jobs, like type-checks and dbt 1.5.4 tests, from Airflow 2.7 to Airflow 2.8, as it seems beneficial to move from 2.7 to 2.8.

Despite pre-installing the providers, we’re noticing that on Python 3.8 and 3.9 with Apache Airflow 2.7, Cosmos still attempts to sync dependencies and continues to encounter deep resolution errors. Therefore, we’re specifically excluding those combinations from running in Unit and Integration test jobs. However, we have Python 3.10 and 3.11 jobs enabled and running fine for Airflow 2.7 in our CI.

## Related Issue(s)

related: #1180 
## Checklist


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,fix-ci]
```
```suggestion
    branches: [main]
```
Remove before merging once we decide to go ahead with the approach in this PR.

tatiana (approved):
Thanks a lot for working on this, @pankajkoti , and coming up with a solution so efficiently!

Issue #1197, opened 2024-09-07
[Bug] Cannot create datasets in projects with model names containing multibyte characters


t0momi219:
### Astronomer Cosmos Version

main (development)

### dbt-core version

1.8

### LoadMode

AUTOMATIC

### ExecutionMode

VIRTUALENV

### InvocationMode

None

### airflow version

2.10.0

### Operating System

Debian GNU/Linux 12 (official airflow docker image)

### Deployment

Docker-Compose

### What happened?

In projects containing models with names like the following, dataset creation fails, and an error occurs during execution.

```txt
└── dbt
    └── my_project
        ├── dbt_packages
        ├── dbt_project.yml
        ├── models
            ├── 日本語名モデル.sql
            └── 日本語名モデル.yml
```
The error message is as follows:

```txt
[2024-09-07T04:10:58.435+0000] {local.py:455} DEBUG - URIs to be converted to Dataset: ['snowflake://***.ap-northeast-1.aws/TEST_DB.DBT.日本語名モデル']
[2024-09-07T04:10:58.437+0000] {taskinstance.py:3301} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.12/site-packages/cosmos/operators/local.py", line 459, in get_datasets
    datasets = [Dataset(uri) for uri in uris]
                ^^^^^^^^^^^^
  File "<attrs generated init airflow.datasets.Dataset>", line 3, in __init__
    _setattr('uri', __attr_converter_uri(uri))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/datasets/__init__.py", line 78, in _sanitize_uri
    raise ValueError("Dataset URI must only consist of ASCII characters")
ValueError: Dataset URI must only consist of ASCII characters
```

### How to reproduce

Add model has multibyte character name.

ex: 日本語名モデル.sql

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

PR #1198, opened 2024-09-07
URL encode dataset names to support multibyte characters

t0momi219:
## Description

In projects containing models with names like the following, dataset creation fails, and an error occurs during execution.
```txt
└── dbt
    └── my_project
        └── models
            ├── 日本語名モデル.sql
            └── 日本語名モデル.yml

```

```log
[2024-09-07T04:10:58.435+0000] {local.py:455} DEBUG - URIs to be converted to Dataset: ['snowflake://***.ap-northeast-1.aws/TEST_DB.TEST_SCHEMA.日本語名モデル']
[2024-09-07T04:10:58.437+0000] {taskinstance.py:3301} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.12/site-packages/cosmos/operators/local.py", line 459, in get_datasets
    datasets = [Dataset(uri) for uri in uris]
                ^^^^^^^^^^^^
  File "<attrs generated init airflow.datasets.Dataset>", line 3, in __init__
    _setattr('uri', __attr_converter_uri(uri))
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/datasets/__init__.py", line 78, in _sanitize_uri
    raise ValueError("Dataset URI must only consist of ASCII characters")
ValueError: Dataset URI must only consist of ASCII characters
```

To support model names with multibyte characters, it might be good to URL encode the names.

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/1197

## Checklist


cosmos/operators/local.py
```diff
@@ -450,7 +451,7 @@ def get_datasets(self, source: Literal["inputs", "outputs"]) -> list[Dataset]:
         uris = []
         for completed in self.openlineage_events_completes:
             for output in getattr(completed, source):
-                dataset_uri = output.namespace + "/" + output.name
+                dataset_uri = output.namespace + "/" + urllib.parse.quote(output.name)
```
pankajkoti, line 454: Can we please add a test case to demonstrate the benefits or the reason for leading to have this change?

tatiana (commented):
Hi @t0momi219 this contribution is very valuable, thank you for working on this!

Please, as suggested by @pankajkoti , could you add a test case?
It would be great if we could have a test that didn't pass this change, when calling `get_datasets`, and that it passes after this change. This would help make sure future changes in Cosmos won't undo this fix.
The right place is probably: https://github.com/astronomer/astronomer-cosmos/blob/main/tests/operators/test_local.py to confirm 

Once we have the test, we can merge this change. If we can do this during this week, we can release it as part of Cosmos 1.7.

t0momi219:
@pankajkoti @tatiana 
Thank you for your review. I plan to implement a test by placing a model with a multibyte character name in the `dev/dags/dbt/simple/` project. If there is a more appropriate directory where this model should be placed, please let me know.
I will write the test by tomorrow at the latest.
Thanks!

tatiana:
Hi @t0momi219 , thank you very much! Sounds like a good plan!

docs/contributing.rst
```diff
@@ -115,6 +115,7 @@ To run the integration tests for the first time, use:
 
     export AIRFLOW_HOME=`pwd`
     export AIRFLOW_CONN_AIRFLOW_DB=postgres://postgres:postgres@0.0.0.0:5432/postgres
```
tatiana, line 117: Thank you so much for updating the contributing guidelines, @t0momi219 !
This one is no longer needed -`AIRFLOW_CONN_EXAMPLE_CONN` replaced `AIRFLOW_CONN_AIRFLOW_DB`, could you update this? We'll merge once it is done 🎉

tatiana (approved):
Thank you very much, @t0momi219 , for identifying, reporting and fixing this issue. I'm sure many members of the community will be able to benefit from this.

Thanks also for adding the missing test.

I made one last comment inline, if you could address it before we merge this PR, it would be awesome! Thank you very much.

docs/contributing.rst
t0momi219: Thank you for your review!
It seems that the missing variable caused an error in my local environment, but I believe it was an issue on my side.
I’ve removed the variable.

pankajkoti (approved):
LGTM, thanks @t0momi219 . Will keep an eye on the CI run & once it succeeds, I will go ahead to merge it

PR #1208, opened 2024-09-18
Fix CI issues - Upgrade actions/upload-artifact & actions/download-artifact to v4 and set min version for packaging

pankajkoti:
I am observing that our CI [jobs](https://github.com/astronomer/astronomer-cosmos/actions/runs/10920015707/job/30308959677?pr=1207) are failing complaining that we're using deprecated version (v2) for `actions/upload-artifact` & `actions/download-artifact`. It is weird that the jobs directly are failed for using deprecating versions, but so is the case that's happening. Hence, to get the CI working I'm upgrading the versions to use v4.

Additionally, we are observing failures for packaging version as discussed in https://github.com/astronomer/astronomer-cosmos/pull/1201#discussion_r1754867454. Hence, also set the min packaging version to `22.0` as per the discussion in that thread

closes: https://github.com/astronomer/astronomer-cosmos/issues/1209

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main,upgrade-gh-action-artifact-upload-download]
+    branches: [main]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,upgrade-gh-action-artifact-upload-download]
```
```suggestion
    branches: [main]
```
Remove current branch added for testing CI before merging the PR

tatiana:
Thanks a lot for working on this, @pankajkoti , and solving the CI issue proactively while we were at the Airflow Summit!

PR #1210, opened 2024-09-19
Fix CI wrt broken coverage upload artifact

pankajkoti:
Apparently the GitHub action https://github.com/actions/upload-artifact released a breaking change https://github.com/actions/upload-artifact/issues/602 (not following semver guidelines) which is affecting our CodeCov job as it's no longer
able to download artifacts that were supposed to be uploaded from our tests in earlier jobs in the CI.

According to suggestion in https://github.com/actions/upload-artifact/pull/607/files, set include include-hidden-files: true, so that our workflow continues to work as it was before.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main,ci-fix]
+    branches: [main]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,ci-fix]
```
```suggestion
    branches: [main]
```
Before merging the PR, remove the current branch added for testing CI

tatiana (approved):

Issue #1119, opened 2024-07-24
Dataset Aliases


pankajkoti:
Description co-authored by @tatiana @pankajastro

Since Cosmos 1.1, it creates Airflow inlets and outlets for every dbt model/seed/snapshot task, which allows end-users to leverage [Airflow Data-aware scheduling](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/datasets.html).

In the past, Cosmos had identified these inlets and outlets using URIs that [were not representative of the dataset being created](https://github.com/astronomer/astronomer-cosmos/issues/305). The one advantage with this approach is that the identifiers could be created during DAG parsing/processing time.

This [changed in the 1.1 release](https://github.com/astronomer/astronomer-cosmos/pull/485), when we decided to adopt the OpenLineage naming convention to describe Airflow Datasets created by Cosmos (inlets/outlets). They became something similar to: "postgres://0.0.0.0:5432/postgres.public.stg_customers". The downside with this approach was: we started using a library `openlineage-integration-common` that can only create the resources URIs after the dbt command was run, since it currently relies on dbt-core artefacts. This means we started creating inlets/outlets during task execution.

A side-effect of this change was that Airflow <= 2.9 was not designed to support setting inlets and outlets during task execution, which resulted in this long-standing issue:
https://github.com/astronomer/astronomer-cosmos/issues/522

Another side effect was that, since we started relying on task execution to determine the Airflow dataset identifier, we didn't expose end-users to a method for easily determining it. More context on https://github.com/astronomer/astronomer-cosmos/issues/1036.

The community very often raises that.

We created an issue in Airflow:
https://github.com/apache/airflow/issues/34206

After several discussions with @uranusjr, he proposed introducing the concept of DatasetAliases to Airflow 2.10. @Lee-W worked on this: https://github.com/apache/airflow/pull/40478

This feature will be released as part of Airlfow 2.10.

The goal of this epic is to leverage Airflow DatasetAliasses in Cosmos, so that:
- users can clearly see datasets created by Cosmos, during task execution, in the Airflow UI
- We can have non-OpenLineage Dataset Aliases that can be added during DAG parsing time. We can expose methods for users than to be able to retrieve these.

Initially planned tasks, more to be added as part of the PoC ticket:
* https://github.com/astronomer/astronomer-cosmos/issues/522
* https://github.com/astronomer/astronomer-cosmos/issues/1135
[...]

Issue #522, opened 2023-09-08
Airflow orphaning unreferenced Dataset URIs in Cosmos 1.1


tatiana:
**Context**

While testing Cosmos 1.1, that implemented #485, I realized that the Dataset URIs were being deleted from the Airflow UI after a few seconds they had been added, and the scheduler would print messages like:
```
 scheduler | [2023-09-08T10:18:34.252+0100] {scheduler_job_runner.py:1742} INFO - Orphaning unreferenced dataset 'postgres://0.0.0.0:5432/postgres.public.stg_customers'
 scheduler | [2023-09-08T10:18:34.252+0100] {scheduler_job_runner.py:1742} INFO - Orphaning unreferenced dataset 'postgres://0.0.0.0:5432/postgres.public.stg_payments'
 scheduler | [2023-09-08T10:18:34.252+0100] {scheduler_job_runner.py:1742} INFO - Orphaning unreferenced dataset 'postgres://0.0.0.0:5432/postgres.public.stg_orders'
 scheduler | [2023-09-08T10:18:34.252+0100] {scheduler_job_runner.py:1742} INFO - Orphaning unreferenced dataset 'postgres://0.0.0.0:5432/postgres.public.customers'
```

I had a conversation with @ashb, one of the original developers of the Dataset feature in Airflow, and he mentioned that Airflow, by design, didn't allow setting Dataset URIs during task execution and that they are being overridden every time the scheduler parses the DAGs.

Ash agreed that it makes sense for Airflow to allow libraries to set Dataset URIs. In the past, there was the concern people would be creating Dataset URIs specific to timestamps, which could lead to other issues in Airflow.

I've logged a ticket in Airflow to support setting Datasets during DAG task executions: https://github.com/apache/airflow/issues/34206

But even if this were part of Airflow 2.8, we need to find a strategy that works for users using Cosmos with older versions of Airflow.

**Recommendation**

Set inlets and outlets Dataset URIs in Cosmos during DAG parsing time when users use `LoadMethod.DBT_LS` or `LoadMethod.DBT_MANIFEST`. We should still use the open lineage Processor class with the dbt `Manifest`, but we may need to change the code so it allows generating events when the `run_results.json` file is not available (to be checked).

Nodes would have two new properties: inlets and outlets, with the correct Dataset URIs.

tatiana:
Follow up with the community in #airflow-dbt:
[...]

PR #1217, opened 2024-09-23
Support using `DatasetAlias` and fix orphaning unreferenced dataset 

tatiana:
**Context**

Cosmos versions between 1.1 and 1.6 supported automatically emitting Airflow Datasets when using `ExecutionMode.LOCAL`. Although the datasets generated by these versions of Cosmos could be utilised for dataset-aware scheduling, the implementation had long-standing issues, as described in #522.

The main problems were:
* Orphaning unreferenced dataset
* Not displaying dataset inlets/outlets in the Airflow UI

These issues were caused by Cosmos defining task outlets and inlets during Airflow task execution, a feature only partially supported before Airflow 2.10: https://github.com/apache/airflow/issues/34206.

**Solution**

Airflow 2.10 has introduced the concept of `DatasetAlias`, as described in the [official docs](https://airflow.apache.org/blog/airflow-2.10.0/#dynamic-dataset-scheduling-through-datasetalias), so operators can dynamically define inlets and outlets during task execution.

This PR uses Airflow `DatasetAlias`, when possible (Airflow 2.10 or above), and does two things:
1. Adds a `DatasetAlias` to every LocalOperator/VirtualenvOperator subclass
2. Dynamically adds `Dataset` as outlets during the `LocalOperator` subclasses execution, associating them to the desired `Dataset` instance.
3. Exposes to users a function to retrieve Cosmos' `DatasetAlias` names, programatically

**Caveats**

* Only works for Airflow 2.10 and above

This feature relies on `DatasetAlias`, only available in Airflow 2.10 and above. If users use previous versions of Airflow, Cosmos behaves like it did before, and the issues described in this task are not solved.

* Unable to leverage DatasetAlias in `airflow dags test`

Although the feature described in this PR works well when scheduling DAGs, triggering them via the UI, or using `airflow dags trigger`, it does not work when users attempt to use `dags test` or `dag.test()`. When trying to test the DAG, these commands fail with an `sqlalchemy.orm. etc.FlushError`. This is a known issue from Airflow 2.10.0 and Airflow 2.10.1 when declaring DatasetAliases, as described in https://github.com/apache/airflow/issues/42495. 

To mitigate this second problem, we've introduced a new Airflow variable, `AIRFLOW__COSMOS__ENABLE_DATASET_ALIAS`, that allows users to disable using dataset aliases when running Cosmos. We'd recommend users who face the `sqlalchemy.orm. etc.FlushError` in their tests to set this configuration to `False` only for running tests - until the issue is solved in Airflow. When this configuration is set to `False`, Cosmos behaves as before the DatasetAlias feature was introduced.

**How this feature was validated**

TODO

**Related tickets**

Closes: #522
Closes: #1119

**Pending**

* Add docs
* Update PR description


cosmos/core/airflow.py
```diff
@@ -29,6 +35,13 @@ def get_airflow_task(task: Task, dag: DAG, task_group: "TaskGroup | None" = None
     if task.owner != "":
         task_kwargs["owner"] = task.owner
 
+    if module_name == "local" and AIRFLOW_VERSION >= Version("2.10"):
+        from airflow.datasets import DatasetAlias
+
+        # ignoring the type because older versions of Airflow raise the follow error in MyPU
```
Lee-W, line 41:
```suggestion
        # ignoring the type because older versions of Airflow raise the follow error in mypy
```

mypy?

cosmos/operators/local.py
```diff
@@ -145,7 +153,19 @@ def __init__(
         self._dbt_runner: dbtRunner | None = None
         if self.invocation_mode:
             self._set_invocation_methods()
-        super().__init__(**kwargs)
+
+        if AIRFLOW_VERSION >= Version("2.10"):
```
tatiana, line 157: only run this if `emit_datasets` is set to true

cosmos/operators/local.py
```diff
@@ -469,20 +488,33 @@ def get_datasets(self, source: Literal["inputs", "outputs"]) -> list[Dataset]:
...
         """
-        Register a list of datasets as outlets of the current task.
+        Register a list of datasets as outlets of the current task, when possible.
+
         Until Airflow 2.7, there was not a better interface to associate outlets to a task during execution.
+        This works before Airflow 2.10 with a few limitations, as described in the ticket:
+        https://github.com/astronomer/astronomer-cosmos/issues/522
+
+        In Airflow 2.10.0 and 2.10.1, we are not able to test Airflow DAGs powered with DatasetAlias.
+        https://github.com/apache/airflow/issues/42495
```
pankajkoti, line 500: Should we mention somewhere in the docstring that we support the power of DatasetAlias in further versions?

pankajkoti (approved):
LGTM, excited for this one! 👏🏽

PR #1224, opened 2024-09-25
Introduce ground work for `ExecutionMode.AIRFLOW_ASYNC`

pankajkoti:
This PR is the groundwork for the implementation of ExecutionMode.AIRFLOW_ASYNC (https://github.com/astronomer/astronomer-cosmos/issues/1120), which - once all other epic tasks are completed - will enable asynchronous execution of dbt resources using Apache Airflow’s deferrable operators. As part of this work, this PR introduces a new option to the enum `ExecutionMode` : `AIRFLOW_ASYNC`. When this execution mode is used, Cosmos now creates a setup task that will pre-compile the dbt project SQL and make it available to the remaining dbt tasks. This PR, however, does not yet leverage Airflow's deferrable operators. If users use  `ExecutionMode.AIRFLOW_ASYNC` they will actually be running `ExecutionMode.LOCAL` operators with this change. The PR (#1230) has a first experimental version of using deferrable operators for task execution.

## Setup task as the ground work for a new Execution Mode: `ExecutionMode.AIRFLOW_ASYNC`:
- Adds a new operator, `DbtCompileAirflowAsyncOperator`, as a root task(analogous to a setup task) in the DAG, running the dbt compile command and uploading the compiled SQL files to a remote storage location for subsequent tasks that fetch these compiled SQL files from the remote storage and run them asynchronously using Airflow's deferrable operators.

## Airflow Configurations:
- `remote_target_path`: Introduces a configurable path to store dbt-generated files remotely, supporting any storage scheme that works with Airflow’s Object Store (e.g., S3, GCS, Azure Blob).
- `remote_target_path_conn_id`: Allows specifying a custom connection ID for the remote target path, defaulting to the scheme’s associated Airflow connection if not set.

## Example DAG for CI Testing:
Introduces an example DAG (`simple_dag_async.py`) demonstrating how to use the new execution mode(The execution like mentioned earlier would still run like Execution Mode LOCAL operators at the moment with this PR alone)
This DAG is integrated into the CI pipeline to run integration tests and aims at verifying the functionality of the `ExecutionMode.AIRFLOW_ASYNC` as and when implementation gets added starting with the experimental implementation in #1230 .

## Unit & Integration Tests:
- Adds comprehensive unit and integration tests to ensure correct behavior.
- Tests include validation for successful uploads, error handling for misconfigured remote paths, and scenarios where `remote_target_path` are not set.

## Documentation:
[...]

Issue #1134, opened 2024-08-01
PoC on `ExecutionMode.LOCAL_AIRFLOW_ASYNC`


tatiana:
Description co-authored by @pankajkoti @pankajastro 

**Context**

As of Cosmos 1.5, we are using dbt core to run transformations in the database. In other words, the Airflow worker node is being blocked during the execution of the SQL transformation. If the transformation takes 6h, we have a Airflow worker node just waiting.

One way to overcome this is to use Airflow deferrable operators, as opposed to using dbt-core commands to execute the transformation. This approach is not new, and it was discussed in Airflow Summit 2023 by [Monzo](https://airflowsummit.org/sessions/2023/airflow-at-monzo-evolving-our-data-platform-as-the-bank-scales/). We (@tatiana ) had a meeting with Monzo to discuss the details afterward, in London.

The overall idea is:
- Pre-compile the dbt project, so we have the pre-compiled SQL
- Execute the pre-compiled SQL using native Airflow operators, enabling the deferable mode

For a first end-to-end, we'll only support:
- SQL models
- `ExecutionMode.LOCAL`
- Databricks

This will allow us to measure the impact of using deferrable operators, and then we can extend to:
- Python models
- Other execution modes
- Other SQL backends

**Acceptance criteria**

* [ ] PoC to try to have this feature end-to-end that
   * Introduce `ExecutionMode.LOCAL_AIRFLOW_ASYNC` (think if we can find a better name)
   * Check how we are populating the `compiled_sql` template field in Cosmos. If this is being populated after running the task, then, we can simplify the first E2E to use `LoadMode.MANIFEST`, and assume the compiled SQL is avaialble for Cosmos.
   * If the dbt profile is not related to Databricks, error
   * if it is related to Databricks, than attempt to use the [DatabricksSubmitRunOperator](https://airflow.apache.org/docs/apache-airflow-providers-databricks/stable/operators/submit_run.html#databrickssubmitrunoperator)  to run the previously compiled sql.
* [ ] Log follow-up tasks to make this PoC production ready and to expand the scope of the first end-to-end (support Python models, support other backend, e.g. Snowflake or BQ)
* [ ] Identify limitations with this approach (e.g. some macros may not work, and how this will play along with Airflow inlets/outlets)

tatiana:
One of the advantages of this feature would also be to give additional information to end-users that Airflow currently exposes, but dbt does not.

[...]

PR #1226, opened 2024-09-25
Draft: dbt compile task as node in dbt graph

pankajkoti:
related: #1134 

Alternative to #1224


pankajkoti:
We decided to go ahead with the approach in https://github.com/astronomer/astronomer-cosmos/pull/1224. Hence closing this PR in favour of that approach.

Issue #1120, opened 2024-07-24
Increase Task Throughput


pankajkoti:
Description co-authored by @tatiana @pankajastro 

As of Cosmos 1.5, we are using dbt core to run transformations in the database. In other words, the Airflow worker node is being blocked during the execution of the SQL transformation. If the transformation takes 6h, we have a Airflow worker node just waiting.

One way to overcome this is to use Airflow deferrable operators, as opposed to using dbt-core commands to execute the transformation. This approach is not new, and it was discussed in Airflow Summit 2023 by [Monzo](https://airflowsummit.org/sessions/2023/airflow-at-monzo-evolving-our-data-platform-as-the-bank-scales/). We (@tatiana ) had a meeting with Monzo to discuss the details afterward, in London.

The overall idea is:
- Pre-compile the dbt project, so we have the pre-compiled SQL
- Execute the pre-compiled SQL using native Airflow operators, enabling the deferable mode

For a first end-to-end, we'll only support:
- SQL models
- `ExecutionMode.LOCAL`
- Databricks

This will allow us to measure the impact of using deferrable operators, and then we can extend to:
- Python models
- Other execution modes
- Other SQL backends
- Support Datasets (it's fine if we don't support in a first release of this, specially if we find a way of leveraging DatasetAlias in all ExecutionModes: #1135)

Tasks initially planned:
1. PoC to try to have this feature end-to-end that
   * Introduce `ExecutionMode.LOCAL_AIRFLOW_ASYNC` (think if we can find a better name)
   * Check how we are populating the `compiled_sql` template field in Cosmos. If this is being populated after running the task, then, we can simplify the first E2E to use `LoadMode.MANIFEST`, and assume the compiled SQL is avaialble for Cosmos.
   * If the dbt profile is not related to Databricks, error
   * if it is related to Databricks, than attempt to use the [DatabricksSubmitRunOperator](https://airflow.apache.org/docs/apache-airflow-providers-databricks/stable/operators/submit_run.html#databrickssubmitrunoperator)  to run the previously compiled sql.

Once we've done a PoC, we can then breakdown into other tasks, so we can split the work.

[...]

PR #1230, opened 2024-09-29
Experimental BQ support to run dbt models with `ExecutionMode.AIRFLOW_ASYNC`

pankajastro:
Enable BQ users to run dbt models (`full_refresh`) asynchronously. This releases the Airflow worker node from waiting while the transformation (I/O) happens in the dataware house, increasing the overall Airflow task throughput (more information: https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/deferring.html). As part of this change, we introduce the capability of not using the dbt command to run actual SQL transformations. This also avoids creating subprocesses in the worker node (`ExecutionMode.LOCAL` with `InvocationMode. SUBPROCESS` and `ExecutionMode.VIRTUALENV`) or the overhead of creating a Kubernetes Pod to execute the actual dbt command (`ExecutionMode.KUBERNETES`). This can avoid issues related to memory and CPU usage.

This PR takes advantage of an already implemented async operator in the Airflow repo by extending it in the Cosmos async operator. It also utilizes the pre-compiled SQL generated as part of the PR https://github.com/astronomer/astronomer-cosmos/pull/1224. It downloads the generated SQL from a remote location (S3/GCS), which allows us to decouple from dbt during task execution. 

## Details

- Expose `get_profile_type` on ProfileConfig: This aids in database selection
- ~Add `async_op_args`: A high-level parameter to forward arguments to the upstream operator (Airflow operator). (This may change in this PR itself)~ The async operator params are process as kwargs in the operator_args parameter
- Implement `DbtRunAirflowAsyncOperator`: This initializes the Airflow Operator, retrieves the SQL query at task runtime from a remote location, modifies the query as needed, and triggers the upstream execute method.

## Limitations

- This feature only works when using Airflow 2.8 and above
- The async execution only works for BigQuery
- The async execution only supports running dbt models (other dbt resources, such as seeds, sources, snapshots, tests, are run using the `ExecutionMode.LOCAL`)
- This will work only if the user provides sets `full_refresh=True` in `operator_args` (which means tables will be dropped before being populated, as implemented in `dbt-core`)
- Users need to use `ProfileMapping` in `ProfileConfig`, since Cosmos relies on having the connection (credentials) to be able to run the transformation in BQ without `dbt-core`
- Users must provide the BQ `location` in `operator_args` (this is a limitation from the `BigQueryInsertJobOperator` that is being used to implement the native Airflow asynchronous support)

## Testing 

We have added a new dbt project to the repository to facilitate asynchronous task execution. The goal is to accelerate development without disrupting or requiring fixes for the existing tests. Also, we have added DAG for end-to-end testing https://github.com/astronomer/astronomer-cosmos/blob/bd6657a29b111510fc34b2baf0bcc0d65ec0e5b9/dev/dags/simple_dag_async.py

## Configuration

Users need to configure the below param to execute deferrable tasks in the Cosmos

- [ExecutionMode: AIRFLOW_ASYNC](https://astronomer.github.io/astronomer-cosmos/getting_started/execution-modes.html)
- [remote_target_path](https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html#remote-target-path)
- [remote_target_path_conn_id](https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html#remote-target-path-conn-id)

Example DAG: https://github.com/astronomer/astronomer-cosmos/blob/bd6657a29b111510fc34b2baf0bcc0d65ec0e5b9/dev/dags/simple_dag_async.py

## Installation

You can leverage async operator support by installing an additional dependency
```
astronomer-cosmos[dbt-bigquery, google]
```

## Documentation 

The PR also document the limitations and uses of Airflow async execution in the Cosmos.

## Related Issue(s)

Related to: https://github.com/astronomer/astronomer-cosmos/issues/1120
Closes: #1134

## Breaking Change?

No

## Notes

This is an experimental feature, and as such, it may undergo breaking changes. We encourage users to share their experiences and feedback to improve it further.

We'd love support and feedback so we can define the next steps.

## Checklist

## Credits

This was a result of teamwork and effort:

Co-authored-by: Pankaj Koti <pankajkoti699@gmail.com>

## Future Work

- Design interface to facilitate the easy addition of new asynchronous databases operators https://github.com/astronomer/astronomer-cosmos/issues/1238
- Improve the test coverage https://github.com/astronomer/astronomer-cosmos/issues/1239
- Address the limitations (we need to log these issues)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,poc-dbt-compile-task]
```
```suggestion
    branches: [main]
```

cosmos/operators/airflow_async.py
```diff
@@ -35,8 +43,81 @@ class DbtSourceAirflowAsyncOperator(DbtSourceLocalOperator):
...
+        remote_model_path = f"{str(remote_target_path).rstrip('/')}/{self.dag_id}/{relative_file_path}"
+
+        print("remote_model_path: ", remote_model_path)
+        object_storage_path = ObjectStoragePath(remote_model_path, conn_id=remote_target_path_conn_id)
+        with object_storage_path.open() as fp:  # type: ignore
+            return fp.read()  # type: ignore
+
+    def drop_table_sql(self) -> None:
+        model_name = self.task_id.split(".")[0]
+        sql = f"DROP TABLE IF EXISTS {self.gcp_project}.{self.dataset}.{model_name};"
```
joppevos, line 90: What if my full_refresh runs for multiple hours, then the table would not be available if we drop it?
would `CREATE OR REPLACE TABLE ` be an option so we only replace it when we have the results ready?

dh-racheldean:
just a question from a curious observer: the current `execute` method, where it DROPs the existing table, then runs the CREATE deferred (i.e. potentially leaving a missing table for hours) - that's just proof of concept, right?

cosmos/operators/airflow_async.py
tatiana: @dh-racheldean @joppevos  thank you very much for the early feedback! We really value your input.

The drop will only happen for full refresh, which does drop the table before creating a new one:
* https://github.com/dbt-labs/dbt-core/blob/5e9f1b515f37dfe6cdae1ab1aa7d190b92490e24/core/dbt/context/base.py#L662-L666
* https://docs.getdbt.com/reference/resource-configs/full_refresh#recommendation
Screenshot 2024-10-01 at 10 13 23

cosmos/operators/local.py
```diff
@@ -1009,5 +1012,5 @@ def __init__(self, **kwargs: str) -> None:
 
 class DbtCompileLocalOperator(DbtCompileMixin, DbtLocalBaseOperator):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
-        kwargs["should_upload_compiled_sql"] = True
+        kwargs["should_upload_compiled_sql"] = False
```
pankajkoti, line 1015:
```suggestion
        kwargs["should_upload_compiled_sql"] = True
```
to update before merging

cosmos/operators/airflow_async.py
tatiana: We're also learning a lot about dbt during this task, @joppevos !

We're trying to get a first version with very, very limited support to collect feedback from the community. It is a proof of concept, and it will be experimental. We thought full refresh was the simplest as a starting point. We'll need more brains and hands to bring this feature to a stable state - this is why your early feedback is so important.

Unfortunately, "dbt compile" does not expose the materializations (e.g. INSERT, REPLACE, CREATE). The `dbt run` command is responsible for decorating the select statements with whatever needs to be done. Another challenge is that currently, the dbt adaptors are responsible for implementing part of this logic; as an example:
https://github.com/dbt-labs/dbt-bigquery/blob/ae0f91c7a9dac303632eba2a99161251be6396f7/dbt/include/bigquery/macros/materializations/incremental_strategy/insert_overwrite.sql

Our solution may involve using `dbtRunner` to handle materializations. The drawback is that this would only work if people install dbt and Airflow in the same Python virtualenv.

cosmos/operators/airflow_async.py
```diff
@@ -35,8 +43,91 @@ class DbtSourceAirflowAsyncOperator(DbtSourceLocalOperator):
...
+        hook.insert_job(configuration=self.configuration, location=self.location, project_id=self.gcp_project)
+
+    def execute(self, context: Context) -> Any | None:
+        if not self.full_refresh:
+            raise CosmosValueError("The async execution only supported for full_refresh")
+        else:
+            self.drop_table_sql()
+
+            sql = self.get_remote_sql()
+            model_name = self.task_id.split(".")[0]
```
pankajkoti, line 120: This fails when we define `DbtTaskGroup` instead of `DbtDag` as the tasks have ID as e.g. `task_group.stg_customers.run` when converted from `DbtTaskGroup` instead of `stg_customers.run` as observed when converted from `DbtDag`

pankajkoti: As a result it creates such a table and keeps overwriting that table for each of the tasks in the task group. dbt tests fail eventually as it cannot find the needed tables like stg_customers, stg_orders, etc.
Screenshot 2024-10-02 at 12 22 53 AM

pankajkoti: Pushed a fix for this in https://github.com/astronomer/astronomer-cosmos/pull/1230/commits/72c6164480c553b12c366bc11fc9d9e049515100

dev/dags/simple_dag_async.py
```diff
@@ -2,38 +2,41 @@
...
+        # test_behavior=TestBehavior.NONE
+    ),
     # normal dag parameters
     schedule_interval=None,
     start_date=datetime(2023, 1, 1),
     catchup=False,
     dag_id="simple_dag_async",
     tags=["simple"],
-    operator_args={"install_deps": True},
+    operator_args={"install_deps": True, "full_refresh": True, "location": "northamerica-northeast1"},
```
pankajkoti, line 40:
```suggestion
    operator_args={"full_refresh": True, "location": "northamerica-northeast1"},
```

pankajkoti:
okay, I think we're close now in making the async DAG run in CI. We would need to update the secret.

 Discovered that we need to pass the gcp_conn_id to the parent class init, otherwise the one that we supply in the profile is lost & the hook tries to find the default credentials which is not the desired behaviour. Fixed in https://github.com/astronomer/astronomer-cosmos/pull/1230/commits/af454a924092b2472b340680ffc6885a5695a1ef
 
 Once, we set the secret it would pass I believe 🤞🏽

cosmos/operators/airflow_async.py
```diff
@@ -1,67 +1,197 @@
...
+    def drop_table_sql(self) -> None:
+        model_name = self.extra_context["dbt_node_config"]["resource_name"]  # type: ignore
+        sql = f"DROP TABLE IF EXISTS {self.gcp_project}.{self.dataset}.{model_name};"
+
+        print("sql: ", sql)
+        print("gcp_project: ", self.gcp_project)
+        print("dataset: ", self.dataset)
+        print("gcp_conn_id: ", self.gcp_conn_id)
+        print("self.configuration: ", self.configuration)
+        print("self.location: ", self.location)
```
pankajkoti, line 153:
```suggestion
```

cosmos/operators/airflow_async.py
```diff
@@ -1,67 +1,197 @@
...
+        remote_target_path_str = str(remote_target_path).rstrip("/")
+
+        if TYPE_CHECKING:
+            assert self.project_dir is not None
+
+        project_dir_parent = str(Path(self.project_dir).parent)
+        relative_file_path = str(file_path).replace(project_dir_parent, "").lstrip("/")
+        remote_model_path = f"{remote_target_path_str}/{dbt_dag_task_group_identifier}/compiled/{relative_file_path}"
+
+        print("remote_model_path: ", remote_model_path)
```
pankajkoti, line 139:
```suggestion
```

tatiana:
Thank you a lot, @pankajkoti , that's outstanding progress!
I updated GH actions earlier today and just released 1.7.0a1 🚀

tatiana:
There are two issues we have to solve before merging this PR

* [ ] 1. CI is getting stuck when trying to run
```
    tests/test_example_dags.py::test_example_dag[simple_dag_async]
```
* [ ] 2. Make compile upstream dependency to the dbt graph parent nodes

pankajkoti:
> * [x]  2. Make compile upstream dependency to the dbt graph parent nodes
> 
> We need to change the function `_add_dbt_compile_task` to not check based on the dbt graph but to identify parent nodes based on `tasks_map.values()` and identify all the tasks that do not have upstream node.

Worked on resolving this in https://github.com/astronomer/astronomer-cosmos/pull/1230/commits/05db6a0cc75f50efc980bc290eda53347e2112d0

Tested the DAG run & built graph with airflow standalone
Screenshot 2024-10-02 at 5 06 18 PM

cosmos/operators/base.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         dbt_cmd_global_flags: list[str] | None = None,
         cache_dir: Path | None = None,
         extra_context: dict[str, Any] | None = None,
+        # configuration: dict[str, Any] | None = None,
```
pankajkoti, line 113:
```suggestion
```

cosmos/operators/base.py
```diff
@@ -140,6 +141,7 @@ def __init__(
         self.cache_dir = cache_dir
         self.extra_context = extra_context or {}
         kwargs.pop("full_refresh", None)  # usage of this param should be implemented in child classes
+        # kwargs["configuration"] = {}
```
pankajkoti, line 144:
```suggestion
```

tatiana:
On the issue:

> There are two issues we have to solve before merging this PR
>  1. CI is getting stuck when trying to run
    tests/test_example_dags.py::test_example_dag[simple_dag_async]

We identified that Cosmos was using its own `dag.test` that was buggy when used with Airflow deferrable operators. We changed to use Airflow's one if users are using AF >= 2.5 and it seems to have resolved the issue:
16a87ea (CI jobs were no longer stuck)

tatiana (approved):
Incredible work, @pankajastro @pankajkoti ! 🎉 

To run dbt transformations without dbt while leveraging Airflow asynchronous processing is far from trivial - and this PR is a significant step in the right direction.

jlaneve:
I haven't read through the code but this feature looks so awesome - it's something we've been talking about since day 1 so it's incredible to see it come to life!

cosmos/operators/airflow_async.py
pankajkoti: I am resolving this conversation for the merge to be allowed. We can unresolve it later again if we would like this to be visible for posterity once the PR is merged.

tatiana:
I just created a few follow-up tickets:
* #1260
* #1261
* #1262
* #1263
* #1264
* #1265

PR #1235, opened 2024-10-03
Try upgrading dbt-core & dbt adapters versions to see if the dependencies resolve

pankajkoti:
Observed sqlite integration job failing to resolve dependencies after adding the dbt-bigquery adapter to the job in the PR #1230 
Hence, try to use newer dbt-core & adapters to see if CI is able to resolve dependencies

### DIFF
--- a/scripts/test/integration-sqlite-setup.sh
+++ b/scripts/test/integration-sqlite-setup.sh
@@ -1,4 +1,4 @@
-pip uninstall -y dbt-core dbt-sqlite openlineage-airflow openlineage-integration-common; \
+pip uninstall -y dbt-core dbt-sqlite dbt-adapters dbt-semantic-interfaces dbt-common dbt-extractor openlineage-airflow openlineage-integration-common; \
 rm -rf airflow.*; \
 airflow db init; \
 pip install 'dbt-core==1.8' 'dbt-sqlite<=1.8' 'dbt-databricks<=1.8' 'dbt-postgres<=1.8'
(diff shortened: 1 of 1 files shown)


scripts/test/integration-sqlite-setup.sh
```diff
@@ -1,4 +1,4 @@
 pip uninstall -y dbt-core dbt-sqlite openlineage-airflow openlineage-integration-common; \
 rm -rf airflow.*; \
 airflow db init; \
-pip install 'dbt-core==1.4' 'dbt-sqlite<=1.4' 'dbt-databricks<=1.4' 'dbt-postgres<=1.4'
+pip install 'dbt-core==1.8' 'dbt-sqlite<=1.8' 'dbt-databricks<=1.8' 'dbt-postgres<=1.8'
```
pankajastro, line 4:
```suggestion
pip install 'dbt-core==1.5' 'dbt-sqlite<=1.5' 'dbt-databricks<=1.5' 'dbt-postgres<=1.5'
```

tatiana: Latest dbt-sqlite adaptor is 1.5 https://pypi.org/project/dbt-sqlite/#history

pankajkoti: yes, we specify `dbt-sqlite<=1.8`, so whenever a newer version is released it would be picked up & should ideally be still compatible with `dbt-core==1.8`. This is my rationale. WDYT?

pankajkoti:
closing in favour of #1236

PR #1237, opened 2024-10-03
Update GCP manifest file path based on new secret update

pankajkoti:
We recently updated our GCP GS connection secret in our CI & that points to a different account.
Update the bucket name available in that account in our example DAG so that it is able to find &
use the needed manifest file with the updated secret in CI.

### DIFF
--- a/dev/dags/cosmos_manifest_example.py
+++ b/dev/dags/cosmos_manifest_example.py
@@ -74,7 +74,7 @@ def cosmos_manifest_example() -> None:
     gcp_gs_example = DbtTaskGroup(
         group_id="gcp_gs_example",
         project_config=ProjectConfig(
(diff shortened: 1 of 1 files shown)


dev/dags/cosmos_manifest_example.py
```diff
@@ -74,7 +74,7 @@ def cosmos_manifest_example() -> None:
     gcp_gs_example = DbtTaskGroup(
         group_id="gcp_gs_example",
         project_config=ProjectConfig(
-            manifest_path="gs://cosmos-manifest-test/manifest.json",
+            manifest_path="gs://cosmos_remote_targe/manifest.json",
```
pankajkoti, line 77:
```suggestion
            manifest_path="gs://cosmos_remote_target/manifest.json",
```

pankajastro (approved):

PR #1240, opened 2024-10-03
Fix Dataset and DatasetAlias scheduling in Airflow 2.10

tatiana:
When tasks inside task groups were emitting events, they were different during task initialization and task execution, resulting in different DatasetAlias representations and scheduling not working as expected.

During task execution, the task ID is modified also to contain the task_group.

We could not catch this before due to inconsistent behaviour between running Airflow using `dags.test()`, `airflow standalone` and `astro-cli`.


cosmos/dataset.py
```diff
@@ -17,7 +17,7 @@ def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_i
             dag_id = task_group.dag_id
         if task_group.group_id is not None:
             task_group_id = task_group.group_id
-            task_group_id = task_group_id.replace(".", "__")
+            task_group_id = task_group_id.split(".")[-1]
```
pankajastro, line 20: Do we need to worry about any backward compatibility issues?

pankajastro (approved):

pankajkoti (approved):
LGTM. Tested this with DbtDag & DbtTaskGroup DAGs with astro-cli.

If we rebase the PR, #1237 that was merged should fix the failing integration tests.

cosmos/dataset.py
tatiana: We haven't released this module yet - we implemented it as part of the `DatasetAlias` that will be released in 1.7.0. Since TaskGroups are unique per DAG, I think it's an opportunity to simplify.

I also predicted that we could have a similar issue with task groups being resolved differently in initialization/execution - so this will hopefully avoid any potential issues in this area.

Issue #1245, opened 2024-10-08
[Feature] Make it easier for end-users to create a DAG for creating/hosting dbt docs


tatiana:
### Description

This idea came from @cmarteepants after she prepared a webinar about Cosmos.

Currently, users have to create a DAG and instantiate manually operators:
https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/dbt_docs.py

Then they also have to set dedicated environment variables to be able to see the docs in Airflow:
https://astronomer.github.io/astronomer-cosmos/configuration/hosting-docs.html

Would there be an easier way of setting this up so these details don't have to be set up in both places?

Would it make sense for us to create a `DbtDocsDag` so users don't have to instantiate an Airflow DAG and add the operators to it? How could we simplify the end-user experience?

### Use case/motivation

Simplify creation and usage of dbt docs.

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

johnmcochran:
I found this issue just now because I'm setting up a demo of astro cosmos locally to show my boss dbt before moving to astronomer. I successfully generated the docs with only a few troubles (found it to be unintuitive that I had to provide an execution_config in the DbtTaskGroup task, yet had to provide just the dbt_executable_path in the DbtDocsS3Operator.

I'm having quite a hard time getting the docs to host properly, even after setting up the variable in airflow for AIRFLOW__COSMOS__DBT_DOCS_CONN_ID and AIRFLOW__COSMOS__DBT_DOCS_DIR, which is surprising to me considering that I know the conn id works since the docs generated into S3 successfully. The docs dir is set to the s3 uri copied straight from S3.

I would have expected the generate docs operator to automatically set up the hosting since the operator already has the aws_conn_id and dropped the docs to S3. Seems redundant needing to create environment variables with the same pieces of information provided to the DbtDocsS3Operator.

cmarteepants:
@johnmcochran that sounds frustrating, but I'm glad you were able to make it work.

[...]

Issue #1139, opened 2024-08-01
Write blog post dbt docs in Airflow with Cosmos


tatiana:
**Context**

Many people from the community need to realise Cosmos can host dbt docs, as implemented by @dwreeves in https://github.com/astronomer/astronomer-cosmos/pull/737.

The goal of this blog post is to give visibility to this feature. Check with @dwreeves if he'd like to author or co-author this.

Some possible topics that could be covered:
- chose a object store ([MinIO](https://min.io/) - [example](https://github.com/astronomer/astro-sdk/pull/1146/files), S3, GCS, Azure blob storage)
- walk through how Cosmos supports users to [generate dbt docs](https://astronomer.github.io/astronomer-cosmos/configuration/generating-docs.html), eg. https://github.com/astronomer/astronomer-cosmos/blob/7889fbe7abf3ad79a14530e91ed90ef6352eba4e/cosmos/operators/local.py#L649
- walk through how Cosmos supports[ hosting the docs](https://astronomer.github.io/astronomer-cosmos/configuration/hosting-docs.html) in the Airflow UI

**Acceptance criteria**
- [ ] Publish the blog post

PR #1250, opened 2024-10-10
Improve dbt Docs Hosting Debugging -- Update dbt_docs_not_set_up.html

johnmcochran:
## Description

[ dbt_docs_dir is None ] directly triggers the [ dbt_docs_net_set_up.html ] page via the cosmos plugin. I had a hard time figuring out why my dbt docs weren't hosting properly considering I had (incorrectly) tried setting up the dbt_docs_dir variable via the Airflow UI. Once I added the environment variable to my Dockerfile, the dbt Docs showed correctly in the airflow UI. I had to inspect the source code for the plugin to find out that cosmos was not correctly seeing the environment variables I had set up.

Adding a note that the "dbt_docs_dir was not found" would make this error page much more helpful from a debugging standpoint for me.

## Related Issue(s)

Related to #1245: Make it easier for end-users to create a DAG for creating/hosting dbt docs
Related to #1139: Write blog post dbt docs in Airflow with Cosmos 

## Breaking Change?

This is my first time creating a PR on a public github, but I think this is as low impact as you can get :)

## Checklist

- [NA ] I have made corresponding changes to the documentation (if required)

I don't know how to test this :( Just trying to save others some time if they have the same problem I had with correctly hosting dbt docs.

### DIFF
--- a/cosmos/plugin/templates/dbt_docs_not_set_up.html
+++ b/cosmos/plugin/templates/dbt_docs_not_set_up.html
@@ -1,6 +1,6 @@
 {% extends base_template %}
 {% block content %}
-    <h1 style="display: block; padding-top: 15px; padding-bottom: 15px">⚠️ Your dbt docs are not set up yet! ⚠️</h1>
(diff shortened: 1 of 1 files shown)


tatiana (approved):
Thank you for the detailed description and for improving this part of Cosmos, @johnmcochran ! It's a valuable contribution.

Issue #1246, opened 2024-10-08
[Bug] DbtVirtualenvBaseOperator uses system-wide dbt instead of virtualenv-specific in v1.7.0


kesompochy:
### Astronomer Cosmos Version

Other Astronomer Cosmos version (please specify below)

### If "Other Astronomer Cosmos version" selected, which one?

1.7.0

### dbt-core version

1.8.7

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

VIRTUALENV

### InvocationMode

SUBPROCESS

### airflow version

2.4.6

### Operating System

 Google Cloud Composer (Linux-based)

### Deployment

Google Cloud Composer

### What happened?

In version 1.7.0, the dbt command is being executed using the local node's Python path instead of the virtualenv path. This causes the operator to use the system-wide dbt installation rather than the one installed in the virtualenv.

The operator should use the Python path from the created virtualenv to execute dbt commands, ensuring that the correct (virtualenv-specific) version of dbt is used.

### Relevant log output

```shell
...
[2024-10-08, 04:33:22 UTC] {virtualenv.py:86} INFO - Trying to run the command:
 ['/opt/python3.8/bin/dbt', 'deps', '--project-dir', '/tmp/tmpmykndlmp', '--profiles-dir', '/tmp/cosmos/profile/9285b6cf4e81fba356567ec2da8448788576db0af33ebb38689f0fbf0760e6e8', '--profile', 'my-profile', '--target', 'test']
...
[2024-10-08, 04:33:44 UTC] {subprocess.py:90} INFO - [0m04:33:44  Running with dbt=1.5.4
...
```

### How to reproduce

1. Set up a DAG using the `DbtVirtualenvBaseOperator` in Cloud Composer.
2. Run the DAG and observe the task logs.
3. In the logs, you should see that the dbt command is being executed with the system Python path instead of the virtualenv path.

### Anything else :)?

This bug was introduced by https://github.com/astronomer/astronomer-cosmos/pull/1200, which I created. I apologize for this bug.
I noticed that `self._py_bin` becomes `None` in the `self.run_subprocess` method, so the command does not change during task execution.

It appears that this issue is caused by `self` being bound to a different instance within the `self.run_subprocess` method. I added the following logger at various points to confirm that the instance id differs at the time of run_subprocess:

```
self.log.info("method called on instance %s", id(self))
```

The logs show:
```
[...]

Issue #958, opened 2024-05-14
[Bug] Cosmos stale temporary directories


tatiana:
**Context**

It seems sometimes Cosmos is creating and not deleting temporary directories.

An example of a report, from 30 April 2024 in the #airflow-dbt  Slack channel:
https://apache-airflow.slack.com/archives/C059CC42E9W/p1714484749579599

```
Robert Jones: Good day I'm running cosmos 1.3.2 on MWAA 2.7.2. I'm using the load_method=LoadMode.DBT_MANIFEST in the RenderConfig. But the tmp is filling up over time with dirs that are mostly symlinks but then there is also a target directories that has the compiled code and graph.gpickle. Does cosmos come with a cleanup mechanism or do I need to build a DAG that cleans the tmp directory?

Here are some examples:
[2024-04-30T12:45:15.478+0000] {{subprocess.py:93}} INFO - /tmp/tmp0x486tay/target:
[2024-04-30T12:45:15.478+0000] {{subprocess.py:93}} INFO - total 38M
[2024-04-30T12:45:15.478+0000] {{subprocess.py:93}} INFO - drwxr-xr-x 4 airflow airflow 4.0K Apr 30 11:37 .
[2024-04-30T12:45:15.478+0000] {{subprocess.py:93}} INFO - drwx------ 4 airflow airflow 4.0K Apr 30 11:36 ..
[2024-04-30T12:45:15.479+0000] {{subprocess.py:93}} INFO - drwxr-xr-x 3 airflow airflow 4.0K Apr 30 11:36 compiled
[2024-04-30T12:45:15.481+0000] {{subprocess.py:93}} INFO - -rw-r--r-- 1 airflow airflow 6.4M Apr 30 11:36 graph.gpickle
[2024-04-30T12:45:15.482+0000] {{subprocess.py:93}} INFO - -rw-r--r-- 1 airflow airflow 843K Apr 30 11:36 graph_summary.json
[2024-04-30T12:45:15.482+0000] {{subprocess.py:93}} INFO - -rw-r--r-- 1 airflow airflow  17M Apr 30 11:37 manifest.json
[2024-04-30T12:45:15.482+0000] {{subprocess.py:93}} INFO - -rw-r--r-- 1 airflow airflow  14M Apr 30 11:36 partial_parse.msgpack
[...]

PR #1252, opened 2024-10-12
Fix DbtVirtualenvBaseOperator to use correct virtualenv Python path

kesompochy:
## Description

This PR addresses an issue where the `DbtVirtualenvBaseOperator` was executing dbt commands using the system-wide Python path instead of the virtualenv path. The root cause was that the self reference in the `run_subprocess` method was bound to a different instance than the one created during initialization, likely due to Airflow's DAG pickling mechanism.
To resolve this, we've refactored the `invoke_dbt` and `handle_exception` methods to be properties. This ensures that they dynamically reference the correct method of the current instance at runtime, rather than being bound to a potentially stale instance from initialization.

## Related Issue(s)

fix https://github.com/astronomer/astronomer-cosmos/issues/1246

This may be related to issue https://github.com/astronomer/astronomer-cosmos/issues/958 in version 1.5.0

## Breaking Change?

No

## Checklist

## Additional Notes

I acknowledge that ideally, a test should be added to reproduce the original issue and verify the fix. However, I found it challenging to create an appropriate test, especially considering that this might require an integration test with Airflow to properly simulate Airflow behavior. If there are suggestions for how to effectively test this scenario, I would greatly appreciate the guidance.

I sincerely apologize for introducing this bug in the first place with PR https://github.com/astronomer/astronomer-cosmos/pull/1200. I understand this has caused inconvenience, and I'm grateful for the opportunity to fix it. I kindly request a thorough review of these changes to ensure we've fully addressed the issue without introducing new problems.


tatiana (commented):
HI @kesompochy thank you very much for working on this.

When we write code, there is always the risk we'll be introducing bugs, and in this particular case, the reviewer is as accountable as the author. So, please, don't blame yourself.

As you mentioned, I really believe we need to write an integration test to try to reproduce the original issue - or a scenario closer to it, that fails before, and passes with this fix.

I'll take sometime to reproduce the issue and try to give ideas on how we could accomplish this. We'll be releasing this in a patch release of Cosmos, if not 1.7.1, then 1.7.2.

tatiana (approved):
@kesompochy I've added a test more representative of the change in #1286. I'll merge your PR and the test I created to validate this change.

kesompochy:
@tatiana, Thank you for reviewing and for adding such a comprehensive test case. Using `dag.test()` in the integration test (https://github.com/astronomer/astronomer-cosmos/pull/1286) is an excellent way to reproduce the pickling issue in a controlled environment.
Thanks to your help and the community's support, I'm glad the fix for the bug I introduced was properly verified. Really appreciate your thorough review!

PR #1275, opened 2024-10-22
Update issue template bug.yml - cosmos version update in the dropdown

pankajkoti:
Housekeeping to use latest released version in the bug report template to select cosmos version.
Cosmos's latest released version in 1.7.0 and hence use the same as the first
dropdown item version that appears to the bug reporters.

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-bug.yml
+++ b/.github/ISSUE_TEMPLATE/01-bug.yml
@@ -15,7 +15,7 @@ body:
       description: What version of Astronomer Cosmos are you running? If you do not see your version in the list, please (ideally) test on
         the latest release or main to see if the issue is fixed before reporting it.
       options:
-        - "1.4.1"
+        - "1.7.0"
         - "main (development)"
         - "Other Astronomer Cosmos version (please specify below)"
       multiple: false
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/01-bug.yml
```diff
@@ -15,7 +15,7 @@ body:
       description: What version of Astronomer Cosmos are you running? If you do not see your version in the list, please (ideally) test on
         the latest release or main to see if the issue is fixed before reporting it.
       options:
-        - "1.4.1"
+        - "1.7.0"
```
tatiana, line 18: It is probably worth adding other versions as well, WDYT?
Would there be a more automatic way of checking the repo versions, as opposed to hard-coding these?

pankajkoti: yes, good question.

I checked and could not seem to find an automated way for it to do within GitHub itself. Perhaps we could add an automation ourselves that takes care of updating the latest version in the dropdown here by checking the latest release, but that would mean the automation would have to create a PR. I am not sure how worth it is at this point of time to invest into building that automation. So perhaps, good to keep it open text to cosmos version instead of providing a dropdown. I pushed such a change in https://github.com/astronomer/astronomer-cosmos/pull/1275/commits/0471b16f97a46858b5c0abe60b4b5d43f509ffc9

Issue #1277, opened 2024-10-23
[Feature] To support task display_name


t0momi219:
### Description

When running models that have names containing multibyte characters, runtime errors occur in Airflow environments where statsd is enabled (MWAA uses this statsd metric for collecting metrics in Cloudwatch).

Related Issue: https://github.com/apache/airflow/issues/18010

To address this, Airflow 2.9 introduced the ability to render tasks using display_name, which allows task names to be rendered separately from their task_id.

Reference: https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/example_dags/example_display_name.html

It would be ideal to have a mechanism that allows users to set a different task_id while using the model name as the display_name when working with models that contain multibyte characters.

### Use case/motivation

The following error occurs in MWAA environment:

```bash
[2024-10-17, 10:27:03 UTC] {taskinstance.py:2865} INFO - Starting attempt 1 of 1
[2024-10-17, 10:27:03 UTC] {validators.py:135} ERROR - Invalid stat name: dag.run_cosmos_dbt.日本語名モデル_run.queued_duration.
Traceback (most recent call last):
  File "/usr/local/airflow/.local/lib/python3.11/site-packages/airflow/metrics/validators.py", line 132, in wrapper
    stat = handler_stat_name_func(stat)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/airflow/.local/lib/python3.11/site-packages/airflow/metrics/validators.py", line 220, in stat_name_default_handler
    raise InvalidStatsNameException(
airflow.exceptions.InvalidStatsNameException: The stat name (dag.run_cosmos_dbt.日本語名モデル_run.queued_duration) has to be composed of ASCII alphabets, numbers, or the underscore, dot, or dash characters.
```

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

PR #1278, opened 2024-10-23
Support rendering tasks with non-ASCII characters

t0momi219:
## Description

When running models that have names containing multibyte characters, runtime errors occur in Airflow environments where statsd is enabled (e.g., MWAA uses this statsd metric for collecting metrics in Cloudwatch).

Related Issue: https://github.com/apache/airflow/issues/18010

To address this, Airflow 2.9 introduced the ability to render tasks using display_name, which allows task names to be rendered separately from their task_id.

Reference: https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/example_dags/example_display_name.html

This PR adds support for display_name, enabling users who use non-ASCII characters as their native language to display task names in their own language, even in environments like MWAA.

### Details
The normalize_task_id parameter is added to RenderConfig.
This option accepts a function to generate a task ID from a node. This allows users to generate arbitrary task IDs from models. If a function is passed to this option, Cosmos will use the model name as the display_name for tasks while rendering them.

```python
def normalize_task_id(node):
    """
    This function takes a node and returns a new task_id.
    """
    if node.name == "ＭＵＬＴＩＢＹＴＥ＿ＭＯＤＥＬ＿ＮＡＭＥ":
        return "MULTIBYTE_MODEL_NAME"

render_config = RenderConfig(
    normalize_task_id=normalize_task_id
)
```

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/1277

## Checklist


t0momi219:
Hi team, ( @tatiana @pankajkoti )
This PR is ready. Could you please review this?

tatiana (commented):
HI @t0momi219, thank you very much for the detailed explanation of the problem and for proposing a fix.

I was surprised with the amount of lines changed to fix the issue. It feels like you tried to do two things at once: fix the problem while refactoring the code. Would it be possible to solve the bug with less changes to the code?

As an example, what if we:
1. Introduced a function (e.g. `normalize_node_name` that takes a `node_name`, normalizes it, handling non-ASCII characters)
2. Replaced the ocurrences of `node.name` by  `normalize_node_name(node.name)`

cosmos/airflow/graph.py
```diff
@@ -155,57 +176,64 @@ def create_task_metadata(
...
-            id=task_id,
-            owner=node.owner,
-            operator_class=calculate_operator_class(
-                execution_mode=execution_mode, dbt_class=dbt_resource_to_class[node.resource_type]
-            ),
-            arguments=args,
-            extra_context=extra_context,
-        )
-        return task_metadata
-    else:
```
tatiana, line 201: This is a critical part of the code, and I feel that the bug fix does not justify changing all these lines.

cosmos/config.py
```diff
@@ -62,6 +62,8 @@ class RenderConfig:
     :param dbt_ls_path: Configures the location of an output of ``dbt ls``. Required when using ``load_method=LoadMode.DBT_LS_FILE``.
     :param enable_mock_profile: Allows to enable/disable mocking profile. Enabled by default. Mock profiles are useful for parsing Cosmos DAGs in the CI, but should be disabled to benefit from partial parsing (since Cosmos 1.4).
     :param source_rendering_behavior: Determines how source nodes are rendered when using cosmos default source node rendering (ALL, NONE, WITH_TESTS_OR_FRESHNESS). Defaults to "NONE" (since Cosmos 1.6).
+    :param airflow_vars_to_purge_dbt_ls_cache: Specify Airflow variables that will affect the LoadMode.DBT_LS cache.
+    :param set_task_id_by_node: A callable that takes a dbt node as input and returns the task ID. This allows users to assign a custom node ID separate from the display name.
```
tatiana, line 66: We received feedback from end-users that Cosmos already has too many configurations. Would it be possible for us to handle non-ASCII following how other libraries handle this, without enforcing users to define a new configuration?

docs/configuration/task-display-name.rst
```diff
@@ -0,0 +1,30 @@
...
+
+You can provide a function to convert the model name to an ASCII-compatible format. The function’s output is used as the TaskID, while the display name on Airflow remains as the original model name.
+
+.. code-block:: python
+
+    from slugify import slugify
+
+
+    def set_task_id_by_node(node):
+        return slugify(node.name)
```
tatiana, line 23: Why not to do this on behalf of the users?

cosmos/airflow/graph.py
t0momi219: Thank you for reviewing this over the weekend. I apologize for the lack of clarity in my pull request.

The purpose of these changes is as follows:

1. Adding a parameter to enable the use of display_name
2. Addressing Ruff errors that occurred as a result of point 1
3. Updating the documentation (I also noticed that there was some missing information about "source_rendering_behavior" and have added that as well).

cosmos/config.py
t0momi219: The reason for not providing users with an option to specify task_id themselves is that automatically generating task_id for names written in non-ASCII characters is a highly challenging task.

For example, while slugify (as mentioned in the documentation) can work in some cases, it’s not suitable for use in actual production code.

Examples:

- slugify converts names based on pronunciation, which makes it difficult to keep Task IDs unique due to homophones. In Japanese, for instance, "accounting" and "finance" are represented by the same word.
- Chinese and Japanese both use similar kanji characters, but these characters have different pronunciations in each language.

In other words, what’s needed is "translation" rather than a mechanical "conversion" like slugify for the sake of automation.

To ensure unique task IDs, some form of mapping would likely be necessary. 
For example: `{ "顧客": "customers", "注文": "orders" }`.

t0momi219: I understand the concern about having too many options for the end users. However, I believe it would be challenging to automatically generate task_id on behalf of the users.

May I please propose adding an option for this purpose? If that’s acceptable, I’d like to work on a revision that minimizes code changes.

docs/configuration/task-display-name.rst
t0momi219: As I mentioned in my previous comment.

cosmos/config.py
tatiana: Hi @t0momi219, Thanks a lot for explaining these differences. I don't speak Chinese nor Japanese, so your explanation was extremely helpful. I now understand the need

Yes, please. It would be amazing if you could make changes to make this customizable without changing the codebase.

pankajkoti: yes, given the explanation I'm inclined to provide & expose it as a configuration. Like that someone who really needs to use it can use.

Also +1 to minimising the code change, easier to review & address the scope of a particular issue. Thanks for working on this @t0momi219

t0momi219:
Hi @tatiana @pankajkoti ,
Thank you for the review, and I appreciate your acceptance of this proposal. I have revised the implementation to avoid impacting the core logic as much as possible.

### PR Changes
- Renamed the added parameter to `normalize_task_id`.
- Modified the code with minimal changes.
- Updated the documentation.

t0momi219:
Hi @tatiana, @pankajkoti 

I am also working on the following PR, and I plan to modify the same generate_task_or_group function in both: https://github.com/astronomer/astronomer-cosmos/pull/1317

Therefore, I’d like to complete this PR first before moving on to the next PR’s changes. If possible, could you please review this PR again? I believe I’ve completed the code adjustments. If there are still any issues, please let me know. Thank you, and I apologize if further modifications are needed.
(Also, I believe that some of the previously failing tests in this PR have now passed thanks to everyone’s improvements. I’m grateful because I wasn’t able to fix it on my own. Thank you.)

pankajastro (commented):
Hey @t0momi219, Could you please rebase it.

cosmos/airflow/graph.py
```diff
@@ -132,7 +132,28 @@ def create_test_task_metadata(
...
+        task_id = resource_suffix
+    elif normalize_task_id:
+        task_id = normalize_task_id(node)
+        args_update["task_display_name"] = f"{node.name}_{resource_suffix}"
+    else:
+        task_id = f"{node.name}_{resource_suffix}"
+    return task_id, args_update
+
+
+  def create_dbt_resource_to_class(test_behavior: TestBehavior) -> dict[str, str]:
```
pankajastro, line 156:
```suggestion
def create_dbt_resource_to_class(test_behavior: TestBehavior) -> dict[str, str]:
```

pankajastro (approved):

Issue #1131, opened 2024-07-31
[Bug] User unable to see the "dbt docs" menu item in Astro


tatiana:
### Astronomer Cosmos Version

Other Astronomer Cosmos version (please specify below)

### If "Other Astronomer Cosmos version" selected, which one?

1.5.1

### dbt-core version

1.7.2

### Versions of dbt adapters

dbt-snowflake==1.7.3
apache-airflow-providers-snowflake==5.1.1
apache-airflow-providers-microsoft-azure==8.2.0
apache-airflow-providers-sendgrid==3.5.0
requests-futures==1.0.1
requests-toolbelt==1.0.0
polars[xlsx2csv]==0.18.15
azure-storage-blob==12.17.0
simple-salesforce==1.12.5
beautifulsoup4==4.12.2

### LoadMode

AUTOMATIC

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

apache-airflow==2.9.1 (astro-runtime:11.3.00

### Operating System

Any

### If a you think it's an UI issue, what browsers are you seeing the problem on?

Chrome

### Deployment

Astronomer

### What happened?

The user has successfully created the dbt docs and stored it in Azure Blob storage by using `DbtDocsAzureStorageOperator`.

They successfully configured the Cosmos plugin to render the docs, by using:

```
AIRFLOW__COSMOS__DBT_DOCS_DIR=wasb://user-bucket/dbt-docs
AIRFLOW__COSMOS__DBT_DOCS_CONN_ID=azure_conn
```

They are not able to see the menu item in the Airflow UI:
![image003](https://github.com/user-attachments/assets/f31c743f-5ea8-49e9-965d-a54fb45f2950)

However, if they access the URL directly, they can see the dbt docs rendered in Airflow:
http://airflow-host/cosmos/dbt_docs

![image001](https://github.com/user-attachments/assets/e9c21965-60eb-4b42-a767-f76a76e9b640)

### Relevant log output

```shell
Apparently, there are no errors in the webserver logs. The only issue is that the menu item is not showing up.
```

### How to reproduce

We have yet to reproduce this problem.
We're currently suspecting this is a permission issue.

### Anything else :)?

Internal slack thread where the problem was discussed: https://astronomer.slack.com/archives/C04EXKRQFD3/p1721637533315019

We'll likely have to add set additional level of permission in  https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/plugin/__init__.py

[...]

PR #1280, opened 2024-10-23
Fix displaying dbt docs as menu item in Astro

tatiana:
As of Cosmos 1.7.0, Astro customers who have the role of "Organization Member" or "Workspace Operator" are not able to see the menu item "dbt Docs" under "Browser". They could access the `/cosmos/dbt_docs` view directly, though. This PR solves this issue.

Closes: #1131 

# About the solution

This PR follows the Astro [documentation](https://www.astronomer.io/docs/learn/using-airflow-plugins#flask-blueprints-and-appbuilder-views) recommendation:
> If you want to use custom menu items in an Airflow environment hosted on Astro, you must give your plugin the necessary permissions. To do this, use the @has_access decorator to give your BaseView method ACTION_CAN_ACCESS_MENU permissions.

By adding:
```
    @has_access([(permissions.ACTION_CAN_ACCESS_MENU, "Custom Menu")]) 
```

# How the fix was tested

In addition to the unit test that checks if all Cosmos plugin endpoints have the expected permissions, I also validated this change with a user with the "Organization Member" role. I confirmed that they could see the menu item and that it redirects to the correct page:

Screenshot 2024-10-24 at 10 24 03


pankajkoti (approved):
The Airflow docstring suggests that the `has_access` decorator is deprecated and we should use alternate `has_access_*` (probably `has_access_view` is what we might need) decorators. We might as well need update to the Astro docs(?)

Screenshot 2024-10-24 at 11 11 58 PM

However, since this has been tested with Astro & it seems to be working, I guess we're safe.

Great find though, thanks for fixing this & sharing the Astro docs for reference 👏🏽

tatiana:
Thanks a lot for the fast review, @pankajkoti!  

> The Airflow docstring suggests that the `has_access` decorator is deprecated and we should use alternate `has_access_*` (probably `has_access_view` is what we might need) decorators. We might as well need update to the Astro docs(?)

I also noticed this but was slightly worried about refactoring this while trying to fix the current bug.

It was a significant interface change. While `has_access` receives a list of tuples (permission & resource) as an argument, the new interface (e.g. `has_access_view`) expects an `AccessView` item. It is unclear if `has_access_view` would be enough to grant menu item permissions:
https://github.com/apache/airflow/blob/6070bb6c354908f6b233f3d92086476d0f62ada0/airflow/www/auth.py#L324

Lastly, I wonder how much of this may change with Airflow 3. I know there are intentions of redefining plugins, so I'm not sure if we should spend time on this now or just focus on the future version once AF3 alphas are released. WDYT?

> However, since this has been tested with Astro and seems to be working, I guess we're safe.

I also asked the CRE team to confirm it works in their tests so that we can have more confidence it now works as expected in Astro Cloud. I will merge this PR, but I suggest we wait for their confirmation before releasing 1.7.1.

pankajkoti:
> Lastly, I wonder how much of this may change with Airflow 3. I know there are intentions of redefining plugins, so I'm not sure if we should spend time on this now or just focus on the future

I fully agree. IMO, it would be wise if we invest little into trying to support more in what we may know is likely to change. So the bare minimum and working fixes is what I would also prefer specifically when it comes to plugins ,🙂

PR #1282, opened 2024-10-24
Fix release to PyPI after `raw` rst directive disabled

tatiana:
We started facing the following error while trying to publish packages to PyPI:

```
ERROR    `long_description` has syntax errors in markup and would not be
         rendered on PyPI.
         line 88: Warning: "raw" directive disabled.
Checking dist/astronomer_cosmos-1.7.1a1.tar.gz: FAILED
```

As seen in https://github.com/astronomer/astronomer-cosmos/actions/runs/11485503895/job/31965756381

I was able to reproduce the Github action error by running:

```
rm -fr dist && python -m build --wheel && twine check dist/*
```

This PR aims to solve the issue without side effects on Scarf metrics.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,7 @@ name: Deploy to PyPi
 on:
   release:
     types: [published]
+    branches: [fix-release]
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -3,6 +3,7 @@ name: Deploy to PyPi
 on:
   release:
     types: [published]
+    branches: [fix-release]
```
tatiana, line 6:
```suggestion
```

pankajkoti (approved):
Nice fix. It's sad how GitHub actions add warnings but essentially push breaking changes & not following semver.

tatiana:
I agree, @pankajkoti . Another observation is that the reStructuredText `img` directive does not support the `referrerpolicy` property. Time will tell how much this will affect Scarf metrics.

PR #1283, opened 2024-10-24
Update the URL for sample dbt docs hosted in Astronomer S3 bucket

pankajkoti:
It was observed by @tatiana that the previous URL 
http://cosmos-docs.s3-website-us-east-1.amazonaws.com/
is no longer functioning and she hosted the docs in an Astronomer S3
bucket available at http://cosmos-demo-dbt-docs.s3-website.eu-north-1.amazonaws.com/

Hence, this PR is quick follow-up to update the references for the newly
hosted docs.

### DIFF
--- a/dev/docker-compose.yaml
+++ b/dev/docker-compose.yaml
@@ -10,7 +10,7 @@ x-airflow-common:
   environment:
     &airflow-common-env
     DB_BACKEND: postgres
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Thanks for the fix, @pankajkoti !

Issue #1253, opened 2024-10-12
[Feature] Store manifest.json state file after each execution


ismailsimsek:
### Description

is it possible to store manifest.json state file after each dbt execution? 

similar to https://github.com/astronomer/astronomer-cosmos/blob/2dd8c7a0dd44ce436b62c3907806d4240fca41f7/cosmos/operators/local.py#L353C5-L353C17

### Use case/motivation

motivation will be using it as a state file and avoiding  seed reloads. 

### Related issues

https://github.com/dbt-labs/dbt-core/issues/7762

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

tatiana:
Hi @ismailsimsek ,
We believe you could accomplish storing the `manifest.json` to a remote object store via the Cosmos callback feature:
https://astronomer.github.io/astronomer-cosmos/configuration/callbacks.html
We are therefore closing this ticket; please feel free to create a new one if you believe anything is missing.

PR #1286, opened 2024-10-29
Minor refactor on `VirtualenvOperators` & add test for PR #1253 & 

tatiana:
Cosmos virtualenv operators are using the system dbt instead of the virtualenv dbt.

Create a test case that illustrates issue #1246. This test fails with Cosmos 1.7 (and the current main branch) and passes when using PR #1252.

This PR also introduces two refactors:
- Reuse the parent class method where applicable, as opposed to re-writing it completely
- Force the Virtualenv invocation mode to be `SUBPROCESS ` since Airflow/Cosmos are not able to import dbt as a library if it is not part of the same Python virtualenv

### DIFF
--- a/tests/operators/test_virtualenv.py
+++ b/tests/operators/test_virtualenv.py
@@ -22,6 +22,7 @@ AIRFLOW_VERSION = Version(airflow.__version__)
 
 DBT_PROJ_DIR = Path(__file__).parent.parent.parent / "dev/dags/dbt/jaffle_shop"
 
+DAGS_FOLDER = Path(__file__).parent.parent.parent / "dev/dags/"
 
 profile_config = ProfileConfig(
     profile_name="default",
@@ -368,7 +369,7 @@ def test_integration_virtualenv_operator(caplog):
     """
     from airflow.models.dagbag import DagBag
 
-    dag_bag = DagBag(dag_folder=None, include_examples=False)
+    dag_bag = DagBag(dag_folder=DAGS_FOLDER, include_examples=False)
     dag = dag_bag.get_dag("example_virtualenv_mini")
 
     dag.test()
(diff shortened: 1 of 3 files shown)


tests/operators/test_virtualenv.py
```diff
@@ -339,3 +356,23 @@ def test__release_venv_lock_current_process(tmpdir):
...
+    Confirm we're using the correct dbt command to run with virtualenv.
+    """
+    from airflow.models.dagbag import DagBag
+
+    dag_bag = DagBag(dag_folder=DAGS_FOLDER, include_examples=False)
+    dag = dag_bag.get_dag("example_virtualenv_mini")
+
+    dag.test()
+
+    assert "Trying to run the command:\n ['/tmp/persistent-venv2/bin/dbt', 'deps'" in caplog.text
```
pankajkoti, line 377: nice testing strategy, thanks for showing this

pankajkoti (approved):

tests/operators/test_virtualenv.py
tatiana: @pankajkoti, my initial attempts involved mocking `run_command`, `run_subprocess`, and even `FullOutputSubprocessResult`. Unfortunately, as described in the original PR (https://github.com/astronomer/astronomer-cosmos/pull/1252#issue-2582912160), all these things behaved very well when we tried to run the DAG using `from tests.utils import test_dag as run_test_dag`, because the issue is a bit deeper in how Airflow pickles/runs things.

Since I could reproduce by running `dags test`, I decided to give the built-in Airflow `dag.test` a try, and I was glad it worked. It was a bit annoying to find a way to test this, but I'm happy we won't have any regressions.

Issue #1129, opened 2024-07-31
Separate out a job in CI that runs integration test on the latest combination of versions


pankajkoti:
Would be nice to see if we can implement separating out a job in the CI that runs integration test on the latest combination of versions so that we can reduce the friction for community contributors to see the integration tests run upfront than having to wait for authorizing CI runs from maintainers. Avoid running expensive tests as part of this job.

pankajkoti:
@dosu not stale

pankajkoti:
consolidating this into https://github.com/astronomer/astronomer-cosmos/issues/1021 and hence, closing

PR #1288, opened 2024-10-30
Add stalebot to label/close stale PRs and issues

tatiana:
As Cosmos grows in popularity, it is important that we automate a few actions:
- labelling issues as stale
- labelling PRs as stale
- closing stale PRs

The goal of this PR is to accomplish this. As a starting point, my suggestion is that we don't mark issues as stale, but I'm open to changing this.

An example of a successful run:
https://github.com/astronomer/astronomer-cosmos/actions/runs/11590467421/job/32268097259

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/stale@v9
         with:
           days-before-issue-stale: 30
-          days-before-issue-close: -1
+          days-before-issue-close: -1  # Never closes issues, more information in https://github.com/actions/stale/
           stale-issue-label: "stale"
           stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
           days-before-pr-stale: 30
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -19,7 +19,8 @@ jobs:
           stale-issue-message: "This issue is stale because it has been open for 90 days with no activity."
           days-before-pr-stale: 60
           days-before-pr-close: 120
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,25 @@
+name: Handle stale issues and PRs
+on:
+  schedule:
+    - cron: "0 11 * * *"
+  push:
+    branches: [stale-bot]
```
tatiana, line 6:
```suggestion
```

pankajkoti: I am guessing https://github.com/astronomer/astronomer-cosmos/issues/1129#issuecomment-2446333216 was result of the push to this branch?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,25 @@
...
+  handle-stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-issue-stale: 90
+          days-before-issue-close: -1
```
pankajkoti, line 17: Just thinking if we should add a link to the stale Action somewhere in this yml for users to quickly check the guide for it or write a comment here that -1 means they are not closed automatically. 

Also, okay if we don't take any action, just sharing what I had a thought of.

pankajkoti (approved):
LGTM. Thanks for adding this! Would definitely help automatically close stale PRs

.github/workflows/stale.yml
tatiana: Thanks, @pankajkoti , addressed in 03eb12c75f666f7e625b08d81ad9b0aee708a9b2

PR #1295, opened 2024-10-31
Fix caching unique id error

tatiana:
Workaround for issues faced by customers in the production.
    
An Astronomer customer using Cosmos 1.5.0 with caching enabled in production faces failures in their DAGs once they change those DAGs or dbt projects using `astro deply --dags`. DAGs that were not changed continue to work.
    
They are using cache based on Airflow Variables.
    
The error they are facing is:

```    
    File /usr/local/lib/python3.11/site-packages/cosmos/dbt/graph.py, line 135, in parse_dbt_ls_output
        unique_id=node_dict[unique_id]
    KeyError: 'unique_id'
```
    
This change aims to add logs and skip nodes that may not contain the necessary fields. This will allow us to identify the root cause.


tatiana:
We identified what was the problem. There was a macro that would print JSONs in dbt ls that were not meant to be handled as dbt nodes.

This was the dbt ls output, before printing the actual nodes:
```
11:20:43  Running with dbt=1.7.6
11:20:45  Registered adapter: bigquery=1.7.2
11:20:45  Unable to do partial parsing because saved manifest not found. Starting full parse.
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
/***************************/ 
Values returned by mac_get_values: 
{}
/***************************/
11:20:55  [WARNING]: Configuration paths exist in your dbt_project.yml file which do not apply to any resources.
11:20:55  Found 85 models, 28 tests, 1 seed, 21 sources, 0 exposures, 0 metrics, 911 macros, 0 groups, 0 semantic models
```

We have two solutions for the customer who faced this issue:
1. Modify the macro print statement to be one-line (e.g. `'Values returned by mac_get_values: {}'`)
2. To use cosmos 1.5.0rc2, which is more resilient to this use-case, and simply skips lines that do not have the expected dbt node keys

In the meantime, I'm going to close this PR and create a new PR against Cosmos main branch, and add a test-case for this.

PR #1297, opened 2024-10-31
Fix unit and integration tests for Python 3.8 & Airflow 2.6

tatiana:
The CI is failing for both unit and integraiton tests for Airflow 2.6 because it is taking a long time to try to resolve dependencies.

Example: https://github.com/astronomer/astronomer-cosmos/actions/runs/11594745046/job/32330864858

This PR aims to resolve the issue.


scripts/test/pre-install-airflow.sh
```diff
@@ -18,19 +21,40 @@ curl -sSL $CONSTRAINT_URL -o /tmp/constraint.txt
...
-uv pip install pydantic --constraint /tmp/constraint.txt
+uv pip install pip --upgrade
 
+uv pip install "apache-airflow==$AIRFLOW_VERSION" --constraint /tmp/constraint.txt
+uv pip install apache-airflow-providers-docker --constraint /tmp/constraint.txt
+uv pip install apache-airflow-providers-postgres --constraint /tmp/constraint.txt
 
-if [ "$AIRFLOW_VERSION" = "2.7" ]; then
+if [ "$AIRFLOW_VERSION" = "2.4" ] || [ "$AIRFLOW_VERSION" = "2.5" ] || [ "$AIRFLOW_VERSION" = "2.6" ]  ; then
+  uv pip install "apache-airflow-providers-amazon[s3fs]" "apache-airflow==$AIRFLOW_VERSION" "urllib3<2"
```
tatiana, line 34: We may not need "s3fs" here, since it was introduced in 2.8

scripts/test/pre-install-airflow.sh
```diff
@@ -18,19 +21,40 @@ curl -sSL $CONSTRAINT_URL -o /tmp/constraint.txt
...
+  uv pip install "apache-airflow-providers-cncf-kubernetes" "apache-airflow==$AIRFLOW_VERSION"
+  uv pip install  "apache-airflow-providers-google<10.11" "apache-airflow==$AIRFLOW_VERSION"
+  uv pip install "apache-airflow-providers-microsoft-azure" "apache-airflow==$AIRFLOW_VERSION"
+  uv pip install pyopenssl --upgrade
+elif [ "$AIRFLOW_VERSION" = "2.7" ] ; then
   uv pip install "apache-airflow-providers-amazon[s3fs]" --constraint /tmp/constraint.txt
```
tatiana, line 40: s3fs is not needed here

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, fix-ci-for-tests.py3.8-2.6]
```
```suggestion
    branches: [main]
```
gentle reminder comment to remove this before merging

cosmos/converter.py
```diff
@@ -181,6 +181,11 @@ def validate_adapted_user_config(
             "RenderConfig.dbt_project_path is required for rendering an airflow DAG from a DBT Graph if no manifest is provided."
         )
 
+    if render_config and render_config.project_path:
+        # Project config is further expected to have project name
+        # even if it had no manifest
+        project_config.project_name = render_config.project_path.stem
```
pankajkoti, line 187: Should we also check here that `project_config.project_name` is NOT already set and only then assign it from `render_config`?

pankajkoti (approved):
These are amazing fixes and great manoeuvring. Certainly this has a massive operational impact & contributors are going to the love the speedy CI now. Kudos to you @tatiana  for being so patient with the acting CI, not giving up & taking it to completion 👏🏽

cosmos/converter.py
tatiana: They are mutually exclusive... Cosmos errors if both of them are set. I can add a comment to this!

tatiana: Created a separate PR with this change, since it solves another underlining issue:
https://github.com/astronomer/astronomer-cosmos/pull/1307

Adding a proper test case for it

PR #1302, opened 2024-11-05
Enable secrect detection precommit check

pankajastro:
related: https://github.com/astronomer/oss-integrations-private/issues/28


pankajkoti (approved):
Thanks for adding this!

Issue #1291, opened 2024-10-30
Update the remote_target_path config docs to reflect what all files are uploaded to remote


pankajkoti:
It’s currently observed that users may assume, after reading the [documentation](https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html#remote-target-path), that all files from the `target` directory of the dbt project are uploaded to the remote target directory. However, this isn’t yet the case. The `remote_target_path` configuration, introduced in PR https://github.com/astronomer/astronomer-cosmos/pull/1224, currently only uploads files from the `compiled` directory within the `target` directory of the dbt project—and only when the execution mode is set to `ExecutionMode.AIRFLOW_ASYNC`. We need to update the documentation for this configuration to clarify this behavior and indicate that additional files will become available as we expand support for uploading them within the operators.

Reference where users assumed https://apache-airflow.slack.com/archives/C059CC42E9W/p1730206906312459. 
The question in that Slack conversation said
```
Hi, folks! I'm trying to configure [remote_target_path](https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html#remote-target-path) but I doesn't send any file to the s3 location that I've set. What else is needed to configure it?
```

**Acceptance Criteria**
- [ ] Docs says what all target directory files are currently uploaded in the remote target path
- [ ] Mention the future work in the docs for supporting more files to be uploaded to remote target path

Issue #1293, opened 2024-10-30
Support uploading more files from the target directory to remote_target_path


pankajkoti:
Currently, the `remote_target_path` [configuration](https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html#remote-target-path), added in PR #1224, only uploads files from the `compiled` directory within the `target` directory of the dbt project—and solely when `ExecutionMode.AIRFLOW_ASYNC` is enabled. However, the `target` directory contains several other files and folders that could benefit users if they were also uploaded to `remote_target_path`.

Beyond the `compiled` directory, the target directory typically includes:

	1.	run/ folder
	2.	graph.gpickle
	3.	graph_summary.json
	4.	manifest.json
	5.	partial_parse.msgpack
	6.	run_results.json
	7.	semantic_manifest.json

A specific request was made in a [Slack conversation](https://apache-airflow.slack.com/archives/C059CC42E9W/p1730206906312459) to have `run_results.json` uploaded and accessible in `remote_target_path`, highlighting its value to users.

We should evaluate the potential benefits of supporting uploads for these additional files and folders and explore enabling this feature across all execution modes, not just `ExecutionMode.AIRFLOW_ASYNC`. Additionally, it may be worthwhile to consider uploading files from the `compiled` directory in other execution modes if it proves beneficial.

We could potentially create sub tasks for each of these files & folders for evaluation of the benefits & supporting to upload those to the `remote_target_path`

joppedh:
@pankajkoti partial_parse would speed-up the runtime. Now on each task run it still needs to parse, even when providing a manifest.json
```
[2024-10-31, 06:03:48 UTC] {logging_mixin.py:188} INFO - 06:03:48  Unable to do partial parsing because saved manifest not found. Starting full parse.
```

tatiana:
@joppedh, on the partial parsing side, have you been able to leverage https://astronomer.github.io/astronomer-cosmos/configuration/partial-parsing.html#partial-parsing? Cosmos should be caching it. But you'd need to leverage `render_config=RenderConfig(enable_mock_profile=False)`.

tatiana:
We introduced support for this via Cosmos callbacks and by declaring standard callbacks in Cosmos that upload all content from a dbt target folder to a remote object store. These feature was significantly improved in Cosmos 1.8.0.

Documentation:
https://astronomer.github.io/astronomer-cosmos/guides/run_dbt/callbacks/callbacks.html

[...]

PR #1305, opened 2024-11-06
[Docs] Add limitation about copying target dir files to remote

pankajkoti:
It’s currently observed that users may assume, after reading the [documentation](https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html#remote-target-path), that all files from the target directory of the dbt project are uploaded to the remote target directory. However, this isn’t yet the case. The `remote_target_path` configuration, introduced in PR https://github.com/astronomer/astronomer-cosmos/pull/1224, currently only uploads files from the compiled directory within the target directory of the dbt project—and only when the execution mode is set to `ExecutionMode.AIRFLOW_ASYNC`. This PR updates the documentation for this configuration to clarify this behavior and indicate that additional files will become available as we expand support for uploading them within the operators.

closes: #1291


pankajkoti:
preview: https://limitation-remote-target-pat.astronomer-cosmos.pages.dev/configuration/cosmos-conf/#remote-target-path

tatiana (commented):
Thanks for updating the docs, @pankajkoti !

It's good that we have a follow up ticket to improve this further: https://github.com/astronomer/astronomer-cosmos/issues/1293

tatiana (approved):

Issue #1309, opened 2024-11-07
[Bug]  after upgrading from 1.6 to 1.7.1, docs does not open in airflow


corsettigyg:
### Astronomer Cosmos Version

1.7.1

### dbt-core version

1.8.1

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

LOCAL

### InvocationMode

SUBPROCESS

### airflow version

2.9.0

### Operating System

linux

### Deployment

Official Apache Airflow Helm Chart

### What happened?

After upgrading cosmos to its latest version, the docs started asking for the user & password. when I provide the default `admin`, I get blocked from the docs

![Image](https://github.com/user-attachments/assets/63b2e546-d898-45ee-8540-5b914da34d18)

### How to reproduce

1. Deploy airflow with dbt-cosmos installed on docker
2. try to access the docs after setting it up in the config

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

tatiana:
Thanks for reporting this bug, @corsettigyg , and I'm sorry for introducing it in the first place.

I hope this will resolve it:
https://github.com/astronomer/astronomer-cosmos/pull/1312

PR #1312, opened 2024-11-11
Fix access to the `dbt docs` menu item outside of Astro cloud

tatiana:
Since Cosmos 1.7.1, users who do not use Astro Cloud started facing 'Access is Denied' when accessing the dbt docs menu item.

This was an unintended side effect of the fix for Astro users to be able to see that menu item, introduced in #1280

Closes: #1309


pankajkoti (approved):
LGTM. Nice fix.

Issue #1316, opened 2024-11-12
[Feature] Task timeout and SLA support


t0momi219:
### Description

In Airflow, both DAGs and tasks can have timeout and SLA times specified. Since dbt models likely have varying expected execution times for each layer, there could be cases where users want to apply SLAs individually to each node.

How about having Cosmos retrieve timeout and SLA times from the node metadata and apply them individually when rendering nodes?

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

pankajkoti:
hi @t0momi219 . Sounds like a great idea. I am definitely up for the idea of timeouts. When it comes to SLA, I'm hearing that they might take away the concept of SLAs in Airflow 3 and come up with something better in later releases of AF 3. So, if you're willing, definitely would encourage the contribution for supporting timeout from node metadata and applying it to AF tasks.

PR #1317, opened 2024-11-12
Add task timeout support.

t0momi219:
## Description

In Airflow, both DAGs and tasks can have timeout specified. Since dbt models likely have varying expected execution times for each layer, there could be cases where users want to apply timeout individually to each node.

How about having Cosmos retrieve timeout from the node metadata and apply them individually when rendering nodes?

Specifically, the expected time is specified in the model's config, which will be read accordingly.

```yml
version: 2
models:
  - name: stg_customers
    config:
      cosmos_task_timeout: 600 # Specify in seconds.
```

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/1316

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-from typing import Any, Callable, Union
-
 from datetime import timedelta
+from typing import Any, Callable, Union
 
 from airflow.models import BaseOperator
 from airflow.models.dag import DAG
@@ -246,7 +245,7 @@ def generate_task_or_group(
         use_task_group=use_task_group,
         source_rendering_behavior=source_rendering_behavior,
         model_timeout=model_timeout,
-        model_sla=model_sla
+        model_sla=model_sla,
     )
 
     # In most cases, we'll  map one DBT node to one Airflow task
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 from typing import Any, Callable, Union
 
+from datetime import timedelta
+
 from airflow.models import BaseOperator
 from airflow.models.dag import DAG
 from airflow.utils.task_group import TaskGroup
@@ -135,6 +137,8 @@ def create_task_metadata(
     dbt_dag_task_group_identifier: str,
     use_task_group: bool = False,
     source_rendering_behavior: SourceRenderingBehavior = SourceRenderingBehavior.NONE,
+    model_timeout: bool = False,
+    model_sla: bool = False,
 ) -> TaskMetadata | None:
     """
     Create the metadata that will be used to instantiate the Airflow Task used to run the Dbt node.
@@ -166,6 +170,11 @@ def create_task_metadata(
             task_id = f"{node.name}_run"
             if use_task_group is True:
                 task_id = "run"
+            if model_timeout and "model_timeout" in node.config.keys():
+                logger.error(f'model_timeout: {node.config["model_timeout"]} in values')
+                args["execution_timeout"] = timedelta(seconds=int(node.config["model_timeout"]))
+            if model_sla and "model_sla" in node.config.keys():
+                args["sla"] = timedelta(seconds=int(node.config["model_sla"]))
         elif node.resource_type == DbtResourceType.SOURCE:
             if (source_rendering_behavior == SourceRenderingBehavior.NONE) or (
                 source_rendering_behavior == SourceRenderingBehavior.WITH_TESTS_OR_FRESHNESS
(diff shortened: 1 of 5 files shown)


cosmos/config.py
```diff
@@ -80,6 +80,8 @@ class RenderConfig:
     enable_mock_profile: bool = True
     source_rendering_behavior: SourceRenderingBehavior = SourceRenderingBehavior.NONE
     airflow_vars_to_purge_dbt_ls_cache: list[str] = field(default_factory=list)
+    model_timeout: bool = False
+    model_sla: bool = False
```
t0momi219, line 84: This PR adds two parameters, but I’m unsure if it’s advisable to add too many parameters to Cosmos without careful consideration.

I’d like to request a review. What do you think? If you have any preferable ideas, please let me know. I’ll make adjustments accordingly and add documentation and tests.

cosmos/airflow/graph.py
```diff
@@ -166,6 +169,11 @@ def create_task_metadata(
             task_id = f"{node.name}_run"
             if use_task_group is True:
                 task_id = "run"
+            if model_timeout and "model_timeout" in node.config.keys():
+                logger.error(f'model_timeout: {node.config["model_timeout"]} in values')
+                args["execution_timeout"] = timedelta(seconds=int(node.config["model_timeout"]))
+            if model_sla and "model_sla" in node.config.keys():
+                args["sla"] = timedelta(seconds=int(node.config["model_sla"]))
```
pankajkoti, line 176: Airflow SLAs are known to have few issues. 

They mentioned in [AF 3 meeting notes](https://cwiki.apache.org/confluence/display/AIRFLOW/Airflow+3+Dev+call%3A+Meeting+Notes) that they would be dropping SLAs of AF 3 & come up with something better in later versions in AF 3+. I am thinking if we should we hold on adding SLA support here? WDYT?
> It was agreed that for now, we will mark it for removal in Airflow 3.0 and adding it back in 3.1. We will assess again in the coming dev calls if something changes. Elad has also offered to help if needed.

pankajkoti (commented):
Good idea. Can we show an example of this in one of the DAGs and potentially also mention about it somewhere in our docs?

cosmos/config.py
pankajkoti: I'm wondering of whether we need this config. Can we not directly use timeouts if they're specified & otherwise not set?

t0momi219: You’re absolutely right. Instead of adding parameters, it would have been better to set Timeout and SLA automatically only when metadata is present.

I wonder why I didn't think of it... thank you very much.

cosmos/airflow/graph.py
t0momi219: I understand well. I’ll consider SLA support to be out of scope. I wasn’t aware of the upcoming features planned for Airflow 3.0. Thank you for letting me know.

t0momi219:
Hi @pankajkoti ,
This PR is ready for review. Could you please check this?

docs/configuration/task-timeout.rst
```diff
@@ -0,0 +1,21 @@
+.. _task-timeout:
+
+Task Timeout
+================
+
+In Airflow, the "execution_timeout" parameter allows you to set the maximum runtime for a Task.
```
pankajkoti, line 6:
```suggestion
In Airflow, the ``execution_timeout`` parameter allows you to set the maximum runtime for a task.
```

docs/configuration/task-timeout.rst
```diff
@@ -0,0 +1,21 @@
...
+With Cosmos, you can specify an execution_timeout for each dbt model converted to a Task.
+This lets users set a threshold for the maximum runtime of a model, triggering a timeout error if the execution exceeds this limit.
+
+By adding cosmos_task_timeout to the config of a dbt model, Cosmos will automatically apply the specified timeout to the Task based on this value.
```
pankajkoti, line 10:
```suggestion
In Cosmos, you can apply an ``execution_timeout`` to each dbt model task by specifying a ``cosmos_task_timeout`` in the model’s configuration, which sets a runtime threshold to trigger a timeout error if exceeded.
```

pankajkoti (approved):
LGTM. some minor suggestion inline. Thanks for dropping the SLA piece. Could you please also update the PR title & description to drop SLA & additionally reflect the renamed config key `cosmos_task_timeout`?

We could merge this once these minor changes are addressed & meantime hopefully @tatiana and/or @pankajastro could also get a chance to review this PR.

docs/configuration/task-timeout.rst
t0momi219: Thank you for the review; I’ve made the necessary corrections.

tatiana:
I believe the interface for this task will change depending on the outcomes of the discussion we're having in #1339

tatiana:
Thank you very much for the contribution, @t0momi219 ! Since #1339 has a more generic implementation, it supports the same behaviour but by using a more extensible interface:
```
version: 2

models:
  - name: name
    description: description
    meta:
      cosmos:
        operator_kwargs:
          execution_timeout: abcd
```

Therefore, I'm closing this PR.

Issue #1319, opened 2024-11-13
[Bug] Fix K8s test on main brach (currently failing)


tatiana:
### Astronomer Cosmos Version

N/A

### dbt-core version

N/A

### Versions of dbt adapters

Cosmos K8s integration tests are failing in our main branch.

This is visible in new PRs, since the issue started happening after the last merge to main:
https://github.com/astronomer/astronomer-cosmos/actions/runs/11796817624/job/32867560902

### LoadMode

AUTOMATIC

### ExecutionMode

AWS_EKS

### InvocationMode

None

### airflow version

N/A

### Operating System

N/A

### Deployment

Official Apache Airflow Helm Chart

### What happened?

K8s tests stopped working in the CI

### Relevant log output

```shell

```

### How to reproduce

Create a branch from main and see the k8s tests failing in the CI

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

pankajkoti:
We merged PR #1320 that checks on the status of the postgres pod -- previously, postgres pod failed & errored as observed in https://github.com/astronomer/astronomer-cosmos/actions/runs/11796817624/job/32867560902 which led to late failure in the K8s tests. It would be nice to observe for a while & if all is well, we can close the issue

pankajkoti:
After discussing with @tatiana we agreed that we could close this ticket & re-open it if we observe any more failures.

PR #1320, opened 2024-11-13
Add postgres pod status check for k8s tests in CI

pankajkoti:
It appears we have a flaky Kubernetes test that failed in PR #1313. As shown in the error log [here](https://github.com/astronomer/astronomer-cosmos/actions/runs/11796817624/job/32867560902?pr=1313#step:7:473), the PostgreSQL pod did not reach the ready state and instead entered an error status. Since the cause of the error status is unclear, this PR introduces a status check for the PostgreSQL pod to ensure it becomes fully running and healthy. If the pod enters an ERROR state, we now run a `kubectl describe` command on the pod to capture the event logs for debugging. The test will also exit with an error code of 1 to prevent further execution.

related: #1319

### DIFF
--- a/scripts/test/kubernetes-setup.sh
+++ b/scripts/test/kubernetes-setup.sh
@@ -43,10 +43,6 @@ while true; do
     sleep 5
   fi
 done
-#while [ "$(kubectl get pod "$POD_NAME" -n default -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}')" != "True" ]; do
-#  echo "Pod $POD_NAME is not ready yet. Waiting..."
-#  sleep 5
-#done
 # Print the name of the PostgreSQL pod
 echo "$POD_NAME"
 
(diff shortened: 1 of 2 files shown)


scripts/test/kubernetes-setup.sh
```diff
@@ -26,6 +26,23 @@ kind load docker-image dbt-jaffle-shop:1.0.0
 # The output is filtered to get the first pod's name
 POD_NAME=$(kubectl get pods -n default -l app=postgres -o jsonpath='{.items[0].metadata.name}')
 
+# Wait for the PostgreSQL pod to be in the 'Running' and 'Ready' state
+echo "Waiting for PostgreSQL pod to be ready..."
+while true; do
```
tatiana, line 31: Do you think there is any risk that we'd have a job running indefinitely with this new approach?

Given our recent experience, GH Actions will timeout after 6h, so it may not be an issue.

tatiana (approved):
This is great, thanks a lot, @pankajkoti ! Crossed fingers this will solve the issue.

scripts/test/kubernetes-setup.sh
pankajkoti: ideally it should reach either to Running or Error state. In both scenarios we have a path forward. yes, and like the default is for 6h, we can explore if we should reduce the timeout in case we observe there are other states reached. WDYT?

pankajkoti: I will go ahead and merge for now to rebase PR #1313 & see how it goes. But please leave a comment if you've and I will try to address that in a subsequent PR. Also, resolving the conversation to unblock the merge :)

Issue #1315, opened 2024-11-12
[Bug] dbt ls throwing error 'mock_value' is not of type 'integer' (TrinoLDAPProfileMapping)


dwolfeu:
### Astronomer Cosmos Version

1.7.1

### dbt-core version

1.8.8

### Versions of dbt adapters

dbt-trino==1.8.3

### LoadMode

AUTOMATIC

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

2.10.2

### Operating System

Debian GNU/Linux 12 (bookworm)

### Deployment

Docker-Compose

### Deployment details

I'm using podman-compose v1.0.6, rather than docker-compose. I access my local dbt project files using a volume mount.

### What happened?

The dag import fails. Looking at the logs (see below), the error is with `dbt ls`, which is using the mock profile and complaining that one of the fields (I guess `port`?) isn't an integer:

```
Runtime Error
  Credentials in profile "cosmos_trino_test", target "foobar" invalid: 'mock_value' is not of type 'integer'
```

### Relevant log output

```shell
[2024-11-12T05:18:13.234+0000] {processor.py:186} INFO - Started process (PID=29) to work on /opt/airflow/dags/cosmos_test_dag.py
[2024-11-12T05:18:13.235+0000] {processor.py:914} INFO - Processing file /opt/airflow/dags/cosmos_test_dag.py for tasks to queue
[2024-11-12T05:18:13.237+0000] {logging_mixin.py:190} INFO - [2024-11-12T05:18:13.237+0000] {dagbag.py:588} INFO - Filling up the DagBag from /opt/airflow/dags/cosmos_test_dag.py
[2024-11-12T05:18:13.253+0000] {logging_mixin.py:190} INFO - [2024-11-12T05:18:13.253+0000] {graph.py:470} INFO - Trying to parse the dbt project using dbt ls cache cosmos_cache__my_cosmos_dag...
[2024-11-12T05:18:13.261+0000] {logging_mixin.py:190} INFO - [2024-11-12T05:18:13.261+0000] {graph.py:476} INFO - Cosmos performance: Cache miss for cosmos_cache__my_cosmos_dag
[2024-11-12T05:18:13.261+0000] {logging_mixin.py:190} INFO - [2024-11-12T05:18:13.261+0000] {graph.py:528} INFO - Trying to parse the dbt project in `/opt/airflow/dags/dbt-project` using dbt ls...
[...]

PR #1322, opened 2024-11-13
Change the mock value of port in TrinoBaseProfileMapping to an integer

dwolfeu:
## Description

The mock value of `port` in `TrinoBaseProfileMapping` has been changed to an integer.

## Related Issue(s)

Closes: #1315

## Checklist

### DIFF
--- a/cosmos/profiles/trino/base.py
+++ b/cosmos/profiles/trino/base.py
@@ -41,6 +41,12 @@ class TrinoBaseProfileMapping(BaseProfileMapping):
         # remove any null values
         return self.filter_null(profile_vars)
 
+    @property
+    def mock_profile(self) -> dict[str, Any]:
+        mock_profile = super().mock_profile
+        mock_profile["port"] = 99999
+        return mock_profile
+
     def transform_host(self, host: str) -> str:
         """Replaces http:// or https:// with nothing."""
         return host.replace("http://", "").replace("https://", "")
--- a/tests/profiles/trino/test_trino_base.py
+++ b/tests/profiles/trino/test_trino_base.py
@@ -84,3 +84,19 @@ def test_profile_args_overrides() -> None:
             "user": "my_login",
             "session_properties": {"my_property": "my_value_override"},
         }
+
+
+def test_mock_profile() -> None:
+    """
+    Tests that the mock_profile values get set correctly. A non-integer value of port will crash dbt ls.
+    """
+    profile_mapping = TrinoBaseProfileMapping(conn_id="mock_conn_id")
+    
+    assert profile_mapping.mock_profile == {
+        "type": "trino",
+        "host": "mock_value",
+        "database": "mock_value",
+        "schema": "mock_value",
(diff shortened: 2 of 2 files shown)


dwolfeu:
I tested this using the [local docker-compose setup](https://astronomer.github.io/astronomer-cosmos/contributing#using-docker-compose-for-local-development), adding 

```
apache-airflow-providers-trino
dbt-trino
```

to `dev/requirements.txt`, and the dag I included to the [bug report](https://github.com/astronomer/astronomer-cosmos/issues/1315).

pankajkoti (commented):
LGTM. Would it be possible to add a test for this that demonstrates how it would fail without this mock_profile method implementation and that with having it succeeds?

cosmos/profiles/trino/base.py
```diff
@@ -41,6 +41,12 @@ def profile(self) -> dict[str, Any]:
         # remove any null values
         return self.filter_null(profile_vars)
 
+    @property
+    def mock_profile(self) -> dict[str, Any]:
```
tatiana, line 45: @dwolfeu, thank you so much for fixing this so quickly!

We're very close to merging this change. As @pankajkoti mentioned, adding a unit test to cover this change would ensure that we don't have a regression in this issue in the future.

You could base on these lines and do a similar check for `mock_profile.`
https://github.com/astronomer/astronomer-cosmos/blob/8ec46d2f680d6d5d4e71402030873c56d001adbc/tests/profiles/trino/test_trino_base.py#L37-L45

Please let us know if you have any questions - we are here to support you to get this to completion.

dwolfeu: I have added a test. It only checks whether the `mock_profile` is set correctly though; it does not do as @pankajkoti suggested, namely testing that `dbt ls` fails without the change and passes with it. I think I would need some help in writing such a test.

Just fyi: I'm on macOS (so Darwin) and I had to change `25beeb54cc4eeabe6198248e286a1cfe` in line 1582 to `6f63493009733a7be34364a6ea3ffd3c` (= the value in line 1584) in my local repo in order to get the unit test to pass.

https://github.com/astronomer/astronomer-cosmos/blob/8ec46d2f680d6d5d4e71402030873c56d001adbc/tests/dbt/test_graph.py#L1581-L1584

tatiana: Thanks for adding the unit test and reporting this issue with the test. I believe, for now, the unit test would already avoid us to have a regression on this - so I'm happy for us to merge this PR.

I logged a separate ticket related to the `hash_dir` test:
https://github.com/astronomer/astronomer-cosmos/issues/1329

tatiana (approved):

pankajkoti:
Thanks a lot @dwolfeu for this great contribution & being so quick at contributing it as well as addressing feedback!

dwolfeu:
No worries. It was a pleasure. Thank you @pankajkoti and @tatiana for your help!

Issue #1268, opened 2024-10-21
Support `dbt clone`


tatiana:
**Context**

Cosmos 1.7 does not support `dbt clone`.

The `dbt clone` command clones selected nodes from the [specified state](https://docs.getdbt.com/reference/node-selection/syntax#establishing-state) to the target schema(s). This command makes use of the clone materialization:

* If your data platform supports zero-copy cloning of tables (Snowflake, Databricks, or BigQuery), and this model exists as a table in the source environment, dbt will create it in your target environment as a clone.
* Otherwise, dbt will create a simple pointer view (select * from the source object)
* By default, dbt clone will not recreate pre-existing relations in the current target. To override this, use the --full-refresh flag.
* You may want to specify a higher number of [threads](https://docs.getdbt.com/docs/running-a-dbt-project/using-threads) to decrease execution time since individual clone statements are independent of one another.

The clone command is useful for:

* blue/green continuous deployment (on data warehouses that support zero-copy cloning tables)
* cloning current production state into development schema(s)
* handling incremental models in dbt Cloud CI jobs (on data warehouses that support zero-copy cloning tables)
* testing code changes on downstream dependencies in your BI tool

https://docs.getdbt.com/reference/commands/clone

**Acceptance criteria**

* [ ] Implement a Cosmos operator that supports `dbt clone` for either Snowflake, Databricks or BQ

Issue #878, opened 2024-03-04
[feature] support for dbt clone or dbt build ? DbtCloneOperator


SPTKL:
Hi there, our cicd runs "write-audit-publish", basically we do 
1) a `dbt build` command to run and test models in staging 
2) a `dbt clone` command to clone models created in step 1 to a production environment. 
Can we implement a `DbtCloneOperator`? If so what's the recommended place? Any suggestions on how to implement? 

Also given the flexibility of cosmos, say if we want to change our process to 
1) for all models, following the order of the dag, `dbt run` the model in staging 
2) `dbt test` the model in staging
3) if test passes, `dbt clone` the model from staging to production

ideally, in the dag, each model should have their own task group with 4 tasks, run, test, snapshot, clone. It's impossible in the dbt cli because `clone` is a different command (but we get run, test, snapshot in `dbt build`). what's the recommended path for implementing something like this ^, I know it's difficult because given our current implementation, the environments are configured using different targets with different dbt profiles.

tatiana:
HI @SPTKL, This is a very useful feature. I'm still learning about the `dbt clone` [command](https://docs.getdbt.com/reference/commands/clone).

Your feature request has similarities with #903 .

One of the challenges is that we may need to introduce Cosmos intermediate storage so that independent tasks, running potentially in different worker nodes/containers / VMs, could share the build artifacts needed by `dbt clone`. This is being discussed in a separate ticket: #894

Would you be interested in contributing to Cosmos in this area?

SPTKL:
Yes, I'm interested!

PR #1326, opened 2024-11-15
Add dbt clone operator

pankajastro:
## Description

This PR introduces the DbtCloneOperator. For more details, refer to the dbt documentation: https://docs.getdbt.com/reference/commands/clone.

## Testing

**Airflow DAG**
```python
from datetime import datetime

from airflow import DAG
from cosmos import DbtSeedLocalOperator, DbtRunLocalOperator, DbtCloneLocalOperator, ProfileConfig

DBT_PROJ_DIR="/usr/local/airflow/dbt/jaffle_shop"

profile_config1=ProfileConfig(
profile_name="bigquery_dev",
target_name="dev",
profiles_yml_filepath="/usr/local/airflow/dbt/jaffle_shop/profiles.yml"
)

profile_config2=ProfileConfig(
profile_name="bigquery_clone",
target_name="dev",
profiles_yml_filepath="/usr/local/airflow/dbt/jaffle_shop/profiles.yml"
)


with DAG("test-id-1", start_date=datetime(2024, 1, 1), catchup=False) as dag:
    seed_operator = DbtSeedLocalOperator(
        profile_config=profile_config1,
        project_dir=DBT_PROJ_DIR,
        task_id="seed",
        dbt_cmd_flags=["--select", "raw_customers"],
        install_deps=True,
        append_env=True,
    )
    run_operator = DbtRunLocalOperator(
        profile_config=profile_config1,
        project_dir=DBT_PROJ_DIR,
        task_id="run",
        dbt_cmd_flags=["--models", "stg_customers"],
        install_deps=True,
        append_env=True,
    )

    clone_operator = DbtCloneLocalOperator(
        profile_config=profile_config2,
        project_dir=DBT_PROJ_DIR,
        task_id="clone",
        dbt_cmd_flags=["--models", "stg_customers", "--state", "/usr/local/airflow/dbt/jaffle_shop/target"],
        install_deps=True,
        append_env=True,
    )

    seed_operator >> run_operator >> clone_operator
``` 

**DBT Profile**
```
bigquery_dev:
  target: dev
  outputs:
    dev:
      type: bigquery
      method: service-account
      project: astronomer-dag-authoring
      dataset: bq_dev
      threads: 4 # Must be a value of 1 or greater
      keyfile: /usr/local/airflow/include/key.json
      location: US


bigquery_clone:
  target: dev
  outputs:
    dev:
      type: bigquery
      method: service-account
      project: astronomer-dag-authoring
      dataset: bq_clone
      threads: 4 # Must be a value of 1 or greater
      keyfile: /usr/local/airflow/include/key.json
      location: US
``` 

**Airflow DAG Run**

Screenshot 2024-11-15 at 6 06 50 PM

**BQ data WH**

Screenshot 2024-11-15 at 6 04 29 PM

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/1268
closes: https://github.com/astronomer/astronomer-cosmos/issues/878

## Breaking Change?
No

## Limitation
- The `dbt clone` command was introduced in dbt-core 1.6.0, so this feature is only available to users with dbt-core version 1.6 or higher https://github.com/dbt-labs/dbt-core/blob/1.6.latest/CHANGELOG.md
- Users should ensure their database is supported for cloning operations.

## Checklist

### DIFF
--- a/cosmos/operators/aws_eks.py
+++ b/cosmos/operators/aws_eks.py
@@ -8,6 +8,7 @@ from airflow.utils.context import Context
 
 from cosmos.operators.kubernetes import (
     DbtBuildKubernetesOperator,
+    DbtCloneKubernetesOperator,
     DbtKubernetesBaseOperator,
     DbtLSKubernetesOperator,
     DbtRunKubernetesOperator,
@@ -157,6 +158,15 @@ class DbtRunOperationAwsEksOperator(DbtAwsEksBaseOperator, DbtRunOperationKubern
     template_fields: Sequence[str] = (
         DbtAwsEksBaseOperator.template_fields + DbtRunOperationKubernetesOperator.template_fields  # type: ignore[operator]
     )
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+class DbtCloneAwsEksOperator(DbtAwsEksBaseOperator, DbtCloneKubernetesOperator):
+    """
+    Executes a dbt core clone command.
+    """
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
--- a/cosmos/operators/azure_container_instance.py
+++ b/cosmos/operators/azure_container_instance.py
@@ -8,6 +8,7 @@ from cosmos.config import ProfileConfig
 from cosmos.operators.base import (
     AbstractDbtBaseOperator,
     DbtBuildMixin,
+    DbtCloneMixin,
     DbtLSMixin,
     DbtRunMixin,
     DbtRunOperationMixin,
@@ -166,4 +167,13 @@ class DbtRunOperationAzureContainerInstanceOperator(DbtRunOperationMixin, DbtAzu
     )
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+class DbtCloneAzureContainerInstanceOperator(DbtCloneMixin, DbtAzureContainerInstanceBaseOperator):
+    """
+    Executes a dbt core clone command.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any):
         super().__init__(*args, **kwargs)
--- a/cosmos/__init__.py
+++ b/cosmos/__init__.py
@@ -30,6 +30,7 @@ from cosmos.log import get_logger
 from cosmos.operators.lazy_load import MissingPackage
 from cosmos.operators.local import (
     DbtBuildLocalOperator,
+    DbtCloneLocalOperator,
     DbtDepsLocalOperator,
     DbtLSLocalOperator,
     DbtRunLocalOperator,
@@ -44,6 +45,7 @@ logger = get_logger(__name__)
 try:
     from cosmos.operators.docker import (
         DbtBuildDockerOperator,
+        DbtCloneDockerOperator,
         DbtLSDockerOperator,
         DbtRunDockerOperator,
         DbtRunOperationDockerOperator,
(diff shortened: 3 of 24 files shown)


cosmos/operators/aws_eks.py
```diff
@@ -160,3 +161,12 @@ class DbtRunOperationAwsEksOperator(DbtAwsEksBaseOperator, DbtRunOperationKubern
...
         super().__init__(*args, **kwargs)
+
+
+class DbtCloneAwsEksOperator(DbtAwsEksBaseOperator, DbtCloneKubernetesOperator):
+    """
+    Executes a dbt core clone command.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
```
tatiana, line 172: What value is the constructor giving us? Would it make sense to do something like this:

```suggestion
class DbtCloneAwsEksOperator(DbtAwsEksBaseOperator, DbtCloneKubernetesOperator):
    """
    Executes a dbt core clone command.
    """
    pass
```

cosmos/operators/azure_container_instance.py
```diff
@@ -167,3 +168,12 @@ class DbtRunOperationAzureContainerInstanceOperator(DbtRunOperationMixin, DbtAzu
...
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
+
+
+class DbtCloneAzureContainerInstanceOperator(DbtCloneMixin, DbtAzureContainerInstanceBaseOperator):
+    """
+    Executes a dbt core clone command.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any):
```
tatiana, line 178: The same feedback given on https://github.com/astronomer/astronomer-cosmos/pull/1326/files#r1843904192 applies here and in the other `DbtClone` operators

tatiana (commented):
@pankajastro it's great to have you back!

Would it make sense to add the example DAG you used to test as one of Cosmos' integration DAGs?

Please, could you add documentation about this feature? I believe we may be lacking docs on how people can use `DbtBuild` operators - we could cover both in the same place.

Last but not least, it may be worth to have a test related to full refresh. Some Cosmos operators support full refresh, others don't. It seems clone supports this, so it would be great to make sure the interfaces are consistent with running models and seeds.

pankajastro:
> @pankajastro it's great to have you back!
> 
> Would it make sense to add the example DAG you used to test as one of Cosmos' integration DAGs?
> 
> Please, could you add documentation about this feature? I believe we may be lacking docs on how people can use `DbtBuild` operators - we could cover both in the same place.
> 
> Last but not least, it may be worth to have a test related to full refresh. Some Cosmos operators support full refresh, others don't. It seems clone supports this, so it would be great to make sure the interfaces are consistent with running models and seeds.

Thank you for the review!

I have added integration tests

For the scope of this PR, I have added docs for the clone operator.

I have updated existing unit tests for the full refresh, but please let me know if we need integration tests for this.

cosmos/operators/aws_eks.py
pankajastro: @tatiana We introduced init in PR [#1162](https://github.com/astronomer/astronomer-cosmos/pull/1162) because the operator import issue was observed for Airflow version 2.10 + Cosmos, and [#1161](https://github.com/astronomer/astronomer-cosmos/issues/1161) was reported in Cosmos. Do you think we should keep this in this PR, and when we work on [#1161](https://github.com/astronomer/astronomer-cosmos/issues/1161), we can clean it up then? That way, we'll be more comfortable.

pankajastro: There was an alternative solution proposed in PR https://github.com/astronomer/astronomer-cosmos/pull/1163, but it started breaking other tests. This may require some investigations.

tatiana: Let's not block the current ticket due to this. I'm happy for us to review this approach in #1161 , as you suggested, @pankajastro !

tatiana (approved):

PR #1328, opened 2024-11-15
Add exempt label option in GH action stale job

pankajastro:
I’ve noticed that the "dosubot" and "stale" jobs are creating unnecessary noise by closing issues and PRs that were not intended to be closed. In this PR, I am introducing the exempt-issue-labels and exempt-pr-labels options for the stale job. These options will ensure that issues and PRs labeled with do-not-stale are excluded from being closed by the "dosubot" and "stale" jobs.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -18,6 +18,8 @@ jobs:
           days-before-pr-stale: 30
           days-before-pr-close: 10
           stale-pr-label: "stale"
+          exempt-issue-labels: "do-not-stale"
+          exempt-pr-labels: "do-not-stale"
           stale-pr-message: "This PR is stale because it has been open for 30 days with no activity."
           close-pr-message: "This PR was closed because it has been inactive for 10 days since being marked as stale."
           repo-token: ${{ secrets.GITHUB_TOKEN }}


pankajkoti (commented):
Are we sure adding these exempt labels here would be respected by dosubot?

pankajastro:
> Are we sure adding these exempt labels here would be respected by dosubot?

I think dosubot currently closes issues/PRs with the stale label. This change ensures that stale jobs won't add the stale label, this will prevent issues and PRs from being closed by dosubot

.github/workflows/stale.yml
```diff
@@ -18,6 +18,8 @@ jobs:
           days-before-pr-stale: 30
           days-before-pr-close: 10
           stale-pr-label: "stale"
+          exempt-issue-labels: "do-not-stale"
+          exempt-pr-labels: "do-not-stale"
```
tatiana, line 22: Based on line 15, the stale bot currently does not close issues.

The source of the problem is Dosubot, which is currently closing issues that we don't want to be closed.  The issue is we're not configuring Dosubot here.

I contacted the Dosubot team to see if we could disable this feature, and they will make this happen:
* Mark as stale, but do not close for all issues

tatiana (approved):
Regardless of Dosubot, this change makes sense - since there are tickets we don't want to be marked as stale. Happy for us to merge it.

Issue #1304, opened 2024-11-05
[CI] Restore integration test / example DAG that loads manifest from gcs and azure tests


tatiana:
As part of #1297, we temporarily commented on two tasks of the manifest example DAG. This issue is a reminder we need to reenable them and fix what is missing for them to work.

PR #1333, opened 2024-11-19
Enable Azure task in the remote store manifest example DAG

pankajkoti:
During the work on PR https://github.com/astronomer/astronomer-cosmos/pull/1297, an issue arose where the Azure remote manifest task began failing after installing providers and packages with constraints. To allow other tests, which were running successfully, to proceed, the task was temporarily disabled. Upon reviewing the GitHub Actions logs from previous successful runs, it was [observed](https://github.com/astronomer/astronomer-cosmos/actions/runs/11573670971/job/32216315282#step:6:250) that the Azure provider version installed was 10.5.1. However, after the refactoring introduced in the PR, the failing actions [showed](https://github.com/astronomer/astronomer-cosmos/actions/runs/11911545582/job/33193301710#step:6:474) azure provider==8.4.0 being installed with the constraints file.

To investigate, I tested locally to identify a working version. While the task failed with 8.4.0, it succeeded with 8.5.0. Analyzing the failure [logs](https://github.com/astronomer/astronomer-cosmos/actions/runs/11911545582/job/33193301710#step:7:467) and reviewing the Azure provider changelog hints that https://github.com/apache/airflow/pull/35820/ is potentially the fix for the failure with our connection setup in the CI that was released in 8.5.0.

Therefore, I propose using azure provider>=8.5.0. However, the Airflow [constraints file](https://raw.githubusercontent.com/apache/airflow/constraints-2.8.0/constraints-3.8.txt) for version 2.8 specifies azure provider==8.4.0, which conflicts with this requirement. To address this, I am making changes to the pre-install script in our CI to install azure provider>=8.5.0 without relying on the constraints file, citing the reasons above.

closes: #1304

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@ openlineage = ["openlineage-integration-common!=1.15.0", "openlineage-airflow"]
 amazon = [
     "apache-airflow-providers-amazon[s3fs]>=3.0.0",
 ]
-google = ["apache-airflow-providers-google"]
+google = ["apache-airflow-providers-google>=10.17.0"]
 microsoft = ["apache-airflow-providers-microsoft-azure"]
 all = [
     "astronomer-cosmos[dbt-all]",
@@ -186,7 +186,7 @@ dependencies = [
     "aenum",
     "apache-airflow-providers-amazon[s3fs]>=3.0.0",
     "apache-airflow-providers-cncf-kubernetes>=5.1.1",
-    "apache-airflow-providers-google>=10.11.0",
+    "apache-airflow-providers-google>=10.17.0",
     "apache-airflow-providers-microsoft-azure",
     "msgpack",
     "openlineage-airflow",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,7 @@ amazon = [
     "apache-airflow-providers-amazon[s3fs]>=3.0.0",
 ]
 google = ["apache-airflow-providers-google>=10.17.0"]
-microsoft = ["apache-airflow-providers-microsoft-azure"]
+microsoft = ["apache-airflow-providers-microsoft-azure>=8.5.0"]
 all = [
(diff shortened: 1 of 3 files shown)


pankajastro (approved):

pyproject.toml
```diff
@@ -70,7 +70,7 @@ amazon = [
     "apache-airflow-providers-amazon[s3fs]>=3.0.0",
 ]
 google = ["apache-airflow-providers-google>=10.17.0"]
-microsoft = ["apache-airflow-providers-microsoft-azure"]
+microsoft = ["apache-airflow-providers-microsoft-azure>=8.5.0"]
```
tatiana, line 73: @pankajkoti would it make sense to try to fix the Airflow constraints files as well?

tatiana (approved):
Thank you very much, @pankajkoti , it's unexpected that the Airflow constraints files are not meeting our needs in this. I left a minor comment.

pyproject.toml
pankajkoti: I believe those constraint files are generated at the time of release and aren't mean to be mutated in order to generate reproducible builds with those constraints. The constraint files ensure that Airflow tests (existing at that time) run fine with those dependencies and the docs mention that we could upgrade/downgrade dependencies as needed. I am afraid we can fix the constraints files at this point of time.

Ref: https://airflow.apache.org/docs/apache-airflow/stable/installation/installing-from-pypi.html#reproducible-airflow-installation

pankajkoti: any thoughts/ideas on this @pankajastro based on your previous experience working with constraints files?

tatiana: Thanks a lot for the explanation, @pankajkoti, it makes sense!

It's a shame - I initially hoped we could leverage the Airflow constraints as they are to solve Cosmos dependencies - but I did realise this didn't quite work on a previous PR.

Issue #1218, opened 2024-09-23
[Bug] Tried to set relationships between tasks in more than one DAG


davidmorosini:
### Astronomer Cosmos Version

Other Astronomer Cosmos version (please specify below)

### If "Other Astronomer Cosmos version" selected, which one?

1.6.0

### dbt-core version

1.7.16

### Versions of dbt adapters

dbt-databricks==1.7.16

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

VIRTUALENV

### InvocationMode

SUBPROCESS

### airflow version

2.8.2

### Operating System

Debian GNU/Linux 12 (bookworm) (Local using docker) 

### Deployment

Other

### What happened?

Hello everyone,

We are trying to update the Cosmos version to `1.6.0` and have encountered a problem. We would like to request your help in understanding what might be happening.

Before describing the issue, please consider the code snippet below. The object `partial_dbt_env` is an instance of `<class 'airflow.models.xcom_arg.PlainXComArg'>`, similar to the example: `{{ task_instance.xcom_pull(task_ids='TASK_ID', dag_id='DAG_NAME', key='return_value') }}`.

```python
from cosmos import DbtTaskGroup, ProjectConfig

project_config = ProjectConfig(
    dbt_project_path=<DBT PROJECT PATH>,
    manifest_path=<MANIFEST PATH>,
    env_vars=partial_dbt_env
)

transform_task_group = DbtTaskGroup(
    group_id=<GROUP ID>,
    project_config=project_config,
    operator_args=<OPERATOR ARGS>,
    profile_config=<PROFILE CONFIG>,
    execution_config=<EXECUTION CONFIG>,
    render_config=<RENDER CONFIG>
)
```

When attempting to update from version `1.5.1` to version `1.6.0` using the code above, we encountered a significant issue. The exception raised is:

```python
airflow.exceptions.AirflowException: Tried to set relationships between tasks in more than one DAG: {<DAG: DAG_NAME>, <DAG: DAG_NAME>}.
```

After investigating, we noticed that the way some assignments are made in the `cosmos/converter.py` code has changed. For example:

- In version `1.5.1` ([see here](https://github.com/astronomer/astronomer-cosmos/blob/astronomer-cosmos-v1.5.1/cosmos/converter.py#L227)), the objects `env_vars` and `dbt_vars` were **assigned by reference**.
[...]

PR #1360, opened 2024-12-02
Fix dag randering for taskflow + DbtTaskGroup combo

pankajastro:
## Description
I believe the deepcopy results in two separate DAG objects—the original object and its deepcopy and the task is referencing to both objects causing failure 
```
airflow.exceptions.AirflowException: Tried to set relationships between tasks in more than one DAG: {<DAG: DAG_NAME>, <DAG: DAG_NAME>}
``` 

Screenshot 2024-12-03 at 2 30 23 AM

DAG Code

```python
from datetime import datetime

from airflow.decorators import task, dag
from cosmos import DbtTaskGroup, ProjectConfig
from include.constants import jaffle_shop_path, venv_execution_config, manifest_path
from include.profiles import airflow_db

@task(task_id="build_partial_dbt_env_vars_operator")
def build_partial_dbt_env():
    # some code
    # This return is for demonstration purposes only
    return {
        "ENV_VAR_NAME": "value",
        "ENV_VAR_NAME_2": False
    }


# partial_dbt_env = build_partial_dbt_env()



@dag(
    schedule_interval="@daily",
    catchup=False,
    tags=["simple"],
)
def simple_task_group1() -> None:
    DbtTaskGroup(
        group_id="transform_task_group",
        project_config=ProjectConfig(
        dbt_project_path=jaffle_shop_path,
        manifest_path=manifest_path,
        env_vars=build_partial_dbt_env()
    ),
        profile_config=airflow_db,
        execution_config=venv_execution_config,
    )

simple_task_group1()
#partial_dbt_env >> transform_task_group
``` 

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/1218

## Checklist

### DIFF
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -229,8 +229,8 @@ class DbtToAirflowConverter:
 
         validate_changed_config_paths(execution_config, project_config, render_config)
 
-        env_vars = copy.deepcopy(project_config.env_vars or operator_args.get("env"))
-        dbt_vars = copy.deepcopy(project_config.dbt_vars or operator_args.get("vars"))
+        env_vars = project_config.env_vars or operator_args.get("vars")
+        dbt_vars = project_config.dbt_vars or operator_args.get("vars")
 
         if execution_config.execution_mode != ExecutionMode.VIRTUALENV and execution_config.virtualenv_dir is not None:
             logger.warning(
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -229,7 +229,7 @@ class DbtToAirflowConverter:
 
         validate_changed_config_paths(execution_config, project_config, render_config)
 
-        env_vars = project_config.env_vars or operator_args.get("vars")
+        env_vars = project_config.env_vars or operator_args.get("env")
         dbt_vars = project_config.dbt_vars or operator_args.get("vars")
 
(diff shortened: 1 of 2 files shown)


pankajastro:
Tested DAG: https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/basic_cosmos_task_group.py on current main and unable to reproduce error mentioned in commit https://github.com/astronomer/astronomer-cosmos/commit/9001c986b60ba12740c49c607f3c3348ef307bd4#diff-8eeeee891e04c96d628b87bbeeed48867e29dfa33244252fc7da61ab06cfd482R4

pankajkoti (approved):

cosmos/converter.py
```diff
@@ -229,8 +229,8 @@ def __init__(
 
         validate_changed_config_paths(execution_config, project_config, render_config)
 
-        env_vars = copy.deepcopy(project_config.env_vars or operator_args.get("env"))
-        dbt_vars = copy.deepcopy(project_config.dbt_vars or operator_args.get("vars"))
+        env_vars = project_config.env_vars or operator_args.get("env")
```
tatiana, line 232: @pankajastro could you add the DAG from the description as an example DAG, so we can have some way to make sure the desired use-case continues working after future chages?

pankajastro: Added test. thanks!

PR #1362, opened 2024-12-03
Save tasks map as DbtToAirflowConverter property

internetcoffeephone:
## Description

Currently, if you want to modify a DAG after it has been rendered,
you have to walk through the dag.dbt_graph, then puzzle the task IDs
and task group IDs together by reverse-engineering your task rendering
strategy.

This is cumbersome and error-prone, hence it makes sense to expose the
mapping from DbtNode to Airflow Task ID as a DAG property. This allows
you to walk the DBT graph while directly accessing any corresponding
Airflow tasks, which makes e.g. adding Airflow sensors upstream of all
source tasks much easier.

## Breaking Change?

No

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -336,7 +336,7 @@ def build_airflow_graph(
     render_config: RenderConfig,
     task_group: TaskGroup | None = None,
     on_warning_callback: Callable[..., Any] | None = None,  # argument specific to the DBT test command
-) -> None:
+) -> dict[str, Union[TaskGroup, BaseOperator]]:
     """
     Instantiate dbt `nodes` as Airflow tasks within the given `task_group` (optional) or `dag` (mandatory).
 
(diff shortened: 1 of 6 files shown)


tatiana (commented):
HI @internetcoffeephone, thank you very much for your contribution.

The feature makes sense, and your implementation is very clean and straightforward.

I only have three requests:
* Add a unitest test confirming this new behaviour;
* Create an example DAG illustrating how you can leverage this feature in `dev/dags`; and
* Either add docs or create an issue so we can address this (we also need to document that `graph` is exposed, so there is debt on our side).

If you can get these changes before 20 December, we'll release this feature as part of Cosmos 1.8.

internetcoffeephone:
@tatiana I finally got around to doing this, hopefully I addressed all three points you mentioned.

Wasn't quite sure what to name the docs entry, I think `Post-rendering DAG customization` is a bit clunky. If you have any better suggestions, I'd be happy to change it.

tatiana (approved):
Thank you very much for addressing the feedback, @internetcoffeephone , we'll release this on 1.8 this week.

Issue #1341, opened 2024-11-26
Enable SQLite adapters tests


pankajastro:
We are running the SQLite adapters tests for dbt-core 1.4, but these tests have recently started failing. The SQLite adapters have since added support for dbt-core 1.5, so I attempted an upgrade, but that also failed and requires further investigation. There are existing issues in the repository related to upgrading the SQLite adapters to various dbt-core versions, some of which have been open for over a year: 
- https://github.com/codeforkjeff/dbt-sqlite/issues

Additionally, there is an open issue calling for a maintainer for the project: https://github.com/codeforkjeff/dbt-sqlite/issues/48

It's also worth noting that dbt-core 1.4 and 1.5 have reached their End of Life (EOL), as outlined here: https://docs.getdbt.com/docs/dbt-versions/core#latest-releases.

The goal of this ticket is to restore the SQLite adapter tests.

tatiana:
The idea with the sqlite adaptor tests was to have a cheap alternative to the Postgres ones. Perhaps we can look into adopting a not so cheap alternative, BQ, or another one - WDYT?

pankajastro:
BQ option looks good to me, and I believe we already have the credentials set up in CI for this

PR #1366, opened 2024-12-05
Refactor `dbt-sqlite` tests to use `dbt-postgres`

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1341

previously ran SQLite tests with dbt-core version 1.4, which has reached its end of life. Recently, these tests have started failing. Since we rely on the repo: https://github.com/codeforkjeff/dbt-sqlite/ for running the tests, and it is no longer actively maintained, this PR performs the necessary cleanup to address the issue

- Point the sqlite test to use Postgres
- Remove CI workflow job: Run-Integration-Tests-Sqlite
- Remove hatch target:  hatch run tests.py3.12-2.10:test-integration-sqlite-setup and  hatch run tests.py3.12-2.10:test-integration-sqlite
- Remove pytest marker `sqlite`
- Remove dbt project simple
- Update manifest file


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, replace-sqlite-to-pg]
```
```suggestion
    branches: [main]
```

tatiana (commented):
Thanks for working on this, @pankajastro , it will be great to no longer depend on an adaptor that stopped being maintained.

Are the example DAGs running as expected?

I had the impression the SQLite tests relied on pre-populated data in the database. Are they running as expected? I was surprised no test is failing - I thought we had an example DAG that relied on this as well.

pankajastro:
> Thanks for working on this, @pankajastro , it will be great to no longer depend on an adaptor that stopped being maintained.
> 
> Are the example DAGs running as expected?
> 
> I had the impression the SQLite tests relied on pre-populated data in the database. Are they running as expected? I was surprised no test is failing - I thought we had an example DAG that relied on this as well.

right, I think that particular DAG got ignored because of

```python
if DBT_VERSION >= Version("1.5.0"):
            file.writelines(["example_cosmos_sources.py\n"])
``` 
I'll fix it

pankajastro:
> Thanks for working on this, @pankajastro , it will be great to no longer depend on an adaptor that stopped being maintained.
> 
> Are the example DAGs running as expected?
> 
> I had the impression the SQLite tests relied on pre-populated data in the database. Are they running as expected? I was surprised no test is failing - I thought we had an example DAG that relied on this as well.

It was indeed not running. 
But now running: https://github.com/astronomer/astronomer-cosmos/actions/runs/12251360416/job/34175888338?pr=1366#step:7:395
post your previous review, I have removed the dbt project simple from repo since it was getting used for sql test only and updated the dbt project jaffle_shop to make this example and other test work

dev/dags/dbt/jaffle_shop/models/orders.sql
```diff
@@ -1,3 +1,9 @@
+{{ config(
+    materialized='view',
+    alias=var('animation_alias', 'orders')
```
tatiana, line 3: Why are we naming an alias to `orders` as `animation_alias`?

dev/dags/dbt/jaffle_shop/profiles.yml
```diff
@@ -16,7 +16,7 @@ postgres_profile:
   outputs:
     dev:
       type: postgres
-      dbname: "{{ env_var('POSTGRES_DATABASE') }}"
+      dbname: "{{ env_var('POSTGRES_DB') }}"
```
tatiana, line 19: Why did we change this?

pankajastro: I pointed out the SQLite test to use this profile, but ENV `POSTGRES_DATABASE` is not set in GitHub action, so I just renamed it the env name is already available in CI. But do let me know if it could have any side effect

dev/dags/dbt/jaffle_shop/models/orders.sql
tatiana: Makes sense, shall we rename `animation_alias` to `orders_alias`?

dev/dags/dbt/jaffle_shop/profiles.yml
tatiana: Thanks for the explanation, @pankajastro , I was confused - but it makes sense!

dev/dags/dbt/jaffle_shop/models/orders.sql
pankajastro: Fix in: https://github.com/astronomer/astronomer-cosmos/pull/1366/commits/8fd03dc8afda10a51b3185c5c5f744b4bf9fe28d

tatiana (approved):
Thanks for refactoring and restoring these tests, @pankajastro !

It's a shame https://github.com/codeforkjeff/dbt-sqlite hasn't been kept up-to-date for over a year now. It doesn't make sense for us to rely on this adaptor. We can review this decision in the future if circumstances change.

PR #1374, opened 2024-12-10
Restore Jaffle Shop so that `basic_cosmos_dag` works as documented

tatiana:
When trying out some changes in Cosmos, I realised that we significantly changed the topology of the jaffle_shop project.

Before, the `basic_cosmos_dag` rendered this way, as described in our [README](https://github.com/astronomer/astronomer-cosmos/blob/main/README.rst):
Screenshot 2024-12-10 at 13 44 47

Now, it renders like this:
Screenshot 2024-12-10 at 13 03 14

Some of the changes:
1. standalone task with a different name (`m u l t i b y t e _run`)
2. `stg_payments` does not depend on `raw_payments_seed`
3. `stg_orders` does not depend on raw_orders_seed` 

The dependency issue (2) and (3) has been happening since a89389de32fd3fc140722ca925d5a28526fd8172

While we need to create new dbt projects / make changes that allow us to validate new features in Cosmos, we should be strict with the Jaffle Shop project used as a reference for our most basic example.


dev/dags/dbt/jaffle_shop/fail.txt
```diff
@@ -0,0 +1,77 @@
+adal==1.2.7
```
pankajkoti, line 1: can we please add a comment here to describe what this file means?

dev/dags/dbt/jaffle_shop/succeed.txt
```diff
@@ -0,0 +1,127 @@
+adal==1.2.7
```
pankajkoti, line 1: similarly can we please add a comment describing the purpose of this file?

pankajkoti (approved):
LGTM, a couple of minor comments inline since I'm not aware of and would like to understand what the fail.txt and succeed.txt files are for :)

dev/dags/dbt/altered_jaffle_shop/dbt_project.yml
```diff
@@ -0,0 +1,26 @@
+name: 'jaffle_shop'
+
+config-version: 2
+version: '0.1'
+
+profile: 'jaffle_shop'
```
pankajastro, line 6: The profile `jaffle_shop` does not exist in profile.yml shall we adjust the name to match one of the profile name in profile.yml?

dev/dags/dbt/altered_jaffle_shop/fail.txt
```diff
@@ -0,0 +1,77 @@
+adal==1.2.7
```
pankajastro, line 1: curious why do we need this file?

dev/dags/dbt/altered_jaffle_shop/packages_.yml
pankajastro, line 1: why do we need this file, given we have `packages.yml`

dev/dags/dbt/altered_jaffle_shop/profiles.yml
```diff
@@ -0,0 +1,24 @@
+default:
```
pankajastro, line 1: Both profile looks very similar, shall we remove one?

pankajastro (approved):

dev/dags/dbt/jaffle_shop/succeed.txt
tatiana: We don't need this, it's a left over, deleting it!

tatiana: Remove them. They were left over files from previous tests! Thanks for pointing out

dev/dags/dbt/jaffle_shop/fail.txt
tatiana: Removed in 5b84560e845ed6c0f8c24ca38697f08bed9a3339, they were left overs!

dev/dags/dbt/altered_jaffle_shop/dbt_project.yml
tatiana: Addressed in 1d23cc7a6a7dd6f3e2c20033b6a4f88ef296f76c

dev/dags/dbt/altered_jaffle_shop/profiles.yml
tatiana: Yes, I've logged a follow-up ticket for this (#1384) since both our profiles share this problem, and we'll need to change the K8s example and docs.

PR #1376, opened 2024-12-10
Pin dbt-databricks version to < 1.9

pankajastro:
Our CI is broken due to a `dbt-databricks` release with conflicting dependencies with `Airflow>=2.8`.

The latest release of`dbt-databricks` has pinned the Pydantic to `"pydantic>=1.10.0, <2"` https://github.com/databricks/dbt-databricks/blob/main/pyproject.toml#L33-L33

This change was added in PR: https://github.com/databricks/dbt-databricks/pull/843  and released on 9 December as part of `dbt-databricks==1.9.0` (https://pypi.org/project/dbt-databricks/1.9.0/).

The `debt-data bricks` conflicts with the neAirflow version, which that requires Pydantic 2.0.

For Airflow < 2.8, things work as expected: https://github.com/astronomer/astronomer-cosmos/actions/runs/12259335118/job/34201293598.

In this PR, I'm pinning dbt-databricks version to < 1.9 to make CI green again.

Follow-up PR: https://github.com/astronomer/astronomer-cosmos/issues/1379

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
     "Jinja2>=3.0.0",
     "msgpack",
     "packaging>=22.0",
-    "pydantic>=1.10.0",
+    # "pydantic>=1.10.0",
     "typing-extensions; python_version < '3.8'",
     "virtualenv",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ dbt-all = [
     "dbt-athena",
     "dbt-bigquery",
     "dbt-clickhouse",
+    # TODO: https://github.com/astronomer/astronomer-cosmos/issues/1379
     "dbt-databricks<1.9",
     "dbt-exasol",
     "dbt-postgres",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,6 @@ dependencies = [
     "Jinja2>=3.0.0",
     "msgpack",
     "packaging>=22.0",
-    # "pydantic>=1.10.0",
     "typing-extensions; python_version < '3.8'",
     "virtualenv",
 ]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
     "Jinja2>=3.0.0",
     "msgpack",
     "packaging>=22.0",
-    "pydantic>=1.10.0",
+    # "pydantic>=1.10.0",
```
pankajastro, line 38:
```suggestion
    "pydantic>=1.10.0",
```

pyproject.toml
```diff
@@ -189,7 +189,6 @@ dependencies = [
     "apache-airflow-providers-microsoft-azure>=8.5.0",
     "msgpack",
     "openlineage-airflow",
-    "pydantic>=1.10.0",
```
pankajastro, line 192: pydantic>=1.10.0

pankajastro, line 192: Not sure if we need this in docs build

pyproject.toml
```diff
@@ -35,7 +35,6 @@ dependencies = [
     "Jinja2>=3.0.0",
     "msgpack",
     "packaging>=22.0",
-    "pydantic>=1.10.0",
```
pankajkoti, line 38: why are we removing this?

tatiana (approved):
Thanks for fixing this and creating the follow up ticket, @pankajastro !

pyproject.toml
pankajastro: looks like force push otherwise, I restored it https://github.com/astronomer/astronomer-cosmos/pull/1376#discussion_r1878376966. thanks for pointing out

pankajastro: Restored it,

Issue #892, opened 2024-03-14
[Feature request] Build operator as a test behavior: `TestBehavior.BUILD`


dwreeves:
In Cosmos 1.4, the "build" operator was added.

This is not currently being used in the DbtTaskGroup / DbtDag API. But there is one use case that can be supported: tests! You can run tests and a model in a single dbt invocation using `dbt build`.

Why would someone want this? A big reason is to reduce the time of DAG runs. Splitting out tests into two Airflow tasks-- one for the node and one for the test-- can slow things down for users who are facing huge bottlenecks in spinning up nodes, e.g. they are not using partial parsing.

This is probably not a simple change because it most likely requires changing some internal assumptions regarding how Cosmos parses DAGs.

I'm very open to someone else taking on this issue, too.

tatiana:
Hi @dwreeves , I like your proposal.
What do you think we introduced this as `TestBehaviour.BUILD`? I believe this is a relatively small change.

PR #1377, opened 2024-12-10
 Add support to `TestBehavior.BUILD`

tatiana:
By default, Cosmos uses `TestBehavior.AFTER_EACH`, creating an Airflow TaskGroup that contains two tasks:
* one to run the model, seed or snapshot
* another to run the tests related to that dbt resource

While many users desire and expect this behaviour, it can also mean additional overhead,  especially in dbt projects with more than 500 models. Each time the `dbt` command is executed, there is an overhead, even when using optimisations such as partial parsing and `dbtRunner`. There is also an overhead on splitting a task into multiple Airflow workers.

Illustrating some numbers with data shared by an Astronomer customer regarding the dbt command execution (between the logs "running dbt with arguments" and "Done."):
* Running `dbt build` for a particular model + its tests: 46s
* Running `dbt run` + `dbt test` individually: 2min15s

This PR introduces a new behaviour, `TestBehavior.BUILD`, where Cosmos can run both the model/seed/snapshot and the associated tests using a single command (`dbt build`). For documentation on the dbt build, check https://docs.getdbt.com/reference/commands/build.

This is an example of how the DAG will render when using this test behaviour when running:
```
 airflow dags test example_cosmos_dbt_build 
```

Screenshot 2024-12-10 at 15 08 45

And this is an example of the output, showing both the model is being run and also the tests, using the build command:
```
[2024-12-10 15:19:23,667] {local.py:405} INFO - Trying to run dbtRunner with:
 ['build', '--models', 'customers', '--full-refresh', '--project-dir', '/var/folders/td/522y78v91d1f5wgh67mj3p0m0000gn/T/tmpghz8naek', '--profiles-dir', '/tmp/profile/ac4e9cde9bc05d574c157e795dcbcc6b60246a73ca1d92d4fc669e90a1e494e0', '--profile', 'default', '--target', 'dev']
 in /var/folders/td/522y78v91d1f5wgh67mj3p0m0000gn/T/tmpghz8naek
[2024-12-10T15:19:23.667+0000] {local.py:405} INFO - Trying to run dbtRunner with:
 ['build', '--models', 'customers', '--full-refresh', '--project-dir', '/var/folders/td/522y78v91d1f5wgh67mj3p0m0000gn/T/tmpghz8naek', '--profiles-dir', '/tmp/profile/ac4e9cde9bc05d574c157e795dcbcc6b60246a73ca1d92d4fc669e90a1e494e0', '--profile', 'default', '--target', 'dev']
 in /var/folders/td/522y78v91d1f5wgh67mj3p0m0000gn/T/tmpghz8naek
15:19:23  Running with dbt=1.8.0
15:19:23  Registered adapter: postgres=1.8.0
15:19:23  Found 5 models, 3 seeds, 20 data tests, 528 macros
15:19:23  
15:19:23  Concurrency: 1 threads (target='dev')
15:19:23  
15:19:23  1 of 4 START sql table model public.customers .................................. [RUN]
15:19:23  1 of 4 OK created sql table model public.customers ............................. [SELECT 100 in 0.04s]
15:19:23  2 of 4 START test not_null_customers_customer_id ............................... [RUN]
15:19:23  2 of 4 PASS not_null_customers_customer_id ..................................... [PASS in 0.02s]
15:19:23  3 of 4 START test relationships_orders_customer_id__customer_id__ref_customers_  [RUN]
15:19:23  3 of 4 PASS relationships_orders_customer_id__customer_id__ref_customers_ ...... [PASS in 0.02s]
15:19:23  4 of 4 START test unique_customers_customer_id ................................. [RUN]
15:19:23  4 of 4 PASS unique_customers_customer_id ....................................... [PASS in 0.02s]
15:19:23  
15:19:23  Finished running 1 table model, 3 data tests in 0 hours 0 minutes and 0.19 seconds (0.19s).
15:19:24  
15:19:24  Completed successfully
15:19:24  
15:19:24  Done. PASS=4 WARN=0 ERROR=0 SKIP=0 TOTAL=4
```

Closes: #892


tests/airflow/test_graph.py
```diff
@@ -227,6 +227,49 @@ def test_build_airflow_graph_with_after_all():
...
+        )
+    topological_sort = [task.task_id for task in dag.topological_sort()]
+    expected_sort = ["seed_parent_seed_build", "parent_model_build", "child_model_build", "child2_v2_model_build"]
+    assert topological_sort == expected_sort
+
+    task_groups = dag.task_group_dict
+    assert len(task_groups) == 0
+
+    assert len(dag.leaves) == 2
+    # assert dag.leaves[0].task_id in ("orders_model_build", "customers_model_build")
```
tatiana, line 269: Uncomment once #1374 is released and this branch is rebased:

```suggestion
    assert dag.leaves[0].task_id in ("orders_model_build", "customers_model_build")
```

dwreeves (approved):
The code makes sense to me, so giving it a LGTM. Thanks for taking this on @tatiana!

---

A thought as I read this, for another day though:

dbt 1.8+ has a new concept called unit tests, which it differentiates from data tests (formerly just "tests"). dbt's preferred way of running things is: `unit tests -> build model -> data tests`.

For the `TestBehavior.BUILD`, we comport with how dbt wants to run things 👍

We don't currently handle dbt unit tests for `TestBehavior.AFTER_ALL` and `TestBehavior.AFTER_EACH` though. 🤔 A natural way to do this would be to have `AFTER_ALL` also mean "before all" when it comes to unit tests, and similarly `AFTER_EACH` means "before each" for unit tests.

Alternatively, these could be decoupled; i.e. "before each, unit test" and "after all, data test," similarly "before all, unit test" and "after each, data test." But for `TestBehavior.BUILD`, decoupling gets weird. 🤷

pankajkoti (approved):
LGTM. Amazing support added and nice optimization. Happy to merge once https://github.com/astronomer/astronomer-cosmos/pull/1374 is merged & this branch is rebased, and the CI passes after.

cosmos/constants.py
```diff
@@ -75,6 +75,7 @@ class TestBehavior(Enum):
     Behavior of the tests.
     """
 
+    BUILD = "build"
```
pankajastro, line 78: shall we update docs: https://github.com/astronomer/astronomer-cosmos/blob/main/docs/configuration/testing-behavior.rst

pankajastro (approved):

cosmos/constants.py
tatiana: Addressed in 20ba9c7ae9fdb68e0d2c0c87f1ab662b5ba96402, with minor fixes in commits afterwards.

You can see the docs rendered in via Cloudflare:
https://585d1321.astronomer-cosmos.pages.dev/configuration/testing-behavior/

tests/airflow/test_graph.py
tatiana: Done in 988ae357005c190f11c51a621b57021b3a7e9006

tatiana:
Thanks a lot for the review and feedback, @dwreeves @pankajastro @pankajkoti!

I added a follow-up action to review what @dwreeves mentioned (unit tests x data tests split): #1386 I temptively added this to Cosmos 2.x, but we may want to handle this before.

PR #1382, opened 2024-12-11
move default_args into kwargs

david-mag:
## Description
When using default argument for `on_error_callback` functions, it is not executed, because its in a sub-dictionary in `kwargs`.


david-mag:
Small update to fix a bug where when using `on_failure_callback` in the `default_args`, callbacks for dbt tests that fail won't be executed.

tatiana (commented):
Hi @david-mag thanks for your contribution, please, could you add a test to cover this use-case?

cosmos/operators/kubernetes.py
```diff
@@ -165,6 +165,9 @@ def __init__(self, on_warning_callback: Callable[..., Any] | None = None, **kwar
         if not on_warning_callback:
             super().__init__(**kwargs)
         else:
+            # Make the following code independent of arguments being defaulted or not.
```
tatiana, line 168: Why are we making this change only to the `DbtTestKubernetesOperator`? would it also make sense in the remaining operators?

david-mag:
added a test case for this

pankajastro (commented):
Hey @david-mag Just for my understanding could you please explain what is on_error_callback and share some DAG example how you are using it?

david-mag:
Oh, my mistake ... I mean `on_failure_callback` 😳 
https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/callbacks.html

tatiana:
This PR became stale, and recently another contributor proposed a more thorough solution for the problem (#1648), so we're closing this PR. Thanks for all the work!

Issue #1367, opened 2024-12-06
Can't filter dbt  nodes by tags when using LoadMode.CUSTOM


yoshimaru46:
## issue

when using `LoadMode.CUSTOM`, tags passed to DbtNode is prefixed with `tags:`.

https://github.com/astronomer/astronomer-cosmos/blob/cfb6bf20ddacb84433af094969602467475fd4bc/cosmos/dbt/graph.py#L656-L666

therefore, `_is_tags_subset` become always false.

https://github.com/astronomer/astronomer-cosmos/blob/cfb6bf20ddacb84433af094969602467475fd4bc/cosmos/dbt/selector.py#L388

## model.sql

```
config(tags=["foo"])
```

## render_config

```
select=["tag:foo"],
```

## _is_tags_subset

- self.config.tags: `set(["foo"])`
- node.tags: `set(["tags:foo"])`

`(set(self.config.tags) <= set(node.tags))` => always false

pankajastro:
Hey @yoshimaru46, would you like to submit a fix?

yoshimaru46:
PR: https://github.com/astronomer/astronomer-cosmos/pull/1387

yoshimaru46:
@pankajastro 

Please review [this PR](https://github.com/astronomer/astronomer-cosmos/pull/1387) at your convenience 🙏

tatiana:
As mentioned in https://github.com/astronomer/astronomer-cosmos/issues/1953, the `LoadMode.CUSTOM` is going to be removed in Cosmos 2.0, so we are not going to fix this, sorry.

PR #1387, opened 2024-12-12
Remove "tags:" prefix from model config selectors

yoshimaru46:
## Description

closes https://github.com/astronomer/astronomer-cosmos/issues/1367#issuecomment-2527350354

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -654,6 +654,8 @@ class DbtGraph:
         for model_name, model in models:
             config = {item.split(":")[0]: item.split(":")[-1] for item in model.config.config_selectors}
             tags = [selector for selector in model.config.config_selectors if selector.startswith("tags:")]
+            # Remove the "tags:" prefix
+            tags = [tag.split(":")[-1] for tag in tags]
             node = DbtNode(
                 unique_id=f"{model.type.value}.{self.project.project_name}.{model_name}",
                 resource_type=DbtResourceType(model.type.value),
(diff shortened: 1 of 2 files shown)


cosmos/dbt/graph.py
```diff
@@ -654,6 +654,8 @@ def load_via_custom_parser(self) -> None:
         for model_name, model in models:
             config = {item.split(":")[0]: item.split(":")[-1] for item in model.config.config_selectors}
             tags = [selector for selector in model.config.config_selectors if selector.startswith("tags:")]
+            # Remove the "tags:" prefix
+            tags = [tag.split(":")[-1] for tag in tags]
```
tatiana, line 658: We should not represent tags in different ways depending on the parsing mode. Either we remove the prefix for all  `LoadMode`s that set this configuration (not only `LoadMode.CUSTOM`, but also `LoadMode.DBT_MANIFEST`), or we remove for none of them.

By changing this here, I'm afraid we may be breaking the behaviour ahead in the selector, for instance:
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/dbt/selector.py#L177

The prefix `"tag:" is currently being used in the `selector` module in at least four places via the `TAG_SELECTOR` constant: https://github.com/astronomer/astronomer-cosmos/blob/110fb0760cb1a21a2a21ca1a06370c44613ba085/cosmos/dbt/selector.py#L20C1-L20C13

Since the `select_nodes` is being used not only by `RenderMode.CUSTOM`, but also by `RenderMode.DBT_MANIFEST`, I believe this change will break the overall behaviour.

I'm surprised no unit tests broke.

tatiana (commented):
@yoshimaru46 thanks for identifying the bug and proposing a fix.

I have shared a concern in-line.

I agree the following line will likely fail in your use case:
https://github.com/astronomer/astronomer-cosmos/blob/110fb0760cb1a21a2a21ca1a06370c44613ba085/cosmos/dbt/selector.py#L422

I have two requests:
1. Could you confirm it also fails if you use `LoadMode.DBT_MANIFEST`
2. Would it make sense to make the change in this method itself, to avoid breaking other parts, as mentioned in my other comment?

Issue #914, opened 2024-04-17
on_warning_callback for source freshness


fabiomx:
It would be good to make the `on_warning_callback` argument available for the `source freshness` command since, similarly to the test command, you can configure the severity (see [freshness](https://docs.getdbt.com/reference/resource-properties/freshness)), and you might need to use a callback to send notifications for the "warn_after" cases or to store its results.

It looks like a similar work to the one done for the test command ([Creating on_warning_callback arg for better warning management](https://github.com/astronomer/astronomer-cosmos/pull/225/)) might be done.

I mention @CorsettiS, as he is the original author of this feature for the test command, and it would be good know his opinion on this.

corsettigyg:
> It would be good to make the `on_warning_callback` argument available for the `source freshness` command since, similarly to the test command, you can configure the severity (see [freshness](https://docs.getdbt.com/reference/resource-properties/freshness)), and you might need to use a callback to send notifications for the "warn_after" cases or to store its results.
> 
> It looks like a similar work to the one done for the test command ([Creating on_warning_callback arg for better warning management](https://github.com/astronomer/astronomer-cosmos/pull/225/)) might be done.
> 
> I mention @CorsettiS, as he is the original author of this feature for the test command, and it would be good know his opinion on this.

hey @fabiomx . it's been a long time I have implemented this feature so I do not remember in details the issues I faced with the freshness check. if I am not wrong, it is because the source freshness has a very specific warn message that it was tricky to deal with regex and at the time of implementation I was only interested in using this feature for tests so I did not put too much thought on it. Probably you just need to create an equivalent regex to extract it and add a new if statement just like [that one](https://github.com/astronomer/astronomer-cosmos/blob/9c340c238f0ebf44e5b6b1844370d3f864236ad5/cosmos/dbt/parser/output.py#L80)

tatiana:
Hi, @fabiomx, this is a very relevant feature! Thanks a lot to @corsettigyg for implementing it for tests!
Would you be interested in contributing to Cosmos with this work?

corsettigyg:
@fabiomx @tatiana it is important to mention that `on_warning_callback` already works for freshness, it just doesn't capture the freshness message. I am in the process of implementing cosmos-dbt in the current company I am working at, and once I am done with that I can work on that feature

tatiana:
That's brilliant, @corsettigyg , thank you very much!

fabiomx:
That's great, @corsettigyg, thanks!

[...]

PR #1400, opened 2024-12-17
Add warning callback on source freshness

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/914

This PR introduced `on_warning_callback` for `dbt source freshness` command i.e on operator `DbtSourceLocalOperator`.

Example DAG
```python
source_rendering_dag = DbtDag(
    project_config=ProjectConfig(
    ),
    operator_args={
        "full_refresh": True,  # used only in dbt commands that support this flag
    },
    render_config=RenderConfig(source_rendering_behavior=SourceRenderingBehavior.ALL),
    # normal dag parameters
    schedule_interval="@daily",
    start_date=datetime(2024, 1, 1),
    catchup=False,
    dag_id="source_rendering_dag",
    default_args={"retries": 2},
    on_warning_callback=lambda context: print(context),
)

```

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -198,7 +198,6 @@ def create_task_metadata(
             else:
                 task_id = f"{node.name}_run"
         elif node.resource_type == DbtResourceType.SOURCE:
-            # if on_warning_callback is not None:
             extra_context["on_warning_callback"] = on_warning_callback
 
             if (source_rendering_behavior == SourceRenderingBehavior.NONE) or (
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -165,6 +165,7 @@ def create_task_metadata(
     use_task_group: bool = False,
     source_rendering_behavior: SourceRenderingBehavior = SourceRenderingBehavior.NONE,
     test_behavior: TestBehavior = TestBehavior.AFTER_ALL,
+    on_warning_callback: Callable[..., Any] | None = None,
 ) -> TaskMetadata | None:
     """
     Create the metadata that will be used to instantiate the Airflow Task used to run the Dbt node.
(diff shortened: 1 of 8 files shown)


cosmos/core/airflow.py
```diff
@@ -35,6 +35,12 @@ def get_airflow_task(task: Task, dag: DAG, task_group: TaskGroup | None = None)
     for k, v in task.airflow_task_config.items():
         task_kwargs[k] = v
 
+    # Set the on_warning_callback of source node in task_kwargs
```
pankajkoti, line 38: The comment mentions source node here, but it appears we set it in all nodes? Should we drop the word source from here in the comment?

cosmos/airflow/graph.py
```diff
@@ -176,17 +177,19 @@ def create_task_metadata(
     :param dbt_dag_task_group_identifier: Identifier to refer to the DbtDAG or DbtTaskGroup in the DAG.
     :param use_task_group: It determines whether to use the name as a prefix for the task id or not.
         If it is False, then use the name as a prefix for the task id, otherwise do not.
+    :param on_warning_callback:
```
pankajkoti, line 180: missing docstring for the param

pankajkoti (commented):
The PR looks good to me. Have a couple of minor comments inline. Is something more to be added to the PR? The description mentions docs to be added. Are we going to add it to the same PR or later?

cosmos/airflow/graph.py
pankajastro: Added docs

cosmos/core/airflow.py
pankajastro: on_warning_callback get set only for node type source but I have removed it

pankajkoti (approved):

cosmos/core/airflow.py
```diff
@@ -35,6 +35,11 @@ def get_airflow_task(task: Task, dag: DAG, task_group: TaskGroup | None = None)
     for k, v in task.airflow_task_config.items():
         task_kwargs[k] = v
 
+    on_warning_callback = task.extra_context.get("on_warning_callback")
```
tatiana, line 38: Do we need to add this here? Why wasn't this necessary when we implemented the Test warning handling?

pankajastro: Thanks! this should not require: fixed in https://github.com/astronomer/astronomer-cosmos/pull/1400/commits/8d7ad6637103772753f3e2a61096f2a5d48f0a9a

tatiana (approved):

Issue #1399, opened 2024-12-17
Test that callback functionality works for ExecutionMode.VIRTUALENV


pankajkoti:
Since the ability to specify a callback argument is already supported in `ExecutionMode.LOCAL` (as demonstrated in [this code reference](https://github.com/astronomer/astronomer-cosmos/blob/0000f80cbdb1e137b423a9297d45380441dd8c9b/cosmos/operators/local.py#L143)), and PR #1389 showcases how to leverage this functionality, we need to verify that this callback support is inherited and functions correctly with `ExecutionMode.VIRTUALENV`, given that `DbtVirtualenvBaseOperator` extends `DbtLocalBaseOperator`.

Related: #1349

PR #1389, opened 2024-12-13
Add helper functions for uploading target directory artifacts to remote cloud storages

pankajkoti:
This PR introduces helper functions that can be passed as callable callbacks for Cosmos tasks to execute post-task execution. These helper functions enable the uploading of artifacts (from the project's target directory) to various cloud storage providers, including AWS S3, Google Cloud Storage (GCS), Azure WASB, and general remote object stores using Airflow’s ObjectStoragePath.

## Key Changes
Adds a `cosmos/io.py` module that includes the following helper functions

1. `upload_artifacts_to_aws_s3`
      - Uploads artifact files from a task’s local target directory to an AWS S3 bucket.
      - Supports dynamically appending DAG metadata (e.g., dag_id, task_id, run_id, and try number) to the uploaded file  paths.
      - Utilizes S3Hook from the airflow.providers.amazon.aws module.

2. `upload_artifacts_to_gcp_gs`

      - Uploads artifact files from a task’s local target directory to a Google Cloud Storage (GCS) bucket.
      - Appends DAG-related context to the GCS object names for better traceability.
      - Leverages GCSHook from airflow.providers.google.cloud.
    
3. `upload_artifacts_to_azure_wasb`
      - Uploads artifact files from a task’s local target directory to an Azure Blob Storage container.
      - Automatically structures blob names with metadata, including dag_id, task_id, and execution details.
      - Utilizes WasbHook from the airflow.providers.microsoft.azure module.
  
4. `upload_artifacts_to_cloud_storage`
      - A generic helper function that uploads artifacts from a task’s local target directory to remote object stores configured via Airflow’s ObjectStoragePath (Airflow 2.8+ feature).
      - Supports custom remote storage configurations such as `remote_target_path` and `remote_target_path_conn_id`.
      - Dynamically constructs file paths that include DAG metadata for clear organization.
     
These helpers functions can be passed as the `callback` argument to `DbtDAG` or to your `Dag` instance as demonstrated in the example DAGs `dev/dags/cosmos_callback_dag.py` and `dev/dags/example_operators.py` correspondingly. You can also pass `callback_args` as shown in the example DAGs. These helper functions are mere examples of how callback functions can be written and passed to your operators/DAGs to be executed after task completions. Taking reference of these helper functions, you can write your own callback function and pass those. 

## Limitations

[...]

PR #1401, opened 2024-12-18
Mention in documentation that the callback functionality is supported in ExecutionMode.VIRTUALENV

pankajkoti:
Following up on PR #1389 wherein we add helper functions to demonstrate how to 
utilise the callback functionality in `ExecutionMode.LOCAL`, I also tested this functionality
with `ExecutionMode.VIRTUALENV` since the VirtualEnv mode operators inherit the Local
mode operators and it does seem to support the callback functionality well. I was well able
to test this by adding the callback arg to the VirtualEnv example DAG we have and see that
the files do get uploaded to remote store. I have added the argument in comments in the example
DAG so that users could take a reference of that, however I have left the argument as commented
so that we do not always keep on uploading the files in our CI runs and given that we already this
in the `example_operators.py` DAG.

The PR also updates the documentation to reflect that the callback functionality is supported in 
the `ExecutionMode.VIRTUALENV` too.

closes: #1399

### DIFF
--- a/dev/dags/example_virtualenv.py
+++ b/dev/dags/example_virtualenv.py
@@ -56,6 +56,16 @@ def example_virtualenv() -> None:
             "py_requirements": ["dbt-postgres"],
             "install_deps": True,
             "emit_datasets": False,  # Example of how to not set inlets and outlets
+            # --------------------------------------------------------------------------
+            # Callback function to upload files using Airflow Object storage and Cosmos remote_target_path setting on
+            # Airflow 2.8 and above
+            # "callback": upload_to_cloud_storage,
+            # --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


dev/dags/example_virtualenv.py
```diff
@@ -56,6 +56,16 @@ def example_virtualenv() -> None:
             "py_requirements": ["dbt-postgres"],
             "install_deps": True,
             "emit_datasets": False,  # Example of how to not set inlets and outlets
+            # --------------------------------------------------------------------------
```
tatiana, line 59: If you could explain why these are commented out, it would be great!

tatiana (approved):

dev/dags/example_virtualenv.py
pankajkoti: added a comment to explain it

PR #1405, opened 2024-12-18
Fix source resource type tests

pankajastro:
`jaffle_shop` does not have a source so let's point the example_source_rendering.py to use the `altered_jaffle_shop` project.


pankajastro:
Created a follow-up issue to see if we can avoid using outdated manifest file

tatiana (approved):

pankajkoti (commented):
Nice fix 🚀

PR #1406, opened 2024-12-18
Fix Snowflake Profile mapping when using AWS default region

tatiana:
When using a Cosmos Snowflake Profile mapping using a Snowflake account set in the AWS default region, Cosmos would fail if the default region was specified in the Airflow connection.

The dbt docs state:
> For AWS accounts in the US West default region, you can use abc123 (without any other segments). For some AWS accounts you will have to append the region and/or cloud platform. For example, abc123.eu-west-1 or abc123.eu-west-2.aws.
https://docs.getdbt.com/docs/core/connect-data-platform/snowflake-setup#account
 
Although it seems that defining the default region would be optional, a Cosmos user reported facing 404 and seeing a dbt error message when attempting to use `SnowflakeUserPasswordProfileMapping` with an Airflow Snowflake connection that defined the region `us-west-2`.
snowflake-404

We solved the issue by removing the region `us-west-2` from the connection.

Since this restriction only applies to AWS and this Snowflake region only exists to AWS, this change seems safe:
Screenshot 2024-12-18 at 18 45 31

### DIFF
--- a/cosmos/profiles/snowflake/user_privatekey.py
+++ b/cosmos/profiles/snowflake/user_privatekey.py
@@ -65,12 +65,7 @@ class SnowflakePrivateKeyPemProfileMapping(SnowflakeBaseProfileMapping):
     @property
     def profile(self) -> dict[str, Any | None]:
         """Gets profile."""
-        profile_vars = {
-            **self.mapped_params,
-            **self.profile_args,
-            # private_key should always get set as env var
-            "private_key": self.get_env_var_format("private_key"),
-        }
-
-        # remove any null values
+        profile_vars = super().profile
+        # private_key should always get set as env var
+        profile_vars["private_key"] = self.get_env_var_format("private_key")
         return self.filter_null(profile_vars)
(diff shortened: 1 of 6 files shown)


cosmos/profiles/snowflake/user_privatekey.py
```diff
@@ -74,11 +74,3 @@ def profile(self) -> dict[str, Any | None]:
 
         # remove any null values
         return self.filter_null(profile_vars)
-
-    def transform_account(self, account: str) -> str:
```
pankajkoti, line 78: Should we also remove this method from the other 3 modules `user_encrypted_privatekey_env_variable.py`, `user_pass.py` and `user_encrypted_privatekey_file.py`?

pankajkoti (approved):
LGTM, great quick fix. I have a question inline and we might need another test to cover the missing line for the CI to pass.

cosmos/profiles/snowflake/user_privatekey.py
tatiana: Done in 8abaee64c08be1d6fa98ea5fd430170f6daa886b

Issue #1379, opened 2024-12-10
Unpin dbt-databricks version


pankajastro:
dbt-databricks has pin the pydantic https://github.com/databricks/dbt-databricks/blob/main/pyproject.toml#L33-L33

Added in PR: https://github.com/databricks/dbt-databricks/pull/843

But the newer version of Airflow requires pydantic 2.0.

For Airflow < 2.8 working fine: https://github.com/astronomer/astronomer-cosmos/actions/runs/12259335118/job/34201293598

Scope of this issue to unpin dbt-databricks version which is done in PR: https://github.com/astronomer/astronomer-cosmos/pull/1376

PR #1409, opened 2024-12-18
Unpin dbt-databricks version

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1379

A new version of dbt-databricks (v1.9.1) has been released, [release notes](https://github.com/databricks/dbt-databricks/releases/tag/v1.9.1). It is expected that PR: https://github.com/databricks/dbt-databricks/pull/874 will resolve the existing dependency issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,8 +44,7 @@ dbt-all = [
     "dbt-athena",
     "dbt-bigquery",
     "dbt-clickhouse",
-    # TODO: https://github.com/astronomer/astronomer-cosmos/issues/1379
-    "dbt-databricks<1.9",
+    "dbt-databricks!=1.9.0",
     "dbt-exasol",
     "dbt-postgres",
     "dbt-redshift",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -44,8 +44,7 @@ dbt-all = [
     "dbt-athena",
     "dbt-bigquery",
     "dbt-clickhouse",
-    # TODO: https://github.com/astronomer/astronomer-cosmos/issues/1379
-    "dbt-databricks<1.9",
+    "dbt-databricks!=1.9.0",
```
pankajkoti, line 47: would be nice to add a comment here on why we don't allow 1.9.0

pankajkoti (approved):
One comment inline, other than that looks good to me

pyproject.toml
pankajastro: Added a comment in 5595e0191f6a4cc6c09162ff2f883f61c0684b12

Issue #1410, opened 2024-12-19
Introduce conditional run for 1000 models in performance tests


pankajkoti:
PR #1403 introduced running 500 & 1000 models in the matrix for the performance tests jobs. However, I'm observing that running although the job for 500 models finished in roughly 12-15 mins, the job for 1000 models runs for about 40+ mins. I suggest we drop running 1000 models with PR and later introduce these 1000 models to run on a conditional basis e.g. when a certain label (run-large-no-models) is added to the PR.

tatiana:
We're closing this issue since we have not been actively using this feature, and we believe the benchmark being developed in https://github.com/astronomer/cosmos-benchmark is more representative of real dbt use-cases.

PR #1411, opened 2024-12-19
Drop running 1000 models in the CI

pankajkoti:
With PR #1403 , we added to the matrix to run tests on 500 & 1000 models. Although the job for
500 models completes in under 15 mins, the job for 1000 models is taking significantly longer ~40+ mins as can be seen here https://github.com/astronomer/astronomer-cosmos/actions/runs/12400943145/job/34619249671. 

I suggest we drop running 1000 models and later add it to the matrix to conditionally run only when a certain label is added to the PR i.e. when we would actually like to benchmark on specific PRs. I have created a follow-up issue for this https://github.com/astronomer/astronomer-cosmos/issues/1410

related: #1410

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -354,7 +354,7 @@ jobs:
       matrix:
         python-version: ["3.11"]
         airflow-version: ["2.8"]
-        num-models: [1, 10, 50, 100, 500, 1000]
+        num-models: [1, 10, 50, 100, 500]
     services:
       postgres:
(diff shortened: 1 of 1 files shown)


tatiana (approved):
Makes sense, @pankajkoti , I wonder if we should change our performance tests to use Astro. That would give a more realistic performance estimate.

pankajkoti:
I am merging this as we have multiple open PRs at the moment and it will help there once rebased.

PR #1412, opened 2024-12-19
To support task display_name

pankajastro:


cosmos/airflow/graph.py
```diff
@@ -188,31 +210,30 @@ def create_task_metadata(
             "dbt_dag_task_group_identifier": dbt_dag_task_group_identifier,
         }
         if test_behavior == TestBehavior.BUILD and node.resource_type in SUPPORTED_BUILD_RESOURCES:
-            task_id = f"{node.name}_{node.resource_type.value}_build"
+            task_id, args = _get_task_id_and_args(node, args, use_task_group, normalize_task_id, "build")
```
pankajastro, line 213: @tatiana Any reason why we include resource type in build task-id?

tatiana: @pankajastro because we can have a build for model, seed and snapshot, and I thought users would like to see it in the Airflow UI some type of differentiation if something is a seed / model / snapshot

PR #1414, opened 2024-12-19
Remove unused docs dependency

pankajastro:
This PR clean-up requires dependency for building and publishing docs


tatiana (approved):
Thanks for cleaning this up, @pankajastro ! This looks much better!

PR #1415, opened 2024-12-19
Support rendering build operator task-id with non-ASCII characters

pankajastro:
In PR [#1278](https://github.com/astronomer/astronomer-cosmos/pull/1278), we introduced support for rendering non-ASCII 
characters in task IDs. However, due to limited access, we 
were unable to implement the same functionality for the 
build operator. This PR aims to extend that functionality 
by adding support for rendering build task IDs with 
non-ASCII characters.

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -138,18 +138,22 @@ def _get_task_id_and_args(
     use_task_group: bool,
     normalize_task_id: Callable[..., Any] | None,
     resource_suffix: str,
+    include_resource_type: bool = False,
 ) -> tuple[str, dict[str, Any]]:
     """
     Generate task ID and update args with display name if needed.
     """
     args_update = args
+    task_display_name = f"{node.name}_{resource_suffix}"
+    if include_resource_type:
+        task_display_name = f"{node.name}_{node.resource_type}_{resource_suffix}"
     if use_task_group:
         task_id = resource_suffix
     elif normalize_task_id:
         task_id = normalize_task_id(node)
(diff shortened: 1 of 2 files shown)


cosmos/airflow/graph.py
```diff
@@ -214,7 +218,7 @@ def create_task_metadata(
         }
 
         if test_behavior == TestBehavior.BUILD and node.resource_type in SUPPORTED_BUILD_RESOURCES:
-            task_id = f"{node.name}_{node.resource_type.value}_build"
+            task_id, args = _get_task_id_and_args(node, args, use_task_group, normalize_task_id, "build", True)
```
pankajkoti, line 221:
```suggestion
            task_id, args = _get_task_id_and_args(node, args, use_task_group, normalize_task_id, "build", include_resource_type=True)
```
since it's a keyword arg, good to reference the arg while passing the value.

tatiana (commented):
Great work, @pankajastro , this was a nice catch and very quick implementation!
Happy for us to merge this once we have unit tests covering this use case (non-ascii model name + build task metadata)

cosmos/airflow/graph.py
pankajastro: Fixed in: ec13fdbeb4d8eee15491c811072763c8ca75fe17

pankajastro:
> Great work, @pankajastro , this was a nice catch and very quick implementation! Happy for us to merge this once we have unit tests covering this use case (non-ascii model name + build task metadata)

Added tests in commit: ec13fdbeb4d8eee15491c811072763c8ca75fe17

tatiana (approved):

Issue #1420, opened 2024-12-23
[Bug] DbtDocsGCSLocalOperator.upload_to_cloud_storage() got an unexpected keyword argument 'context'


tuantran0910:
### Astronomer Cosmos Version

1.8.0

### dbt-core version

1.9.0

### LoadMode

AUTOMATIC

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

2.10.3

### Operating System

linux

### Deployment

Official Apache Airflow Helm Chart

### What happened?

Since `astronomer-cosmos` version `v1.8.0` has been released recently, I have found a new bug related to `dbt docs`

In the file `cosmos/operators/local.py`, the `context` is passed as parameter to the `callback` function:

```python
class DbtLocalBaseOperator(AbstractDbtBaseOperator):

      # Other logic 

      def run_command(
          self,
          cmd: list[str],
          env: dict[str, str | bytes | os.PathLike[Any]],
          context: Context,
      ) -> FullOutputSubprocessResult | dbtRunnerResult:

              # Other logic

              if self.callback:
                  self.callback_args.update({"context": context})
                  self.callback(tmp_project_dir, **self.callback_args)
```

The class `DbtDocsCloudLocalOperator` has the method `upload_to_cloud_storage` and assign it as the `callback` function:

```python
class DbtDocsCloudLocalOperator(DbtDocsLocalOperator, ABC):
    """
    Abstract class for operators that upload the generated documentation to cloud storage.
    """

    def __init__(
        self,
        connection_id: str,
        bucket_name: str,
        folder_dir: str | None = None,
        **kwargs: Any,
    ) -> None:
        """Initializes the operator."""
        self.connection_id = connection_id
        self.bucket_name = bucket_name
        self.folder_dir = folder_dir

        super().__init__(**kwargs)

        # override the callback with our own
        self.callback = self.upload_to_cloud_storage

    @abstractmethod
    def upload_to_cloud_storage(self, project_dir: str) -> None:
        """Abstract method to upload the generated documentation to cloud storage."""
```

Here the `upload_to_cloud_storage` method is missing the parameter `context`, then it causes the error: `TypeError: DbtDocsGCSLocalOperator.upload_to_cloud_storage() got an unexpected keyword argument 'context'`

### Relevant log output

```shell
  File "/opt/airflow/dag.py", line 68, in docs_dag
    DbtDocsGCSOperator(
[...]

PR #1422, opened 2024-12-23
Add `kwargs` param in DocsOperator method upload_to_cloud_storage

pankajastro:
This PR addresses an issue where the `callback` callable parameter's
handling was modified in PR https://github.com/astronomer/astronomer-cosmos/pull/1389 leading to a change in 
the way keyword arguments like context were passed to the callable.

However, the Docs Operator has its own implementation of the callback callable, 
which only expects a single parameter, project_dir. As a result, passing extra 
keyword arguments like context is causing a mismatch and resulting in the error described in #1420.

This PR add kwargs param in upload_to_cloud_storage for Docs operators

We have integration tests for these operator but look like CI does have not required setup and it get ignored.
https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/dbt_docs.py

related: #1420

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+1.8.1a1 (2024-12-23)
+--------------------
+
+Bug Fixes
+
+* Add kwargs param in DocsOperator method upload_to_cloud_storage by @pankajastro in #1422
+
 1.8.0 (2024-12-20)
 --------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ Changelog
 
 Bug Fixes
 
-* Add kwargs param in DocsOperator method upload_to_cloud_storage by @pankajastro in #1422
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+1.8.1a1 (2024-12-23)
+--------------------
+
+Bug Fixes
+
+* Add kwargs param in DocsOperator method upload_to_cloud_storage by @pankajastro in #1422
```
pankajkoti, line 9:
```suggestion
* Add ``kwargs`` param in DocsOperator method ``upload_to_cloud_storage`` by @pankajastro in #1422
```

pankajkoti (approved):
Thanks for the quick fix @pankajastro . Indeed, it appears that the example DAG we have was silently not running the tasks as it could not find the connections. I have created a PR #1428 to adapt to our existing CI connections and run this example DAG and as can be seen in the action https://github.com/astronomer/astronomer-cosmos/actions/runs/12500718837/job/34877247783?pr=1428 it failed with the same reason as in #1420. 

I will rebase my PR after merging this one to verify CI passes after running the docs DAG.

PR #1432, opened 2024-12-27
Install providers with airflow

pankajkoti:


tatiana (approved):
Thanks for sorting this, @pankajkoti !

PR #1449, opened 2025-01-13
update openlineage to avoid deprecation warning

CorsettiS:
## Description

openlineage suggests using event_v2 as the new import call. This can be seen in the source code of openlineage itself for the [run](https://github.com/OpenLineage/OpenLineage/blob/main/client/python/openlineage/client/run.py) method.

## Related Issue(s)

UserWarning: {message : DeprecationWarning('This module is deprecated. Please use `openlineage.client.event_v2`.'), category : 'DeprecationWarning', filename : '/home/airflow/.local/lib/python3.10/site-packages/openlineage/client/filter.py', lineno : 9, line : None}

## Breaking Change?

No

## Checklist

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -46,7 +46,7 @@ else:
 if TYPE_CHECKING:
     from airflow.datasets import Dataset  # noqa: F811
     from dbt.cli.main import dbtRunner, dbtRunnerResult
-    from openlineage.client.run import RunEvent
+    from openlineage.client.event_v2 import RunEvent
 
 
 from sqlalchemy.orm import Session


cosmos/operators/local.py
```diff
@@ -46,7 +46,7 @@
 if TYPE_CHECKING:
     from airflow.datasets import Dataset  # noqa: F811
     from dbt.cli.main import dbtRunner, dbtRunnerResult
-    from openlineage.client.run import RunEvent
+    from openlineage.client.event_v2 import RunEvent
```
pankajastro, line 49: If we are planning to do this, we may have to upgrade open lineage dependency

tatiana: It is important we support the old version of the OL package and the new, perhaps using a try/catch exception statement.

Also, if possible, we'd add some test. Perhaps we could have a subset of the tests using an older version of OL, eg.:
https://github.com/astronomer/astronomer-cosmos/blob/main/.github/workflows/test.yml#L270

tatiana (commented):
This is really helpful, @CorsettiS , thank you very much for pointing this out and submitting a fix. I left some minor feedback in-line

cosmos/operators/local.py
CorsettiS: @pankajkoti fair enough, I have added the try-except condition. It seems like a very small case to add subsets of tests on top of it IMO tho

pankajkoti: thanks @CorsettiS . Can we then add `  # pragma: no cover` to ignore those lines from test coverage so that the CI gets green?

pankajkoti (approved):

Issue #1444, opened 2025-01-09
[Bug] Relationship tests show up with TestBehaviour.NONE


roydobbe:
### Astronomer Cosmos Version

1.8.1

### dbt-core version

1.8.1

### Versions of dbt adapters

dbt-synapse 1.8.1

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

VIRTUALENV

### InvocationMode

SUBPROCESS

### airflow version

2.10.4

### Operating System

-

### Deployment

Astronomer

### What happened?

After upgrading to Cosmos 1.8.1, relationship tests show up in our TaskGroup even when we set TestBehaviour to NONE

These tasks dont' seem to do anything according to the logs

Example:

```python
DbtTaskGroup(
        group_id=f"dbt_{select.replace(':', '_')}",
        project_config=ProjectConfig(
            "/usr/local/airflow/dags/dbt/dwh",
            manifest_path="/usr/local/airflow/dags/dbt/dwh/target/manifest.json",
            dbt_vars=dbt_vars,
        ),
        profile_config=profile_config(conn_id=conn_id, threads=dbt_threads),
        execution_config=ExecutionConfig(
            dbt_executable_path=f"{os.environ['AIRFLOW_HOME']}/dbt_synapse_venv/bin/dbt",
        ),
        operator_args={
            "install_deps": True,
            "queue": queue,
            "pool": pool,
            "full_refresh": full_refresh,
            "dbt_cmd_global_flags": ["--cache-selected-only"],
            "callback": upload_run_results_to_blob,
        },
        render_config=RenderConfig(
            test_behavior=TestBehaviour.NONE,
            select=[select],
            load_method=LoadMode.DBT_MANIFEST,
        ),
    )
```
![Image](https://github.com/user-attachments/assets/1630c29b-79b8-4028-9271-5ec61052a4bf)

tests are set up as followed:

```yaml
  - name: id_xxxx
    tests:
      - not_null
      - relationships:
          to: ref("xxxx")
          field: id_xxx
          config:
            severity: warn
```

### Relevant log output

```shell
[...]

PR #1433, opened 2024-12-27
Fix rendering dbt tests with multiple parents

tatiana:
If these two circumstances are met:
1. The dbt project has tests that rely on multiple parent models and;
2. The `DbtDag` or `DbtTaskGroup` use `TestBehavior.AFTER_EACH` (default) or `TestBehavior.BUILD`

Cosmos 1.8.0 and previous versions would attempt to run the same test multiple times after each parent model run, likely failing if any of the parents hadn't been run yet.

This PR aims to fix this behaviour by not running tests with multiple dependencies within each task group / build task - and by adding those tests to run only once and after all parents have run.

# Related issues

Closes: #978
Closes: #1365

This change also sets the ground for adding support to tests that don't have any dependencies, a problem discussed in the following tickets:

* #959
* #1242
* #1279

# How to reproduce

There are two steps to reproduce this problem:

1. To create a representative dbt project
2. To create a Cosmos `DbtDag` that uses this dbt project to reproduce the original problem

## Representative dbt project

We created a dbt project named `multiple_parents_test` that has a test called`custom_test_combined_model` that depends on two models:
* combined_model
* model_a

The expectation from a user perspective is that, since the `combined_model` depends on `model_a`, that the  `multiple_parents_test` will only be run after both models were run, once.

Definitions of the test:
```
{% test custom_test_combined_model(model) %}
WITH source_data AS (
    SELECT id FROM {{ ref('model_a') }}
),
combined_data AS (
    SELECT id FROM {{ model }}
)
SELECT
    s.id
FROM
    source_data s
LEFT JOIN
    combined_data c
    ON s.id = c.id
WHERE
    c.id IS NULL
{% endtest %}
```

By running the following `dbt build` command, we confirm that the test depends on both models:

```
dbt build --select "+custom_test_combined_model_combined_model_"
11:59:29  Running with dbt=1.8.2
11:59:29  Registered adapter: postgres=1.8.1
11:59:29  Found 3 models, 6 data tests, 414 macros
11:59:29  
11:59:30  Concurrency: 4 threads (target='dev')
11:59:30  
11:59:30  1 of 9 START sql view model public.model_a ..................................... [RUN]
11:59:30  2 of 9 START sql view model public.model_b ..................................... [RUN]
11:59:30  1 of 9 OK created sql view model public.model_a ................................ [CREATE VIEW in 0.18s]
11:59:30  2 of 9 OK created sql view model public.model_b ................................ [CREATE VIEW in 0.18s]
[...]

PR #1463, opened 2025-01-14
 Fix rendering detached tests in `TestBehavior.NONE` and `AFTER_ALL`

tatiana:
As part of #1433 (Cosmos 1.8.1), we introduced a bug: detached test nodes were being incorrectly displayed when using:
  - `TestBehavior.NONE` (when no tests should be run)
  - `TestBehavior.AFTER_ALL` (when all tests should be run by the end of the DAG, regardless of their type)
   
 This PR solves this issue.
    
 Closes: #1444

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -409,6 +409,27 @@ def _get_dbt_dag_task_group_identifier(dag: DAG, task_group: TaskGroup | None) -
     return dag_task_group_identifier
 
 
+def calculate_detached_nodes(
+    nodes: dict[str, DbtNode],
+    test_behavior: TestBehavior,
+    detached_nodes: dict[str, DbtNode],
+    detached_from_parent: dict[str, list[DbtNode]],
+) -> None:
+    """
+    Given the nodes that represent a dbt project and the test_behavior, calculate detached test nodes
+    (test nodes that have multiple dependencies and should run independently).
+
+    Change in-place the dictionaries detached_nodes (detached node ID : node) and detached_from_parent (parent node ID that
+    is upstream to this test and the test node).
+    """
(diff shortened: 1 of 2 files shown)


cosmos/airflow/graph.py
```diff
@@ -409,6 +409,27 @@ def _get_dbt_dag_task_group_identifier(dag: DAG, task_group: TaskGroup | None) -
...
+
+    Change in-place the dictionaries detached_nodes (detached node ID : node) and detached_from_parent (parent node ID that
+    is upstream to this test and the test node).
+    """
+    if test_behavior in (TestBehavior.BUILD, TestBehavior.AFTER_EACH):
+        for node_id, node in nodes.items():
+            if is_detached_test(node):
+                detached_nodes[node_id] = node
+                for parent_id in node.depends_on:
+                    detached_from_parent[parent_id].append(node)
```
pankajkoti, line 430:
```suggestion
    if test_behavior not in (TestBehavior.BUILD, TestBehavior.AFTER_EACH):
        return
    for node_id, node in nodes.items():
        if not is_detached_test(node):
            continue
        detached_nodes[node_id] = node
        for parent_id in node.depends_on:
            detached_from_parent[parent_id].append(node)
```

suggestion to reduce a couple of levels of complexity if you like it based on https://medium.com/swlh/return-early-pattern-3d18a41bba8 but I know everyone may not feel the same way :)

pankajkoti (approved):
LGTM. Minor suggestion if you'd like to consider otherwise also happy with as is now.

cosmos/airflow/graph.py
tatiana: Thanks, @pankajkoti ! Since it's within our accepted complexity levels, I suggest we keep, but it is certainly worth to keep this in mind if we want to reduce the complexity level further.

Issue #1440, opened 2025-01-06
[Bug]


ms32035:
### Astronomer Cosmos Version

1.8.1

### dbt-core version

1.9.0

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

KUBERNETES

### InvocationMode

None

### airflow version

2.10.4

### Operating System

Debian GNU/Linux 12 (bookworm)

### Deployment

Official Apache Airflow Helm Chart

### What happened?

Looks like for #1433 in case of some tests provided by packages like dbt_utils or dbt_expectations the names generated are absurdly long 
and result in:
```
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 968, in __init__
    validate_key(task_id)
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/utils/helpers.py", line 55, in validate_key
    raise AirflowException(f"The key has to be less than {max_length} characters")
airflow.exceptions.AirflowException: The key has to be less than 250 characters
```

Additionally, it seems that tests are generated as their own task regardless of the `test_behaviour` setting. Specifically, these are generated when the value is `AFTER_ALL`

### Relevant log output

```shell

```

### How to reproduce

Create a test using `dbt_utils.relationships_where` where table and column names are long

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

pankajkoti:
Thanks for reporting the issue here. 

@ms32035 @jtrobles-cdd just curious was it only Cosmos version upgrade you have or also Airflow version upgrade?

Also, since Airflow is limiting the task key to be less than 250 keys, how would you propose we could curtail the task key at the same time make it identifiable of what node it's representing. Any ideas here could help us solve the issue in a user-friendly manner. 

@ms32035 regarding 

> it seems that tests are generated as their own task regardless of the test_behaviour setting. Specifically, these are generated when the value is AFTER_ALL

That's an expected behaviour as is described in the docs https://github.com/astronomer/astronomer-cosmos/blob/main/docs/configuration/testing-behavior.rst, you can refer on that page to the section for `Example when changing the behavior to use TestBehavior.AFTER_ALL:`

jtrobles-cdd:
> did you only upgrade Cosmos, or was there also an upgrade to Airflow?

Only Cosmos. I've been using Airflow 2.10.4 (with Astro Runtime 12.6.0) for many weeks without issues.

[...]

PR #1464, opened 2025-01-14
Fix detached test tasks names so they do not exceed 250 chars

tatiana:
Since we introduced detached test tasks in #1433 (released in 1.8.0), users started facing issues due to very long task names exceeding Airflow's limits.
    
Example of Python traceback reported by user:
```
      File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 968, in __init__
        validate_key(task_id)
      File "/home/airflow/.local/lib/python3.12/site-packages/airflow/utils/helpers.py", line 55, in validate_key
        raise AirflowException(f"The key has to be less than {max_length} characters")
    airflow.exceptions.AirflowException: The key has to be less than 250 characters
```
    
This PR fixes this issue. In case the name exceeds Airflow's limit (250 ATM), it will name the detached test using:
- "detached_{incremental unique number}_test"

We also considered naming the new test using:
- "parent1_parent2_..._test" - but that may not solve the issue, especially in circumstances where the same parents may have multiple detached nodes. Perhaps in future, we could bundle them in a single task.

Closes: #1440

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -430,6 +430,13 @@ def identify_detached_nodes(
                     detached_from_parent[parent_id].append(node)
 
 
+def calculate_detached_node_name(node: DbtNode) -> str:
+    """
+    Given a detached test node, calculate its name.
+    """
+    return f"{node.resource_name.split('.')[0]}_test"
+
+
 def build_airflow_graph(
     nodes: dict[str, DbtNode],
     dag: DAG,  # Airflow-specific - parent DAG where to associate tasks and (optional) task groups
@@ -503,6 +510,21 @@ def build_airflow_graph(
         if task_or_group is not None:
             logger.debug(f"Conversion of <{node.unique_id}> was successful!")
             tasks_map[node_id] = task_or_group
+
+    # Handle detached test nodes
+    for node_id, node in detached_nodes.items():
+        datached_node_name = calculate_detached_node_name(node)
+        test_meta = create_test_task_metadata(
+            datached_node_name,
+            execution_mode,
+            test_indirect_selection,
+            task_args=task_args,
+            on_warning_callback=on_warning_callback,
+            render_config=render_config,
+            node=node,
+        )
+        test_task = create_airflow_task(test_meta, dag, task_group=task_group)
+        tasks_map[node_id] = test_task
 
     # If test_behaviour=="after_all", there will be one test task, run by the end of the DAG
     # The end of a DAG is defined by the DAG leaf tasks (tasks which do not have downstream tasks)
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -410,6 +410,13 @@ def _get_dbt_dag_task_group_identifier(dag: DAG, task_group: TaskGroup | None) -
     return dag_task_group_identifier
 
 
+def should_create_detached_nodes(test_behavior: TestBehavior) -> bool:
+    """
+    Decide if we should calculate / insert detached nodes into the graph.
+    """
+    return test_behavior in (TestBehavior.BUILD, TestBehavior.AFTER_EACH)
+
+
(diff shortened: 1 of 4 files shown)


cosmos/airflow/graph.py
```diff
@@ -409,6 +410,21 @@ def _get_dbt_dag_task_group_identifier(dag: DAG, task_group: TaskGroup | None) -
     return dag_task_group_identifier
 
 
+def calculate_detached_node_name(node: DbtNode, counter: list[int] = [0]) -> str:
```
pankajkoti, line 413:
```suggestion
def calculate_detached_node_name(node: DbtNode, counter: list[int] | None = None) -> str:
```

cosmos/airflow/graph.py
```diff
@@ -409,6 +410,21 @@ def _get_dbt_dag_task_group_identifier(dag: DAG, task_group: TaskGroup | None) -
...
+    """
+    Given a detached test node, calculate its name. It will either be:
+     - the name of the test with a "_test" suffix, if this is smaller than 250
+     - or detached_{an incremental number}_test
+    """
```
pankajkoti, line 418:
```suggestion
    """
    if counter is None:
        counter = [0]
```

pankajkoti (approved):
LGTM. Minor suggestion in-line to avoid initialising list in the args.

cosmos/airflow/graph.py
tatiana: @pankajkoti this cannot be `None`, because the method relies on the signature to be able to be incremental

cosmos/airflow/graph.py
pankajkoti: ahh I see. Would it be possible to initialise the counter to [0] in the calling function before calling for the first time and then reusing the list somehow while calling it in the subsequent calls? I am not sure if there could be some side-effects due to this. Meaning, the counter would keep on incrementing always across parses of all the DAGs until a restart of the worker, no? 

Also happy to keep as is and observe and perhaps we could refactor later if needed.

cosmos/airflow/graph.py
pankajkoti: thanks. Wrote my response in https://github.com/astronomer/astronomer-cosmos/pull/1464#discussion_r1916490212

cosmos/airflow/graph.py
tatiana: That's a very good question, @pankajkoti .

The counter is restarted each time a new process is created.

From a scheduler DAG parsing perspective, the `DagFileProcessorManager` process executes an infinite loop that determines which files need to be processed, and it spawns separate `DagFileProcessorProcess` processes that are started to convert an individual file into one or more DAG objects. So we should be safe:
https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/dagfile-processing.html

From a worker node perspective, each time a task scheduled, we're running a new process to reparse the DAG and then execute it, so I also believe we're safe.

pankajkoti: yes, thanks for sharing the link Tatiana. I believe with that we're safe.

tatiana: BTW: just giving additional context, I checked with @ashb, and he confirmed that this applies to Airflow 2.x, and it may apply to Airflow 3.x, but he pointed out - and we all agree - that we're using an internal implementation detail of Airflow that may change.

I've logged a follow up ticket for us to review and improve this further: #1469

tatiana: I also refactored so we don't rely on Python's [default mutable object behaviour](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil) in 7335ef6bb9f45104ddf5d1e873d3482c8f8e5fb3

pankajkoti (approved):

PR #1466, opened 2025-01-14
Fix custom "+tag:some_tag" selector issue related to tests tag inheritance

tatiana:
The selector method `_should_include_node` changes test tasks to inherit tags from their parent nodes. While this behaviour is acceptable and desirable in some cases, it can cause problems using graph selectors with tags.

This PR improves the test coverage, narrows down the problem and fixes the problem reported by Astronomer customers. More details below.

A user reported that they see the correct `DbtDag` when using Cosmos 1.8.1 with:

* `LoadMode.DBT_LS`
* `RenderConfig(selector="accounts_marts")`

Where the selector `accounts_marts` is defined as:

```
    - name: accounts_marts
      description: Run Accounts models
      definition:
        intersection:
          - '+tag:accounts'
          - '+tag:datamart'
          - '+tag:stratus'
```

The expected behaviour includes:
- 164 Airflow tasks
- 152 Local run tasks
- 12 Snapshot tasks

However, when they attempt to run the same `DbtDag` using:

* `LoadMode.DBT_MANIFEST`
* `RenderConfig(select=["+tag:accounts,+tag:datamart,+tag:stratus"])`

Their `DbtDag` seems to have the wrong subset of nodes.

They reported:
- 197 Airflow tasks
- 183 Local run tasks
- 14 Snapshot tasks

This pull request aims to reproduce and fix this issue.

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -498,7 +498,6 @@ def select_nodes(
     exclude = exclude or []
     if not select and not exclude:
         return nodes
-
     validate_filters(exclude, select)
     subset_ids = apply_select_filter(nodes, project_dir, select)
     if select:
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -174,8 +174,6 @@ class GraphSelector:
 
         elif TAG_SELECTOR in self.node_name:
             tag_selection = self.node_name[len(TAG_SELECTOR) :]
-            # The following normally wuld return 20
-            # now it is returning 210
             root_nodes.update({node_id for node_id, node in nodes.items() if tag_selection in node.tags})
 
         elif CONFIG_SELECTOR in self.node_name:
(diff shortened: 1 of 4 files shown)


cosmos/dbt/selector.py
```diff
@@ -175,6 +174,8 @@ def filter_nodes(self, nodes: dict[str, DbtNode]) -> set[str]:
 
         elif TAG_SELECTOR in self.node_name:
             tag_selection = self.node_name[len(TAG_SELECTOR) :]
+            # The following normally wuld return 20
+            # now it is returning 210
```
pankajkoti, line 178: Do we need this comment?

pankajkoti (approved):
Looks great to me! Fantastic breakdown and analysis in identifying and addressing this fix 👏🏽

cosmos/dbt/selector.py
tatiana, line 178:
```suggestion
```

tatiana: Thanks for mentioning, @pankajkoti , we certainly don't need this - removed in 59dad89c130dd5b0f6b43e7c82709e9b4428a1bf

PR #1150, opened 2024-08-12
Extend Cosmos custom selector to support + when using paths and tags

mvictoria:
## Description

This PR resolves #1141 where dbt projects loaded with DBT_MANIFEST were unable to parse using the GraphSelector.

Changes:
- updated GraphSelector to take into account path and tag dbt selector methods
- update SelectorConfig to use regex to parse selection statement to handle graph and path statements

ToDo:
- ~add tests covering change~
- ~enable config dbt selector method~
- refactor

## Related Issue(s)

closes #1141
closes #823 

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1467, opened 2025-01-14
Improve Cosmos ``select`` docs to include latest graph operator support

tatiana:
To include the changes introduced in the PR #1150.

### DIFF
--- a/docs/configuration/selecting-excluding.rst
+++ b/docs/configuration/selecting-excluding.rst
@@ -17,7 +17,7 @@ The ``select`` and ``exclude`` parameters are lists, with values like the follow
 - ``config.materialized:table``: include/exclude models with the config ``materialized: table``
 - ``path:analytics/tables``: include/exclude models in the ``analytics/tables`` directory. In this example, `analytics/table` is a relative path, but absolute paths are also supported.
 - ``+node_name+1`` (graph operators): include/exclude the node with name ``node_name``, all its parents, and its first generation of children (`dbt graph selector docs <https://docs.getdbt.com/reference/node-selection/graph-operators>`_)
-- ``+/path/to/model_g+`` (graph operators): include/exclude all the nodes in the path ``/path/to/model_g``, their parents and children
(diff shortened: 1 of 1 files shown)


docs/configuration/selecting-excluding.rst
```diff
@@ -17,6 +17,9 @@ The ``select`` and ``exclude`` parameters are lists, with values like the follow
 - ``config.materialized:table``: include/exclude models with the config ``materialized: table``
 - ``path:analytics/tables``: include/exclude models in the ``analytics/tables`` directory
 - ``+node_name+1`` (graph operators): include/exclude the node with name ``node_name``, all its parents, and its first generation of children (`dbt graph selector docs <https://docs.getdbt.com/reference/node-selection/graph-operators>`_)
+- ``+/path/to/model_g+`` (graph operators): include/exclude all the nodes in the path ``/path/to/model_g``, their parents and children
```
pankajkoti, line 20: just trying to understand here if the path needs a leading slash and what would it mean. a leading slash would usually mean it's an absolute path, no? does it have to be a relative path `+path/to/model_g+` instead of the current `+/path/to/model_g+` to not include the leading slash?

pankajkoti (approved):
LGTM. One question in-line for my understanding.

docs/configuration/selecting-excluding.rst
tatiana: @pankajkoti This example uses an absolute path, but it should also work with relative paths if users do not add the leading slash:
https://github.com/astronomer/astronomer-cosmos/blob/967dd17bc91fe297a58c54f87c07fad5bb921d05/cosmos/dbt/selector.py#L95-L97

tatiana, line 20:
```suggestion
- ``+/path/to/model_g+`` (graph operators): include/exclude all the nodes in the absolute path ``/path/to/model_g``, their parents and children. Relative paths are also supported.
```

tatiana: I made some improvements to the descriptions so this is clearer: 9a5eaf8b758dd4872f86eff695cb290b36b00920, 0f08af2602b08ac1bf18fc164c5b5729de4bca0a

Issue #1469, opened 2025-01-15
Refactor detached tests implementation to group based on test parents


tatiana:
Since Cosmos 1.8.1, particularly change #1433, if a test has multiple parents, we create an independent test task for each.
Our original approach was to name the task based on the test name, which led to issue #1440.

This bug affected many users, so we released a quick fix, #1464, that would change the test name to "detached_{incremental unique number}_test" if the name were longer than 250 characters. This implementation, however, relies on an internal Airflow implementation detail: that a new process always parses the DAG, as described in https://github.com/astronomer/astronomer-cosmos/pull/1464/files#r1916648300. This is true for Airflow 2.x and should remain True in Airflow 3.0 - but it is an internal implementation detail.

The goal with this ticket is to refactor the detached test implementation so we only have a single detached task for each group of parents. This way, we could name the task either:
a) "parent1_parent2_..._test"
b) or "<deterministic_hash([parent1, parent2])>_test", if (a) is bigger than 250 chars

filipe-alves-pleo:
It would be great if the refactoring could handle the relationship tests as a special case either:

1. Relationship tests for the same model are grouped under a single task like `model1_test`
2. Relationship tests are not treated as detached tasks but as a special case in the model's test task (instead of creating additional tasks), and required model dependencies added to the test task directly.

Note: To Identify the correct model to which the test belongs, use the property`$.nodes[*].attached_node` in the dbt manifest.

### Reasoning:

- Relationship tests are a common practice hence they come in big numbers, adding each one as a separate task creates overhead for the DAG execution and adverse impact in the UI usability. This is different from other multi-parent tests that are more specific (and limited in numbers), and also not clearly defined as belonging to only one of the parent nodes.
- Relationship tests are effectively testing the child model (akin to a FK consistency check in a database), it makes sense to have them centralised in a single test task bunching together all of them per model for readability.

### Final thoughts:
This concept could be expanded to any other test type containing the `attached_node` property. Those are the cases where the test is explicitly defined in a single model although need to reference a second model as part of the test.

alt-theo:
bump, can we include the additional suggestion from @filipe-alves-pleo in the next milestone as well?

pankajkoti:
[...]

PR #1470, opened 2025-01-15
Allow users to opt-in or out (default) of detached test nodes

tatiana:
Since we introduced detached test tasks to fix a customer issue in #1433 (release 1.8.1), we changed how Cosmos renders DAGs, with the chance that Cosmos significantly changed how it renders a dbt project - even when users did not change their `DbtDag` or `DbtTaskGroup` configuration. This is unacceptable in a micro release - and for this reason we're reverting this change and making the feature opt-in.

PR #1433 led to issues such as #1464, reported by multiple Cosmos users, and also issues that did not become Github issues, such as an Astro customer who reported that when they upgraded to Cosmos 1.8.1, the number of tasks increased dramatically. One problem with #1433 was that it did not empower users to opt in or out of having detached test nodes, solving the problem for some but causing problems for many.

This PR aims to solve this problem by introducing a new property to `RenderConfig`: `should_detach_multiple_parents_tests`. We are reverting the Cosmos DAG rendering to what it was in 1.8.0 and before: by default, it will not detach tests with multiple parents. Users must opt-in for this behaviour if and when they want to.

We understand this may be perceived as a breaking change by some, but it is the correct way to move forward and avoid causing further disruption.

Ultimately, we'll review the current implementation, as described in #1469, but with this change, we're documenting the current behaviour and empowering users.


cosmos/config.py
```diff
@@ -64,6 +64,7 @@ class RenderConfig:
     :param source_rendering_behavior: Determines how source nodes are rendered when using cosmos default source node rendering (ALL, NONE, WITH_TESTS_OR_FRESHNESS). Defaults to "NONE" (since Cosmos 1.6).
     :param airflow_vars_to_purge_dbt_ls_cache: Specify Airflow variables that will affect the LoadMode.DBT_LS cache.
     :param normalize_task_id: A callable that takes a dbt node as input and returns the task ID. This allows users to assign a custom node ID separate from the display name.
+    :param should_detach_multiple_parents_tests: A boolean for users to control if they want to run in separate tasks tests that depend on multiple parents
```
pankajkoti, line 67:
```suggestion
    :param should_detach_multiple_parents_tests: A boolean that allows users to decide whether to run tests with multiple parent dependencies in separate tasks.
```

docs/configuration/testing-behavior.rst
```diff
@@ -111,3 +111,75 @@ When at least one WARN message is present, the function passed to ``on_warning_c
...
+
+While the standard behavior of Cosmos works for many cases, there are a few scenarios when the test fails unless both models
+run. To overcome this issue, starting in Cosmos 1.8.2, we introduced the parameter
+``should_detach_multiple_parents_tests`` in ``RenderConfig``. By default, it is ``False``. If it is set to ``True`` and
+``TestBehavior` is ``AFTER_EACH`` or ``BUILD``, Cosmos will identify all the test nodes that depend on multiple parents
+and will create a standalone test task for each of them.
+
+Cosmos will attempt to name this task after the test original name. Since some test names can be very long (over 250 characters)
+and Airflow does not support IDs longer than 250 characters; Cosmos will name them "detached_0_test", incrementing
+0 as needed.
```
pankajkoti, line 133:
```suggestion
Cosmos will attempt to name this task after the test's original name. However, since some test names can exceed 250 characters and Airflow does not support IDs longer than this limit, Cosmos will assign names like “detached_0_test,” incrementing the number as needed.
```

pankajkoti (approved):
Everything looks great to me. Thanks for the quick fix and tidying up/making it configurable around this issue!

Issue #1271, opened 2024-10-21
[async] Evaluate possibility of supporting macros when using `ExecutionMode.AIRFLOW_ASYNC`


tatiana:
**Context**

Many dbt macros are calculated during runtime. We should evaluate if/how we could overcome this limitation when using  `ExecutionMode.AIRFLOW_ASYNC`. If we can't, we should at least document this limitation.

**Acceptance criteria**

* [ ] Have a proposal to handle macros when using `ExecutionMode.AIRFLOW_ASYNC`
* [ ] Update documentation

Issue #1266, opened 2024-10-21
[async] Evaluate the possiblity of using dbt itself to create the full SQL command


tatiana:
**Context**

When implementing #1230, we realised that the `dbt compile` command outputs the select statements related to models and transformations, but not necessarily the remaining relevant parts of the query (including creates, updates, inserts, drops).

This logic lives partially in dbt-core code and partially in the dbt adaptors of interest.

**Acceptance criteria**

* [ ] Analyse the possibility of, during the `dbt compile` - or somewhere related, to a setup task - to pre-create the full queries that we want to run with the async operators afterwards.

tatiana:
Take into account if we'd be able to solve other related tickets by using this approach:
- https://github.com/astronomer/astronomer-cosmos/issues/1261
- https://github.com/astronomer/astronomer-cosmos/issues/1265

pankajastro:
**DBT Compile**
- This only generates the `select` statement for model, test, and analysis
- The `–empty` flag is used to avoid expensive read on data warehouse but still we get `select` query only 
> it's often useful to execute the underlying select statement to find the source of the bug
https://docs.getdbt.com/reference/commands/compile

**DBT Run**
- There is no `--dry-run` option in `dbt run` command 
- `–empty` flag: The `dbt run` command support `–empty` flag. The `--empty` flag, when used with the dbt run command, offers a behaviour similar to a dry-run by generating and running SQL statements without loading data into/from the data warehouse. https://docs.getdbt.com/reference/commands/run#the---empty-flag

In conclusion, while the --empty flag offers a lightweight, dry-run-like behaviour. So we may be able to use it in compile task to generate sql query but if we do this, we must keep in mind that
- The compile task will create table/view etc in data warehouse
- The overall run command may not be idempotent

pankajastro:
@pankajkoti and I discussed this.

**1. Empty Flag in Run Command:**

- The `--empty` flag in the dbt run command introduces additional query elements related to limits and parsing, which can make the query more prone to bugs.

Example

with empty
```
...
orders as (
    select * from (select * from "postgres"."postgres"."stg_orders" where false limit 0) _dbt_limit_subq_stg_orders
)
...
```

without empty

```
...
orders as (
    select * from "postgres"."postgres"."stg_orders"
),
...
```

**2. Empty Flag in Run Command to Create Table/View + Compile command to get full query**

- The `--empty` flag when used in the dbt run command can create a table or view without data
[...]

Issue #1265, opened 2024-10-21
[async] Support running models without `--full-refresh` when using `ExecutionMode.AIRFLOW_ASYNC`


tatiana:
**Context**

As of Cosmos 1.7, the `ExecutionMode.AIRFLOW_ASYNC` only allows running models with  `--full-refresh`. We should support also support `full_refresh: false` for models of especially large datasets, which you would never want dbt to fully drop and recreate.

**Acceptance criteria"**

* [ ] The `ExecutionMode.AIRFLOW_ASYNC` mode also supports running models without full refresh

Issue #1264, opened 2024-10-21
[async] Emit datasets when using `ExecutionMode.AIRFLOW_ASYNC`


tatiana:
**Context**

As of Cosmos 1.7, we introduced `ExecutionMode.AIRFLOW_ASYNC`. However, when using this mode, users are not benefiting from Cosmos automatic Airflow dataset emission both from an OpenLineage perspective and also from an Airflow dataset perspective.

**Acceptance criteria**

* [ ] When using `ExecutionMode.AIRFLOW_ASYNC` with BQ, Cosmos should emit Datasets the same way it does when using `ExecutionMode.LOCAL `

pankajkoti:
Since the async operator now extends the `DbtRunLocalOperator` as part of the work in PR #1474, the functionality is already available through `DbtRunLocalOperator`. Therefore, if we proceed with this approach, no additional work may be required to support this. We will monitor this issue alongside PR #1474 and validate once the work in that PR is completed.

PR #1474, opened 2025-01-21
Create and run accurate SQL statements when using `ExecutionMode.AIRFLOW_ASYNC`

pankajkoti:
# Overview

This PR introduces a reliable way to extract SQL statements run by `dbt-core` so Airflow asynchronous operators can use them. It fixes the experimental BQ implementation of `ExecutionMode.AIRFLOW_ASYNC` introduced in Cosmos 1.7 (#1230).

Previously, in #1230, we attempted to understand the implementation of how `dbt-core` runs `--full-refresh` for BQ, and we hard-coded the SQL header in Cosmos as an experimental feature. Since then, we realised that this approach was prone to errors (e.g. #1260) and that it is unrealistic for Cosmos to try to recreate the logic of how `dbt-core` and its adaptors generate all the SQL statements for different operations, data warehouses, and types of materialisation.

With this PR, we use `dbt-core` to create the complete SQL statements without `dbt-core` running those transformations. This enables better compatibility with various `dbt-core` features while ensuring correctness in running models.

The drawback of the current approach is that it relies on monkey patching, a technique used to dynamically update the behaviour of a piece of code at run-time. Cosmos is monkey patching `dbt-core` adaptors methods at the moment that they would generally execute SQL statements - Cosmos modifies this behaviour so that the SQL statements are writen to disk without performing any operations to the actual data warehouse.

The main drawback of this strategy is in case dbt changes its interface. For this reason, we logged the follow-up ticket https://github.com/astronomer/astronomer-cosmos/issues/1489 to make sure we test the latest version of dbt and its adapters and confirm the monkey patching works as expected regardless of the version being used.  That said, since the method being monkey patched is part of the `dbt-core` interface with its adaptors, we believe the risks of breaking changes will be low.

The other challenge with the current approach is that every Cosmos task relies on the following:
1. `dbt-core` being installed alongside the Airflow installation
2. the execution of a significant part of the `dbtRunner` logic

We have logged a follow-up ticket to evaluate the possibility of overcoming these challenges: #1477

## Key Changes

1. Mocked BigQuery Adapter Execution:
    - Introduced `_mock_bigquery_adapter()` to override `BigQueryConnectionManager.execute`, ensuring SQL is only written to the `target` directory and skipping execution in the warehouse.
    - The generated SQL is then submitted using Airflow’s BigQueryInsertJobOperator in deferrable mode.
4. Refactoring `AbstractDbtBaseOperator`:
    - Previously, `AbstractDbtBaseOperator` inherited `BaseOperator`, causing conflicts when used with `BigQueryInsertJobOperator` with our`EXECUTIONMODE.AIRFLOW_ASYNC` classes and the interface built in #1483 
    - Refactored to `AbstractDbtBase` (no longer inheriting `BaseOperator`), requiring explicit `BaseOperator` initialization in all derived operators.
    - Updated the below existing operators to consider this refactoring needing derived classes to initialise `BaseOperator`:
        - `DbtAzureContainerInstanceBaseOperator`
        - `DbtDockerBaseOperator`
        - `DbtGcpCloudRunJobBaseOperator`
        - `DbtKubernetesBaseOperator`
5. Changes to dbt Compilation Workflow
    - Removed `_add_dbt_compile_task`, which previously pre-generated SQL and uploaded it to remote storage and subsequent task downloaded this compiled SQL for their execution.
    - Instead, `dbt run` is now directly invoked in each task using the mocked adapter to generate the full SQL.
    - A future [issue](https://github.com/astronomer/astronomer-cosmos/issues/1477) will assess whether we should reintroduce a compile task using the mocked adapter for SQL generation and upload, reducing redundant dbt calls in each task.

## Issue updates
The PR fixes the following issues:
1. closes: #1260 
    - Previously, we only supported --full-refresh dbt run with static SQL headers (e.g., CREATE/DROP TABLE).
    - Now, we support dynamic SQL headers based on materializations, including CREATE OR REPLACE TABLE, CREATE OR REPLACE VIEW, etc.
2. closes: #1271 
    - dbt macros are evaluated at runtime during dbt run invocation using mocked adapter, and this PR lays the groundwork for supporting them in async execution mode.
3. closes: #1265 
    - Now, large datasets can avoid full drops and recreations, enabling incremental model updates.
6. closes: #1261 
    - 	Previously, only tables (--full-refresh) were supported; this PR implements logic for handling different materializations that dbt supports like table, view, incremental, ephemeral, and materialized views.
7. closes: #1266 
    - Instead of relying on dbt compile (which only outputs SELECT statements), we now let dbt generate complete SQL queries, including SQL headers/DDL statements for the queries corresponding to the resource nodes and state of tables/views in the backend warehouse
8. closes: #1264 
    - We support emitting datasets for `EXECUTIONMODE.AIRFLOW_ASYNC` too with this PR

## Example DAG showing `EXECUTIONMODE.AIRFLOW_ASYNC` deferring tasks and the dynamic query submitted in the logs

Screenshot 2025-02-04 at 1 02 42 PM

## Next Steps & Considerations:
- It's acknowledged that using mock patching may have downsides, however, this currently seems the best approach to achieve our goals. It's understood and accepted the risks associated with this method. To mitigate them, we are expanding our test coverage to include all currently supported dbt adapter versions in our test matrix in #1489. This will ensure compatibility across different dbt versions and helps us catch potential issues early.
- Further validation of different dbt macros and materializations with `ExecutionMode.AIRFLOW_ASYNC` by seeking feedback from users by testing alpha https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.9.0a5 created with changes from this PR.
- https://github.com/astronomer/astronomer-cosmos/issues/1477, Compare the efficiency of generating SQL dynamically vs. pre-compiling and uploading SQL via a separate task.
- Add compatibility across all major cloud datawarehouse backends (dbt adapters).


cosmos/operators/local.py
```diff
@@ -475,8 +485,16 @@ def run_command(
...
+                if return_sql and sql_context:
+                    profile_type = sql_context["profile_type"]
+                    mock_adapter_callable = PROFILE_TYPE_MOCK_ADAPTER_CALLABLE_MAP.get(profile_type)
+                    if not mock_adapter_callable:
+                        raise CosmosValueError(
+                            f"Mock adapter callable function not available for profile_type {profile_type}"
+                        )
+                    mock_adapter_callable()
 
+                self.log.debug("Using environment variables keys: %s", env.keys())
```
pankajkoti, line 497:
```suggestion
```

cosmos/operators/local.py
```diff
@@ -475,8 +485,16 @@ def run_command(
 
                 full_cmd = cmd + flags
 
-                self.log.debug("Using environment variables keys: %s", env.keys())
+                if return_sql and sql_context:
```
pankajkoti, line 488:
```suggestion
                self.log.debug("Using environment variables keys: %s", env.keys())
                if return_sql and sql_context:
```

cosmos/airflow/graph.py
```diff
@@ -371,32 +370,6 @@ def generate_task_or_group(
     return task_or_group
 
 
-def _add_dbt_compile_task(
```
tatiana, line 374: As we agreed, I've added a task to follow up on this:
https://github.com/astronomer/astronomer-cosmos/issues/1477

Ideally, we'd compare both approaches before 1.9, but I'm happy with us to go ahead with approach of #1474 in 1.9 - it is already a significant improvement when compared to 1.7.0, since it solves the existing SQL query issue.

cosmos/mocked_dbt_adapters.py
```diff
@@ -0,0 +1,21 @@
...
+    import agate
+    from dbt.adapters.bigquery.connections import BigQueryAdapterResponse, BigQueryConnectionManager
+    from dbt_common.clients.agate_helper import empty_table
+
+    def execute(  # type: ignore[no-untyped-def]
+        self, sql, auto_begin=False, fetch=None, limit: Optional[int] = None
+    ) -> Tuple[BigQueryAdapterResponse, agate.Table]:
+        return BigQueryAdapterResponse("mock_bigquery_adapter_response"), empty_table()
+
+    BigQueryConnectionManager.execute = execute
```
tatiana, line 16: We may have to mock a different method when using `dbt build`. That said, this is outside of the scope of fixing the original but this PR solves, so I've logged a follow up ticket:
https://github.com/astronomer/astronomer-cosmos/issues/1476

cosmos/operators/airflow_async.py
```diff
@@ -60,13 +55,11 @@ class DbtSourceAirflowAsyncOperator(DbtBaseAirflowAsyncOperator, DbtSourceLocalO
...
 
 
-class DbtRunAirflowAsyncOperator(BigQueryInsertJobOperator):  # type: ignore
+class DbtRunAirflowAsyncOperator(BigQueryInsertJobOperator, DbtRunLocalOperator):  # type: ignore
 
     template_fields: Sequence[str] = (
         "full_refresh",
         "project_dir",
-        "gcp_project",
-        "dataset",
```
tatiana, line 69: Why did we remove these as templated fields?

cosmos/operators/airflow_async.py
```diff
@@ -60,13 +55,11 @@ class DbtSourceAirflowAsyncOperator(DbtBaseAirflowAsyncOperator, DbtSourceLocalO
     pass
 
 
-class DbtRunAirflowAsyncOperator(BigQueryInsertJobOperator):  # type: ignore
+class DbtRunAirflowAsyncOperator(BigQueryInsertJobOperator, DbtRunLocalOperator):  # type: ignore
```
tatiana, line 58: It would be great if we can see the SQL statement in the operator Templated Field, like we can do with the Cosmos Local operators. If this is too much work, I'm happy for it to be done in a follow up PR/ticket.

tatiana (commented):
@pankajkoti I'm very excited that we now have a more reliable way of calculating the full dbt SQL query. This approach fixes #1260 and solves many of the async tickets we have open.

Monkey-patching always carries a risk, but it is worth it at this stage.

It would be great if - either as part of this PR - or as a priority follow-up PR, we have an efficient way of testing that the monkey patching works in multiple versions of dbt, including the latest releases, and that the transformation is not being executed when we run the dbt command. I believe this must be done before we release this feature in 1.9.0

I've logged two follow-up tickets that are relevant:
- one is to consider the re-introduction of the compile task, if that means we can avoid having dbt installed in all dbt worker nodes, and executing the command in most Cosmos tasks: https://github.com/astronomer/astronomer-cosmos/issues/1477
- the other is to support `TestBehavior.BUILD` https://github.com/astronomer/astronomer-cosmos/issues/1476

It would be great if these could be accomplished before 1.9.0 release, but I'm also happy with us sticking to approach if time does not allow further analysis / implementation.

pankajkoti:
cc: @joppevos for visibility on the ongoing work

cosmos/operators/airflow_async.py
```diff
@@ -61,19 +61,16 @@ def __init__(  # type: ignore
         extra_context: dict[str, object] | None = None,
         **kwargs,
     ) -> None:
```
pankajkoti, line 63:
```suggestion
    ) -> None:

```

cosmos/operators/airflow_async.py
```diff
@@ -61,19 +61,16 @@ def __init__(  # type: ignore
...
-
         # Cosmos attempts to pass many kwargs that async operator simply does not accept.
         # We need to pop them.
         clean_kwargs = {}
         non_async_args = set(inspect.signature(AbstractDbtBaseOperator.__init__).parameters.keys())
         non_async_args |= set(inspect.signature(DbtLocalBaseOperator.__init__).parameters.keys())
         non_async_args -= {"task_id"}
```
pankajkoti, line 69:
```suggestion
        non_async_args -= {"task_id"}

```

cosmos/operators/airflow_async.py
```diff
@@ -18,7 +18,7 @@
     DbtTestLocalOperator,
 )
 
-_SUPPORTED_DATABASES = ["bigquery"]
+_SUPPORTED_DATABASES = [BIGQUERY_PROFILE_TYPE]
```
pankajkoti, line 21:
```suggestion
_SUPPORTED_DATABASES = ["bigquery"]
```

cosmos/airflow/graph.py
pankajkoti: I agree, thanks for creating the follow-up task.

cosmos/mocked_dbt_adapters.py
pankajkoti: I agree, thanks for creating the follow-up ticket.

cosmos/operators/airflow_async.py
pankajkoti: we have them back now

cosmos/operators/airflow_async.py
pankajkoti: this din't work trivially for me. I have logged a follow-up ticket to work on it https://github.com/astronomer/astronomer-cosmos/issues/1490

cosmos/operators/local.py
```diff
@@ -419,12 +420,21 @@ def _cache_package_lockfile(self, tmp_project_dir: Path) -> None:
...
+
     def run_command(
         self,
         cmd: list[str],
         env: dict[str, str | bytes | os.PathLike[Any]],
         context: Context,
-    ) -> FullOutputSubprocessResult | dbtRunnerResult:
+        return_sql: bool = False,
+        sql_context: dict[str, Any] | None = None,
+    ) -> FullOutputSubprocessResult | dbtRunnerResult | str:
```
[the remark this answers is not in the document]
pankajkoti: yes, thanks for the suggestion. I have tried to incorporate these suggestion, we have what's suggested in option "a"

cosmos/operators/local.py
```diff
@@ -647,11 +669,15 @@ def get_openlineage_facets_on_complete(self, task_instance: TaskInstance) -> Ope
         )
 
     def build_and_run_cmd(
-        self, context: Context, cmd_flags: list[str] | None = None
+        self,
```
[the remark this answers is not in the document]
pankajkoti: yes, done so.

tests/operators/test_base.py
```diff
@@ -173,5 +175,21 @@ def test_dbt_mixin_add_cmd_flags_run_operator(args, expected_flags):
...
     assert base_operator.append_env is False
+
+
+def test_abstract_dbt_base_is_not_airflow_base_operator():
+    AbstractDbtBase.__abstractmethods__ = set()
+    base_operator = AbstractDbtBase(task_id="fake_task", project_dir="fake_dir")
+    assert not isinstance(base_operator, BaseOperator)
+
+
+def test_abstract_dbt_base_init_no_super():
```
pankajkoti, line 189: I added a test so that we do not reintroduce back the conflicting `super.init()` call in `AbstractDbtBase`

cosmos/dbt_adapters/__init__.py
pankajastro, line 1: shall we change this to private module some options
- `_adapters`
- `_adapters/mock`

cosmos/dbt_adapters/bigquery.py
```diff
@@ -0,0 +1,33 @@
...
+
+def _mock_bigquery_adapter() -> None:
+    from typing import Optional, Tuple
+
+    import agate
+    from dbt.adapters.bigquery.connections import BigQueryAdapterResponse, BigQueryConnectionManager
+    from dbt_common.clients.agate_helper import empty_table
+
+    def execute(  # type: ignore[no-untyped-def]
+        self, sql, auto_begin=False, fetch=None, limit: Optional[int] = None
```
pankajastro, line 16:
```suggestion
        self, sql, auto_begin=False, fetch=None, limit: int | None = None
```

cosmos/operators/local.py
```diff
@@ -629,11 +675,17 @@ def get_openlineage_facets_on_complete(self, task_instance: TaskInstance) -> Ope
...
+        context: Context,
+        cmd_flags: list[str] | None = None,
+        run_as_async: bool = False,
```
pankajastro, line 681: The `run_as_async` may sound misleading since we do not run query in this case, right?
shall we rename to `mock_query_execution` or something else?

pankajastro (approved):
LGTM 👍

cosmos/dbt_adapters/__init__.py
pankajkoti: yes, we can, also will wait for @tatiana's feedback on this one.

cosmos/dbt_adapters/bigquery.py
pankajkoti: I copied the signature from the dbt_adapter as is defined there since we wish to mock that function. I think it's have the signature as is there to avoid any type mismatch errors. WDYT?

CHANGELOG.rst
```diff
@@ -18,6 +18,7 @@ Features
 * Allow users to opt-out of ``dbtRunner`` during DAG parsing with ``InvocationMode.SUBPROCESS`` by @tatiana in #1495. Check out the `documentation <https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html#how-to-run-dbt-ls-invocation-mode>`_.
 * Add structure to support multiple db for async operator execution by @pankajastro in #1483
 * Support overriding the ``profile_config`` per dbt node or folder using config by @tatiana in #1492. More information `here <https://astronomer.github.io/astronomer-cosmos/profiles/#profile-customise-per-node>`_.
+* Use dbt to generate the full SQL and support different materializations for BQ for ``ExecutionMode.AIRFLOW_ASYNC`` by @pankajkoti in #1474
```
tatiana, line 21:
```suggestion
* Create and run accurate SQL statements when using `ExecutionMode.AIRFLOW_ASYNC` by @pankajkoti in #1474
```

cosmos/dbt_adapters/__init__.py
tatiana: With this PR, we have dbt adaptor/data warehouse code spread in three different parts of Cosmos code-base:
- `profiles/bigquery`
- `_asynchronous/bigquery.py`
- `dbt_adapters/bigquery.py`

I had given similar feedback in @pankajastro's PR https://github.com/astronomer/astronomer-cosmos/pull/1483#discussion_r1926660077, and I think by introducing `dbt_adapters` we're making things even more spread in the codebase.

My strong recommendation is that we modify the following:
- remove the recently introduced `dbt_adapters` completely
- move the necessary code to `_asynchronous/bigquery.py`

The monkey patching being performed by Cosmos, at the moment, is specific to the context of `_asynchronous` processing. Since the package is not part of the official interface, we can change this in future if we want to.

cosmos/dbt_adapters/__init__.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+
+from typing import Any
+
+from cosmos.constants import BIGQUERY_PROFILE_TYPE
+from cosmos.dbt_adapters.bigquery import _associate_bigquery_async_op_args, _mock_bigquery_adapter
+
+PROFILE_TYPE_MOCK_ADAPTER_CALLABLE_MAP = {
```
tatiana, line 8: We should not do this type of mapping again.

I had given @pankajastro similar feedback on https://github.com/astronomer/astronomer-cosmos/pull/1474/files#r1941988073

He implemented the "automatic mapping":
https://github.com/astronomer/astronomer-cosmos/blob/c344eb4750f0269ecd267326373814abf56ec18e/cosmos/operators/_asynchronous/base.py#L16-L36

By doing https://github.com/astronomer/astronomer-cosmos/pull/1474/files#r1941988073 my hope is that this is removed.

tatiana (approved):
@pankajkoti Congratulations on the outstanding work in this PR and on your patience in addressing and fixing each bug that popped up during the development of this feature. I can't wait to see this feature used in production.

In addition to the feedback that I gave previously, there are two change requests:
- This comment: https://github.com/astronomer/astronomer-cosmos/pull/1474/files#r1941988073
- And this comment: https://github.com/astronomer/astronomer-cosmos/pull/1474/files#r1941991943

Given the size of this PR and all the challenges already overcome, I do not want my design requests to block its merging. So, your PR is approved. However, please create a follow-up ticket and prioritise it over any other work so the interfaces can be simplified as soon as possible. Other tasks planned for the 1.9 release will depend on these interface changes, so please prioritise them over any other work so we can wrap this up.

cosmos/dbt_adapters/__init__.py
pankajkoti: thanks @pankajastro and @tatiana . I have created a follow-up task to work on immediately on this https://github.com/astronomer/astronomer-cosmos/issues/1508

cosmos/dbt_adapters/__init__.py
pankajkoti: thanks @tatiana . I have created a follow-up task to work on immediately on this https://github.com/astronomer/astronomer-cosmos/issues/1508

CHANGELOG.rst
```diff
@@ -18,6 +18,7 @@ Features
 * Allow users to opt-out of ``dbtRunner`` during DAG parsing with ``InvocationMode.SUBPROCESS`` by @tatiana in #1495. Check out the `documentation <https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html#how-to-run-dbt-ls-invocation-mode>`_.
 * Add structure to support multiple db for async operator execution by @pankajastro in #1483
 * Support overriding the ``profile_config`` per dbt node or folder using config by @tatiana in #1492. More information `here <https://astronomer.github.io/astronomer-cosmos/profiles/#profile-customise-per-node>`_.
+* Create and run accurate SQL statements when using ``ExecutionMode.AIRFLOW_ASYNC`` by @pankajkoti in #1474
```
pankajkoti, line 21:
```suggestion
* Create and run accurate SQL statements when using ``ExecutionMode.AIRFLOW_ASYNC`` by @pankajkoti, @tatiana and @pankajastro in #1474
```

tatiana:
We'd love feedback from the community on this feature!

To try it out: `astronomer-cosmos==1.9.0a5`

tatiana:
@joppevos any chance you could give us feedback if this solved the initial bug we had in the query?

joppevos:
> @joppevos any chance you could give us feedback if this solved the initial bug we had in the query?

Thank you both for this feature 😊 I will test it very soon on our projects

jroach-astronomer:
Just so it's all in one place, here is a refactor of this PR: https://github.com/astronomer/astronomer-cosmos/pull/1509

Issue #865, opened 2024-02-27
Support dbt programatic invocation for dbt ls load mode


jbandoro:
In #850 support was added for execution mode to use `dbtRunner` programatic invocation of dbt and there was a substantial performance benefit over running dbt in a subprocess.

Cosmos could similarly use the dbt runner for `LoadMode.DBT_LS` parsing. The [RenderConfig.validate_dbt_command](https://github.com/astronomer/astronomer-cosmos/blob/1168f58c68285b7a0f6ea9c07ce7be06ff1e5722/cosmos/config.py#L84-L85) could be used to check if `dbtRunner` can be imported.

The [result](https://docs.getdbt.com/reference/programmatic-invocations#reusing-objects) of `dbtRunner.invoke(["ls", ...])` will be a manifest object so we wouldn't have to parse the ls output to a file.

tatiana:
This is a great feature, @jbandoro ! Do you think you may have the bandwidth next month to help us implement it for Cosmos 1.6? Otherwise, we may have to delay it for a feature release.

> The [result](https://docs.getdbt.com/reference/programmatic-invocations#reusing-objects) of dbtRunner.invoke(["ls", ...]) 
will be a manifest object so we wouldn't have to parse the ls output to a file.

The `manifest.json` does not include the filtered nodes—it contains the whole dbt project. I don't think we should use it for `dbt ls`—otherwise, we'd have to rely on Cosmos custom filters, not leveraging the benefits of using `dbt ls`.

PR #1480, opened 2025-01-21
Move `DbtRunner` related functions into `dbt/runner.py` module

tatiana:
Previously, @jbandoro added support for Cosmos to use `dbtRunner` to execute dbt commands in the Airflow worker nodes when using `ExecutionMode.LOCAL` instead of Python's subprocess #850. This change significantly improved Cosmos resource utilisation in worker nodes.

The first step to implementing support to use `dbtRunner` during DAG parsing (ticket #865) when using `LoadMode.DBT_LS` is to extract `dbtRunner` methods somewhere it can be reused by `graph.py` and is not specific to `operators.py`.

This PR does the following:
- Create `dbt/runner.py`
- Move all `dbtRunner`related functions to this module, refactoring them as needed
- Refactor `operators/local.py` to use the newly introduced methods

Related: #865

### DIFF
--- a/cosmos/dbt/runner.py
+++ b/cosmos/dbt/runner.py
@@ -1,0 +1,119 @@
+from __future__ import annotations
+
+import sys
+from functools import lru_cache
+from typing import TYPE_CHECKING
+
+from cosmos.dbt.project import change_working_directory, environ
+from cosmos.exceptions import CosmosDbtRunError
+from cosmos.log import get_logger
+
+if "pytest" in sys.modules:
+    # We set the cache limit to 0, so nothing gets cached by default when
+    # running tests
+    cache = lru_cache(maxsize=0)
+else:  # pragma: no cover
+    try:
+        # Available since Python 3.9
+        from functools import cache
+    except ImportError:
+        cache = lru_cache(maxsize=None)
+
+
+logger = get_logger(__name__)
+
+if TYPE_CHECKING:  # pragma: no cover
+    from dbt.cli.main import dbtRunner, dbtRunnerResult
+
+
+# TODO: Change operators/local.py invoke_dbt to use this
+# TODO: Change operators/local.py _discover_invocation_mode to use this
+@cache
+def is_available() -> bool:
+    """
+    Checks if the dbt runner is available (if dbt-core is installed in the same Python virtualenv as Airflow."
+    """
+    try:
+        from dbt.cli.main import dbtRunner  # noqa
+    except ImportError:
+        return False
+    return True
+
+
+# TODO: Change operators/local.py to use this in run_dbt_runner
+@cache
+def get_runner() -> dbtRunner:
+    """
+    Retrieves a dbtRunner instance.
+    """
+    from dbt.cli.main import dbtRunner
+
+    return dbtRunner()
+
+
+# TODO: Change operators/local.py run_dbt_runner to use this
+def run_command(command: list[str], env: dict[str, str], cwd: str) -> dbtRunnerResult:
+    """
+    Invokes the dbt command programmatically.
+    """
+    # Exclude the dbt executable path from the command
+    cli_args = command[1:]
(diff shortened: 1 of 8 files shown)


cosmos/dbt/runner.py
```diff
@@ -0,0 +1,111 @@
...
+logger = get_logger(__name__)
+
+if TYPE_CHECKING:  # pragma: no cover
+    from dbt.cli.main import dbtRunner, dbtRunnerResult
+
+
+@cache
+def is_available() -> bool:
+    """
+    Checks if the dbt runner is available (if dbt-core is installed in the same Python virtualenv as Airflow."
```
pankajkoti, line 32:
```suggestion
    Checks if the dbt runner is available (if dbt-core is installed in the same Python virtualenv as Airflow)."
```

cosmos/dbt/runner.py
```diff
@@ -0,0 +1,111 @@
...
+    from dbt.cli.main import dbtRunner
+
+    return dbtRunner()
+
+
+def run_command(command: list[str], env: dict[str, str], cwd: str) -> dbtRunnerResult:
+    """
+    Invokes the dbt command programmatically.
+    """
+    # Exclude the dbt executable path from the command
```
pankajkoti, line 55: not your change, but can we confirm how does `command` look like here since we have this comment? is this excluding the dbt executable path or the `run` part in the command prefix?

cosmos/operators/local.py
```diff
@@ -393,24 +384,12 @@ def run_subprocess(self, command: list[str], env: dict[str, str], cwd: str) -> F
 
     def run_dbt_runner(self, command: list[str], env: dict[str, str], cwd: str) -> dbtRunnerResult:
         """Invokes the dbt command programmatically."""
-        try:
-            from dbt.cli.main import dbtRunner
-        except ImportError:
+        if not dbt_runner.is_available():
             raise ImportError(
```
pankajkoti, line 388: Should we change this error type since the ImportError here seems a bit off?

cosmos/operators/local.py
```diff
@@ -51,18 +51,17 @@
 
 from sqlalchemy.orm import Session
 
+import cosmos.dbt.runner as dbt_runner
 from cosmos.config import ProfileConfig
 from cosmos.constants import (
     OPENLINEAGE_PRODUCER,
 )
 from cosmos.dbt.parser.output import (
-    extract_dbt_runner_issues,
```
pankajkoti, line 59: Can we mark this method as deprecated and log a message saying that it will be removed in 2.0.0 ?

cosmos/operators/local.py
```diff
@@ -51,18 +51,17 @@
...
     extract_freshness_warn_msg,
     extract_log_issues,
-    parse_number_of_warnings_dbt_runner,
```
pankajkoti, line 62: same question about possibility of marking this method as deprecated and up for removal in 2.0.0?

cosmos/dbt/runner.py
tatiana: Thanks, @pankajkoti , done in 650a99d55ea9f56f9b9f17155d8f94904bfa0237

cosmos/dbt/runner.py
tatiana: I improved the docs in a7b49b30bdd0fa9feb9b92b6d4deccad94c65c8e

cosmos/operators/local.py
tatiana: Great suggestion; I did this in d921ed9ae077f61719b03375edffe4b586b9c5d1

cosmos/operators/local.py
tatiana: Changed in fc38c1536c5b71eb664726c50c7d5619a56b7535

pankajastro (approved):

PR #1484, opened 2025-01-23
Use `dbtRunner` in the DAG Processor when using `LoadMode.DBT_LS` if `dbt-core` is available

tatiana:
This PR significantly improves Cosmos resource utilisation in the scheduler DAG Processor and in the worker nodes when dynamically converting dbt workflows into Airflow DAGs using the `LoadMode.DBT_LS`. 

It introduces support to use `dbtRunner` during DAG parsing if `dbt-core` and its adaptors are in the same Python virtualenv as Airflow.

This change is particularly relevant given the way Airflow (2.x) parses DAGs not only as part of the scheduler loop but also whenever a task executes:

Screenshot 2025-01-23 at 13 51 40

> Diagram extracted from the talk @pankajkoti and I gave at Airflow Summit 2024
> https://airflowsummit.org/sessions/2024/overcoming-performance-hurdles-in-integrating-dbt-with-airflow/)

When using `LoadMode.DBT_LS`, Cosmos runs `dbt ls` whenever Airflow parses the DAG (in case of a cache miss).

Suppose there is a Cosmos `DbtDag` with 200 concurrent tasks. If the dbt project changes, when Airflow and Cosmos attempt to parse the `DbtDag`, they will invalidate the dbt ls cache. If 200 Cosmos tasks execute concurrently when there is a cache miss, they will all run the same `dbt ls` command. Until Cosmos 1.8, Cosmos would always create a subprocess for each command. If 200 tasks execute in a worker node, this would represent 400 processes attempting to run concurrently, leading to a vast resource CPU - and potentially memory - spike and Out of Memory (OOM) errors. While this change does not avoid the 200 tasks attempting to run `dbt ls` concurrently, it avoids each of them creating an additional subprocess - optimising the resource utilisation.

This change is heavily influenced by changes (#850) previously made by @jbandoro, who added support for Cosmos to use `dbtRunner` to execute dbt commands in the Airflow worker nodes when using `ExecutionMode.LOCAL` instead of Python's subprocess.

Closes: #865

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -17,6 +17,7 @@ from typing import TYPE_CHECKING, Any, Dict, Optional
 
 from airflow.models import Variable
 
+import cosmos.dbt.runner as dbt_runner
 from cosmos import cache, settings
 from cosmos.cache import (
     _configure_remote_cache_dir,
@@ -158,11 +159,8 @@ def is_freshness_effective(freshness: Optional[dict[str, Any]]) -> bool:
     return False
 
 
-def run_command(command: list[str], tmp_dir: Path, env_vars: dict[str, str]) -> str:
+def run_command_with_subprocess(command: list[str], tmp_dir: Path, env_vars: dict[str, str], log_dir: Path) -> str:
     """Run a command in a subprocess, returning the stdout."""
-    command = [str(arg) if arg is not None else "<None>" for arg in command]
-    logger.info("Running command: `%s`", " ".join(command))
-    logger.debug("Environment variable keys: %s", env_vars.keys())
     process = Popen(
         command,
         stdout=PIPE,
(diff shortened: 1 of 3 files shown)


cosmos/dbt/graph.py
```diff
@@ -597,11 +647,12 @@ def load_via_dbt_ls_without_cache(self) -> None:
                     self.profile_config.target_name,
                 ]
 
-                self.log_dir = Path(env.get(DBT_LOG_PATH_ENVVAR) or tmpdir_path / DBT_LOG_DIR_NAME)
                 self.target_dir = Path(env.get(DBT_TARGET_PATH_ENVVAR) or tmpdir_path / DBT_TARGET_DIR_NAME)
                 env[DBT_LOG_PATH_ENVVAR] = str(self.log_dir)
```
pankajkoti, line 651: Will `self.log_dir` still be `None` at this point of time? And would we be altering the env var to None always in that case? Is that our desired behavior?

With this, in the below line it would always resort & set `self.log_dir` to `tmpdir_path/DBT_LOG_DIR_NAME`, no?

pankajkoti (approved):
Looks all good to me, hoping the results will be amazing with this very cool optimisation.

I just have a quick question in-line regarding setting the log_dir.

cosmos/dbt/graph.py
tatiana: Nice catch, @pankajkoti , I fixed this in 7db1de03833ee44e434a91f023e6105d6de0582a

tatiana:
This feature is available to be tested in [1.9.0a3](https://pypi.org/project/astronomer-cosmos/1.9.0a3/)

PR #1492, opened 2025-01-27
Support overriding the ``profile_config`` per dbt node or folder using config

tatiana:
Allow users to override `profile_config` per operator or folder using config.  Extends the work introduced by @wornjs in #1339, when we started supporting operator-level arguments defined in the dbt project/model YAML.

This PR includes a generic implementation that introduces the `[meta][cosmos][profile_config]` section, at the same level as `operator_kwargs`, and ensures these configs override the original profile config.

Example overriding a `profile_config` that uses `profiles_yml_filepath`:
```
      +meta:
        cosmos:
          operator_kwargs:
              (...)
          profile_config:
             profile_name: non_default_profile
             target_name: stage
             profiles_yml_filepath: /some/dir/profiles.yml
```

Example of an interface when using `profile_mapping`:
```
      +meta:
        cosmos:
          operator_kwargs:
            (...)    
          profile_config:
             profile_name: non_default_profile
             target_name: stage
             profile_mapping:
               conn_id: non_default_connection
               profile_args:
                 schema: non_default_schema   
```


tatiana:
@wornjs @dwreeves @pankajastro @pankajkoti  would love your thoughts. Please look at the description and the current/ possible alternative implementation. I'm happy to hear other ideas as well! I'd rather take a bit longer and have an interface that makes sense.

cosmos/airflow/graph.py
```diff
@@ -96,6 +97,22 @@ def exclude_detached_tests_if_needed(
         task_args["exclude"] = exclude  # type: ignore
 
 
+def override_profile_if_needed(task_kwargs: dict[str, Any], profile_kwargs_override: dict[str, Any]) -> None:
```
tatiana, line 100:
```suggestion
def _override_profile_if_needed(task_kwargs: dict[str, Any], profile_kwargs_override: dict[str, Any]) -> None:
```

dev/dags/dbt/jaffle_shop/dbt_project.yml
```diff
@@ -24,3 +24,13 @@ models:
...
+
+
+seeds:
+  jaffle_shop:
+      +meta:
+        cosmos:
+          profile_config:
+            profile_name: postgres_profile
+            profile_mapping:
+              threads: 2
```
pankajastro, line 36: Shall we document this configurations

pankajastro (approved):

dev/dags/dbt/jaffle_shop/dbt_project.yml
tatiana: Thanks, @pankajastro , documented in 602b64b85ce1739645d790a4f73d46e5fd3d0827

PR #1495, opened 2025-01-28
Allow users to opt-out `dbtRunner` during DAG parsing

tatiana:
While speaking to a customer about #1484, they mentioned they have the following setup:
- `dbt-databricks` installed in the same Python virtualenv as Cosmos/Airflow
- `dbt-bigquery` installed in a separate Python virtualenv using Astro Dockerfile

And run DAGs using both with the same image. This means 1.9.0a3 breaks them since they use `LoadMode.DBT_LS` and only `debt-data bricks` can be parsed. This means that we have to add support to allow users to opt-in / out of using the `dbtRunner` during DAG parsing - similar to what was done for task execution, in `ExecutionConfig`.

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -70,6 +70,7 @@ class RenderConfig:
     emit_datasets: bool = True
     test_behavior: TestBehavior = TestBehavior.AFTER_EACH
     load_method: LoadMode = LoadMode.AUTOMATIC
+    invocation_mode: InvocationMode = InvocationMode.DBT_RUNNER
     select: list[str] = field(default_factory=list)
     exclude: list[str] = field(default_factory=list)
     selector: str | None = None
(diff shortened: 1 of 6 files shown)


pankajkoti (approved):
LGTM. 

I guess would be nice to highlight in our documentation somewhere that starting Cosmos 1.9, we're optimising to use dbtRunner by default for parsing and also show them how they can change the default with the corresponding param in their DAG. WDYT?

cosmos/config.py
```diff
@@ -70,6 +70,7 @@ class RenderConfig:
     emit_datasets: bool = True
     test_behavior: TestBehavior = TestBehavior.AFTER_EACH
     load_method: LoadMode = LoadMode.AUTOMATIC
+    invocation_mode: InvocationMode = InvocationMode.DBT_RUNNER
```
pankajastro, line 73: shall we add this in docstring?

pankajastro (approved):

tatiana:
Thanks for the feedback, @pankajastro & @pankajkoti !

I've added a breaking change notice in the changelog: 14b0a2c431b428e4a247634b119ae99cfe3b485a

As well as docs in 53429d9ee442311c8f73130a4ce809f02fb20b37

cosmos/config.py
tatiana: Done in 622be6cd4281fc44e82578187affe2e3884f979f

Issue #1458, opened 2025-01-13
Bug: `operator_args={“install_deps”: True}` when dependencies are already installed


tatiana:
A customer reported that if you put True in the `operator_args={“install_deps”}` and the dependencies are already installed, they see Cosmos is raising the same exception as #1457.

The packages can't be installed, but the task continues, and the dbt run command is executed successfully.

Issue #1457, opened 2025-01-13
Bug when using `RenderConfig(dbt_deps=True)` and `operator_args={"install_deps": False}`


tatiana:
Customer is facing this error when `RenderConfig(dbt_deps=True)` and `operator_args={"install_deps": False}`:
![Image](https://github.com/user-attachments/assets/495e46c8-419d-48d3-9bb4-e8955ce4c43e)

This issue does not happen if both of them are False.

PR #1505, opened 2025-02-04
Gracefully error when users set imcompatible `RenderConfig.dbt_deps` and `operator_args` `install_deps`

tatiana:
Customer is facing this error when `RenderConfig(dbt_deps=True)` and `operator_args={"install_deps": False}`:
Image

This issue does not happen if both of them are `False`.

Closes: #1458
Closes: #1457


cosmos/converter.py
```diff
@@ -96,8 +97,21 @@ def validate_arguments(
...
+        has_non_empty_dependencies = execution_config.project_path and has_non_empty_dependencies_file(
+            execution_config.project_path
+        )
+        if (
+            has_non_empty_dependencies
+            and (
+                render_config.load_method == LoadMode.DBT_LS
+                or (render_config.load_method == LoadMode.AUTOMATIC and not project_config.is_manifest_available())
+            )
+            and (render_config.dbt_deps != task_args.get("install_deps", True))
```
pankajkoti, line 111: Can we reconfirm the case when the user has not explicitly set `install_deps` in their task, what value comes here for `install_deps`?

I am thinking that when user has not specified `install_deps` in their `task_args`, assuming it would be False by default and at the same time if they have specified `render_config.dbt_deps=False`, we would raise this error and this should not be the case, no? 

I am thinking if we could rather change this to,
```
if render_config.dbt_deps and task_args.get("install_deps") ==  False
```

tatiana: @pankajkoti the current "default" value of `install_deps` is https://github.com/astronomer/astronomer-cosmos/blob/002c91934e75c97c109ef82c6612b07368c4c371/cosmos/operators/local.py#L185

I agree we should improve this; perhaps I could isolate and unify the logic as a function, e.g. "calculate_default_install_deps`, and use it in both places?

pankajkoti: > the current "default" value of install_deps is True at

Wouldn’t the current default value of `install_deps` be False? I mean at https://github.com/astronomer/astronomer-cosmos/blob/002c91934e75c97c109ef82c6612b07368c4c371/cosmos/operators/local.py#L145

Yes, since we’re validating arguments before initializing the operators, it looks like setting the default value is deferred until the operators are actually instantiated. This could cause some confusion because if the user hasn’t explicitly set install_deps in their DAG operators, the key won’t exist in task_args at this stage, meaning task_args.get("install_deps") would return None.

I agree that having a common function to determine the default value would be beneficial, as it could be used in both places. Additionally, we’d need to account for cases where the user provides a custom value for a specific operator, ensuring that we correctly handle overrides while still computing the default value when necessary.

tatiana: > Wouldn’t the current default value of install_deps be False? I mean at

From a function signature definition, yes, `False` is the default value. But from an initialization perspective, we're changing the "default" behaviour
https://github.com/astronomer/astronomer-cosmos/blob/002c91934e75c97c109ef82c6612b07368c4c371/cosmos/operators/local.py#L185

That means that even if users explicitly set `True`, the `install_ops` may become `False`. 

> Additionally, we’d need to account for cases where the user provides a custom value for a specific operator, ensuring that we correctly handle overrides while still computing the default value when necessary.

Yes, I believe we should not allow users -as of now - to override this property. Otherwise they will end up seeing the original exception reported by our customer. WDYT?

pankajkoti: yes, probably we could brainstorm this more together

tatiana, line 111:
```suggestion
            and (render_config.dbt_deps != task_args.get("install_deps", False))
```

tatiana: Thanks a lot, @pankajkoti , it was great to have the call, I was parallelising many things and unable to see your point. I just the made the change we agreed!

pankajkoti (approved):
LGTM. Thanks for the discussion together! Happy to be merged once CI passes.

Issue #1508, opened 2025-02-05
Refactor dbt adapter mocking modules


pankajkoti:
Address the following suggestions received for PR #1474 :

https://github.com/astronomer/astronomer-cosmos/pull/1474#discussion_r1941988073

https://github.com/astronomer/astronomer-cosmos/pull/1474#discussion_r1941991943

PR #1509, opened 2025-02-05
Refactor to consolidate async dbt adapter code

pankajkoti:
This PR addresses the feedback in PR #1474 suggesting to consolidate dbt adapter methods
in the _asynchronous module and additionally avoiding keeping dictionary type maps for adapter 
callables. 

closes: #1508


cosmos/helpers.py
```diff
@@ -0,0 +1,13 @@
+import importlib
```
tatiana, line 1: @pankajkoti could we rename this module to `utils/importer.py`?

Also, it may be worth to log a ticket so we refactor other parts of the code that could leverage this, WDYT?

tatiana (approved):
Looks great, @pankajkoti , thank you very much for refactoring this part of the code. I left a minor feedback on the `helper.py` module name

cosmos/helpers.py
pankajkoti: Thanks Tatiana for the suggestion. I have created a follow-up ticket to reuse this method in other places of our existing codebase.

PR #1521, opened 2025-02-07
Add `ProjectConfig.install_dbt_deps` & change operator `install_deps=True` as default

tatiana:
Cosmos errors if users set different values in `RenderConfig.dbt_deps` and `operator_args={"install_deps"}`, and since #1505, it started raising a more graceful error to users.

The problem is that since the default value for `RenderConfig.dbt_deps` was `True` and the default value for `operator_args={"install_deps"=False}` was False, it meant that by default, Cosmos would raise an exception if users wrote a simple DAG like:
```
    dag = DbtDag(
        project_config=ProjectConfig(
        ),
        dag_id="basic_cosmos_dag",
    )
```

Which was a poor experience. This PR aims to solve this.

The documentation will be updated in a separate PR.

Related to: #1458
Related to: #1457

### DIFF
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -223,8 +223,9 @@ def override_configuration(
     if execution_config.invocation_mode:
         operator_args["invocation_mode"] = execution_config.invocation_mode
 
-    if "install_deps" not in operator_args:
-        operator_args["install_deps"] = project_config.install_dbt_deps
+    if execution_config in (ExecutionMode.LOCAL, ExecutionMode.VIRTUALENV):
+        if "install_deps" not in operator_args:
+            operator_args["install_deps"] = project_config.install_dbt_deps
 
 
 class DbtToAirflowConverter:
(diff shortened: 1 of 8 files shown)


cosmos/converter.py
```diff
@@ -196,6 +203,31 @@ def validate_changed_config_paths(
         )
 
 
+def override_configuration(
```
pankajastro, line 206: Do we need some separate tests for this function, or is it getting covered by some different tests?

pankajastro (approved):
lgtm

pankajkoti (approved):
LGTM

cosmos/converter.py
tatiana: @pankajastro @pankajkoti could we add a follow up ticket so I can address this after? Due to illness I have not been able to improve the test coverage and we should not release 1.9 without this

pankajkoti: yes, created a follow-up ticket for this https://github.com/astronomer/astronomer-cosmos/issues/1540

PR #1526, opened 2025-02-08
Filter dbt and non-dbt kwrgs correctly for async operator

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1527


pankajkoti (approved):
LGTM.

I guess would also be nice to add an arg `install_deps: True` in the async example DAG too here https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/simple_dag_async.py#L40 so that we test the end to end execution of it as well.

Issue #1489, opened 2025-01-24
Run async DAGs over multiple versions of dbt adapters to ensure monkeypatch works correctly


pankajkoti:
With the work being done in PR #1474, add a job in CI to test async DAGs across multiple versions of dbt adapters. This ensures that the monkeypatching applied to the adapters works correctly across different versions.

The job should:
	1.	Include a matrix of supported dbt adapter versions.
	2.	Run the async DAGs for each version in the matrix.
	3.	Validate that the monkeypatched behavior is consistent and functional across all versions.

This will help catch potential compatibility issues early and ensure robust support for async DAGs in varying dbt adapter environments.

PR #1535, opened 2025-02-12
Add CI job to test multiple dbt versions for the async DAG

pankajkoti:
This PR introduces a new CI job named `Run-Integration-Tests-DBT-Async` to ensure compatibility of the async example DAG with multiple dbt versions. It achieves this by adding a third dimension to the `pyproject.toml` matrix, enabling the CI to run the DAG across a list of dbt versions.

Additionally, this PR includes a new test file: `tests/test_async_example_dag.py`. While we already have `tests/test_example_dags.py`, certain dbt versions have shown parsing issues with some example DAGs, which can cause CI failures unrelated to the async DAG. To prevent this, the new file is a modified copy that exclusively tests `simple_async_dag`, with other DAGs ignored via `.airflowignore`. This ensures that the CI job focuses on validating the async DAG without being affected by unrelated parsing errors.

closes: #1489

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main,run-async-on-dbt-versions]
+    branches: [main]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,7 @@ jobs:
           architecture: "x64"
 
       - run: pip3 install hatch
-      - run: hatch run tests.py3.9-2.8:type-check
+      - run: hatch run tests.py3.9-2.8-1.9:type-check
 
   Run-Unit-Tests:
     runs-on: ubuntu-latest
@@ -343,6 +343,7 @@ jobs:
     needs: Authorize
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: [ "3.11" ]
         airflow-version: [ "2.10" ]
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -307,11 +311,81 @@ jobs:
...
+          DATABRICKS_WAREHOUSE_ID: mock
+          DATABRICKS_TOKEN: mock
+          POSTGRES_HOST: localhost
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+          POSTGRES_SCHEMA: public
+          POSTGRES_PORT: 5432
+          AIRFLOW__COSMOS__REMOTE_TARGET_PATH: "s3://cosmos-remote-cache/target_compiled/"
+          AIRFLOW__COSMOS__REMOTE_TARGET_PATH_CONN_ID: aws_s3_conn
```
tatiana, line 340: Why did we add all these environment variables in this particular job?

tests/test_async_example_dag.py
```diff
@@ -0,0 +1,65 @@
+from __future__ import annotations
```
tatiana, line 1: Could you add a note in the PR description about why we needed to create a new job? This information can be valuable in the future if we want to understand some of our decisions.

tatiana (commented):
Hi @pankajkoti , I believe this is great, thank you very much!

I left some comments inline and, please, could you log a ticket to cover this part of the original ticket:
3. Validate that the monkeypatched behavior is consistent and functional across all versions.

So we can actually confirm that:
- The setup task does not actually perform the queries in BigQuery
Making sure the mock works as expected for the desired dbt versions?

Also, if you could log a ticket for us to add a compatibility document, explaining which versions of dbt/airflow/python are tested with Cosmos, it could be great.

tests/test_async_example_dag.py
pankajkoti: added a note

pankajkoti:
> Hi @pankajkoti , I believe this is great, thank you very much!
> 
> I left some comments inline and, please, could you log a ticket to cover this part of the original ticket: 3. Validate that the monkeypatched behavior is consistent and functional across all versions.
> 
> So we can actually confirm that:
> 
> * The setup task does not actually perform the queries in BigQuery
>   Making sure the mock works as expected for the desired dbt versions?
> 
> Also, if you could log a ticket for us to add a compatibility document, explaining which versions of dbt/airflow/python are tested with Cosmos, it could be great.

Thanks. Created a couple of tickets:
1. https://github.com/astronomer/astronomer-cosmos/issues/1605
2. https://github.com/astronomer/astronomer-cosmos/issues/1606

tatiana (approved):

Issue #1273, opened 2024-10-21
[async] Fallback to using dbt-core to run transformation if async fails


tatiana:
There will be scenarios when the async transformation may fail (either due to a macro, or some other misconfiguration) when using `ExecutionMode.AIRFLOW_ASYNC`.

In those cases, Cosmos should fallback to run them using `dbt-core` in the worker node similar to `ExecutionMode.LOCAL`

tatiana:
Since we're now using `dbt-core` to generate the query, we'll close this for now. We can revisit this in the future.

PR #1538, opened 2025-02-13
Remove the fallback operator for async task

pankajastro:
Since we decided to close: https://github.com/astronomer/astronomer-cosmos/issues/1273 the fallback for async operator class does not makes sense


cosmos/operators/_asynchronous/base.py
```diff
@@ -18,7 +18,7 @@ def _create_async_operator_class(profile_type: str, dbt_class: str) -> Any:
 
     The function constructs a class path string for an asynchronous operator, based on the provided `profile_type` and
     `dbt_class`. It attempts to import the corresponding class dynamically and return it. If the class cannot be found,
-    it falls back to returning the `DbtRunLocalOperator` class.
+    it raise an error.
```
pankajkoti, line 21:
```suggestion
    it raises an error.
```

pankajkoti (approved):

PR #1543, opened 2025-02-17
Fix dbt project parsing `dbt_vars` behavior

AlexandrKhabarov:
## Description

DbtToAirflowConverter can pass dbt_vars to DbtGraph with help of ProjectConfig or operator_args. If operator_args is used in DbtToAirflowConverter then it will lead to the issue with absence of dbt_vars in dbt ls command (faced rendering issue during usage of cosmos with project level variables)

## Breaking Change?

No breaking changes 

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -376,9 +376,7 @@ class DbtGraph:
         """
         Change args list in-place so they include dbt vars, if they are set.
         """
-        if self.project.dbt_vars:
-            cmd_args.extend(["--vars", json.dumps(self.project.dbt_vars, sort_keys=True)])
-        elif self.dbt_vars:
+        if self.dbt_vars:
             cmd_args.extend(["--vars", json.dumps(self.dbt_vars, sort_keys=True)])
 
     @cached_property
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -378,6 +378,8 @@ class DbtGraph:
         """
         if self.project.dbt_vars:
             cmd_args.extend(["--vars", json.dumps(self.project.dbt_vars, sort_keys=True)])
+        elif self.dbt_vars:
+            cmd_args.extend(["--vars", json.dumps(self.dbt_vars, sort_keys=True)])
 
     @cached_property
     def dbt_ls_args(self) -> list[str]:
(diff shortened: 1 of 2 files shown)


pankajastro (approved):

cosmos/dbt/graph.py
```diff
@@ -378,6 +378,8 @@ def _add_vars_arg(self, cmd_args: list[str]) -> None:
         """
         if self.project.dbt_vars:
             cmd_args.extend(["--vars", json.dumps(self.project.dbt_vars, sort_keys=True)])
+        elif self.dbt_vars:
+            cmd_args.extend(["--vars", json.dumps(self.dbt_vars, sort_keys=True)])
```
pankajkoti, line 382:
```suggestion
        if self.dbt_vars:
            cmd_args.extend(["--vars", json.dumps(self.dbt_vars, sort_keys=True)])
```
This should yield the same result, right?

@tatiana Was there a specific reason we checked for `dbt_vars` only in `self.project` instead of directly using `self.dbt_vars` while working on https://github.com/astronomer/astronomer-cosmos/pull/1114?

@AlexandrKhabarov While Tatiana provides insights on this, logically speaking, wouldn’t it make more sense to use the variables set at the project level during parsing/graph building rather than the ones set for the operator(I would incline to think that operator args would be used for each of the operator tasks during execution)? What do you think?

tatiana: @pankajkoti I believe it was a mistake to not just use `self.dbt_vars`, given the logic in https://github.com/astronomer/astronomer-cosmos/blob/56b386ec0aff009fc9b898df921e6cbb4a5551ac/cosmos/converter.py#L272

We probably should avoid using `self.project.dbt_vars` in this class and rely only on `self.dbt_vars

tatiana:
@AlexandrKhabarov, sorry for the delay in giving feedback; this week has been quite intense at work, and I wanted to give enough attention to this review. Please let us know your thoughts on committing @pankajkoti 's suggestion - I would love to have this released as part of Cosmos 1.9.1, with the other bugfixes

tatiana (approved):

tatiana:
Thank you very much, @AlexandrKhabarov. We'll release this as part of Cosmos 1.9.1 this week.

Issue #1533, opened 2025-02-11
Run async compile task using a Python virtualenv


tatiana:
Alpha testers of 1.9.0a5 reported facing issues when trying to install `dbt-bigquery` in the same virtualenv as Airflow.  For this reason, this ticket is essential. Although we found a workaround to make this work:

```
FROM quay.io/astronomer/astro-runtime:12.6.0-python-3.11

USER root

RUN pip install dbt-bigquery

USER astro
```

This worked because we are not installing the dependencies in the requirements.txt` and pip doesn't realise the conflict is there.

Follow up #1477

1. Check if we can monkey patch the dbt run when using ExecutionMode.VIRTUALENV
2. Replace using dbtRunner with the virtualenv call

tatiana:
What we'd like to attempt:
1. Have the compile task inherit from the virtualenv operator
2. Expose in `ExecutionConfig.async_requirements` that receives a list of packages that should be installed in the virtualenv
3.  Once the virtualenv is created, change the dbt entry point script `(...)/bin/dbt` that looks like:
```
# -*- coding: utf-8 -*-
import re
import sys
from dbt.cli.main import cli
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(cli())
```

So that it invokes the relevant monkey patching code

https://github.com/astronomer/astronomer-cosmos/blob/968b80eb7a6e572076eec2ab2ec95697e29a3501/cosmos/operators/_asynchronous/bigquery.py#L33-L38

^ The most risky part of this implementation is (3). To mitigate risks, I suggest we first attempt to do 3 by hardcoding the monkey patch there and see if this works.

PR #1544, opened 2025-02-18
Extend VirtualEnv operator and mock dbt adapters for setup & teardown tasks in ExecutionMode.AIRFLOW_ASYNC

pankajkoti:
## Summary

This PR extends the `DbtRunVirtualenvOperator` to the setup and teardown tasks for `ExecutionMode.AIRFLOW_ASYNC`. It ensures that dbt adapters are installed in the virtualenv created by the `DbtRunVirtualenvOperator` and are properly mocked within the virtual environment.

## Key Changes

✅ **Extending SetupAsyncOperator and TeardownAsyncOperator with DbtRunVirtualenvOperator and mocking dbt adapters**
- These operators inherit from `DbtRunVirtualenvOperator` and override `run_subprocess`.
- The operators extract the mock function from the appropriate `cosmos.operators._asynchronous` module.
- The extracted function is injected dynamically into the dbt CLI entry script within the virtual environment, so that mocked dbt adapters get called before executing dbt commands.

✅ **Decoupled dbt Adapters from Airflow Environment**
- With this change, dbt adapters (e.g., `dbt-bigquery`, `dbt-postgres`) no longer need to be installed in the same environment as the Airflow installation.
- The `ExecutionConfig` now exposes `async_py_requirements` which can be set by DAG authors, ensuring that the necessary dbt dependencies are installed inside the virtual environment for `ExecutionMode.AIRFLOW_ASYNC`.

closes: #1533 

---------

Co-authored-by: Pankaj Singh <98807258+pankajastro@users.noreply.github.com>

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -389,17 +389,17 @@ class ProfileConfig:
 @dataclass
 class ExecutionConfig:
     """
-    Contains configuration about how to execute dbt.
+        Contains configuration about how to execute dbt.
 
-    :param execution_mode: The execution mode for dbt. Defaults to local
-    :param invocation_mode: The invocation mode for the dbt command. This is only configurable for ExecutionMode.LOCAL.
-    :param test_indirect_selection: The mode to configure the test behavior when performing indirect selection.
-    :param dbt_executable_path: The path to the dbt executable for runtime execution. Defaults to dbt if available on the path.
-    :param dbt_project_path: Configures the DBT project location accessible at runtime for dag execution. This is the project path in a docker container for ExecutionMode.DOCKER or ExecutionMode.KUBERNETES. Mutually Exclusive with ProjectConfig.dbt_project_path
-    :param virtualenv_dir: Directory path to locate the (cached) virtual env that
-    should be used for execution when execution mode is set to `ExecutionMode.VIRTUALENV`
-    :param async_py_requirements:  A list of Python packages to install when `ExecutionMode.AIRFLOW_ASYNC` is used. This parameter is required only if both `enable_setup_async_task` and `enable_teardown_async_task` are set to `True`. 
-Example: `["dbt-postgres==1.5.0"]` 
+        :param execution_mode: The execution mode for dbt. Defaults to local
+        :param invocation_mode: The invocation mode for the dbt command. This is only configurable for ExecutionMode.LOCAL.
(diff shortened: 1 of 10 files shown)


cosmos/config.py
```diff
@@ -398,6 +398,8 @@ class ExecutionConfig:
     :param dbt_project_path: Configures the DBT project location accessible at runtime for dag execution. This is the project path in a docker container for ExecutionMode.DOCKER or ExecutionMode.KUBERNETES. Mutually Exclusive with ProjectConfig.dbt_project_path
     :param virtualenv_dir: Directory path to locate the (cached) virtual env that
     should be used for execution when execution mode is set to `ExecutionMode.VIRTUALENV`
+    :param async_py_requirements: In the case Execu....Creates a virtual environment with the specified dependencies. Example:
+           ["dbt-postgres==1.5.0"]
```
pankajastro, line 402:
```suggestion
    :param async_py_requirements:  A list of Python packages to install when `ExecutionMode.AIRFLOW_ASYNC` is used. This parameter is required only if both `enable_setup_async_task` and `enable_teardown_async_task` are set to `True`. 
Example: `["dbt-postgres==1.5.0"]` 
```
need to change maybe something like

cosmos/config.py
```diff
@@ -398,6 +398,8 @@ class ExecutionConfig:
     :param dbt_project_path: Configures the DBT project location accessible at runtime for dag execution. This is the project path in a docker container for ExecutionMode.DOCKER or ExecutionMode.KUBERNETES. Mutually Exclusive with ProjectConfig.dbt_project_path
     :param virtualenv_dir: Directory path to locate the (cached) virtual env that
     should be used for execution when execution mode is set to `ExecutionMode.VIRTUALENV`
+    :param async_py_requirements: In the case Execu....Creates a virtual environment with the specified dependencies. Example:
```
pankajastro, line 401: Also, we have a parameter `py_requirements` similar to this, but it is part of the `operator_args` param, so I'm not sure if we should experiment here at least in docstring.

pankajastro (approved):

cosmos/config.py
pankajkoti: yes, thanks. added experimental alongside `ExecutionMode.AIRFLOW_ASYNC`

PR #1559, opened 2025-02-24
Implement dbt exposure selector

### LINKED ISSUE #1551: [Feature] `exposure:` selectors
### Description

A limited set of dbt selectors are implemented in https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/dbt/selector.py:
```
PATH_SELECTOR = "path:"
TAG_SELECTOR = "tag:"
CONFIG_SELECTOR = "config."
SOURCE_SELECTOR = "source:"
RESOURCE_TYPE_SELECTOR = "resource_type:"
```

dbt support many more selector methods as documented here: https://docs.getdbt.com/reference/node-selection/methods. I would be particularly interested in an implementation of `exposure`, which hopefully could be very similar to `source`. Maybe the same could be said of of `metric` and `semantic_model`?

### Use case/motivation

This could be particularly useful to build DAGs with selectors like: `+exposure:exposure_name` in order to select an exposure and all of its ancestors.

I think this is a feature that could make for many who are using exposure in order to trigger the update of all the corresponding models.

For my use-case, I intend to use it combined with the node_converter functionality in order to automate the publication of some dashboards and datasets.

### Related issues

_No response_

ghjklw:
## Description

Implement dbt exposure selector, following exactly the same pattern as source.

## Related Issue(s)

Closes #1551 

## Breaking Change?

No

## Checklist

### DIFF
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -139,6 +139,7 @@ class DbtResourceType(aenum.Enum):  # type: ignore
     SEED = "seed"
     TEST = "test"
     SOURCE = "source"
+    EXPOSURE = "exposure"
 
     @classmethod
     def _missing_value_(cls, value):  # type: ignore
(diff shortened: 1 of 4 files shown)


tatiana (approved):
This is a great feature, thank you for extending Cosmos selectors, @ghjklw !
Since it is a new feature, we'll be releasing this in Cosmos 1.10

PR #1563, opened 2025-02-25
Implement `no_retries_on_test_failure` on `DbtTestLocalOperator`

### LINKED ISSUE #1537: [Feature] raise AirflowFailException in order to avoid retrying dbt test tasks on failure
### Description

Currently, all dbt tasks generated by Cosmos may inherit the same `retries` setting through the `default_args` parameter. However, unlike `dbt run` tasks, it seems unreasonable for `dbt test` tasks to be retried upon failure.

If a data test fails, rerunning it will likely result in the same failure, making retries unnecessary and incurring additional time and resource costs.

To optimize execution, how about introducing a parameter like `no_retries_on_test_failure: bool = True`? When enabled, this would ensure that `dbt test` tasks raise an `AirflowFailException` instead of an `AirflowException`, preventing them from being retried while keeping the existing retry behavior for other dbt tasks.

### Use case/motivation

To optimize execution, dbt test tasks should raise an `AirflowFailException` instead of an `AirflowException` when they fail. This will prevent them from being retried while keeping other dbt tasks, such as `dbt run`, unaffected.

### Related issues

_No response_

okayhooni:
## Description
Added `no_retries_on_test_failure` parameter to `DbtTestLocalOperator` with a default value of `True`.
This parameter raises `AirflowFailException`, ensuring that failed DBT tests are not retried unnecessarily, even if a common retries value of 1 or more is set in `DbtDag`/`DbtTaskGroup`.

Unlike the `on_warning_callback` parameter, it does not seem necessary to propagate `no_retries_on_test_failure` through the `converter` and `airflow/graph` modules. Instead, similar to the `callback` parameter in `AbstractDbtLocalBase`, it can be passed via `operator_args` when needed.

## Related Issue(s)
closes https://github.com/astronomer/astronomer-cosmos/issues/1537

## Breaking Change?
Nope.

## Checklist
- [] I have made corresponding changes to the documentation (if required) -> Not needed in this case.

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -929,8 +929,10 @@ class DbtTestLocalOperator(DbtTestMixin, DbtLocalBaseOperator):
             result = self.build_and_run_cmd(context=context, cmd_flags=self.add_cmd_flags())
         except CosmosDbtRunError as e:
             if self.no_retries_on_test_failure:
-                self.log.error("DBT test failed and `no_retries_on_test_failure=True`, "
-                               "raising AirflowFailException to prevent retries.")
+                self.log.error(
+                    "DBT test failed and `no_retries_on_test_failure=True`, "
+                    "raising AirflowFailException to prevent retries."
+                )
                 raise AirflowFailException(e)
             raise
 
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -16,7 +16,7 @@ from urllib.parse import urlparse
 import airflow
 import jinja2
 from airflow import DAG
-from airflow.exceptions import AirflowException, AirflowSkipException
+from airflow.exceptions import AirflowException, AirflowFailException, AirflowSkipException
 from airflow.models import BaseOperator
 from airflow.models.taskinstance import TaskInstance
 from airflow.utils.context import Context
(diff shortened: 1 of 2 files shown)


pankajastro (approved):
LGTM

pankajkoti (commented):
Hi @okayhooni, thanks a lot for the report and your contribution!

Apologies for the delayed review—I was out last week and have been catching up this week.

After discussing this internally with @tatiana, we feel it would be better to unify this configuration with our existing interface for similar scenarios. For reference, in PR [#1492](https://github.com/astronomer/astronomer-cosmos/pull/1492/), we added support for overriding the profile per dbt node. Similarly, we think it would be more consistent to allow this configuration to be overridden at the test level for specific nodes and at the `dbt_project.yml` level for global application.

For example, similar to [this configuration](https://github.com/astronomer/astronomer-cosmos/blob/e31fc5b4f2b98b6792ec84bf14aca63a32ac7e95/dev/dags/dbt/jaffle_shop/dbt_project.yml#L29), we could support something like:
```
tests:
  jaffle_shop:
      +meta:
        cosmos:
          retries: 0
```

Another point to consider is that test failures may also occur due to connectivity issues, such as being unable to connect to the datasource. In such cases, it might make sense to retry the task. If we can determine the actual reason for the failure—whether it was due to a failed dbt check or a connectivity issue—then we could apply this configuration accordingly (perhaps by analyzing the logs).

Additionally, we might not want to disable retries by default, as it maybe considered breaking change for existing versions. Right now, tasks may retry for various reasons beyond dbt test failures, such as network interruptions.

Looking forward to your thoughts on this!

okayhooni:
@pankajkoti I see your point, and it makes sense.

In our pipeline's use case, data testing is just an additional safeguard, and the test results should not impact the main batch flow. Because of this, retries for test tasks should generally be avoided, regardless of the reason for failure.

However, as you mentioned, a more precise approach would be to parse the error message from the `dbt test` command output to determine whether the failure was truly due to a data-related test issue. I'll work on improving the code to incorporate this logic when I have time..!

pankajkoti:
Hi @okayhooni, following up to see if you'd get a chance to iterate on this PR? I think it would be a great addition. If you'd like us to take up the issue, please let us know!

pankajkoti:
Moving this to the 1.16.0 milestone

Issue #1569, opened 2025-02-27
[Bug] on_warning_callback not working with BUILD test mode


corsettigyg:
### Astronomer Cosmos Version

1.8.2

### dbt-core version

1.8

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

2.8.2

### Operating System

macOS

### Deployment

Docker-Compose

### What happened?

When `TestBehavior.BUILD ` is enabled, the argument `on_warning_callback` does not work for the rendered node. As pointed out by @tatiana, all operators have to accommodate this change, since the `on_warning_callback`  feature essentially requires to handle the logs of the operator, which requires a different approach 
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/kubernetes.py#L120-L128 
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/docker.py#L96-L104
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/gcp_cloud_run_job.py#L126-L134
https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/azure_container_instance.py#L104-L112

I have added a PR to fix the issue for the [LocalOperator](https://github.com/astronomer/astronomer-cosmos/pull/1571)

### Relevant log output

```shell

```

### How to reproduce

1. use `TestBehavior.BUILD `  with some cosmos DAG that contains a dbt model with warns
2. you will see that the on_warning_callback is not being used at all

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

PR #1571, opened 2025-02-27
Support `on_warning_callback` with `TestBehavior.BUILD` and `ExecutionMode.LOCAL`

corsettigyg:
## Description

as of now, when we set TestBehavior.BUILD, we are not leveraging the method on_warning_callback that is available for Test nodes and Source Nodes. I have added the parsing to DbtBuildLocalOperator in order to fix it. I tested it locally and I got positive results

## Related Issue(s)

https://github.com/astronomer/astronomer-cosmos/issues/1569

## Breaking Change?

Should not be

## Checklist

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   pages:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,7 +57,7 @@ repos:
           - --py37-plus
           - --keep-runtime-typing
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.9.6
+    rev: v0.9.7
     hooks:
       - id: ruff
         args:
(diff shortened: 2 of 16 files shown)


tatiana (commented):
@corsettigyg thanks a lot for improving this.

Ideally, we'd also have feature parity / fix this in the other executors, as an example:
- https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/kubernetes.py#L120-L128 (which even has a generic class to handle warnings)
- https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/docker.py#L96-L104
- https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/gcp_cloud_run_job.py#L126-L134
- https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/azure_container_instance.py#L104-L112
- https://github.com/astronomer/astronomer-cosmos/blob/0811e46d80814a3e86cbb87bfe4623ce592e540d/cosmos/operators/aws_ecs.py#L122

So their build operators also can receive `on_warning_callback`.

I understand these changes are outside the scope of the current PR, but could you please log a GitHub issue so we can keep track of them?

I'm happy for us to merge this PR once the checks pass.

tatiana (approved):

corsettigyg:
@tatiana Good catch ! I renamed the name of this PR to specify that it only accommodates the LocalOperator EM. There is an issue logged [here](https://github.com/astronomer/astronomer-cosmos/issues/1569) that I have created.

Since the CI is all green, happy to contribute :)

PR #1582, opened 2025-03-03
remove unused instance arguments from async Bigquery operator

### LINKED ISSUE #1581: [Bug] ASYNC execution mode tries to get connection during parsing
### Astronomer Cosmos Version

1.9.0

### dbt-core version

1.8.0

### What happened?

When parsing a DAG using ExecutionMode. AIRFLOW_ASYNC it tries to get the given connection id.
We don't want this, since it slow downs parsing and we don't have access to the connect ids in all environments
```
airflow.exceptions.AirflowNotFoundException: The conn_id `foo` isn't defined
```

### How to reproduce

Create a profile config with a non existing conn_id and try to parse the DAG

```python
# dummy
    async_py_requirements=["dbt-bigquery==1.8.1", "dbt-core==1.8.1"],
)

    profile_name=dbt_profile.profile_name,
    target_name=dbt_profile.target,
        conn_id="foo",
    )
)

```

joppevos:
## Description

Removes unused Bigquery async arguments that trigger a validation of the profile at parsing time, which we want to avoid.
## Related Issue(s)

closes #1581
## Checklist

### DIFF
--- a/cosmos/operators/_asynchronous/bigquery.py
+++ b/cosmos/operators/_asynchronous/bigquery.py
@@ -57,7 +57,7 @@ def _configure_bigquery_async_op_args(async_op_obj: Any, **kwargs: Any) -> Any:
 
 class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtLocalBase):  # type: ignore[misc]
 
-    template_fields: Sequence[str] = ("gcp_project", "dataset", "location", "compiled_sql")
+    template_fields: Sequence[str] = ("location", "compiled_sql")
     template_fields_renderers = {
         "compiled_sql": "sql",
     }
@@ -73,9 +73,6 @@ class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtL
         self.project_dir = project_dir
         self.profile_config = profile_config
         self.gcp_conn_id = self.profile_config.profile_mapping.conn_id  # type: ignore
-        profile = self.profile_config.profile_mapping.profile  # type: ignore
-        self.gcp_project = profile["project"]
-        self.dataset = profile["dataset"]
         self.extra_context = extra_context or {}
         self.configuration: dict[str, Any] = {}
         self.dbt_kwargs = dbt_kwargs or {}


pankajkoti (approved):
Looks good to me. I believe this breaks nothing, but it’s just that with this we would no longer see the project and dataset values in the rendered templates in the UI. 

Would like to hear if we feel those are crucial @tatiana @pankajastro ?

tatiana (commented):
Hi, @joppevos! 

Thanks a lot for trying out this feature and fixing this. After making the changes you shared in this PR, did `ExecutionMode.AIRFLOW_ASYNC` work as expected?

Is there any chance we could add some tests so that we don't have regressions in the future?

pankajkoti (commented):
@joppevos instead of removing those template fields, would it be possible to store those values during the execution of the task, similar to how we store the compile_sql at https://github.com/astronomer/astronomer-cosmos/blob/c84cb1e81df8060819562f0a6f44d8d7c00578c3/cosmos/operators/_asynchronous/bigquery.py#L160

Probably we could rename the store_compile_sql method to store_template_fields that takes care of these fields as well

joppevos:
>After making the changes you shared in this PR, did `ExecutionMode.AIRFLOW_ASYNC` work as expected?
> Is there any chance we could add some tests so that we don't have regressions in the future?

The ASYNC works as expected. I'll have a look at the tests

pankajkoti (approved):
LGTM. One test is failing, I guess we need to update the method name in the patch for test_execute_complete test. Could you please check it? Thanks for the quick addressing of the suggestion.

tests/operators/_asynchronous/test_bigquery.py
```diff
@@ -134,15 +132,15 @@ def test_store_compiled_sql(mock_rendered_ti, mock_get_remote_sql, profile_confi
     mock_task_instance.task = operator
     mock_context = {"ti": mock_task_instance}
 
-    operator._store_compiled_sql(mock_context, session=mock_session)
+    operator._store_template_fields(mock_context, session=mock_session)
```
tatiana, line 136: @joppevos would it make sense to add some checks here for `operator.gcp_project` and `operator.dataset`?
Happy for us to merge after this. We'll release it as part of Cosmos 1.9.1

PR #1592, opened 2025-03-07
fix: container_name is null for ecs integration

nicor88:
### TL/DR
* pas `container_name` to kwargs
* use a default value for **aws_conn_id**

### Long version

The current implementation of ECS integration implies passing the `container_name` as part of the operator_args.
e.g. 
```
operator_args = {
   "container_name": "main",
   ...
}
```
Anyhow, this lead to errors like this:
```
[2025-03-07, 16:40:34 UTC] {ecs.py:515} INFO - EcsOperator overrides: {'containerOverrides': [{'name': None, 'command': ['dbt', '--no-partial-parse', 'run', '--models', 'my_first_dbt_model'], 'environment': [{'name': 'AIRFLOW_CTX_DAG_OWNER', 'value': '***'}, {'name': 'AIRFLOW_CTX_DAG_ID', 'value': 'example_cosmos'}, {'name': 'AIRFLOW_CTX_TASK_ID', 'value': 'dbt_task_group.my_first_dbt_model.run'}, {'name': 'AIRFLOW_CTX_EXECUTION_DATE', 'value': '2025-03-07T16:40:31.716155+00:00'}, {'name': 'AIRFLOW_CTX_TRY_NUMBER', 'value': '1'}, {'name': 'AIRFLOW_CTX_DAG_RUN_ID', 'value': 'manual__2025-03-07T16:40:31.716155+00:00'}, {'name': 'EXTRA_VAR', 'value': 'extra_value'}]}]}
```
The container name is None, leading to a failure in how boto3 invokes the container.

The issue was due to the fact that `container_name` was not passed to the kwargs, therefore, the container_name was not set properly to the value that was set to cosmos.

### Full logs
<pre>
2025-03-10, 12:49:41 UTC] {ecs.py:512} INFO - Running ECS Task - Task definition: dbt - on cluster dbt
[2025-03-10, 12:49:41 UTC] {ecs.py:515} INFO - EcsOperator overrides: {'containerOverrides': [{'name': None, 'command': ['dbt', '--no-partial-parse', 'run', '--models', 'my_first_dbt_model'], 'environment': [{'name': 'AIRFLOW_CTX_DAG_OWNER', 'value': '***'}, {'name': 'AIRFLOW_CTX_DAG_ID', 'value': 'example_cosmos'}, {'name': 'AIRFLOW_CTX_TASK_ID', 'value': 'dbt_task_group.my_first_dbt_model.run'}, {'name': 'AIRFLOW_CTX_EXECUTION_DATE', 'value': '2025-03-10T12:49:28.878404+00:00'}, {'name': 'AIRFLOW_CTX_TRY_NUMBER', 'value': '1'}, {'name': 'AIRFLOW_CTX_DAG_RUN_ID', 'value': 'manual__2025-03-10T12:49:28.878404+00:00'}, {'name': 'EXTRA_VAR', 'value': 'extra_value'}]}]}
[2025-03-10, 12:49:41 UTC] {base.py:84} INFO - Retrieving connection 'aws_default'
[2025-03-10, 12:49:44 UTC] {credentials.py:1147} INFO - Found credentials in environment variables.
[2025-03-10, 12:49:44 UTC] {taskinstance.py:3313} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/taskinstance.py", line 768, in _execute_task
    result = _execute_callable(context=context, **execute_callable_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/taskinstance.py", line 734, in _execute_callable
    return ExecutionCallableRunner(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/utils/operator_helpers.py", line 252, in run
    return self.func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/cosmos/operators/base.py", line 278, in execute
    self.build_and_run_cmd(context=context, cmd_flags=self.add_cmd_flags())
  File "/home/airflow/.local/lib/python3.12/site-packages/cosmos/operators/aws_ecs.py", line 98, in build_and_run_cmd
    result = EcsRunTaskOperator.execute(self, context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 424, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/providers/amazon/aws/operators/ecs.py", line 526, in execute
    self._start_task()
  File "/home/airflow/.local/lib/python3.12/site-packages/airflow/providers/amazon/aws/operators/ecs.py", line 626, in _start_task
    response = self.client.run_task(**run_opts)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/botocore/client.py", line 569, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/botocore/client.py", line 980, in _make_api_call
    request_dict = self._convert_to_request_dict(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/botocore/client.py", line 1047, in _convert_to_request_dict
    request_dict = self._serializer.serialize_to_request(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/botocore/validate.py", line 381, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter overrides.containerOverrides[0].name, value: None, type: <class 'NoneType'>, valid types: <class 'str'>
</pre>
  

## Related Issue(s)
I didn't created any issue - but I just thought to propose a fix.

## Breaking Change?

It does because user have to use `dbt_container_name` with ECS, but it's currently broken.

## Checklist

### DIFF
--- a/cosmos/operators/aws_ecs.py
+++ b/cosmos/operators/aws_ecs.py
@@ -72,7 +72,6 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
         )
         logger.info('Container name: {}'.format(self.container_name))
         super().__init__(
-            container_name=self.container_name,
             **kwargs)
         # In PR #1474, we refactored cosmos.operators.base.AbstractDbtBase to remove its inheritance from BaseOperator
         # and eliminated the super().__init__() call. This change was made to resolve conflicts in parent class
@@ -116,7 +115,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
         self.overrides = {
             "containerOverrides": [
                 {
-                    "name": self.container_name,
+                    "name": self.container_name or 'main',
                     "command": self.command,
                     "environment": [{"name": key, "value": value} for key, value in self.environment_variables.items()],
                 }
--- a/cosmos/operators/aws_ecs.py
+++ b/cosmos/operators/aws_ecs.py
@@ -40,7 +40,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
     """
 
     template_fields: Sequence[str] = tuple(
-        list(AbstractDbtBase.template_fields) + list(EcsRunTaskOperator.template_fields) + 'dbt_container_name'
+        list(AbstractDbtBase.template_fields) + list(EcsRunTaskOperator.template_fields) + list("dbt_container_name")
     )
 
     intercept_flag = False
@@ -50,7 +50,6 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
         # arguments required by EcsRunTaskOperator
         cluster: str,
         task_definition: str,
-
         dbt_container_name: str,
         #
         aws_conn_id: str = DEFAULT_CONN_ID,
@@ -105,7 +104,6 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
         # For the first round, we're going to assume that the command is dbt
         # This means that we don't have openlineage support, but we will create a ticket
         # to add that in the future
-        logger.debug('Container name: {}'.format(self.dbt_container_name))
         self.dbt_executable_path = "dbt"
         dbt_cmd, env_vars = self.build_cmd(context=context, cmd_flags=cmd_flags)
         self.environment_variables = {**env_vars, **self.environment_variables}
@@ -120,6 +118,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
                 }
             ]
         }
+
 
 class DbtBuildAwsEcsOperator(DbtBuildMixin, DbtAwsEcsBaseOperator):
     """
--- a/cosmos/operators/aws_ecs.py
+++ b/cosmos/operators/aws_ecs.py
@@ -70,6 +70,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
                 "overrides": None,
             }
         )
+        logger.info('Container name: {}'.format(self.container_name))
         super().__init__(
             container_name=self.container_name,
             **kwargs)
--- a/cosmos/operators/aws_ecs.py
+++ b/cosmos/operators/aws_ecs.py
@@ -62,6 +62,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
         self.profile_config = profile_config
         self.command = command
         self.environment_variables = environment_variables or DEFAULT_ENVIRONMENT_VARIABLES
+        self.container_name = container_name
         kwargs.update(
             {
                 "aws_conn_id": aws_conn_id,
--- a/cosmos/operators/aws_ecs.py
+++ b/cosmos/operators/aws_ecs.py
@@ -21,6 +21,7 @@ from cosmos.operators.base import (
 
 logger = get_logger(__name__)
 
+DEFAULT_CONN_ID = "aws_default"
 DEFAULT_ENVIRONMENT_VARIABLES: dict[str, str] = {}
 
 try:
@@ -47,11 +48,11 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
     def __init__(
         self,
         # arguments required by EcsRunTaskOperator
-        aws_conn_id: str,
         cluster: str,
         task_definition: str,
         container_name: str,
         #
+        aws_conn_id: str = DEFAULT_CONN_ID,
         profile_config: ProfileConfig | None = None,
         command: list[str] | None = None,
         environment_variables: dict[str, Any] | None = None,
@@ -69,7 +70,9 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
                 "overrides": None,
             }
         )
-        super().__init__(**kwargs)
+        super().__init__(
+            container_name=self.container_name,
+            **kwargs)
         # In PR #1474, we refactored cosmos.operators.base.AbstractDbtBase to remove its inheritance from BaseOperator
         # and eliminated the super().__init__() call. This change was made to resolve conflicts in parent class
         # initializations while adding support for ExecutionMode.AIRFLOW_ASYNC. Operators under this mode inherit
@@ -103,6 +106,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
         # For the first round, we're going to assume that the command is dbt
         # This means that we don't have openlineage support, but we will create a ticket
         # to add that in the future
+        logger.info('Container name: {}'.format(self.container_name))
         self.dbt_executable_path = "dbt"
         dbt_cmd, env_vars = self.build_cmd(context=context, cmd_flags=cmd_flags)
         self.environment_variables = {**env_vars, **self.environment_variables}
@@ -117,7 +121,6 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
                 }
             ]
         }
-
 
 class DbtBuildAwsEcsOperator(DbtBuildMixin, DbtAwsEcsBaseOperator):
     """
--- a/cosmos/operators/aws_ecs.py
+++ b/cosmos/operators/aws_ecs.py
@@ -22,6 +22,7 @@ from cosmos.operators.base import (
 logger = get_logger(__name__)
 
 DEFAULT_CONN_ID = "aws_default"
+DEFAULT_CONTAINER_NAME = "dbt"
 DEFAULT_ENVIRONMENT_VARIABLES: dict[str, str] = {}
 
 try:
@@ -50,7 +51,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
         # arguments required by EcsRunTaskOperator
         cluster: str,
         task_definition: str,
-        container_name: str,
+        container_name: str = DEFAULT_CONTAINER_NAME,
         #
         aws_conn_id: str = DEFAULT_CONN_ID,
         profile_config: ProfileConfig | None = None,
--- a/cosmos/operators/aws_ecs.py
+++ b/cosmos/operators/aws_ecs.py
@@ -50,7 +50,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
         # arguments required by EcsRunTaskOperator
         cluster: str,
         task_definition: str,
-        container_name: str,
+        dbt_container_name: str,
         #
         aws_conn_id: str = DEFAULT_CONN_ID,
         profile_config: ProfileConfig | None = None,
@@ -61,7 +61,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
         self.profile_config = profile_config
         self.command = command
         self.environment_variables = environment_variables or DEFAULT_ENVIRONMENT_VARIABLES
-        self.container_name = container_name
+        self.dbt_container_name = dbt_container_name
         kwargs.update(
             {
                 "aws_conn_id": aws_conn_id,
@@ -70,9 +70,7 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
                 "overrides": None,
             }
         )
-        logger.info('Container name: {}'.format(self.container_name))
-        super().__init__(
-            **kwargs)
+        super().__init__(**kwargs)
         # In PR #1474, we refactored cosmos.operators.base.AbstractDbtBase to remove its inheritance from BaseOperator
         # and eliminated the super().__init__() call. This change was made to resolve conflicts in parent class
         # initializations while adding support for ExecutionMode.AIRFLOW_ASYNC. Operators under this mode inherit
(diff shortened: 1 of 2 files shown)


cosmos/operators/aws_ecs.py
```diff
@@ -60,12 +61,12 @@ def __init__(
         self.profile_config = profile_config
         self.command = command
         self.environment_variables = environment_variables or DEFAULT_ENVIRONMENT_VARIABLES
-        self.container_name = container_name
         kwargs.update(
             {
                 "aws_conn_id": aws_conn_id,
                 "task_definition": task_definition,
                 "cluster": cluster,
+                "container_name": container_name,
```
tatiana, line 69: @nicor88 Would it make sense for us to define this value dynamically based on other task properties, such as the task name, so end-users wouldn't need to define it manually?

Do we foresee the need to trigger different runs of the same task simultaneously? If yes, ideally, the name would contain the DAG run or some other differentiator while still respecting the rules: https://docs.aws.amazon.com/mediastore/latest/ug/containers-rules-for-names.html

tatiana (commented):
Hi @nicor88,

Thank you very much for reporting the issue and fixing it.

This seems to be an info message:
> [2025-03-07, 16:40:34 UTC] {ecs.py:515} INFO - EcsOperator overrides: {'containerOverrides': [{'name': None, 'command': ['dbt', '--no-partial-parse', 'run', '--models', 'my_first_dbt_model'], 'environment': [{'name': 'AIRFLOW_CTX_DAG_OWNER', 'value': '***'}, {'name': 'AIRFLOW_CTX_DAG_ID', 'value': 'example_cosmos'}, {'name': 'AIRFLOW_CTX_TASK_ID', 'value': 'dbt_task_group.my_first_dbt_model.run'}, {'name': 'AIRFLOW_CTX_EXECUTION_DATE', 'value': '2025-03-07T16:40:31.716155+00:00'}, {'name': 'AIRFLOW_CTX_TRY_NUMBER', 'value': '1'}, {'name': 'AIRFLOW_CTX_DAG_RUN_ID', 'value': 'manual__2025-03-07T16:40:31.716155+00:00'}, {'name': 'EXTRA_VAR', 'value': 'extra_value'}]}]}

Could you share the actual Boto error you faced, with stacktrace? It may be worth to share the versions being used as well (both of Boto, as well as the AWS Airflow provider.

Based on the ECS documentation, it does seem that the container name is mandatory: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/create-task-definition.html#json-validate-for-create

> 12. For each container to define in your task definition, complete the following steps. For Name, enter a name for the container.

That said, it also seems that the Airflow operator handles this:
>:param container_name: The name of the container to fetch logs from. If not set, the first container is used.

https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_modules/airflow/providers/amazon/aws/operators/ecs.html#EcsRunTaskOperator

@nicor88 would it make sense for us to calculate the container dynamically? If this field is mandatory, I'm worried how it would work with `DbtTaskGroup` and `DbtDag`.

@aoelvp94 @CarlosGitto please, could you help reviewing this change? Did you ever face a similar issue, do you have any recommendations?

cosmos/operators/aws_ecs.py
nicor88: I'm not entirely sure that will work with ECS.

The ECS task definition must include the container name - I'm pasting a piece of terraform code to show you how an ECS definition is created:


```
resource "aws_ecs_task_definition" "dbt" {
  family = "dbt"
  execution_role_arn = aws_iam_role.dbt_ecs_execution_role.arn
  # task_role_arn = aws_iam_role.dbt_ecs_task_role.arn

  requires_compatibilities = ["FARGATE"]
  network_mode             = "awsvpc"
  # seems that more CPU is needed to speed up compilation
  cpu                      = 512 # 256
  memory                   = 1024 # 512
  container_definitions = jsonencode([
    {
      name      = "main"
      image     = "${aws_ecr_repository.dbt.repository_url}:latest"
      essential = true
      logConfiguration = {
        logDriver = "awslogs"
        options = {
          "awslogs-group"         = "${aws_cloudwatch_log_group.dbt_log_group.name}"
          "awslogs-region"        = "eu-west-1"
          "awslogs-stream-prefix" = "ecs"
        }
      },
      environment = [
        {
          name  = "STAGE",
          value = "dev"
        }
      ],
      secrets = [
        {
          name      = "EXAMPLE",
          valueFrom = "arn:aws:secretsmanager:eu-west-1:abc:secret:airflow-dbt/dbt-cdf:example::"
        }
      ]
    }])
}

```

As you can see the container definition contains the "container_name" and how environment variables needs to be set for it and seems mandatory.

tatiana: Thanks a lot, @nicor88 ! I need to learn more about ECS. For now, we'll merge your fix, and I hope we can have some proper integration tests on all execution models of Cosmos sooner than later.

tatiana:
@nicor88 thank you very much for your explanation. We still need to solve the tests, our team will try to get the CI checks to pass before we merge this PR, so it can be released in 1.9.1.

tests/operators/test_aws_ecs.py
```diff
@@ -185,6 +185,7 @@ def test_dbt_aes_ecs_overrides_parameter():
 
     assert "containerOverrides" in actual_overrides
     actual_container_overrides = actual_overrides["containerOverrides"][0]
+    assert actual_container_overrides["name"] == "my-dbt-container-name"
```
marianore-muttdata, line 188:
```suggestion
    assert actual_container_overrides["name"] == "dbt"
```
Maybe to maintain consistency with `DEFAULT_CONTAINER_NAME`?

cosmos/operators/aws_ecs.py
```diff
@@ -60,12 +62,12 @@ def __init__(
         self.profile_config = profile_config
         self.command = command
         self.environment_variables = environment_variables or DEFAULT_ENVIRONMENT_VARIABLES
-        self.container_name = container_name
```
marianore-muttdata, line 63: Hey! Just a thing I noted. If we delete this class property, the call at [L116](https://github.com/astronomer/astronomer-cosmos/pull/1592/files#diff-def7397dc271f28ca3cb2d7c56a2cc3953c4cd1c08c3dfaf69328d43d4083acaR116) would fail, right?
I'm not sure if not passing the defined name will affect the container initialization (i.e. being created with another name), as I cannot see any assignation of the `container_name` to the container in the base class [EcsRunTaskOperator](https://github.com/apache/airflow/blob/main/providers/amazon/src/airflow/providers/amazon/aws/operators/ecs.py).
Did you manage to validate these last changes with different container names?

cosmos/operators/aws_ecs.py
```diff
@@ -47,11 +49,11 @@ class DbtAwsEcsBaseOperator(AbstractDbtBase, EcsRunTaskOperator):  # type: ignor
     def __init__(
         self,
         # arguments required by EcsRunTaskOperator
-        aws_conn_id: str,
         cluster: str,
         task_definition: str,
-        container_name: str,
+        container_name: str = DEFAULT_CONTAINER_NAME,
         #
+        aws_conn_id: str = DEFAULT_CONN_ID,
```
marianore-muttdata, line 56:
```suggestion
        aws_conn_id: str = DEFAULT_CONN_ID,
        #
```

marianore-muttdata (commented):
Great catch. Just to add a bit of context, we found that [this PR](https://github.com/apache/airflow/pull/43138) added the changes that overrides the `container_name` functionality.
It's present since the version 9.1.0 of the `apache-airflow-providers-amazon` lib ([ref](https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/changelog.html#id6))

cosmos/operators/aws_ecs.py
nicor88: I did - the class not fail because we assign the container_name as part of the kargs that are passed then to the class.

tests/operators/test_aws_ecs.py
nicor88: Acutally I thought about that, but I believe that this test needs to stay like this, because we want to test out if the actual container value is passed.
therefore we could either add another test, in which we do not set the container_name, and check if it's dbt, and leave this one as it is.

cosmos/operators/aws_ecs.py
nicor88: @marianore-muttdata the trick happen [here](https://github.com/astronomer/astronomer-cosmos/blob/87665dcf49efb82d4bcf7af605c2c834f2214466/cosmos/operators/aws_ecs.py#L83-L87)
container_name it's available as part of DbtAwsEcsBaseOperator, therefore self.container_name actually exists, but it's just passed via kargs.

PS: I tested this branch via an real integration test in my local airflow environment.

As @tatiana mentioned will be good to put this type of integration tests in place in astronomer-cosmos ci.

marianore-muttdata: Right, it takes the value from the base class 👌 
Looks good then 🚀

cosmos/operators/aws_ecs.py
nicor88: @tatiana let me know if you want to have chat to understand a bit more how ECS works - I already shared with few engineers out there some of my learning.

Regarding integrations tests, absolutely, we should be able to those, but I believe that will require some infra in your side - anyhow if necessary, happy to help.

cosmos/operators/aws_ecs.py
nicor88: Honestly I went crazy with understanding that as well.

pankajastro: This param was recently introduced in airflow-providers-amazon. So I think it makes sense to keep this attribute https://github.com/apache/airflow/pull/46152

cosmos/operators/aws_ecs.py
```diff
@@ -60,12 +62,12 @@ def __init__(
         self.profile_config = profile_config
         self.command = command
         self.environment_variables = environment_variables or DEFAULT_ENVIRONMENT_VARIABLES
```
pankajastro, line 64:
```suggestion
        self.environment_variables = environment_variables or DEFAULT_ENVIRONMENT_VARIABLES
        self.container_name = container_name
```

cosmos/operators/aws_ecs.py
nicor88: @pankajastro just to make it clear, the param is kept, it's just passed in another way

nicor88:
@pankajastro maybe you missed this: https://github.com/astronomer/astronomer-cosmos/pull/1592#discussion_r1993804118

but not sure why you put back `self.container_name=container_name`, the container name is passed via kwargs
Therefore https://github.com/nicor88/astronomer-cosmos/blob/78bf46e6b49d7c57df226f06f22a033e83e3cdbb/cosmos/operators/aws_ecs.py#L65 it's redundant IMHO.

pankajastro:
> @pankajastro maybe you missed this: [#1592 (comment)](https://github.com/astronomer/astronomer-cosmos/pull/1592#discussion_r1993804118)
> 
> but not sure why you put back `self.container_name=container_name`, the container name is passed via kwargs Therefore https://github.com/nicor88/astronomer-cosmos/blob/78bf46e6b49d7c57df226f06f22a033e83e3cdbb/cosmos/operators/aws_ecs.py#L65 it's redundant IMHO.

What version of airflow-providers-amazon provider were you using for your testing?

nicor88:
>What version of airflow-providers-amazon provider were you using for your testing?

apache-airflow-providers-amazon[s3fs]==9.4.0

pankajastro:
> @pankajastro `apache-airflow-providers-amazon==9.4.0`, which should be latest. for what I can see [here](https://pypi.org/project/apache-airflow-providers-amazon/)

Hmm, could you please test it with an older version like 9.0.0, if it works, happy to remove?  I believe since this param was introduced in `__init__` of `EcsRunTaskOperator` recently, it would fail for an older version of the provider

pankajkoti:
> @pankajastro maybe you missed this: [#1592 (comment)](https://github.com/astronomer/astronomer-cosmos/pull/1592#discussion_r1993804118)
> 
> but not sure why you put back `self.container_name=container_name`, the container name is passed via kwargs Therefore https://github.com/nicor88/astronomer-cosmos/blob/78bf46e6b49d7c57df226f06f22a033e83e3cdbb/cosmos/operators/aws_ecs.py#L65 it's redundant IMHO.

hi @nicor88 , like @pankajastro mentioned https://github.com/apache/airflow/pull/46152 was only release recently in Airflow Amazon provider release 9.4.0, that is not available in previous versions of the Amazon provider. 

Our CI runs tests on multiple versions of Apache Airflow and we install the provider versions as per the constraints file available for that version of the Airflow installation. Hence, as you can see the PR tests failed earlier at https://github.com/astronomer/astronomer-cosmos/actions/runs/13830044214/job/38692107021#step:6:1083 for Airflow version 2.6 which installed [Amazon provider 8.20.0](https://github.com/astronomer/astronomer-cosmos/actions/runs/13830044214/job/38692107021#step:5:337) that does not have container_name property being assigned via kwargs in our case.

nicor88:
Thanks for the context @pankajastro and @pankajkoti 

My point was simply that the self.container_name=container_name was redundant, but that was based on the latest aws providers, as we need to be back-compatible, then I suppose that leave in it's totally fine - actually a must, thanks for the full context.

Issue #1595, opened 2025-03-10
Run Async DAG without setup/teardown task in CI


pankajastro:
Currently, The DAG run only with the config when the setup/teardown task is enabled.
Aim here is to run the DAG also without setup/teardown task

PR #1599, opened 2025-03-12
Run async in CI DAG without setup/teardown task

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1595

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, test_without_setup_task]
+    branches: [main]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, test_without_setup_task]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
--- a/tests/operators/_asynchronous/test_bigquery.py
+++ b/tests/operators/_asynchronous/test_bigquery.py
@@ -54,7 +54,7 @@ def test_dbt_run_airflow_async_bigquery_operator_base_cmd(profile_config_mock):
 @patch.object(DbtRunAirflowAsyncBigqueryOperator, "build_and_run_cmd")
 def test_dbt_run_airflow_async_bigquery_operator_execute(mock_build_and_run_cmd, profile_config_mock, monkeypatch):
     """Test execute calls build_and_run_cmd with correct parameters."""
-    monkeypatch.setattr("cosmos.operators._asynchronous.bigquery.enable_setup_async_task", False)
+    monkeypatch.setattr("cosmos.operators._asynchronous.bigquery.settings.enable_setup_async_task", False)
     operator = DbtRunAirflowAsyncBigqueryOperator(
         task_id="test_task",
         project_dir="/path/to/project",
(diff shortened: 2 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, test_without_setup_task]
```
```suggestion
    branches: [main]
```

tests/operators/_asynchronous/test_bigquery.py
```diff
@@ -54,7 +54,7 @@ def test_dbt_run_airflow_async_bigquery_operator_base_cmd(profile_config_mock):
 @patch.object(DbtRunAirflowAsyncBigqueryOperator, "build_and_run_cmd")
 def test_dbt_run_airflow_async_bigquery_operator_execute(mock_build_and_run_cmd, profile_config_mock, monkeypatch):
     """Test execute calls build_and_run_cmd with correct parameters."""
-    monkeypatch.setattr("cosmos.operators._asynchronous.bigquery.enable_setup_async_task", False)
+    monkeypatch.setattr("cosmos.operators._asynchronous.bigquery.settings.enable_setup_async_task", False)
```
tatiana, line 57: Why did we use `monkeypatch.setattr` instead of the `patch` decorator, that is most commonly used in Cosmos?
```
@patch("cosmos.operators._asynchronous.bigquery.settings.enable_setup_async_task", return_value=False)
```

tests/operators/test_local.py
```diff
@@ -1444,7 +1444,7 @@ def test_mock_dbt_adapter_unsupported_profile_type():
 def test_async_execution_without_start_task(mock_read_sql, mock_bq_execute, monkeypatch):
     from airflow.providers.google.cloud.operators.bigquery import BigQueryInsertJobOperator
 
-    monkeypatch.setattr("cosmos.operators.local.enable_setup_async_task", False)
+    monkeypatch.setattr("cosmos.operators.local.settings.enable_setup_async_task", False)
```
tatiana, line 1447: Same comment here as https://github.com/astronomer/astronomer-cosmos/pull/1599/files#r1992806219

tests/test_example_dags.py
```diff
@@ -133,3 +125,31 @@ def test_example_dag(session, dag_id: str):
...
+
+
+@pytest.mark.skipif(
+    AIRFLOW_VERSION < Version("2.8") or AIRFLOW_VERSION in PARTIALLY_SUPPORTED_AIRFLOW_VERSIONS,
+    reason="See PR: https://github.com/apache/airflow/pull/34585 and Airflow 2.9.0 and 2.9.1 have a breaking change in Dataset URIs, and Cosmos errors if `emit_datasets` is not False",
+)
+@pytest.mark.integration
+def test_async_example_dag_without_setup_task(session, monkeypatch):
+    monkeypatch.setattr("cosmos.operators.local.settings.enable_setup_async_task", False)
+    monkeypatch.setattr("cosmos.operators.local.settings.enable_teardown_async_task", False)
```
tatiana, line 152: Same comment here as https://github.com/astronomer/astronomer-cosmos/pull/1599/files#r1992806219

tatiana (commented):
Hi @pankajastro , thank you very much for improviing the tests in this area!

Why did we use `monkeypatch` instead of the `patch` decorator, in other parts of Cosmos?

You can find some comments inline:
https://github.com/astronomer/astronomer-cosmos/pull/1599/files#r1992806219

It would be great if we could make this behaviour consistent.

tests/operators/_asynchronous/test_bigquery.py
pankajastro: The initial idea was to set this value at runtime, but since I have changed the import path, the `patch` should also work. Pushed a fix in d4f545836f83fefcf193fc7ce0ff92f4c0d82b27

pankajkoti (approved):
LGTM

PR #1600, opened 2025-03-13
Improve unit tests speed from 89s to 14s

tatiana:
A few tests, such as `test_configure_remote_target_path_no_remote_target`, were taking a long time when using `hatch run tests.py3.9-2.9:test-cov`.

Time to run this command before these changes: 89.34s
Time to run this command after these changes: 14.50s

Also fix unittest that was failing locally.

### DIFF
--- a/tests/operators/test_local.py
+++ b/tests/operators/test_local.py
@@ -1457,6 +1457,7 @@ def test_async_execution_without_start_task(mock_read_sql, mock_bq_execute, monk
     mock_bq_execute.assert_called_once()
 
 
+@pytest.mark.integration
 @pytest.mark.skipif(not AIRFLOW_IO_AVAILABLE, reason="Airflow did not have Object Storage until the 2.8 release")
 @patch("pathlib.Path.rglob")
 @patch("cosmos.operators.local.AbstractDbtLocalBase._construct_dest_file_path")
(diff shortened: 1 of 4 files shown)


pankajastro (approved):

tests/operators/test_local.py
```diff
@@ -1457,6 +1457,7 @@ def test_async_execution_without_start_task(mock_read_sql, mock_bq_execute, monk
     mock_bq_execute.assert_called_once()
 
 
+@pytest.mark.integration
```
pankajkoti, line 1460: do we know what part of this test needs an integration? i am hoping if we could mock something more that maybe missing to be mocked and keep this as a unit test? Maybe we could add a TODO for this to recheck later.

tests/test_io.py
```diff
@@ -66,12 +66,14 @@ def test_upload_artifacts_to_azure_wasb(dummy_kwargs):
         assert hook_instance.load_file.call_count == 2
 
 
+@pytest.mark.integration
```
pankajkoti, line 69: same question as above. do we know what part of this test needs an integration? i am hoping if we could mock something more that maybe missing to be mocked and keep this as a unit test? Maybe we could add a TODO for this to recheck later.

pankajkoti (approved):

tests/test_io.py
tatiana: With the new mocks, this became a unit test 🎉

tests/operators/test_local.py
tatiana: This will need a little bit more work. This is the new ticket: #1608

PR #1601, opened 2025-03-13
Fix path selector when `manifest.json` was created in MS Windows

tatiana:
Users who generated the `manifest.json` using MS Windows and attempted to use Cosmos path selectors after, such as `path:models/edr/run_results' were unable to do so, because the paths in Windows were different from the selector:

```
    "model.elementary.model_run_results": {
      "database": "FDH_DEV_DB",
      "schema": "MONITORING",
      "name": "model_run_results",
      "resource_type": "model",
      "package_name": "elementary",
      "path": "edr\\run_results\\model_run_results.sql",
      "original_file_path": "models\\edr\\run_results\\model_run_results.sql",
      "unique_id": "model.elementary.model_run_results",
      "fqn": [
        "elementary",
        "edr",
        "run_results",
        "model_run_results"
      ],
```

As observed in this example, the property `original_file_path` used the `\\` character as a divider in the path, but the selector checked using the Posix notation.

Since Cosmos implements path selectors using: path_selection in str(node.file_path), we have to normalize the input for the filter to work.

This issue only happened when using `LoadMode.DBT_MANIFEST` and not `LoadMode.DBT_LS` since dbt normalizes this internally when handling selectors as part of this command line.

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -46,6 +46,13 @@ from cosmos.log import get_logger
 logger = get_logger(__name__)
 
 
+def _normalize_path(path: str) -> str:
+    """
+    Converts a potentially Windows path string into a Posix-friendly path.
+    """
+    return Path(path.replace("\\", "/")).as_posix()
+
+
 class CosmosLoadDbtException(Exception):
     """
     Exception raised while trying to load a `dbt` project as a `DbtGraph` instance.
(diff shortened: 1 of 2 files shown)


pankajkoti (approved):
LGTM, thanks for normalising the path here for the selectors to work well on Windows generated manifests 🙌🏽

PR #1602, opened 2025-03-13
Fix select behaviour using `LoadMode.MANIFEST` and a path with star

tatiana:
Let's say the dbt project has a file_path "gen2/models/parent.sql"
```
parent_node = DbtNode(
    unique_id=f"{DbtResourceType.MODEL.value}.{SAMPLE_PROJ_PATH.stem}.parent",
    resource_type=DbtResourceType.MODEL,
    depends_on=[grandparent_node.unique_id, another_grandparent_node.unique_id],
    file_path=SAMPLE_PROJ_PATH / "gen2/models/parent.sql",
    tags=["has_child", "is_child"],
    config={"materialized": "view", "tags": ["has_child", "is_child"]},
)
```

When using Cosmos 1.9.0 with `LoadMode.MANIFEST` and trying to use:

```
RenderConfig(select="gen2/models/*")
```

The selector would not return any results.

It would still work with `LoadMode.DBT_LS`.

The goal of this PR is to solve this issue.

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -175,7 +175,7 @@ class GraphSelector:
         # Index nodes by name, we can improve performance by doing this once
         # for multiple GraphSelectors
         if PATH_SELECTOR in self.node_name:
-            path_selection = self.node_name[len(PATH_SELECTOR) :]
+            path_selection = self.node_name[len(PATH_SELECTOR) :].rstrip("*")
             root_nodes.update({node_id for node_id, node in nodes.items() if path_selection in str(node.file_path)})
 
         elif TAG_SELECTOR in self.node_name:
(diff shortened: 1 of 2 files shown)


cosmos/dbt/selector.py
```diff
@@ -175,7 +175,7 @@ def filter_nodes(self, nodes: dict[str, DbtNode]) -> set[str]:
         # Index nodes by name, we can improve performance by doing this once
         # for multiple GraphSelectors
         if PATH_SELECTOR in self.node_name:
-            path_selection = self.node_name[len(PATH_SELECTOR) :]
+            path_selection = self.node_name[len(PATH_SELECTOR) :].rstrip("*")
```
pankajastro, line 178: Just for my understanding, why does the path contain `*` regex matching?

pankajkoti (approved):
LGTM

cosmos/dbt/selector.py
pankajkoti: I think it's more the case that users may specify a `*` in the path selector and we want to support that

tatiana: Since we're checking if the path defined in the selector is "in" the dbt resource path string, the path stops working if users happen to pass * in Cosmos. Since this is a common practice when using a few dbt filters. and by removing `*` we seem to have the same/expected behaviour, I introduced this change.

PR #1604, opened 2025-03-13
Add test case that fully covers recent select issue

tatiana:
Makes sure the fixes:
- Fix path selector when manifest.json was created in MS Windows (#1601)
- Fix select behaviour using LoadMode.MANIFEST and a path with star (#1602)

Work from an end-to-end perspective, solving Astro customer's original issue.


pankajkoti (approved):
Nice cover!

PR #1617, opened 2025-03-18
Support filtering by config meta nested properties

tatiana:
Allow users to use `select` and `exclude` based on `config.meta` when using `LoadMode.MANIFEST`.

An Astro Customer request reported in: https://astronomer.slack.com/archives/C027CNLBGG7/p1741967575692289?thread_ts=1741881283.263679&cid=C027CNLBGG7.

> For the selector issue, it has to do with the model tagging I mentioned above. Right now, with tags, if a user set the following tags in the `dbt_project.yml`:
> ```
> models:
>   +tags=["core_models", "daily"]
> ```
> but they have a model under models/ that needs to run monthly, so they set the following in the model config block:
> ```
> {{
>   config(
>    tags=["monthly"]
>  )
>}}
>```
> That specific model ends up with the following tags: ["core_models", "daily", "monthly"] . This is because tags are additive when applied this way. With the way our Cosmos configs are setup (daily dag provided as an example):
>```render_config=RenderConfig(
>  load_method=LoadMode.DBT_MANIFEST,
>  select=["tag:core_models", "tag:daily"],
>  exclude=[
>    "tag:hourly",
>    "tag:weekly",
>    "tag:monthly",
>    "tag:static",
>    "tag:analyst_scheduled",
>  ],
>)
> ```
> This would prevent the task that's supposed to run monthly from ever being scheduled.
> Since selects are additive, we can't just have:
>
> ```render_config=RenderConfig(
>  load_method=LoadMode.DBT_MANIFEST,
>  select=["tag:core_models", "tag:daily"],
>)
>```
> as this would pull all models with either core_models or daily tags, which isn't our desired goal.
> In order to resolve this, we are discussing avenues to define the dbt_environment and frequency in a more unique way. Our current thought process is to utilize the meta section of the model config and set the following for models:
>
>```models:
>  <model-dir>:
>    +meta: {
>      "dbt_environment": "<dbt_environment>",
>      "frequency": "<frequency>"
>    }
>```
>This way, we can define defaults in the dbt_project.yml, but then override at a lower granularity when need be. However, in order for this to work with cosmos, we need the ability to select models based on this config, something like config:meta.dbt_environment and config:meta.frequency. Based on our limited testing, this didn't seem possible with `DBT_MANIFEST` load method, but did seem possible with `DBT_LS`

Closes: #1619

### DIFF
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -231,9 +231,9 @@ class GraphSelector:
 
         elif CONFIG_SELECTOR in self.node_name:
             config_selection_key, config_selection_value = self.node_name[len(CONFIG_SELECTOR) :].split(":")
-            # currently tags, materialized, and schema are the only supported config keys
+            # currently tags, materialized, schema and meta are the only supported config keys
             # logic is separated into two conditions because the config 'tags' contains a
-            # list of tags, but the config 'materialized', and 'schema' contain strings
+            # list of tags, the config 'materialized' & 'schema' contain strings and meta contains dictionaries
             if config_selection_key == "tags":
                 root_nodes.update(
                     {
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -258,7 +258,7 @@ class GraphSelector:
                     {
                         node_id
                         for node_id, node in nodes.items()
-                        if _check_nested_value_in_dict(node.config, config_selection_key)
+                        if _check_nested_value_in_dict(node.config, f"{config_selection_key}:{config_selection_value}")
                     }
                 )
             else:
--- a/cosmos/dbt/selector.py
+++ b/cosmos/dbt/selector.py
@@ -504,7 +504,7 @@ class NodeSelector:
                 config.pop(key)
         return True
 
-    def _should_include_based_on_config(self, node: DbtNode, config: dict[Any, Any]) -> bool:
+    def _should_include_based_on_non_meta_and_non_tag_config(self, node: DbtNode, config: dict[Any, Any]) -> bool:
         node_non_meta_or_tag_config = {
             key: value
             for key, value in node.config.items()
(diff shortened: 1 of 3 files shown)


cosmos/dbt/selector.py
```diff
@@ -221,7 +253,16 @@ def filter_nodes(self, nodes: dict[str, DbtNode]) -> set[str]:
...
-
+            elif config_selection_key.startswith(CONFIG_META_PATH):
+                root_nodes.update(
+                    {
+                        node_id
+                        for node_id, node in nodes.items()
+                        if _check_nested_value_in_dict(node.config, f"{config_selection_key}:{config_selection_value}")
+                    }
+                )
+            else:
```
pankajkoti, line 264: It seems now we would still be logging a warning even if the `config_selection_key` is either `materialized` or `schema` whereas earlier that was not the case, no?

cosmos/dbt/selector.py
```diff
@@ -451,6 +493,31 @@ def select_nodes_ids_by_intersection(self) -> set[str]:
...
+        # Deal with meta properties
+        selector_meta_patterns = {key: value for key, value in config.items() if key.startswith(CONFIG_META_PATH)}
+        if selector_meta_patterns:
+            for key, value in selector_meta_patterns.items():
+                if not _check_nested_value_in_dict(node.config, f"{key}:{value}"):
+                    return False
+                config.pop(key)
+        return True
+
+    def _should_include_based_on_config(self, node: DbtNode, config: dict[Any, Any]) -> bool:
```
pankajkoti, line 507:
```suggestion
    def _should_include_based_on_non_meta_and_non_tag_config(self, node: DbtNode, config: dict[Any, Any]) -> bool:
```
I know that could be ugly name but wondering if we could call something like that that since meta & tag are still part of config looking from high level, no?

pankajkoti (approved):
LGTM. A couple of questions in-line.

cosmos/dbt/selector.py
tatiana: Hm, I don't think this will be the case because of:
https://github.com/astronomer/astronomer-cosmos/blob/929e309d95bb9a8c6b859681231b486b7121217c/cosmos/dbt/selector.py#L245-L255

PR #1623, opened 2025-03-18
[Bug fix] no test task on upstream only with detached test

### LINKED ISSUE #1622: [Bug] `should_detach_multiple_parents_tests=True` incorrectly converts fully detached upstream nodes into task groups with non-functional test tasks
### Astronomer Cosmos Version

latest

### dbt-core version

1.8.8

### Versions of dbt adapters

_No response_

### LoadMode

AUTOMATIC

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

2.8.4

### Operating System

_No response_

### Deployment

Other 3rd-party Helm chart

### Deployment details

(Airflow deployment with our own in-house Helm chart)

### What happened?

When `should_detach_multiple_parents_tests=True`, all test tasks are included in task groups, even when the upstream nodes are fully detached.

This results in the creation of task groups that contain test tasks, but these test tasks do not perform any actual operations.

### Relevant log output

```shell
Trying to run dbtRunner with:
 ['test', '--exclude', 'poc_test_with_mng_all', ...]
...
Running with dbt=1.8.8
...
Nothing to do. Try checking your model configs and model specification args
...
```

### How to reproduce

- Set should_detach_multiple_parents_tests=True.
- Execute a DAG with dbt model which has downstream test, but is fully detached.
- Observe that the node is still converted into a task group containing test tasks that do nothing.

## Expected Behavior
- If all test tasks for an upstream node are detached, that node should not be converted into a separate task group containing its own test tasks.
- Task groups should only be created when they contain meaningful and executable test tasks.

### Contact Details

_No response_

okayhooni:
## Description
- Currently, `should_detach_multiple_parents_tests=True` incorrectly converts fully detached upstream nodes into task groups with non-functional test tasks
- Modify the logic to ensure that an upstream node is not converted into a task group with unnecessary dbt-test task when all of its test tasks are detached.

## Related Issue(s)
closes #1622

## Breaking Change?
no

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -864,9 +864,13 @@ class DbtGraph:
         Updates in-place:
         * self.filtered_nodes
         """
+        from cosmos.airflow.graph import is_detached_test, should_create_detached_nodes  # to bypass cyclic import
+
         for _, node in list(self.nodes.items()):
             if node.resource_type == DbtResourceType.TEST:
-                for node_id in node.depends_on:
-                    if node_id in self.filtered_nodes:
-                        self.filtered_nodes[node_id].has_test = True
+                for upstream_node_id in node.depends_on:
+                    if upstream_node_id in self.filtered_nodes and (
+                        should_create_detached_nodes(self.render_config) and not is_detached_test(node)
+                    ):
+                        self.filtered_nodes[upstream_node_id].has_test = True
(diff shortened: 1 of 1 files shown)


tatiana:
Hi @okayhooni did you have any chance of working on this further? We'll release Cosmos 1.10 on the 30th of April, and we'd love to have this bug fix included.

Issue #1621, opened 2025-03-18
Be able to parse and run a DAG containing one Cosmos Local operator in Airflow 3


tatiana:
**Context**

We are facing several issues when attempting to run Cosmos with Airflow 3. To advance, we're reducing the scope of the problem so we can iterate quicker.

Once #1620 is completed, we want to run  in Airflow 3  a simple DAG built with parts of Cosmos using the following configuration:

* `RenderConfig` should use `LoadMode.DBT_MANIFEST`
* `ProfileConfig` with Profile YAML

Example:
```
import os
from datetime import datetime
from pathlib import Path
from typing import Any

from airflow import DAG

from cosmos.config import ProfileConfig
from cosmos.operators.local import DbtCloneLocalOperator

DEFAULT_DBT_ROOT_PATH = Path(__file__).parent / "dbt"
DBT_ROOT_PATH = Path(os.getenv("DBT_ROOT_PATH", DEFAULT_DBT_ROOT_PATH))
DBT_PROJ_DIR = Path("/Users/pankaj/Documents/astro_code/airflow3_dev/jaffle_shop")
DBT_PROFILE_PATH = DBT_PROJ_DIR / "profiles.yml"
DBT_ARTIFACT = DBT_PROJ_DIR / "target"

profile_config = ProfileConfig(
    profile_name="default",
    target_name="dev",
    profiles_yml_filepath=DBT_PROFILE_PATH,
)

with DAG("example_operators", start_date=datetime(2024, 1, 1), catchup=False) as dag:
    clone_operator = DbtCloneLocalOperator(
        profile_config=profile_config,
        project_dir=DBT_PROJ_DIR,
        task_id="clone",
        dbt_cmd_flags=["--models", "stg_customers", "--state", DBT_ARTIFACT],
        install_deps=True,
        append_env=True,
    )

    clone_operator

```

**Acceptance criteria**

* [ ] Any developer should be able to run this DAG, so we should not have hard-coded path values
* [ ] The DAG should be parsed without errors
* [ ] The dbt clone command should be run without errors

pankajastro:
The DAG parsing is working on the branch: https://github.com/astronomer/astronomer-cosmos/tree/run_operator

I'm able to run DAG 
```
import os
from datetime import datetime
from pathlib import Path

from airflow import DAG

from cosmos.config import ProfileConfig
[...]

PR #1642, opened 2025-03-25
Run cosmos operator in Airflow3

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1621

- Add a cosmos DAG in AF3 tooling 
- Add a check to avoid a db call if AF3 is installed
- Fix setup.sh script to install local cosmos changes in virtual env

Screenshot 2025-03-25 at 2 12 19 PM
Screenshot 2025-03-25 at 2 12 31 PM
Screenshot 2025-03-25 at 2 12 44 PM
Screenshot 2025-03-25 at 2 12 55 PM
Screenshot 2025-03-25 at 2 14 53 PM

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -775,6 +775,8 @@ class DbtLocalBaseOperator(AbstractDbtLocalBase, BaseOperator):
                 base_operator_kwargs["outlets"] = [
                     DatasetAlias(name=get_dataset_alias_name(dag_id, task_group_id, self.task_id))
                 ]  # type: ignore
+        if "task_id" in base_operator_kwargs:
+            base_operator_kwargs.pop("task_id")
         BaseOperator.__init__(self, task_id=self.task_id, **base_operator_kwargs)
 
 
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -388,7 +388,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
             else:
                 self.freshness = ""
 
-            _store_freshness_json()
+        _store_freshness_json()
 
     def run_subprocess(self, command: list[str], env: dict[str, str], cwd: str) -> FullOutputSubprocessResult:
         self.log.info("Trying to run the command:\n %s\nFrom %s", command, cwd)
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -31,12 +31,11 @@ from cosmos.cache import (
     _get_latest_cached_package_lockfile,
     is_cache_package_lockfile_enabled,
 )
-from cosmos.constants import FILE_SCHEME_AIRFLOW_DEFAULT_CONN_ID_MAP, InvocationMode
+from cosmos.constants import _AIRFLOW3_VERSION, FILE_SCHEME_AIRFLOW_DEFAULT_CONN_ID_MAP, InvocationMode
 from cosmos.dataset import get_dataset_alias_name
 from cosmos.dbt.project import get_partial_parse_path, has_non_empty_dependencies_file
 from cosmos.exceptions import AirflowCompatibilityError, CosmosDbtRunError, CosmosValueError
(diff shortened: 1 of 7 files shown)


cosmos/operators/local.py
```diff
@@ -742,17 +760,24 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+
+                # ignoring the type because older versions of Airflow raise the follow error in mypy
+                # error: Incompatible types in assignment (expression has type "list[DatasetAlias]", target has type "str")
+                dag_id = kwargs.get("dag")
+                task_group_id = kwargs.get("task_group")
+                base_operator_kwargs["outlets"] = [
+                    DatasetAlias(name=get_dataset_alias_name(dag_id, task_group_id, self.task_id))
+                ]  # type: ignore
+        if "task_id" in base_operator_kwargs:
+            base_operator_kwargs.pop("task_id")
```
pankajastro, line 779: CI is not happy without this: https://github.com/astronomer/astronomer-cosmos/actions/runs/14054743769/job/39351655874

cosmos/settings.py
```diff
@@ -66,3 +66,5 @@ def convert_to_boolean(value: str | None) -> bool:
 enable_telemetry = conf.getboolean("cosmos", "enable_telemetry", fallback=True)
 do_not_track = convert_to_boolean(os.getenv("DO_NOT_TRACK"))
 no_analytics = convert_to_boolean(os.getenv("SCARF_NO_ANALYTICS"))
+
+_AIRFLOW3_VERSION = Version("3.0.0a1")
```
tatiana, line 70: This is a constant, not a user-configurable value, so it shouldn't be in `settings.py`. Should we move it to `constants.py`?

Also, why `a1`? Probably `Version("3")` is more accurate

cosmos/operators/local.py
tatiana: It seems there was some issue in the dependencies, but the tests are working now, right?

tatiana (approved):
Looks great, thanks for sharing this, @pankajastro , it will help us parallelising the work.

Please, if you could address the feedback in the comment before merging, it would be great.

cosmos/settings.py
pankajastro: Yes, constants.py would be the right place for this. Thanks!

> Also, why a1? Probably Version("3") is more accurate

Since we have just the alpha/beta version available at the moment, if we will keep `Version("3")` it would cause a problem in dev env because `3.0.0b2 < 3`

```
>>> v1 = Version("3.0.0b2")
>>> v2 = Version("3")
>>> v1 < v2
True
```

cosmos/operators/local.py
pankajastro: yes, tests are working fine now

cosmos/operators/local.py
```diff
@@ -464,7 +477,8 @@ def _handle_datasets(self, context: Context) -> None:
         outlets = self.get_datasets("outputs")
         self.log.info("Inlets: %s", inlets)
         self.log.info("Outlets: %s", outlets)
-        self.register_dataset(inlets, outlets, context)
+        if AIRFLOW_VERSION < _AIRFLOW3_VERSION:
```
pankajkoti, line 480: I feel it maybe better to introduce these conditions in the respective method and based on the Airflow version, for now we could decide on whether to call the inner function or not. Later on, we could call Airflow version specific inner functions within those functions to handle both versions. WDYT?

pankajkoti (commented):
Sorry for the late review. I just have initiated a discussion inline on the placement of Airflow version check conditionals. Please check

cosmos/operators/local.py
pankajastro: I thought it, but I found this approach easier for handling decorator calls like provide_session. However, I'll give it some more thought and, if possible, will create a follow-up PR.

Issue #1629, opened 2025-03-21
[Bug] DbtTaskGroup in KubernetesOperator Executes Immediately Despite Airflow Task Dependencies


dinesh0002:
### Astronomer Cosmos Version

1.8.2

### dbt-core version

1.9.1

### Versions of dbt adapters

dbt==1.0.0.38.25
dbt-snowflake==1.9.1

### LoadMode

DBT_LS_FILE

### ExecutionMode

KUBERNETES

### InvocationMode

None

### airflow version

2.10.3

### Operating System

Debian

### Deployment

Docker-Compose

### What happened?

When using DbtTaskGroup from dbt-cosmos within a KubernetesOperator, the dbt models execute immediately at the start of the DAG run, even though the task group is explicitly defined to run after upstream tasks. Despite setting dependencies (e.g., start_task >> ... >> process_layer >> ... >> end_task), the DbtTaskGroup ignores the workflow order and runs at the beginning.

process_layer = DbtTaskGroup(
            project_config=ProjectConfig(dbt_project_path=PROJECT_DIR),
            render_config=RenderConfig(load_method=LoadMode.DBT_LS_FILE, dbt_ls_path=DBT_LS_DIR),
            execution_config=ExecutionConfig(execution_mode=ExecutionMode.KUBERNETES),
            operator_args={
                "name": di_context["name"],
                "namespace": di_context["namespace"],
                "image": di_context["image"],
                "image_pull_policy": "Always",
                "is_delete_operator_pod": True,
                "service_account_name": di_context["name"],
                "retries": 3,
                "depends_on_past": False,
                "do_xcom_push": True,
                "get_logs": True,
                "secrets": [secret_all_keys],
                "env_from": [config_map_env_from],
                "env_vars": {
                    "DBT_TARGET": os.getenv("DI_ENV"),
                    "PROJECT_DIR": str(PROJECT_DIR)  
                },
            },
        )

![Image](https://github.com/user-attachments/assets/812074b9-864c-44ab-8ef0-7f9d68e4b031)

### Relevant log output

```shell

```

### How to reproduce

1) Define a DAG with a sequence of tasks, including a DbtTaskGroup configured for Kubernetes execution.

2) Set dependencies such that the DbtTaskGroup (process_layer) runs after upstream tasks (e.g., start_task).

3) Trigger the DAG. Observe that the DbtTaskGroup executes immediately, bypassing the defined dependency order.

### Are you willing to submit PR?

[...]

PR #1648, opened 2025-03-28
Kubernetes Operator argument fix

johnhoran:
## Description
Cosmos 1.9.0 introduced a change such that `AbstractDbtBase` no longer inherits from airflow `BaseOperator` and I believe this introduced two bugs.

- In the past you could give arguments using airflows `default_args` and these would end up being passed to cosmos.  So for example we were specifying the `project_dir` as a default arg, and in 1.9.0 these projects were broken because AbstractDbtBase was not longer getting a project_dir specified.   So the first change is to check both kwargs and default_args for the values AbstractDbtBase requires.  I've only made this change for the kubernetes operator, but I'd argue it should be done for the local a docker ones too.
- The existing code tries to inspect `KubernetesPodOperator` for the args that it requires, but notably this approach misses arguments that are required by `BaseOperator`, e.g. the task_group.

## Checklist

### DIFF
--- a/cosmos/operators/kubernetes.py
+++ b/cosmos/operators/kubernetes.py
@@ -5,7 +5,7 @@ from abc import ABC
 from os import PathLike
 from typing import Any, Callable, Sequence
 
-from airflow.models import TaskInstance
+from airflow.models import TaskInstance, BaseOperator
 from airflow.utils.context import Context, context_merge
 
 from cosmos.config import ProfileConfig
@@ -58,20 +58,31 @@ class DbtKubernetesBaseOperator(AbstractDbtBase, KubernetesPodOperator):  # type
 
     def __init__(self, profile_config: ProfileConfig | None = None, **kwargs: Any) -> None:
         self.profile_config = profile_config
-        super().__init__(**kwargs)
+
         # In PR #1474, we refactored cosmos.operators.base.AbstractDbtBase to remove its inheritance from BaseOperator
         # and eliminated the super().__init__() call. This change was made to resolve conflicts in parent class
         # initializations while adding support for ExecutionMode.AIRFLOW_ASYNC. Operators under this mode inherit
         # Airflow provider operators that enable deferrable SQL query execution. Since super().__init__() was removed
         # from AbstractDbtBase and different parent classes require distinct initialization arguments, we explicitly
         # initialize them (including the BaseOperator) here by segregating the required arguments for each parent class.
-        base_operator_args = set(inspect.signature(KubernetesPodOperator.__init__).parameters.keys())
-        base_kwargs = {}
-        for arg_key, arg_value in kwargs.items():
-            if arg_key in base_operator_args:
-                base_kwargs[arg_key] = arg_value
-        base_kwargs["task_id"] = kwargs["task_id"]
-        KubernetesPodOperator.__init__(self, **base_kwargs)
+        default_args = kwargs.get("default_args", {})
+        base_args = {}
+        for arg in set(inspect.signature(AbstractDbtBase.__init__).parameters.keys()):
+            if arg in kwargs:
+                base_args[arg] = kwargs[arg]
+            elif arg in default_args:
+                base_args[arg] = default_args[arg]
+
+        AbstractDbtBase.__init__(self, **base_args)
+        operator_args = {}
+        for arg in set((
+            *inspect.signature(KubernetesPodOperator.__init__).parameters.keys(),
+            *inspect.signature(BaseOperator.__init__).parameters.keys()
+        )):
+            if arg in kwargs:
+                operator_args[arg] = kwargs[arg]
+
+        KubernetesPodOperator.__init__(self, **operator_args)
 
     def build_env_args(self, env: dict[str, str | bytes | PathLike[Any]]) -> None:
         env_vars_dict: dict[str, str] = dict()
--- a/dev/dags/dbt_docs.py
+++ b/dev/dags/dbt_docs.py
@@ -41,13 +41,11 @@ with DAG(
     schedule_interval="@daily",
     doc_md=__doc__,
     catchup=False,
-    default_args={
-        "retries": 2,
-        "project_dir": DBT_ROOT_PATH / "jaffle_shop",
-    },
+    default_args={"retries": 2},
 ) as dag:
     generate_dbt_docs_aws = DbtDocsS3Operator(
         task_id="generate_dbt_docs_aws",
+        project_dir=DBT_ROOT_PATH / "jaffle_shop",
         profile_config=profile_config,
         connection_id="aws_s3_conn",
         bucket_name="cosmos-ci-docs",
--- a/dev/dags/dbt_docs.py
+++ b/dev/dags/dbt_docs.py
@@ -41,11 +41,13 @@ with DAG(
     schedule_interval="@daily",
     doc_md=__doc__,
     catchup=False,
-    default_args={"retries": 2},
+    default_args={
+        "retries": 2,
+        "project_dir": DBT_ROOT_PATH / "jaffle_shop",
+    },
 ) as dag:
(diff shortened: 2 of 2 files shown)


tatiana (commented):
Thanks a lot @johnhoran for reporting and proposing a fix for this change.

There are a few integration tests that are not passing, could you take a look, please? Since this PR now changes more than the K8s Operator, it may be worth to rename the PR.

I also think it would be essential to have a DAG in `dev/dags` that uses the default args, so we have a complete end-to-end test.

johnhoran:
> There are a few integration tests that are not passing, could you take a look, please?

@tatiana I'm wondering what the right path is here, I feel like maybe I'm letting this PR blow up a little.  The failure is because I tried to follow the convention in airflow that all arguments must be consumed by the operator, so the location keyword in https://github.com/astronomer/astronomer-cosmos/blob/b3535dfd44ad09b5557629470cd101c68c66ea0d/tests/operators/test_airflow_async.py#L56, for example, could be modified to 
```
DbtDag(
        project_config=ProjectConfig(dbt_project_path=DBT_PROJECTS_ROOT_DIR / DBT_PROJECT_NAME),
            async_py_requirements=["dbt-bigquery"],
        ),
        schedule_interval=None,
        catchup=False,
        dag_id="simple_dag_async",
        default_args={"location": "us", "install_deps": True},
)
```
The DAG produced here has both operators that consume the location keyword, like `DbtRunAirflowAsyncOperator`, and ones that don't like `SetupAsyncOperator`.  

But at that point I'm introducing a breaking change.  I do like the idea of maintaining that convention, and the idea of subtracting from the args that are passed to downstream operator feels a bit more resilient than trying to figure out what args are mentioned in the __init__ and just passing those because its always possible that a downstream operator might consume an arg from kwargs instead of explicitly mentioning it in the init method.

tatiana:
@johnhoran This part of the code is currently quite convoluted, and something we want to improve sooner rather than later.

@pankajkoti and @pankajastro recently worked on this part of the code, and a change was merged yesterday.  @johnhoran would you like to rebase and see if things have improved in any way? Otherwise, one of them may be able to jump in and give some advice!

cosmos/operators/aws_ecs.py
```diff
@@ -25,7 +25,10 @@
 DEFAULT_CONTAINER_NAME = "dbt"
 DEFAULT_ENVIRONMENT_VARIABLES: dict[str, str] = {}
 
+from airflow.models import BaseOperator
```
pankajkoti, line 28:
```suggestion
try:  # Airflow 3
    from airflow.sdk.bases.operator import BaseOperator
except ImportError:  # Airflow 2
    from airflow.models import BaseOperator
```
we're currently adding support for Airflow 3 in parallel, and it would be nice to add the altered path for BaseOperator in Airflow3 if possible in this PR. This was first added in https://github.com/astronomer/astronomer-cosmos/pull/1657/files for your reference. Similar suggestion for other places in the PR.

cosmos/operators/aws_ecs.py
```diff
@@ -68,25 +71,41 @@ def __init__(
                 "aws_conn_id": aws_conn_id,
                 "task_definition": task_definition,
                 "cluster": cluster,
-                "container_name": container_name,
                 "overrides": None,
             }
         )
-        super().__init__(**kwargs)
+
+        if provider_version >= "9.3.0":
```
pankajkoti, line 78: Do we observe a failure on having this kwarg on versions < 9.3.0?

Also, we need to convert these versions into packaging.version.Version instances similar to https://github.com/astronomer/astronomer-cosmos/blob/faa8f5899e15336db4a0f15d4e277b3e0c4901f3/cosmos/operators/local.py#L98

as otherwise string comparisons don't work as expected. 
e.g. when the provider_version = '10.0.0', '10.0.0' > '9.3.0' evaluates to False instead of the expected True value

cosmos/operators/aws_ecs.py
```diff
@@ -68,25 +71,41 @@ def __init__(
...
+        for arg in {*inspect.signature(AbstractDbtBase.__init__).parameters.keys()}:
+            try:
+                base_kwargs[arg] = kwargs[arg]
+                if arg not in operator_args:
+                    operator_kwargs.pop(arg)
+            except KeyError:
+                try:
+                    base_kwargs[arg] = default_args[arg]
+                except KeyError:
+                    pass
```
pankajkoti, line 106:
```suggestion
            if arg in kwargs:
                base_kwargs[arg] = kwargs[arg]
                if arg not in operator_args:
                    operator_kwargs.pop(arg, None)
            elif arg in default_args:
                base_kwargs[arg] = default_args[arg]
```

This should yield the same result, but without the need for nested try-except KeyError handling. WDYT? If it looks fine, then similar suggestion for other places in the PR.

cosmos/operators/base.py
```diff
@@ -146,7 +147,6 @@ def __init__(
         self.dbt_cmd_global_flags = dbt_cmd_global_flags or []
         self.cache_dir = cache_dir
         self.extra_context = extra_context or {}
-        kwargs.pop("full_refresh", None)  # usage of this param should be implemented in child classes
```
pankajkoti, line 149: @tatiana do you recall if we had a reason for doing this? also please notice we're adding the full_refresh as a kwarg above. Just wished to draw your attention on this change to see if you're aware of/recall of why we had this earlier.

cosmos/operators/gcp_cloud_run_job.py
```diff
@@ -80,17 +81,30 @@ def __init__(
                 "project_id": project_id,
                 "region": region,
                 "job_name": job_name,
-                "command": command,
-                "environment_variables": environment_variables,
```
pankajkoti, line 84: can you please share why we're removing these?

pankajkoti (commented):
Thanks for observing these issues and helping us fix those @johnhoran 🙇🏽 

I have some questions and suggestions in-line. Looking forward to your thoughts there and getting this PR in soon!

hiimseeking:
Will also fix #1629

cosmos/operators/aws_ecs.py
johnhoran: Good catch. Yeah prior to version 9.3.0 the container name wasn't an accepted parameter.  So on some of the tests which use older versions of the AWS package it was causing failures ones I required all parameters to be consumed.

cosmos/operators/aws_ecs.py
johnhoran: I don't feel very strongly about it, but I've heard the try except is preferred because its a single lookup to the dictionary instead of checking if the key exists and then getting the value for it.

cosmos/operators/base.py
johnhoran: Looking at https://github.com/astronomer/astronomer-cosmos/pull/590 which first introduced it, it sounds like it might be a similar thing to the core question here.  The operator_args is passed to all operators but not all operators need all the arguments, whereas operators consume from default_args selectively.  In the case of "full_refresh", the test operators didn't consume it.

johnhoran:
> I believe it would be nice to keep it consistent to how it is currently. 

Okay, I've re-factored the PR to just fix the operator argument

johnhoran:
Hi @pankajkoti sorry this has gone around so many times.  The test is passing on my end now with airflow2.4, so hopefully this is the last run.

pankajkoti:
> Hi @pankajkoti sorry this has gone around so many times. The test is passing on my end now with airflow2.4, so hopefully this is the last run.

thanks @johnhoran . I re-triggered the CI for running integration tests. We can definitely re-trigger more if need be. And please don't be sorry :), all you're doing is helping the community with the fix and also investing your time here, that's much appreciated 🙇🏽

pankajkoti (approved):
LGTM. Thanks a lot @johnhoran for fixing these 🙇🏽 👏🏽

tatiana (approved):

Issue #1512, opened 2025-02-06
[Feature] Extend `EXECUTION_MODE.AIRFLOW_ASYNC` for `dbt-snowflake` adapter


pankajkoti:
### Description

Continuing the work introduced in Cosmos 1.7 to support executing dbt project resources in a deferrable fashion using Airflow provider operators, this feature aims to extend EXECUTION_MODE.AIRFLOW_ASYNC to the dbt-Snowflake adapter.

Initially, this approach was implemented for BigQuery and refined through PRs [#1474](https://github.com/astronomer/cosmos/pull/1474) and [#1483](https://github.com/astronomer/cosmos/pull/1483) to ensure robustness. The next step is to explore how we can extend this functionality to Snowflake, leveraging its capabilities to execute dbt models asynchronously while integrating seamlessly with Airflow’s deferrable operators.

Explore how we can utilize existing deferrable operators from the Apache Airflow Snowflake provider or astronomer-providers. If suitable operators are unavailable, consider adding the necessary components to the Cosmos codebase to enable deferrable execution efficiently and accelerate development.

### Use case/motivation

This enhancement will improve efficiency, reduce resource consumption, and provide users with a more scalable and optimized way to run dbt projects on Snowflake within Airflow.

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

jroach-astronomer:
@pankajkoti, do you mind assigning this issue to me? Thanks!

pankajkoti:
@jroach-astronomer assigned to you :)

jroachgolf84:
Hey all - I am starting to chip away at this PR, and I have a few questions. In https://github.com/astronomer/cosmos/pull/1474, we've mocked the `BigQueryConnectionManager.execute` method by returning a `BiqQueryAdapterResponse` and an `empty_table`. Why is the case?

In Snowflake, there is no `SnowflakeAdapterResponse`. Instead, I've used the `AdapterResponse`, and mocked the SnowflakeConnectionManager.execute` method. Thoughts?

tatiana:
Great progress, @jroach-astronomer !

> Hey all - I am starting to chip away at this PR, and I have a few questions. In https://github.com/astronomer/astronomer-cosmos/pull/1474, we've mocked the BigQueryConnectionManager.execute method by returning a BiqQueryAdapterResponse and an empty_table. Why is the case?

[...]

PR #1650, opened 2025-03-31
Extend `EXECUTION_MODE.AIRFLOW_ASYNC` for `dbt-snowflake` adapter

ghost:
## Description

Draft PR for extended the `EXECUTION_MODE.AIRFLOW_ASYNC` for the `dat-snowflake` adapter.

## Related Issue(s)

addresses: #1512

## Breaking Change?

Will confirm once PR is reviewed.

## Checklist

### DIFF
--- a/cosmos/operators/_asynchronous/snowflake.py
+++ b/cosmos/operators/_asynchronous/snowflake.py
@@ -59,11 +59,9 @@ def _configure_snowflake_async_op_args(async_op_obj: Any, **kwargs: Any) -> Any:
         raise CosmosValueError("Keyword argument 'sql' is required for Snowflake Async operator.")
 
     async_op_obj.sql = sql
-
     return async_op_obj
 
 
-# This is no longer a SnowflakeOperator, meaning the SQLExecuteQueryOperator needs to be used.
 class DbtRunAirflowAsyncSnowflakeOperator(SnowflakeSqlApiOperator, AbstractDbtLocalBase):  # type: ignore[misc]
     template_fields: Sequence[str] = ("compiled_sql", "freshness")
     template_fields_renderers = {"compiled_sql": "sql", "freshness": "json"}
@@ -145,7 +143,6 @@ class DbtRunAirflowAsyncSnowflakeOperator(SnowflakeSqlApiOperator, AbstractDbtLo
                 "The `profile_config.profile`_mapping attribute must be defined to use `ExecutionMode.AIRFLOW_ASYNC`"
             )
 
-        # self.gcp_project = profile["project"]  # This should only needed for BigQuery
         self._refresh_template_fields(context=context, session=session)
 
 
--- a/cosmos/dataset.py
+++ b/cosmos/dataset.py
@@ -35,7 +35,6 @@ def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_i
     return "__".join(identifiers_list)
 
 
-# TODO: This should live somewhere else...
 def configure_datasets(operator_kwargs: dict[str, Any], task_id: str) -> None:
     """
     Sets the outlets for the operator by creating a DatasetAlias.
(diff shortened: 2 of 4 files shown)


jroach-astronomer:
@tatiana, @pankajkoti, do you mind taking a look at this PR? This is only a draft, and I don't actually have something that's running. However, I want to make sure that I'm on the right track.

I'm also running into the following (see below). Any idea what that might be?

```
Compilation Error
  agate.mapped_sequence.MappedSequence object has no element 1
  > in macro list_relations_without_caching (macros/adapters/metadata.sql)
  > called by <Unknown>
```

cosmos/operators/_asynchronous/base.py
```diff
@@ -34,7 +39,70 @@ def _create_async_operator_class(profile_type: str, dbt_class: str) -> Any:
...
+    dag_id = operator_kwargs.get("dag")
+    task_group_id = operator_kwargs.get("task_group")
+    operator_kwargs["outlets"] = [
+        DatasetAlias(name=get_dataset_alias_name(dag_id, task_group_id, task_id))
+    ]  # type: ignore
+
+    return operator_kwargs
+
+
+def get_remote_sql(async_context: dict[str, Any], project_dir: str) -> str:
```
pankajkoti, line 63: It might make sense to move this method to the `io` module https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/io.py and re-use everywhere (refactoring in BigQuery can be a separate PR), but for now, we could at least add this method there

cosmos/operators/_asynchronous/snowflake.py
```diff
@@ -0,0 +1,154 @@
...
+    sql = kwargs.get("sql")
+
+    if not sql:
+        raise CosmosValueError("Keyword argument 'sql' is required for Snowflake Async operator.")
+
+    async_op_obj.sql = sql
+    return async_op_obj
+
+
+# This is no longer a SnowflakeOperator, meaning the SQLExecuteQueryOperator needs to be used. Does this make this
```
pankajkoti, line 64: Is this comment relevant. I see we're already using the SnowflakeSqlApiOperator which seems to have the deferrable support.

pankajkoti (commented):
The implementation and approach look solid to me, @jroach-astronomer, and align well with the direction we’re heading 👏🏽 

on the error you're seeing, can you provide more details on when and where you're getting this error (is it while running the DAG or during parsing)? any additional traceback that you see?

cosmos/operators/_asynchronous/snowflake.py
```diff
@@ -0,0 +1,154 @@
...
+    @provide_session
+    def _store_template_fields(self, context: Context, session: Session = NEW_SESSION) -> None:
+        if not settings.enable_setup_async_task:
+            self.log.info("SQL cannot be made available, skipping registration of compiled_sql template field")
+            return
+
+        sql = get_remote_sql(async_context=self.async_context, project_dir=self.project_dir).strip()
+
+        self.log.debug("Executed SQL is: %s", sql)
+        self.compiled_sql = sql  # TODO: What is this being used for?
```
pankajkoti, line 143: This is a templated field for the class and it gets showed in the UI (task details rendered fields section) for the users to see what SQL query was being executed as part of the task run

pankajkoti (commented):
The implementation and approach look solid to me, @jroach-astronomer, and align well with the direction we’re heading 👏🏽 

on the error you're seeing, can you provide more details on when and where you're getting this error (is it while running the DAG in the monkey-patched method or during DAG parsing)? any additional traceback that you see?

cosmos/operators/_asynchronous/snowflake.py
jroach-astronomer: Will change that, thanks!

jroach-astronomer:
> on the error you're seeing, can you provide more details on when and where you're getting this error (is it while running the DAG in the monkey-patched method or during DAG parsing)? any additional traceback that you see?

I am seeing this error when triggering the DAG (the first Task fails with this exception). When I run again, I'll provide more details.

PR #1656, opened 2025-04-09
Improve AF3 tooling

pankajastro:
- Unpin Airflow and provider version and install it with --pre arg
- Use Cosmos from the root without installing it

### DIFF
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -7,9 +7,19 @@ from abc import ABCMeta, abstractmethod
 from pathlib import Path
 from typing import Any, Sequence, Tuple
 
+import airflow
 import yaml
 from airflow.utils.context import Context, context_merge
-from airflow.utils.operator_helpers import context_to_airflow_vars
+from packaging.version import Version
+
+from cosmos.constants import _AIRFLOW3_VERSION
+
+AIRFLOW_VERSION = Version(airflow.__version__)
+
+if AIRFLOW_VERSION < _AIRFLOW3_VERSION:
+    from airflow.utils.operator_helpers import context_to_airflow_vars
+else:
+    from airflow.sdk.execution_time.context import context_to_airflow_vars  # type: ignore
 from airflow.utils.strings import to_boolean
 
 from cosmos.dbt.executable import get_system_dbt
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -7,19 +7,14 @@ from abc import ABCMeta, abstractmethod
 from pathlib import Path
 from typing import Any, Sequence, Tuple
 
-import airflow
 import yaml
 from airflow.utils.context import Context, context_merge
-from packaging.version import Version
 
-from cosmos.constants import _AIRFLOW3_VERSION
-
-AIRFLOW_VERSION = Version(airflow.__version__)
-
-if AIRFLOW_VERSION < _AIRFLOW3_VERSION:
+try:
     from airflow.utils.operator_helpers import context_to_airflow_vars
-else:
+except ImportError:  # pragma: no cover
     from airflow.sdk.execution_time.context import context_to_airflow_vars  # type: ignore
+
 from airflow.utils.strings import to_boolean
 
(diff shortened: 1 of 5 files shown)


scripts/airflow3/env.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/bash
 
-set -e
-set -x
+#set -e
+#set -x
```
pankajastro, line 4:
```suggestion
set -e
```

cosmos/operators/base.py
```diff
@@ -7,9 +7,19 @@
...
+from packaging.version import Version
+
+from cosmos.constants import _AIRFLOW3_VERSION
+
+AIRFLOW_VERSION = Version(airflow.__version__)
+
+if AIRFLOW_VERSION < _AIRFLOW3_VERSION:
+    from airflow.utils.operator_helpers import context_to_airflow_vars
+else:
+    from airflow.sdk.execution_time.context import context_to_airflow_vars  # type: ignore
```
pankajkoti, line 22: Just a point of discussion. How about try -- except ImportError?

scripts/airflow3/requirements.txt
```diff
@@ -1,9 +1,7 @@
---find-links https://dist.apache.org/repos/dist/dev/airflow/3.0.0b2/
-apache-airflow==3.0.0b2
-apache-airflow-task-sdk==1.0.0b2
-apache-airflow-providers-standard==1.0.0b2
-apache-airflow-providers-fab==2.0.0b2
-apache-airflow-providers-celery==3.11.0b2
```
pankajkoti, line 6: do we not need celery?

scripts/airflow3/setup.sh
```diff
@@ -4,26 +4,17 @@
 set -e
 
 # Create a UV virtual environment named 'env' (you can change this as needed)
-echo "Creating UV virtual environment at $(pwd)/tools"
-uv venv venv --directory "$(pwd)/scripts/airflow3"
+echo "Creating virtual environment at $(pwd)/tools"
+python3 -m venv "$(pwd)/scripts/airflow3/venv"
```
tatiana, line 8: We'll probably need to make this change to more places, but it would make our local development experience easier, so we can easily spot if we're on AF2 or AF3:
```suggestion
python3 -m venv "$(pwd)/scripts/airflow3/venv-af3"
```

scripts/airflow3/requirements.txt
```diff
@@ -1,9 +1,7 @@
...
-apache-airflow-providers-cncf-kubernetes==10.4.0b2
+apache-airflow
+apache-airflow-task-sdk
+apache-airflow-providers-standard
+apache-airflow-providers-fab
```
pankajkoti, line 4: can we check that this does not reinstall airflow? and if it does it installs airflow 3?

scripts/airflow3/requirements.txt
pankajkoti, line 6: do we not need celery provider?

scripts/airflow3/setup.sh
```diff
@@ -4,26 +4,17 @@
...
 
 # Activate the virtual environment
 echo "Activating virtual environment..."
 source "$(pwd)/scripts/airflow3/venv/bin/activate"
 
 # Install dependencies in the virtual environment
 echo "Installing dependencies..."
-uv pip install -r "$(pwd)/scripts/airflow3/requirements.txt"
+pip3 install --pre -r "$(pwd)/scripts/airflow3/requirements.txt"
```
tatiana, line 16: Once we need to re-add the providers, we can leverage the constraints file eg.
```
--constraints https://raw.githubusercontent.com/apache/airflow/refs/tags/constraints-3.0.0rc1.post4/constraints-3.9.txt
```

Since this is not needed now, we probably can do it only once we need

scripts/airflow3/requirements.txt
pankajastro: Should not reinstall unless there are conflicts because we are not enforcing it with command args

PR #1663, opened 2025-04-11
Fix `test_indirect_selection` flag to be propagated in case of `TestBehavior.BUILD`

corsettigyg:
## Description

At my current company, we are using  `TestBehavior.BUILD` to build our DAGs and during development I have noticed that for `TestBehavior.BUILD`, the flag `test_indirect_selection` is not being propagated to the task, which is not the case if we are using the classic **run >> test** approach of cosmos.

By tweaking a little bit the graph.py file, i found a fix that worked fine when I tested it.

` ['build', '--indirect-selection', 'buildable', '--models', '***', '--project-dir', ***', '--profiles-dir', '***', '--profile', '***', '--target', 'dev']`

## Breaking Change?

Should not be, since we are just propagating the flag to the required parts.

## Checklist


tatiana (approved):
Thanks a lot for reporting and fixing this, @corsettigyg !

PR #1665, opened 2025-04-14
Check only major version until AF3 stable release

pankajastro:
When running the Cosmos DAG with runtime-image AF 3 using astro-cli, it seems to be trying to access the database because some condition checks are unintentionally returning true

```python
>>> from packaging.version import Version
>>> _AIRFLOW3_VERSION1 = Version("3.0.0a1")
>>> _AIRFLOW3_VERSION2 = Version("3.0.0.dev202504070002")
>>> _AIRFLOW3_VERSION2 < _AIRFLOW3_VERSION1
True
>>>
```


pankajkoti (approved):
Nice

Issue #1630, opened 2025-03-24
[Feature] Customer request on dbt deps pre-install + update if needed


tatiana:
### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

PR #1668, opened 2025-04-15
Support running `dbt deps` incrementally to pre-defined `dbt_packages` during DAG parsing

tatiana:
Support running `dbt deps` incrementally to pre-calculated `dbt_packages` during DAG parsing. This was a use case requested by an Astro customer.

Before this change, Cosmos supported two types of configuration:
* If users choose `RenderConfig. dbt_deps=False` or `ProjectConfig.install_dbt_deps=False`, Cosmos would create a symbolic link for the user's pre-defined `dbt_packages` (background:  #488, #600, #730)
* If users choose `RenderConfig. dbt_deps=True` or `ProjectConfig.install_dbt_deps=True` (default), Cosmos would ignore any user-predefined ' dbt_packages` and do a run `dbt deps` from scratch from a temporary folder.

An Astronomer customer requested to reuse the defined initially `dbt_packages` directory and run `dbt deps` (incrementally).

We do not run dbt commands directly in the original dbt project folder with Cosmos because some users use read-only filesystems (#414). We also decided to use symbolic links instead of copying the directory due to performance issues (#488). Since we did not want to introduce a breaking change in a minor Cosmos release by changing the existing Cosmos 1.x behaviour to meet this new use case, this PR supports:
* Copying the dbt deps related files (dbt packages folder and symbolic link) to the Cosmos temporary folder; and
* Running `dbt deps`.

So this is not a breaking change, users must opt into this behaviour by either:
- Changing individual `DbtDag` or `DbtTaskGroup` instances, using `ProjectConfig.copy_dbt_packages=True` (new configuration) and `RenderConfig. dbt_deps=True` or `ProjectConfig.install_dbt_deps=True`; or
- Changing the behaviour globally, by setting the Airflow configuration either via the environment variable `AIRFLOW__COSMOS__DEFAULT_COPY_DBT_PACKAGES_VALUE=True`. or via the `airflow.cfg`:
```
[cosmos]
default_copy_dbt_packages_value=True
```

The following two missing parts are being added as part of a separate PR:
- Mimic this behaviour during task execution;
- Update documentation to be representative of both changes.

Depends on #1669 

Related to: #1630


cosmos/dbt/graph.py
```diff
@@ -522,7 +528,7 @@ def load(
             LoadMode.CUSTOM: self.load_via_custom_parser,
             LoadMode.DBT_LS: self.load_via_dbt_ls,
             LoadMode.DBT_LS_FILE: self.load_via_dbt_ls_file,
-            LoadMode.DBT_LS_CACHE: self.load_via_dbt_ls_cache,
```
pankajkoti, line 525: why are  we changing this?

cosmos/dbt/graph.py
```diff
@@ -635,6 +641,59 @@ def run_dbt_deps(self, dbt_cmd: str, dbt_project_path: Path, env: dict[str, str]
...
+        #  | C         | False            | True              | True                                | should copy dbt deps files   |
+        #  | D         | True             | True              | True                                | should copy & run `dbt deps` |
+        #
+        # Use cases description:
+        # A. High performance and deps may become outdated: Users run `dbt deps` outside of Cosmos and give pre-generated dbt packages. Dependencies may become outdated.
+        # B. Low performance and up-to-date deps: Cosmos always run `dbt deps` from scratch, every time the DAG is parsed (every time the DAG is parsed).
+        # C. (Non-practical) Middle performance and deps may become outdated: Users manage `dbt deps` outside of Cosmos and give pre-generated dbt packages. Dependencies may become outdated. More expensive than A, similar behaviour.
+        # D. Middle performance and up-to-date deps: Users run `dbt deps` outside of Cosmos and give pre-generated dbt packages. Cosmos runs dbt deps taking into account those user-generated files.
```
pankajkoti, line 690: I am trying to understand more here "Cosmos runs dbt deps taking into account those user-generated files."

What happens here? Do we or dbt take into account the version/checksum to handle installation of packages incrementally?

cosmos/dbt/project.py
```diff
@@ -34,12 +36,36 @@ def has_non_empty_dependencies_file(project_path: Path) -> bool:
...
+    for relative_path in dbt_packages_paths:
+        src_path = source_folder / relative_path
+        dst_path = target_folder / relative_path
+
+        os.makedirs(os.path.dirname(dst_path), exist_ok=True)
+
+        if src_path.is_dir():
+            shutil.copytree(src_path, dst_path, dirs_exist_ok=True)
+        else:
+            shutil.copy2(src_path, dst_path)
```
pankajkoti, line 58: Trying to understand what is the advantage of shutil.copy2 over shutil.copy here. Could you please share?

pankajkoti (approved):
LGTM. Some minor questions for my understanding in-line

cosmos/settings.py
```diff
@@ -29,6 +29,7 @@
 enable_cache_profile = conf.getboolean("cosmos", "enable_cache_profile", fallback=True)
 dbt_profile_cache_dir_name = conf.get("cosmos", "profile_cache_dir_name", fallback="profile")
 virtualenv_max_retries_lock = conf.getint("cosmos", "virtualenv_max_retries_lock", fallback=120)
+default_copy_dbt_packages_value = conf.getboolean("cosmos", "default_copy_dbt_packages_value", fallback=False)
```
pankajastro, line 32: Shall we document it here: https://github.com/astronomer/astronomer-cosmos/blob/main/docs/configuration/cosmos-conf.rst

cosmos/dbt/project.py
tatiana: `copy2` preserves the original file metadata - it probably does't make a difference here

pankajastro (approved):

cosmos/settings.py
tatiana: Yes, @pankajastro ! I'm creating a follow-up PR with documentation changes

cosmos/dbt/graph.py
pankajastro: I think If `dbt_packages` is missing, it will use `package-lock.yml` (if present) to install the correct package versions.

cosmos/dbt/project.py
```diff
@@ -34,12 +36,36 @@ def has_non_empty_dependencies_file(project_path: Path) -> bool:
...
+    for relative_path in dbt_packages_paths:
+        src_path = source_folder / relative_path
+        dst_path = target_folder / relative_path
+
+        os.makedirs(os.path.dirname(dst_path), exist_ok=True)
+
+        if src_path.is_dir():
+            shutil.copytree(src_path, dst_path, dirs_exist_ok=True)
+        else:
+            shutil.copy(src_path, dst_path)
```
tatiana, line 58:
```suggestion
            shutil.copy2(src_path, dst_path)
```

tatiana: Reverting due to failing test: https://github.com/astronomer/astronomer-cosmos/actions/runs/14494584744/job/40658978949?pr=1668

`copy2` handles the scenario where the original file doesn't exist - something that we care about

tuantran0910:
Hi, this feature is great! I just have a question — will it still work when using the KubernetesExecutor in Airflow? Since each task runs in its own Pod with a separate environment, and the Pod is removed after the task completes (along with any temporary files), wouldn't that mean dependencies have to be reinstalled from scratch on every run?

PR #1670, opened 2025-04-16
Support running `dbt deps` incrementally to pre-defined `dbt_packages` during task execution

tatiana:
# Motivation

Support running `dbt deps` incrementally to pre-calculated `dbt_packages` during the task execution. This was a use case requested by an Astro customer.

Before this change, Cosmos supported two types of configuration:
* If users choose `operator_args={"install_deps": False}` or `ProjectConfig.install_dbt_deps=False`, Cosmos would create a symbolic link for the user's pre-defined `dbt_packages` (background:  #488, #600, #730)
* If users choose `operator_args={"install_deps": True}` or `ProjectConfig.install_dbt_deps=True` (default), Cosmos would ignore any user-predefined ' dbt_packages` and do a run `dbt deps` from scratch from a temporary folder.

An Astronomer customer requested to reuse the defined initially `dbt_packages` directory and run `dbt deps` (incrementally).

# Implementation

We do not run dbt commands directly in the original dbt project folder with Cosmos because some users use read-only filesystems (#414). We also decided to use symbolic links instead of copying the directory due to performance issues (#488). Since we did not want to introduce a breaking change in a minor Cosmos release by changing the existing Cosmos 1.x behaviour to meet this new use case, this PR supports:
* Copying the dbt deps related files (dbt packages folder and symbolic link) to the Cosmos temporary folder; and
* Running `dbt deps`.

So this is not a breaking change, users must opt into this behaviour by using `ProjectConfig.copy_dbt_packages=True` (new configuration) and `operator_args={"install_dbt_deps": True}` or `ProjectConfig. install_dbt_deps =True` and one of the following:
- Changing the operator to receive the argument `copy_dbt_packages=True` 
- Changing individual `DbtDag` or `DbtTaskGroup` instances to use the new configuration `ProjectConfig.copy_dbt_packages=True` 
- Changing the behaviour globally, by setting the Airflow configuration either via the environment variable `AIRFLOW__COSMOS__DEFAULT_COPY_DBT_PACKAGES_VALUE=True` or via the `airflow.cfg`:
```
[cosmos]
default_copy_dbt_packages_value=True
```

# How this was tested

To validate the end-to-end behaviour, we run the following dag from `dev/dags`:
```
airflow dags test dbt_deps_example
```

# Related tickets

This is a follow-up to #1668 and #1669.

I'll make a follow-up PR covering the documentation.

Closes: #1630

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -667,7 +667,7 @@ class DbtGraph:
         dbt project paths and this is cheaper than copying those folders.
 
         The current settings make sense for Cosmos 1.x and allow users to set things in different ways, while being backwards
-        compatible. We should review thisf or Cosmos 2.x.
+        compatible. We should review this for Cosmos 2.x.
         """
 
         should_not_create_dbt_deps_symbolic_link = self.should_install_dbt_deps or self.project.copy_dbt_packages
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -39,7 +39,13 @@ from cosmos.constants import (
     SourceRenderingBehavior,
 )
 from cosmos.dbt.parser.project import LegacyDbtProject
-from cosmos.dbt.project import copy_dbt_packages, create_symlinks, environ, get_partial_parse_path, has_non_empty_dependencies_file
+from cosmos.dbt.project import (
+    copy_dbt_packages,
+    create_symlinks,
+    environ,
+    get_partial_parse_path,
+    has_non_empty_dependencies_file,
+)
 from cosmos.dbt.selector import select_nodes
 from cosmos.log import get_logger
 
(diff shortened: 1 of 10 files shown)


cosmos/dbt/graph.py
```diff
@@ -528,7 +528,7 @@ def load(
             LoadMode.CUSTOM: self.load_via_custom_parser,
             LoadMode.DBT_LS: self.load_via_dbt_ls,
             LoadMode.DBT_LS_FILE: self.load_via_dbt_ls_file,
-            LoadMode.DBT_LS_CACHE: self.load_via_dbt_ls_cache,
+            LoadMode.DBT_LS_CACHE: self.load_via_dbt_ls,
```
pankajkoti: I believe this was unintentional
```suggestion
            LoadMode.DBT_LS_CACHE: self.load_via_dbt_ls_cache,
```

pankajkoti (approved):

Issue #1634, opened 2025-03-24
Validate Cosmos `ExecutionMode.AIRFLOW_ASYNC` operators with Airflow 3


tatiana:
Similar to #1621

Issue #1633, opened 2025-03-24
Validate Cosmos `ExecutionMode.VIRTUALENV` operators with Airflow 3


tatiana:
Similar to #1621

pankajkoti:
PR #1674 added env variables to airflow 3 [env.sh](https://github.com/astronomer/astronomer-cosmos/blob/5ae181e02e7696d82c51fb4f628abbdc25e41c8e/scripts/airflow3/env.sh) files to not use setup task for the async DAG as it appears the ExecutionMode.VIRTUALENV is currently not compatible with Airflow 3. The two issues I observed are:
1. Import path for prepare_virtualenv needs to adjusted in the cosmos/operators/virtualenv.py module
```
try:  # Airflow 3
    from airflow.providers.standard.utils.python_virtualenv import prepare_virtualenv
except ImportError:  # Airflow 2
    from airflow.utils.python_virtualenv import prepare_virtualenv
```
2. When the task executes, it complained about `distutils` not being available. We need to investigate and workaround it.

PR #1674, opened 2025-04-16
Add Airflow 3 compatibility for BiqQuery deferrable support (ExecutionMode.AIRFLOW_ASYNC)

pankajkoti:
## Changes

- Remove SQLAlchemy session dependency and calls to DB ORM
- Fix base class inheritance for proper object initialisation wrt Airflow 3
- Add symlink in the airflow3 dags for the simple_dag_async.py example DAG
- Set environment variables in airflow3 env to not use setup task for the async DAG until we resolve the Virtualenv execution mode
- Add dbt-bigquery and Google provider to requirements.txt in Airflow 3 setup required for the example DAG

closes: #1634


cosmos/operators/_asynchronous/bigquery.py
```diff
@@ -110,6 +107,10 @@ def __init__(
         self.compiled_sql = ""
         self.gcp_project = ""
         self.dataset = ""
+        DbtRunAirflowAsyncBigqueryOperator.__bases__ = (
```
tatiana, line 110: It would be great if we added a comment on why we're redefining `__bases__` twice, potentially with some example of error message that lead to this solution

scripts/airflow3/env.sh
```diff
@@ -18,3 +18,8 @@ export AIRFLOW__CORE__DAGBAG_IMPORT_ERROR_TRACEBACK_DEPTH=10
 export AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT=300
 # export AIRFLOW__LOGGING__LOGGING_LEVEL=DEBUG
 export AIRFLOW_REPO_DIR="$PWD/../airflow"
+
+
+# Until we fix the ExecutionMode.VIRTUALENV
+export AIRFLOW__COSMOS__ENABLE_SETUP_ASYNC_TASK=False
```
tatiana, line 24: It is probably worth to add a comment on https://github.com/astronomer/astronomer-cosmos/issues/1633 about this

tatiana (approved):
Excellent work, @pankajkoti !

dev/dags/simple_dag_async.py
```diff
@@ -24,7 +24,7 @@
 simple_dag_async = DbtDag(
     # dbt/cosmos-specific parameters
     project_config=ProjectConfig(
-        DBT_ROOT_PATH / "original_jaffle_shop",
+        DBT_ROOT_PATH / "jaffle_shop",
```
pankajastro, line 27: Curious, why are we changing the project?

cosmos/operators/_asynchronous/bigquery.py
pankajkoti: yes, added comments to explain those. please check.

scripts/airflow3/env.sh
pankajkoti: added a comment there https://github.com/astronomer/astronomer-cosmos/issues/1633#issuecomment-2812428496

dev/dags/simple_dag_async.py
pankajkoti: I observed we have the same project across all our example DAGs, so altered it for consistency.

pankajastro: Look like `jaffle_shop` does not bigquery profile which may require here? https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/dbt/jaffle_shop/profiles.yml

pankajastro: ohh it uses profile mapping

pankajastro (approved):

dev/dags/simple_dag_async.py
pankajkoti, line 27:
```suggestion
        DBT_ROOT_PATH / "original_jaffle_shop",
```

pankajkoti: based on the discussions, I reverted the project to use `original_jaffle_shop`. We can work on it in a separate PR to get rid of original_jaffle_shop.

PR #1676, opened 2025-04-17
Remove original_jaffle_shop dbt project

pankajkoti:
PR #1230 introduced the `original_jaffle_shop` project for `ExecutionMode.AIRFLOW_ASYNC`, likely to support early customisations for async execution. Since Cosmos 1.9, we’ve shifted to relying on dbt to generate the SQL, making those customisations unnecessary. As this project isn’t used in any other example DAG, it now seems redundant. I’ve updated the test and example DAG to use `jaffle_shop` instead, which is consistently used across most of our examples.

This is a follow-up PR from the discussion in https://github.com/astronomer/astronomer-cosmos/pull/1674#discussion_r2048690393

related: https://github.com/astronomer/astronomer-cosmos/issues/1677


dev/dags/simple_dag_async.py
```diff
@@ -24,7 +24,7 @@
 simple_dag_async = DbtDag(
     # dbt/cosmos-specific parameters
     project_config=ProjectConfig(
-        DBT_ROOT_PATH / "original_jaffle_shop",
+        DBT_ROOT_PATH / "jaffle_shop",
```
pankajkoti, line 27: Apart from file deletions by removing the `dbt/original_jaffle_shop` folder in the PR, this is one of the two changes we're introducing in this PR. Adding this comment for helping reviewers.

tests/operators/test_airflow_async.py
```diff
@@ -28,7 +28,7 @@
 from cosmos.profiles import get_automatic_profile_mapping
 
 DBT_PROJECTS_ROOT_DIR = Path(__file__).parent.parent.parent / "dev/dags/dbt"
-DBT_PROJECT_NAME = "original_jaffle_shop"
+DBT_PROJECT_NAME = "jaffle_shop"
```
pankajkoti, line 31: Apart from file deletions by removing the `dbt/original_jaffle_shop` folder in the PR, this is second of the two changes we're introducing in this PR. Adding this comment for helping reviewers.

tatiana (approved):
Looks great, @pankajkoti , thanks for doing this cleaning up - and having a dedicated PR for this change.

PR #1687, opened 2025-04-22
Fix or ignore type check error

pankajastro:


cosmos/operators/virtualenv.py
```diff
@@ -34,7 +34,11 @@
 )
 
 if TYPE_CHECKING:
-    from airflow.utils.context import Context  # pragma: no cover
+    if TYPE_CHECKING:
+        try:
+            from airflow.sdk.definitions.context import Context
+        except ImportError:
+            from airflow.utils.context import Context  # type: ignore[attr-defined]
```
pankajastro, line 41:
```suggestion
    try:
        from airflow.sdk.definitions.context import Context
    except ImportError:
        from airflow.utils.context import Context  # type: ignore[attr-defined]
```

pankajkoti (commented):
Can we add pragma:no-cover blocks for the type check so as to not drop the project coverage?

pankajkoti (approved):
LGTM, thanks for fixing these

PR #1691, opened 2025-04-23
Try dropping Airflow dependency in mypy pre-commit

pankajkoti:


pankajkoti:
appears mypy needs apache-airflow in the isolated environment it creates as otherwise it reports the errors as seen in the CI failures. Likely the fixes being made in #1687 is what we would need due to the nature of how mypy works with its isolated environments.

Issue #1587, opened 2025-03-04
Evaluate how to support dbt docs in Airflow 3


tatiana:
Airflow 3 introduces several UI changes. This ticket aims to validate how the Cosmos dbt docs plugin behaves in Airflow 3, using similar strategies to https://github.com/astronomer/astronomer-cosmos/issues/1488.

Some of the things we expect to change:

1. We should not use Fab, only Flask. We should replace the following:
   - https://github.com/astronomer/astronomer-cosmos/blob/46912b6db088d096103e8b0a80c32ee77179203b/cosmos/plugin/__init__.py#L272
   - https://github.com/astronomer/astronomer-cosmos/blob/46912b6db088d096103e8b0a80c32ee77179203b/cosmos/plugin/__init__.py#L208-L212

2. As of today, Airflow does not support custom menu items. We should also look into this.

tatiana:
Very important: @pankajkoti Before we start working on this, we should align with @bbovenzi @pierrejeambrun to confirm what is ready and what is not.

We should probably remove the Fab dependency as a starting point, but we need to see how much we can expose. We may have to wait for AF 3.1

bbovenzi:
2. We do allow custom menu items as links! As long as you adjust the url in the plugin to point to the dbt docs then the AF3 UI should display it as an external link.

There is also a way to force plugins to load as iframes inside the "Security" tab.

Otherwise, we will be expanding on plugins in 3.1

bbovenzi:
Based on our conversation. The new external_views iframe plugins should work. We just need to make sure that the UI already handles managing url state between the iframe view and the main UI.

pankajkoti:
> Based on our conversation. The new external_views iframe plugins should work. We just need to make sure that the UI already handles managing url state between the iframe view and the main UI.

Thanks @bbovenzi and @pierrejeambrun for taking the time yesterday to discuss this over our call yesterday. In our call, I missed discussing another problem on the possibility of how we could support multiple dbt projects to be supported in the docs view. I am following up with you over Slack to discuss the same.

PR #1646, opened 2025-03-28
Add Airflow 3 tests matrix entry

pankajkoti:
This PR adds Airflow 3 matrix entries to the various tests that we run in our CI including unit, integration, dbt-async, performance and k8s tests.

It also makes tweaks to the setup scripts for airflow version specific db commands. Since, "airflow db init" no longer works with Airflow 3, the PR replaces its usage with airflow db migrate along with needed refactoring to those scripts.

While the PR attempts to fix few of the failing tests to add compatibility for Airflow 3, I am  skipping remaining failing tests to be run on Airflow 3. I have created independent issues to resolve each of them and have also added TODO comments linking to those issues in the TODO comment. 

related: #1631

PR #1692, opened 2025-04-24
Disable CosmosPlugin for AF3

pankajkoti:
While I am not seeing any explicit errors with having the CosmosPlugin in an Airflow 3 env, I think it's still wise that we disable it for now for Airflow 3 until we work on #1587 

closes: #1685


cosmos/plugin/__init__.py
```diff
@@ -1,273 +1,11 @@
-import os.path as op
```
pankajkoti, line 1: all of the existing content of this file has been copied over as-is to the new `plugin_impl.py` file in the same directory path

pankajkoti:
I tested that the plugin continues to work with AF2
Screenshot 2025-04-24 at 8 37 43 PM

pankajkoti:
This could help in the failing test for AF3 in https://github.com/astronomer/astronomer-cosmos/actions/runs/14662043590/job/41148487115?pr=1646 for PR #1646

cosmos/plugin/__init__.py
tatiana: Do you know why Git didn't consider this a rename, but instead treated it as a new file?

Perhaps we could rename this just to `plugin/plugin.py` - its a bit ungly, but I feel impl is not adding much. Happy with alternatives as well :)

tatiana (approved):
Thanks, @pankaj, really excited for the AF3 support!

PR #1693, opened 2025-04-24
Allow multiple callbacks

corsettigyg:
## Description

During some experimentation with [callbacks](https://astronomer.github.io/astronomer-cosmos/configuration/callbacks.html#example-using-callbacks-with-a-single-operator) I have noticed that we can only use a single callback per time, which can be very limiting if I want to add custom callbacks together with the default ones provided by cosmos.

This PR is very simple and aims to allow multiple callbacks to be passed as a list of functions, that will then be executed sequentially. 

## Breaking Change?

Should not be

## Checklist

### DIFF
--- a/tests/operators/test_local.py
+++ b/tests/operators/test_local.py
@@ -1562,3 +1562,23 @@ def test_test_clone_project(create_symlinks_mock, copy_dbt_packages_mock, caplog
     assert f"Cloning project to writable temp directory {tmp_dir_path} from {project_dir}" in caplog.text
     assert "Copying dbt packages to temporary folder." in caplog.text
     assert "Completed copying dbt packages to temporary folder." in caplog.text
+
+@patch("cosmos.operators.local.AbstractDbtLocalBase.store_freshness_json")
+@patch("cosmos.operators.local.AbstractDbtLocalBase.store_compiled_sql")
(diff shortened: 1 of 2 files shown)


corsettigyg:
@pankajkoti tried adding the tests

pankajkoti (approved):
LGTM. Thanks for adding the test @corsettigyg . I will wait for a while before merging the PR in case some else would like to take a review of the PR too. We're aiming to release Cosmos 1.10 sometime next week and would include this change in the release

tests/operators/test_local.py
```diff
@@ -1562,3 +1562,26 @@ def test_test_clone_project(create_symlinks_mock, copy_dbt_packages_mock, caplog
...
+        task_id="my-task",
+        project_dir="my/dir",
+        callback=multiple_callbacks,
+        callback_args={"arg1": "value1"},
+    )
+
+    context = {"dag_run": MagicMock(), "task": MagicMock()}
+    operator._handle_post_execution("/tmp/project_dir", context)
+
+    for callback_fn in multiple_callbacks:
```
tatiana, line 1586: In future, it may be worth doing  individual asserts without a for in the test.

Usually, the more logic (if, for and while) we add to a test, the less reliable it may be - and we may need to create tests for tests

PR #1695, opened 2025-04-24
add MonteCarlo callback example in documentation for importing dbt artifacts

corsettigyg:
## Description

Added an extra example on how to import dbt artifacts using montecarlo integration with dbt-core as a callback. The implementation is quite challenging considering the lack of documentation provided by montecarlo itself for the python SDK and after struggling to implement something functional, I believe it is worth having it in the cosmos documentation since the line above the code-block  mentions montecarlo, suggesting that many users would be interested in that.

## Breaking Change?

No

## Checklist

### DIFF
--- a/docs/configuration/callbacks.rst
+++ b/docs/configuration/callbacks.rst
@@ -103,27 +103,6 @@ Below, find an example of a callback method that raises an exception if the quer
 Users can use the same approach to call the data observability platform `montecarlo <https://docs.getmontecarlo.com/docs/dbt-core>`_ or other services.
 
 .. code-block:: python
-
-    from pycarlo.core import Client, Session, Query
-    from pycarlo.features.dbt.dbt_importer import DbtImporter
-    from cosmos.log import get_logger
-    from cosmos.exceptions import CosmosValueError
-
-    logger = get_logger(__name__)
-
-    def get_resource_id(client):
-        """Get the resource ID of the first warehouse connected to the user's account"""
-        query = Query()
-        query.get_user().account.warehouses.__fields__("name", "connection_type", "uuid")
-        warehouses = client(query).get_user.account.warehouses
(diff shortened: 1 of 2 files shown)


docs/configuration/callbacks.rst
```diff
@@ -102,6 +102,76 @@ Below, find an example of a callback method that raises an exception if the quer
...
 Users can use the same approach to call the data observability platform `montecarlo <https://docs.getmontecarlo.com/docs/dbt-core>`_ or other services.
 
+.. code-block:: python
+    def montecarlo_import_artifacts(
+        project_dir: str,
+        mcd_id: str,
+        mcd_token: str,
+        job_name: str,
+        project_name: str = "default-project",
+        resource_id: str = None,
```
pankajkoti, line 112: can we also add an example of how to specify and pass these callback args in the DAG when using this callback function. I think that would be a great help.

We tried to do something similar previously in https://github.com/astronomer/astronomer-cosmos/blob/fb6246d9797eb4f68079851661cbd9e8c04ab8ff/dev/dags/cosmos_callback_dag.py#L50

I was also thinking if we could add this example in the example DAG, but I guess we don't have a Montecarlo subscription and hence the token credentials. So would be nice to leave an example here on how to use it in their DAG. WDYT?

pankajkoti (approved):
This looks like a great addition, thanks for adding this. 

Just have one suggestion in-line if you could please take a look.

docs/configuration/callbacks.rst
corsettigyg: I would say that the documentation makes it pretty explicit on how to pass the callback args using "callback_args", so I decided to not include it. the args have description in the function itself to make it straightforward. I can add it for sure, but my idea was to keep it as simple as the `error_if_slow` function that is defined above it

PR #1697, opened 2025-04-25
Make wording and command consistent in the contributing doc

pankajkoti:
I observed the following discrepancies between the wording and the command in our contributing doc:
1. In the example wording we say for Python 3.10, but the command actually used 3.8
2. In the example wording we say for Airflow 2.5, but the command actually used 2.4

This PR tries to make them consistent and while we're at it, making it consistent to use Python 3.9 and Airflow 2.10.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -91,11 +91,11 @@ We currently use `hatch <https://github.com/pypa/hatch>`_ for building and distr
 
 The tool can also be used for local development. The `pyproject.toml <https://github.com/astronomer/astronomer-cosmos/blob/main/pyproject.toml>`_ file currently defines a matrix of supported versions of Python, Airflow and dbt-core for which a user can run the tests against.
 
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -91,11 +91,11 @@ We currently use `hatch <https://github.com/pypa/hatch>`_ for building and distr
 
 The tool can also be used for local development. The `pyproject.toml <https://github.com/astronomer/astronomer-cosmos/blob/main/pyproject.toml>`_ file currently defines a matrix of supported versions of Python, Airflow and dbt-core for which a user can run the tests against.
 
-For instance, to run the tests using Python 3.10, `Apache Airflow® <https://airflow.apache.org/>`_ 2.5 and `dbt-core <https://github.com/dbt-labs/dbt-core/>`_ 1.9, use the following:
```
pankajkoti, line 94: we said Python 3.10 and Airflow 2.5 here previously and in the command actually used Python 3.8 and Airflow 2.4

pankajastro (approved):

tatiana (approved):
Thanks, @pankajkoti !

PR #1698, opened 2025-04-25
Set CosmosPlugin to None for airflow 3

pankajkoti:
Interestingly, while working on PR #1692, I didn’t encounter the following error initially. However, something in my environment must have changed—possibly related to the Airflow 3 setup—because I’m now seeing a plugin load error in the `airflow standalone` logs when using a Hatch-managed Airflow 3 environment:

```
dag-processor | [2025-04-25T18:51:24.537+0530] {plugins_manager.py:262} ERROR - Failed to import plugin cosmos
dag-processor | Traceback (most recent call last):
dag-processor | File "/Users/pankajkoti/Library/Application Support/hatch/env/virtual/astronomer-cosmos/D9FI7Men/tests.py3.9-3.0-1.9/lib/python3.9/site-packages/airflow/plugins_manager.py", line 254, in load_entrypoint_plugins
dag-processor | plugin_class = entry_point.load()
dag-processor | File "/Users/pankajkoti/Library/Application Support/hatch/env/virtual/astronomer-cosmos/D9FI7Men/tests.py3.9-3.0-1.9/lib/python3.9/site-packages/importlib_metadata/__init__.py", line 213, in load
dag-processor | return functools.reduce(getattr, attrs, module)
dag-processor | AttributeError: module 'cosmos.plugin' has no attribute 'CosmosPlugin'
```

While this doesn’t block DAG execution or general Airflow usage, I believe it’s still worthwhile to iterate on PR #1692 and explicitly set the plugin to `None` for Airflow 3.x. This avoids the plugin load error altogether.

With that change in place, the error is resolved, and the environment works seamlessly with both Airflow 2 and Airflow 3. I’ve also added a basic sanity check to ensure that CosmosPlugin is either None or properly initialized, depending on the Airflow version in use.

related: #1685

### DIFF
--- a/cosmos/plugin/__init__.py
+++ b/cosmos/plugin/__init__.py
@@ -1,11 +1,19 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
 from airflow import __version__ as airflow_version
 from packaging import version
 
 from cosmos.constants import _AIRFLOW3_MAJOR_VERSION
 
+if TYPE_CHECKING:
+    from .plugin_impl import CosmosPlugin as _CosmosPluginType
+
+CosmosPlugin: _CosmosPluginType | None = None
 # The plugin is only loaded if the Airflow version is less than 3.0. This is because the plugin is incompatible with
 # Airflow 3.0 and above. Once the compatibility issue is resolved as part of
 # https://github.com/astronomer/astronomer-cosmos/issues/1587, the import statement can be moved outside of the
 # conditional block.
 if version.parse(airflow_version).major < _AIRFLOW3_MAJOR_VERSION:
-    from .plugin_impl import CosmosPlugin as CosmosPlugin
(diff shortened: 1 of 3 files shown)


cosmos/plugin/__init__.py
```diff
@@ -1,11 +1,19 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
 from airflow import __version__ as airflow_version
 from packaging import version
 
 from cosmos.constants import _AIRFLOW3_MAJOR_VERSION
 
+if TYPE_CHECKING:
```
tatiana, line 10: Should we add a pragma no cover here?

tatiana (approved):

cosmos/plugin/__init__.py
pankajkoti: added so, thanks.

Issue #1710, opened 2025-04-28
Performance test failing AttributeError: 'DbtDag' object has no attribute 'task_count' AF 3 CI


pankajkoti:
Action Link: https://github.com/astronomer/astronomer-cosmos/actions/runs/14703034847/job/41256457936?pr=1646#step:7:439

Traceback trimmed:
```
=========================== short test summary info ============================
FAILED tests/perf/test_performance.py::test_perf_dag - AttributeError: 'DbtDag' object has no attribute 'task_count'
```

related: #1646 
related: #1631

PR #1714, opened 2025-04-28
Use a different way to get tasks count for asserting test_perf_dag

pankajkoti:
The PR makes a change in the assert of test_perf_dag to use `len(dag.tasks)` which works with both AF2 and AF3 for getting the task count in the DAG. We observed that task_count attribute is unavailable on DAG object in AF3 like it's available in AF3.

closes: #1710

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,8 +3,8 @@ name: test
 on:
   push: # Run on pushes to the default branch
     branches: [main, add-airflow-3-test-matrix-entry]
-   pull_request_target: # Also run on pull requests originated from forks
-     branches: [main]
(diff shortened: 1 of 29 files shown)


pankajkoti:
TODO: Rebase after #1646 is merged.

tatiana (approved):
Looks great, thanks, @pankajkoti !

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, issue-1710-enable-af3-perf-dag]
```
```suggestion
    branches: [main]
```
Remove before merging

Issue #1708, opened 2025-04-28
DBT Async test job failing in CI while setting up AF 3


pankajkoti:
Action Link: https://github.com/astronomer/astronomer-cosmos/actions/runs/14703034847/job/41256451634?pr=1646#step:7:295

Traceback:
```
AIRFLOW_VERSION=$(airflow version)
+ airflow version
Traceback (most recent call last):
  File "/home/runner/.local/share/hatch/env/virtual/astronomer-cosmos/Za_bFbg4/tests.py3.11-3.0-1.6/bin/airflow", line 4, in <module>
    from airflow.__main__ import main
  File "/home/runner/.local/share/hatch/env/virtual/astronomer-cosmos/Za_bFbg4/tests.py3.11-3.0-1.6/lib/python3.11/site-packages/airflow/__main__.py", line 37, in <module>
    from airflow.cli import cli_parser
  File "/home/runner/.local/share/hatch/env/virtual/astronomer-cosmos/Za_bFbg4/tests.py3.11-3.0-1.6/lib/python3.11/site-packages/airflow/cli/cli_parser.py", line 39, in <module>
    from airflow.api_fastapi.app import get_auth_manager_cls
  File "/home/runner/.local/share/hatch/env/virtual/astronomer-cosmos/Za_bFbg4/tests.py3.11-3.0-1.6/lib/python3.11/site-packages/airflow/api_fastapi/app.py", line 35, in <module>
    from airflow.api_fastapi.execution_api.app import create_task_execution_api_app
  File "/home/runner/.local/share/hatch/env/virtual/astronomer-cosmos/Za_bFbg4/tests.py3.11-3.0-1.6/lib/python3.11/site-packages/airflow/api_fastapi/execution_api/app.py", line 27, in <module>
    from cadwyn import (
  File "/home/runner/.local/share/hatch/env/virtual/astronomer-cosmos/Za_bFbg4/tests.py3.11-3.0-1.6/lib/python3.11/site-packages/cadwyn/__init__.py", line 3, in <module>
    from .applications import Cadwyn
  File "/home/runner/.local/share/hatch/env/virtual/astronomer-cosmos/Za_bFbg4/tests.py3.11-3.0-1.6/lib/python3.11/site-packages/cadwyn/applications.py", line 29, in <module>
    from cadwyn._utils import DATACLASS_SLOTS, same_definition_as_in
[...]

PR #1715, opened 2025-04-28
Reinstall airflow dependency on pydantic>=2.11 for dbt adapater versions 1.6 & 1.9

pankajkoti:
`apache-airflow-core 3.0.0` requires `pydantic>=2.11.0`, but the installed dbt adapters in case of version 1.6 and 1.9 install
`pydantic 1.10.22` which make it incompatible.
With `pydantic 1.10.22` we get the below error
```
File "<...>/hatch/env/virtual/astronomer-cosmos/D9FI7Men/tests.py3.11-3.0-1.6/lib/python3.11/site-packages/cadwyn/_utils.py", line 5, in <module>
     from pydantic._internal._decorators import unwrap_wrapped_function
ModuleNotFoundError: No module named 'pydantic._internal'
```
Hence, we re-install pydantic with the required minimum version after installing dbt adapters in case of 1.6 & 1.9 to be compatible with Airflow 3 dependencies.

closes: #1708


scripts/test/integration-dbt-async.sh
```diff
@@ -4,13 +4,29 @@ set -v
 set -x
 set -e
 
+echo "Current shell: $SHELL"
+ps -p $$
+
```
pankajkoti, line 9:
```suggestion
```

pyproject.toml
```diff
@@ -181,7 +181,7 @@ test-integration = 'sh scripts/test/integration.sh'
 test-kubernetes = "sh scripts/test/integration-kubernetes.sh"
 test-kubernetes-setup = "sh scripts/test/kubernetes-setup.sh"
 test-integration-dbt-1-5-4 = 'sh scripts/test/integration-dbt-1-5-4.sh'
-test-integration-dbt-async = 'sh scripts/test/integration-dbt-async.sh  {matrix:dbt}'
+test-integration-dbt-async = 'bash scripts/test/integration-dbt-async.sh  {matrix:dbt}'
```
pankajkoti, line 184: `sh` does not support the conditional logic with syntax `[[` and `]]` used to test dbt version 1.6 & 1.9  and hence using bash which supports that.

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,enable-ci-dbt-async-1.6-1.9-af3]
```
```suggestion
    branches: [main]
```
Remove before merging

tatiana (approved):
Looks great, thanks, @pankajkoti !
Dependecies resolution can be a nightmare

Issue #1707, opened 2025-04-28
FAILED tests/operators/test_virtualenv.py::test_integration_virtualenv_operator AF 3 CI


pankajkoti:
Action Link: https://github.com/astronomer/astronomer-cosmos/actions/runs/14703034847/job/41256451294?pr=1646#step:7:4029

Not attaching traceback as it's too long. Please refer to the above GitHub action link

Trimmed traceback:
```
)\nFileNotFoundError: [Errno 2] No such file or directory: \'/tmp/persistent-venv2\'\nINFO     airflow.models.dagrun.DagRun:dagrun.py:1145 Marking run <DagRun example_virtualenv_mini @ None: manual__2025-04-28T08:00:05.465435+00:00_J68AclrN, state:running, queued_at: None. run_type: manual> failed\nINFO     airflow.models.dagrun.DagRun:dagrun.py:1224 DagRun Finished: dag_id=example_virtualenv_mini, logical_date=None, run_id=manual__2025-04-28T08:00:05.465435+00:00_J68AclrN, run_start_date=2025-04-28 08:00:05.465435+00:00, run_end_date=2025-04-28 08:00:05.727756+00:00, run_duration=0.262321, state=failed, run_type=manual, data_interval_start=None, data_interval_end=None,\n' = <_pytest.logging.LogCaptureFixture object at 0x7f122e3cfac0>.text
```

related: #1646 
related: #1631

PR #1718, opened 2025-04-28
Fix `test_integration_virtualenv_operator`

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1707

For AF3, it appears that the render_template is attempting to resolve the virtualenv_dir templated field path, which leads to an error during execution.

```
[...]

PR #1721, opened 2025-04-29
Change type of virtualenv_dir argument from path to string

pankajastro:
closes: #1707

### DIFF
--- a/cosmos/operators/virtualenv.py
+++ b/cosmos/operators/virtualenv.py
@@ -76,14 +76,16 @@ class DbtVirtualenvBaseOperator(DbtLocalBaseOperator):
         py_requirements: list[str] | None = None,
         pip_install_options: list[str] | None = None,
         py_system_site_packages: bool = False,
-        virtualenv_dir: Path | None = None,
+        virtualenv_dir: str | None = None,
         is_virtualenv_dir_temporary: bool = False,
         **kwargs: Any,
     ) -> None:
         self.py_requirements = py_requirements or []
         self.pip_install_options = pip_install_options or []
         self.py_system_site_packages = py_system_site_packages
-        self.virtualenv_dir = virtualenv_dir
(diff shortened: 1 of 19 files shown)


cosmos/operators/virtualenv.py
```diff
@@ -76,14 +76,16 @@ def __init__(
         py_requirements: list[str] | None = None,
         pip_install_options: list[str] | None = None,
         py_system_site_packages: bool = False,
-        virtualenv_dir: Path | None = None,
+        virtualenv_dir: str | None = None,
```
tatiana, line 79: @pankajastro this could be seen as a breaking change - what motivated it?

pankajkoti: I pushed that change @tatiana in https://github.com/astronomer/astronomer-cosmos/pull/1721/commits/1f1947c8c2fe3b3eb52a9c23aa04ae4c8dbd7572. We had a discussion regarding this in the other PR at https://github.com/astronomer/astronomer-cosmos/pull/1718#discussion_r2065555001.

Apparently Airflow 3 is evaluating & validating that the `Path` exists as part of its rending as mentioned by Pankaj in the other PR's description.

tatiana: @pankajkoti I just read the thread - would it be possible for this to be a Path and for us to convert to a string if Airflow 3 only, as part of the operator implementation?

pankajkoti: we need to try out though, when does the rendering evaluate and if we have control over it(meaning if the rendering happens even before the operator initializes/executes and if yes, if there's a method to override the type for it).

pankajkoti: will try it out later today to dig into this.

pankajkoti:
After rebasing, this PR shows a lot of files changed although the changes are already in main and would make it difficult to review. I created a fresh PR to reflect only the changes being made #1724 

Hence, I am closing this PR.

PR #1724, opened 2025-04-29
Ensure virtualenv_dir path exists

pankajkoti:
Airflow3 pre-emptively tries to validate rendered fields and also validates that the path exists in case of Path like objects specified in rendered fields. Hence, we ensure that the path specified is created first.

closes: #1707 

Co-authored-by: Pankaj Singh <98807258+pankajastro@users.noreply.github.com>


tatiana (approved):
Thanks for fixing this @pankajkoti and @pankajastro !

Issue #1706, opened 2025-04-28
Example DAGs failing for AttributeError: 'DAG' object has no attribute 'test' AF3 CI


pankajkoti:
Action Link: https://github.com/astronomer/astronomer-cosmos/actions/runs/14703034847/job/41256451294?pr=1646#step:7:4031

Failing tests
```
FAILED tests/test_example_dags.py::test_example_dag[basic_cosmos_task_group_different_owners] - AttributeError: 'DAG' object has no attribute 'test'
FAILED tests/test_example_dags.py::test_example_dag[cosmos_profile_mapping] - AttributeError: 'DAG' object has no attribute 'test'
FAILED tests/test_example_dags.py::test_example_dag[user_defined_profile] - AttributeError: 'DAG' object has no attribute 'test'
FAILED tests/test_example_dags.py::test_example_dag[example_taskflow] - AttributeError: 'DAG' object has no attribute 'test'
FAILED tests/test_example_dags.py::test_example_dag[cosmos_manifest_example] - AttributeError: 'DAG' object has no attribute 'test'
FAILED tests/test_example_dags.py::test_example_dag[example_cosmos_cleanup_dag] - AttributeError: 'DAG' object has no attribute 'test'
FAILED tests/test_example_dags.py::test_example_dag[basic_cosmos_task_group] - AttributeError: 'DAG' object has no attribute 'test'
```

related: #1646 
related: #1631

PR #1726, opened 2025-04-29
Enable example DAGs to run in CI that were disabled in PR #1646

pankajkoti:
Airflow 3.0 does not currently support calling `.test()` on DAGs defined using the `@dag` decorator, as the decorator now returns an object that does not expose the `.test()` method. There is an open issue https://github.com/apache/airflow/issues/45549 around this area. This PR updates the affected example DAGs to explicitly use `airflow.models.DAG` instead of the `@dag` decorator, restoring compatibility with `dag.test()` for local testing and CI.

This also re-enables the example DAGs that were disabled in PR #1646.

closes: #1706

### DIFF
--- a/dev/dags/basic_cosmos_task_group.py
+++ b/dev/dags/basic_cosmos_task_group.py
@@ -6,7 +6,7 @@ import os
 from datetime import datetime
 from pathlib import Path
 
-from airflow.decorators import dag
+from airflow import DAG
 from airflow.operators.empty import EmptyOperator
 
 from cosmos import DbtTaskGroup, ExecutionConfig, ProfileConfig, ProjectConfig, RenderConfig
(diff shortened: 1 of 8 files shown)


tatiana (approved):

> Airflow 3.0 does not currently support calling .test() on DAGs defined using the @dag decorator
Is this a design decision or a bug? It would be great to add a link the GH issue if it is a bug.

pankajkoti:
> > Airflow 3.0 does not currently support calling .test() on DAGs defined using the @dag decorator
> 
> Is this a design decision or a bug? It would be great to add a link the GH issue if it is a bug.

Yes, @tatiana definitely not an intentional decision but I guess it's an effect of resources getting migrated to TaskSDK. The team seems to be aware of it that it does not currently work (I had reported a similar issue previously in Astronomer Slack) and also has a ticket for this: https://github.com/apache/airflow/issues/45549

@kaxil also [pointed in Astronomer Slack](https://astronomer.slack.com/archives/C07HG5KUJCS/p1745941219492719?thread_ts=1745941009.675399&cid=C07HG5KUJCS) out one more way to do it currently while dag.test support gets ready 

```
from airflow.models.dag import DAG

dag = .... # dag from SDK

DAG.from_sdk_dag(dag).test()
```

Since we need to modify the DAGs anyway, I think it’s fine to keep the current PR as is and keep this insight in mind for next time.

I will merge this PR now so we could rebase it and run more example DAGs in our ongoing PRs, but please leave a comment if you've additional feedback and I can address it the next iteration.

PR #1735, opened 2025-05-02
Re-enable integration tests durations to troubleshoot performance degradation

tatiana:
There is an apparent degradation in the speed when running our integration tests in AF3 compared to AF2.

We're re-enabling these metrics so that we can analyse them.

As an example, during the run https://github.com/astronomer/astronomer-cosmos/actions/runs/14774756080:
- Run-Integration-Tests(3.9,2.10,1.9) took 12m 40s
- Run-Integration-Tests(3.9,3.0,1.9) took 41m 6s


pankajkoti (approved):
Thanks for adding this!

PR #1741, opened 2025-05-04
Refactor `AIRFLOW_ASYNC` so that the path in the remote object store is specific per DAG run

### LINKED ISSUE #1613: (Bug) Refactor `AIRFLOW_ASYNC` so that the path in the remote object store is  specific per DAG run
As of Cosmos 1.9.0 and 1.9.1, the remote object store path is based on the DAG and TaskGroup names but not based on DagRun. This means that if there are two DAG runs of the same DAG, and if the teardown runs for the first DAG run, it will affect the tasks attempting to be run by the second DAG Run.

If we make the path in the remote object store specific per DAG Run, we can avoid this error.

This problem only happens when we use the setup with the async run.

tuantran0910:
## Description

Refactor `AIRFLOW_ASYNC` so that the path in the remote object store is specific per DAG run. The format of remote model path will be:

```python
# test_cosmos/simple_dag_async/run/jaffle_shop/models/example/my_first_dbt_model.sql
remote_model_path = f"{remote_target_path_str}/{dbt_dag_task_group_identifier}/{run_id}/run/{relative_file_path}"
```

## Related Issue(s)

Closes #1613


pankajastro (commented):
Thanks a lot for the contribution, @tuantran0910! I've triggered the CI for the integration tests.

pankajastro:
Currently, the teardown task deletes SQL files through multiple calls, and the goal was to reduce the risk of deleting unexpected files when the task runs in parallel. Now that we're uploading based on the run-id, would it be worthwhile to delete the entire run-id folder directly, rather than iterating over each file? What are your thoughts?

https://github.com/astronomer/astronomer-cosmos/blob/9a334a8fbaddb518960fb4444aa7ee93625d340e/cosmos/operators/local.py#L338

This may also apply to the setup task.

tuantran0910:
> Currently, the teardown task deletes SQL files through multiple calls, and the goal was to reduce the risk of deleting unexpected files when the task runs in parallel. Now that we're uploading based on the run-id, would it be worthwhile to delete the entire run-id folder directly, rather than iterating over each file? What are your thoughts?
> 
> https://github.com/astronomer/astronomer-cosmos/blob/9a334a8fbaddb518960fb4444aa7ee93625d340e/cosmos/operators/local.py#L338
> 
> This may also apply to the setup task.

Hmm, I think it's a good idea to delete the entire folder `run-id` for optimizing the workflow. I will implement this later.

tuantran0910:
Hi @pankajastro, can you checkout the logic to delete run-id specific directory at my new commit https://github.com/astronomer/astronomer-cosmos/pull/1741/commits/2bf72583ab9cf5e65b75b36aec2eafc4188e5545. Thanks a lot :D

cosmos/operators/local.py
```diff
@@ -335,17 +336,28 @@ def _upload_sql_files(self, tmp_project_dir: str, resource_type: str) -> None:
...
+            run_id = self.extra_context["run_id"]
+            run_dir_path_str = f"{str(dest_target_dir).rstrip('/')}/{dag_task_group_identifier}/{run_id}"
+            run_dir_path = ObjectStoragePath(run_dir_path_str, conn_id=dest_conn_id)
+
+            if run_dir_path.exists():
+                run_dir_path.rmdir(recursive=True)
+                self.log.info("Deleted remote run directory: %s", run_dir_path_str)
+            else:
+                self.log.debug("Remote run directory does not exist, skipping deletion: %s", run_dir_path_str)
+        except Exception as e:
```
tatiana, line 359: Did you experience any particular exception? It would be great to capture the specific exception. This line captures all exceptions - including the ones that we cannot or perhaps should not recover from. This can make debugging tricky, so we tend to avoid it in the Cosmos codebase.

In this particular case, if the teardown task can't delete the files, should we just log the error and let the Airflow task pass? It may be better to let the exception be, fail the task, and let the user address any potential issues on their end.

Users who prefer to not have an tear down can set the Cosmos configuration [enable_teardown_async_task](https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html#enable-teardown-async-task).

tests/test_async_remote_target_dir_specific_run_id.py
```diff
@@ -0,0 +1,115 @@
+from pathlib import Path
```
tatiana, line 1: Since this test validates the method `_construct_dest_file_path` of the `AbstractDbtLocalBase` class, that is available in `cosmos.operators.local` module, please,  move these tests to:
https://github.com/astronomer/astronomer-cosmos/blob/main/tests/operators/test_local.py

To simplify the code/test maintenance, we have a test module per code module.

tatiana (commented):
This looks great, @tuantran0910 ! Thank you very much for improving this part of Cosmos and making it more reliable.

I've given some feedback in-line. Also, it seems there is an integration test failing, could you take a look please.

cosmos/operators/local.py
tuantran0910: Oh, that’s my mistake. We actually don’t need this try/except block, since we always ensure the remote target path exists. If it doesn't, it should be created during the setup async task. So, I’ll go ahead and remove the try/except statement.

Can you check via my new commit https://github.com/astronomer/astronomer-cosmos/pull/1741/commits/377de4f83065ad937b12bca354235065d2175d68 ? :D

tests/test_async_remote_target_dir_specific_run_id.py
tuantran0910: I have moved them into `tests/operators/test_local.py`. 

Can you check my new commits:
- https://github.com/astronomer/astronomer-cosmos/pull/1741/commits/377de4f83065ad937b12bca354235065d2175d68
- https://github.com/astronomer/astronomer-cosmos/pull/1741/commits/3dae779691d650f799f2d2fac34720bace698673

tuantran0910:
I am going to fix the tests later :D

tatiana:
@tuantran0910 thanks a lot! We're aiming to release 1.10.1 on Friday - if you get the chance to fix the tests before, we'll release as part of this milestone.

tuantran0910:
> @tuantran0910 thanks a lot! We're aiming to release 1.10.1 on Friday - if you get the chance to fix the tests before, we'll release as part of this milestone.

Hi @tatiana, I have fixed it in my newest commit. Can you try running the integration tests to see whether it has been solved ?

tuantran0910:
Hi @tatiana, @pankajastro, I think this PR is ready to review again and hopefully merging. Can you check it again :D ? Thanks!

tuantran0910:
Hi @tatiana, can you enable the integration tests :D ? I will rebase the main branch after making sure that all the test has passed :D

tuantran0910:
Hi @tatiana, I think this PR is ready to merge. Can you check it the last time :D Thank you a lot

cosmos/operators/local.py
```diff
@@ -336,17 +337,25 @@ def _upload_sql_files(self, tmp_project_dir: str, resource_type: str) -> None:
         elapsed_time = time.time() - start_time
         self.log.info("SQL files upload completed in %.2f seconds.", elapsed_time)
 
-    def _delete_sql_files(self, tmp_project_dir: Path, resource_type: str) -> None:
+    def _delete_sql_files(self) -> None:
+        """Deletes the entire run-specific directory from the remote target."""
         dest_target_dir, dest_conn_id = self._configure_remote_target_path()
-        source_run_dir = Path(tmp_project_dir) / f"target/{resource_type}"
-        files = [str(file) for file in source_run_dir.rglob("*") if file.is_file()]
+        if not dest_target_dir or not dest_conn_id:
```
tatiana, line 343: @tuantran0910 It seems we didn't write a test that covers this use case. Could you please add it?

tuantran0910:
> @tuantran0910 Thank you for all the patience on getting this to work. There is a minor request regarding code coverage - if you could address this as part of this PR, it would be great. If you can't do this before the release of 1.10.1, please, could you do a follow-up PR?

Hey @tatiana, I will try to push a quick commit to fix the coverage within an hour. Thank you for pointing that.

tuantran0910:
Hi @tatiana, I have added the test in https://github.com/astronomer/astronomer-cosmos/pull/1741/commits/fe92ef0d86d5e0f50a10d485df4ce259a939c700. Thank you

pankajkoti (approved):
Thanks for improvising this @tuantran0910 . This is certainly a very valuable contribution 👏🏽 

I tested this with an example async DAG running in Airflow UI and it works smoothly as intended for. We will include this fix in the upcoming release 1.10.1.

PR #1749, opened 2025-05-07
Fix dbt ls invocation method log message

tatiana:
We were incorrectly logging that we were using dbtRunner even if we were using subprocess because we had duplicated logic to decide what was being used to log and to execute the actual code.

Co-authored-by: Daniel Standish <15932138+dstandish@users.noreply.github.com>

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -261,12 +261,13 @@ def run_command(
 ) -> str:
     """Run a command either with dbtRunner or Python subprocess, returning the stdout."""
 
-    runner = "dbt Runner" if dbt_runner.is_available() else "Python subprocess"
+    use_dbt_runner = invocation_mode == InvocationMode.DBT_RUNNER and dbt_runner.is_available()
+    runner = "dbt Runner" if use_dbt_runner else "Python subprocess"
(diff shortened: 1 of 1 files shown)


pankajastro (approved):

cosmos/dbt/graph.py
```diff
@@ -261,12 +261,13 @@ def run_command(
 ) -> str:
     """Run a command either with dbtRunner or Python subprocess, returning the stdout."""
 
-    runner = "dbt Runner" if dbt_runner.is_available() else "Python subprocess"
+    use_dbt_runner = invocation_mode == InvocationMode.DBT_RUNNER and dbt_runner.is_available()
```
pankajkoti, line 264: looks like the variable assignment to `invocation_mode` maybe now redundant here? don't see it getting used in the method

tatiana: @pankajkoti so sorry, I missed your comment before merging. `invocation_mode` is an argument in the scope of this function. How do you think we could improve this?

pankajkoti: sorry I had misread the expression. We are not actually assigning to invocation_mode but using it for evaluation. Apologies, please ignore 🙇🏽

PR #1750, opened 2025-05-07
Default RenderConfig’s InvocationMode to SUBPROCESS to avoid task hangs observed with dbt Runner on Airflow 2

pankajkoti:
We have received user reports of tasks hanging after upgrading to Cosmos 1.9.0. While the issue is not observed on Airflow 3, I was able to reproduce it on Airflow 2. Through investigation, it seems the issue stems from the use of dbt Runner during DAG processing, which was introduced citing optimisations described in [PR #1484](https://github.com/astronomer/astronomer-cosmos/pull/1484). Switching to a subprocess-based invocation resolves the hang and allowed tasks to progress normally.

Changes in this PR:
 - Default `RenderConfig.invocation_mode` to `InvocationMode.SUBPROCESS`.
 - Temporarily disable related test cases that are only relevant when using `DBT_RUNNER`. These tests currently fail due to the adjusted default and will be re-enabled once the issue is resolved.

Reference issues:
- [Cosmos Issue #1751](https://github.com/astronomer/astronomer-cosmos/issues/1751) — Track investigation and resolution of the dbt Runner hang on Airflow 2.

Next steps:
Once we identify and resolve the root cause of the hanging behaviour in Airflow 2 task execution with DBT_RUNNER for dag processing, we can safely revert the default back and re-enable the corresponding tests.

related: #1751

### DIFF
--- a/cosmos/__init__.py
+++ b/cosmos/__init__.py
@@ -6,7 +6,7 @@ Astronomer Cosmos is a library for rendering dbt workflows in Airflow.
 Contains dags, task groups, and operators.
 """
 
-__version__ = "1.10.0"
+__version__ = "1.10.1a1"
 
 
 from cosmos.airflow.dag import DbtDag
--- a/cosmos/__init__.py
+++ b/cosmos/__init__.py
@@ -6,7 +6,7 @@ Astronomer Cosmos is a library for rendering dbt workflows in Airflow.
 Contains dags, task groups, and operators.
 """
 
-__version__ = "1.10.1a1"
+__version__ = "1.10.1a4"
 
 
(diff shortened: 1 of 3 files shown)


pankajkoti:
closing in favour of the alternative solution proposed in #1760

cosmos/__init__.py
```diff
@@ -6,7 +6,7 @@
 Contains dags, task groups, and operators.
 """
 
-__version__ = "1.10.0"
+__version__ = "1.10.1a1"
```
pankajkoti, line 9:
```suggestion
__version__ = "1.10.1a4"
```

PR #1756, opened 2025-05-08
Allow to set task display name as a user-defined function

corsettigyg:
## Description

Currently cosmos allows to override the task_id value with the param [normalize_task_id](https://astronomer.github.io/astronomer-cosmos/configuration/task-display-name.html), but there is no process in place to override the `display_name` .

At my current company we have some logic in place using the task_id and corresponding suffixes, but we would like to still change how tasks are displayed in the airflow UI, plus from a user's perspective it is sometimes confusing to have the suffix being displayed in the airflow UI instead of just having the name of the model itself. It would be nice to have a way to keep the task_id as it is while overriding the display_name instead so if a user is already using the task_ids for some custom logic, it is possible to change how the tasks are displayed in the UI without causing issues.

## Breaking Change?

Should not be

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -157,6 +157,8 @@ dev/dags/.airflowignore
 # Local Airflow standalone
 airflow.cfg
 airflow.db
+airflow.db-shm
+airflow.db-wal
 standalone_admin_password.txt
 simple_auth_manager_passwords.json.generated
 webserver_config.py


corsettigyg:
a small disclaimer: I took the liberty to rename `task_display_name` to `task_name` in the code since it was not exactly correct, and since this PR aims to separate both concepts, `task_name` is a more generic name in case that the `display_name` is not set 😄

PR #1759, opened 2025-05-08
Support applying `node_converter` at a task level instead of task group level

### LINKED ISSUE #1758: [Bug] Node_conversion on resource_type test is broken
### Astronomer Cosmos Version

1.10.0

### dbt-core version

1.9.4

### Versions of dbt adapters

dbt-bigquery==1.9.1

### LoadMode

DBT_LS

### ExecutionMode

LOCAL

### InvocationMode

DBT_RUNNER

### airflow version

2.10.5+astro.2

### Operating System

Debian GNU/Linux 12

_No response_

### Deployment

Astronomer

### Deployment details

_No response_

### What happened?

While working with cosmos, I found some behavior of node_converters that seems unexpected to me, and possibly a bug:
1. When using AFTER_EACH, if you use a conversion function on Model or Source, it converts the ENTIRE task group, including both the model/source task AND the test task. I would expect only the model / source to be converted, not the tasks because the selection is specified by DbtResourceType and test is not being specified.
2. When using a conversion function on resource type Test, it very much does not work as expected:
  a. When using any test_behavior (except NONE): It does NOT convert detached tests with multiple parents
  b. When using any test_behavior (except NONE): It creates separate tasks for non-detached tests (single parent) and converts those.  
  c. When using AFTER_EACH: it does NOT convert the non-detached test tasks in the task groups (single parent tests).  This is in addition to the behavior mentioned in 2.b. so now each test has its own task AND it's still covered in the task_group.test task.  
  d When using AFTER_ALL: it does NOT convert the final test task.  This is addition to the behavior mentioned in 2.b so now each test has its own task AND it's still covered in the project_test task.  

### Relevant log output

```shell

```

### How to reproduce

1. Set test_behavior to AFTER_EACH and add a node_converter for DbtResourceType("model") 
2. Set a node_coverter for DbtResourceType("test") and play around with the different test behaviors 

### Anything else :)?

I have a proposed fix to convert the task instead of the "node":
- this is more consistent with node converter depending on the resource type, which seems to align better with individual tasks rather than task groups of two different resource types 
- this works better with different test_behaviors 
- this doesn't work if someone wants to convert entire task groups, but NOT the detached tests, but that seems like an edge-case 

Note: this is a large change to how node_converters work and is almost definitely a breaking change.  

### Contact Details

_No response_

anyapriya:
## Description

Modifies the `node_converter` logic to be used not only at a task group level, but also at a task level.

These are the reasons why someone may want to use `node_converter` per task instead of `task_group`:
- This is more consistent with the node converter depending on the resource type, which seems to align better with individual tasks rather than task groups of two different resource types
- This works better with different test_behaviors
- This supports converting type test nodes 

Closes #1758 

## Breaking Change?

This change was made non-breaking by introducing a new `RenderConfig` argument `node_conversion_by_task_group`, which is set to True by default.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,7 +57,7 @@ repos:
           - --py37-plus
           - --keep-runtime-typing
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.2
+    rev: v0.14.3
     hooks:
       - id: ruff
         args:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,15 +1,87 @@
 Changelog
 =========
 
-1.11.0a1 (2025-06-23)
+1.11.0 (2025-10-29)
 ---------------------
 
-Feature
+Features
 
+* Introduce ``ExecutionMode.WATCHER`` to reduce DAG run time by 1/5 in several PRs. Learn more about it `here <https://astronomer.github.io/astronomer-cosmos/getting_started/watcher-execution-mode.html#watcher-execution-mode>`_. This feature was implemented via multiple PRs, including:
+  * Expose new execution mode by @tatiana @pankajastro @pankajkoti in #1999
+  * Add ``DbtProducerWatcherOperator`` for the proposed ``ExecutionMode.WATCHER`` by @pankajkoti in #1982
+  * Add ``DbtConsumerWatcherSensor`` for the proposed ``ExecutionMode.WATCHER`` by @pankajastro in #1998
+  * Push producer's task completion status to XCOM by @pankajkoti in #2000
+  * Add default priority_weight for ``DbtProducerWatcherOperator`` by @pankajkoti in #1995
+  * Add sample dbt events for the dbt watcher execution mode by @pankajkoti in #1952
+  * Add ``compiled_sql`` as a template fields on ```ExecutionMode.WATCHER``` when using ``run_results.json`` by @pankajastro in #2070
+  * Set ``push_run_results_to_xcom`` kwargs correctly for invocation mode subprocess and Watcher mode by @pankajastro in #2067
+  * Store compiled SQL as template field for dbt callback events in ``ExecutionMode.WATCHER`` by @pankajkoti in #2068
+  * Add initial documentation for ``ExecutionMode.WATCHER`` by @tatiana in #2046
+  * Support running ``State.UPSTREAM_FAILED`` tasks when WATCHER consumer upstream tasks fail by @tatiana in #2062
+  * Fail sensor tasks immediately if the ``ExecutionMode.WATCHER`` producer task fails by @pankajastro in #2040
+  * Add ``WATCHER`` to GitHub issue template by @tatiana in #2056
+  * Add support for ``TestBehavior.AFTER_ALL`` with ``ExecutionMode.WATCHER`` by @pankajastro in #2049
+  * Add support for ``TestBehavior.NONE`` with ``ExecutionMode.WATCHER``  by @pankajastro in #2047
+  * Fix ``ExecutionMode.WATCHER`` behaviour with ``DbtTaskGroup`` by @tatiana in #2044
+  * Fix Cosmos behaviour when using watcher with ``InvocationMode.DBT_RUNNER`` by @tatiana in #2048
+
+* Add Airflow 3 plugin for dbt docs with multiple dbt projects support by @pankajkoti in #2009, check the `documentation <https://astronomer.github.io/astronomer-cosmos/configuration/hosting-docs.html>`_.
 * Initial support to ``dbt Fusion`` by @tatiana in #1803. `More details here. <https://astronomer.github.io/astronomer-cosmos/configuration/dbt-fusion>`_.
+* Support to prune sources without downstream references in dbt projects by @corsettigyg in #1988
+* Allow to set task display name as a user-defined function by @corsettigyg in #1761
+* Add dbt project's hash to dag docs to support dag versioning in Airflow 3 by @pankajkoti in #1907
+* feat: Add Jinja templating support for ``dbt_cmd_flags`` by @skillicinski in #1899
+* Add Scarf metric to collect the execution mode uses by @pankajastro in #1981
+* Support Airflow 3.1 by @tatiana in #1980
+* Add MySQL profile mapping by @Lee2532 in #1977
+* Add sqlserver profile mapping by @pankajastro in #1737
 
-(many other features, pending details)
+Enhancement
 
+* Use XCom to store sql when using ``ExecutionMode.AIRFLOW_ASYNC`` by @pankajastro in #1934
+* Refactor ``AIRFLOW_ASYNC`` teardown so it doesn't install the virtualenv by @pankajastro in #1938
+* Reuse the virtual env for ``AIRFLOW_ASYNC`` setup task by @pankajastro in #1939
+* Improve dataset/asset experience in Cosmos by @tatiana in #2030
+* Add ``downstreams`` to ``DbtNode`` by @wornjs in #2028
+
+Bug fixes
+
+* Fix tags extraction by @ms32035 in #1915
+* Fix task flow operator args by @anyapriya in #2024
+
+Documentation
+
+* Add documentation for Airflow 3 Plugin supporting dbt docs for multiple dbt projects by @pankajkoti in #2063
+* Add Cosmos Deferrable Operator Guide by @pankajastro in #1922
+* Add dbt Fusion documentation by @tatiana in #1824 #1830
+* Update dbt-fusion.rst to explicitly highlight it is in alpha by @tatiana in https://github.com/astronomer/astronomer-cosmos/pull/1838
+* Fix a bunch of docs build errors and warnings by @pankajkoti in https://github.com/astronomer/astronomer-cosmos/pull/1886
+* Add docs note for param virtualenv_dir for async execution mode by @pankajastro in #1969
+* Use pepy.tech downloads badge in README by @pankajkoti in #1920
+* Correct the default value of ``cache_dir`` by @seokyun.ha in #2027
+* Improve ``ExecutionMode.WATCHER`` docs by @tatiana in #2071
+
+Others
+
+* Promote @corsettigyg to committer by @tatiana in #1985
+* Add @pankajkoti and @pankajastro to ``contributors.rst`` by @tatiana in #1983
+* Update setup script for airflow3 script by @dwreeves in #2023
+* Prevent pytest from trying to test classes that aren't actually tests by @anyapriya in #2032
+* Fix ``dag.test()`` for Airflow 3.1+ by syncing DAG to database by @kaxil in #2037
+* Disable Scarf in CI by @pankajastro in #2016
+* Fix failing dbt Fusion tests when run in parallel in CI by @pankajkoti in https://github.com/astronomer/astronomer-cosmos/pull/1896
+* Fix MyPy issues related to ``ObjectStoragePath`` in main branch by @tatiana in #2012
+* Cleanup example dbt event JSON dictionaries kept for XCOM reference by @pankajkoti in #1997
+* Bump min hatch version that includes fixes for click>=8.3.0 by @pankajkoti in #1996
+* Use official postgres image from Docker hub for kubernetes setup by @pankajkoti in #1986
+* Use click<8.3.0 for hatch as click 8.3 breaks hatch by @pankajkoti in #1987
+* Pin Airflow version in type check CI job by @pankajastro in #2003
+* Improve comments after feedback on #1948 by @tatiana in #1963
+* Fix running tests with dbt Fusion 2.0.0 preview versions by @tatiana in #1948
+* Test hardening of dbt node having tags as unset or missing by @pankajkoti in #1918
+* Fix Sphinx issue in the main branch by @tatiana in #2064
+* pre-commit autoupdate in #2065, #2043, #2033, #2019, #1990, #2019, #2008, #1941, #1935, #1924
+* GitHub dependabot update in #2051, #2050, #2038, #2022, #1947, #1955, #1946, #1944, #1945, #1928, #1921, #1917
 
 1.10.2 (2025-08-08)
 ---------------------
--- a/cosmos/__init__.py
+++ b/cosmos/__init__.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 from cosmos import settings
 
-__version__ = "1.11.0a13"
+__version__ = "1.11.0"
 
 if not settings.enable_memory_optimised_imports:
     from cosmos.airflow.dag import DbtDag
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -6,7 +6,7 @@ import logging
 import zlib
 from datetime import timedelta
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable, List, Sequence, Union
+from typing import TYPE_CHECKING, Any, Callable, List, Union
 
 import airflow
 from packaging.version import Version
@@ -303,7 +303,7 @@ class DbtConsumerWatcherSensor(BaseSensorOperator, DbtRunLocalOperator):  # type
 
         return event_json.get("data", {}).get("run_result", {}).get("status")
 
-    def _get_status_from_run_results(self, ti: Any) -> Any:
+    def _get_status_from_run_results(self, ti: Any, context: Context) -> Any:
         compressed_b64_run_results = ti.xcom_pull(task_ids=self.producer_task_id, key="run_results")
 
         if not compressed_b64_run_results:
@@ -323,6 +323,10 @@ class DbtConsumerWatcherSensor(BaseSensorOperator, DbtRunLocalOperator):  # type
             return None
 
         logger.info("Node Info: %s", run_results_str)
+        self.compiled_sql = node_result.get("compiled_code")
+        if self.compiled_sql:
+            self._override_rtif(context)
+
         return node_result.get("status")
 
     def _get_producer_task_state(self, ti: Any) -> Any:
@@ -356,7 +360,7 @@ class DbtConsumerWatcherSensor(BaseSensorOperator, DbtRunLocalOperator):  # type
         if use_events:
             status = self._get_status_from_events(ti, context)
         else:
-            status = self._get_status_from_run_results(ti)
+            status = self._get_status_from_run_results(ti, context)
 
         if status is None:
 
@@ -411,7 +415,7 @@ class DbtSourceWatcherOperator(DbtSourceLocalOperator):
     Executes a dbt source freshness command, synchronously, as ExecutionMode.LOCAL.
     """
 
-    template_fields: Sequence[str] = DbtSourceLocalOperator.template_fields
+    template_fields: tuple[str, ...] = DbtConsumerWatcherSensor.template_fields
 
 
 class DbtRunWatcherOperator(DbtConsumerWatcherSensor):
--- a/cosmos/plugin/airflow3.py
+++ b/cosmos/plugin/airflow3.py
@@ -10,23 +10,34 @@ from typing import Any, Generator, Optional
 from unittest.mock import patch
 from urllib.parse import urlsplit
 
+import airflow
 from airflow.configuration import conf
 from airflow.plugins_manager import AirflowPlugin
 from airflow.sdk import ObjectStoragePath
 from fastapi import FastAPI
 from fastapi.responses import HTMLResponse, JSONResponse, Response
+from packaging.version import Version
 
 from cosmos.constants import AIRFLOW_OBJECT_STORAGE_PATH_URL_SCHEMES
 from cosmos.plugin.snippets import IFRAME_SCRIPT
+
+# Airflow version gating: External views feature for the plugins used here (CosmosAF3Plugin) exist only in >= 3.1
+AIRFLOW_VERSION = Version(airflow.__version__)
+
+
+def ensure_airflow_version_supported() -> None:
+    if AIRFLOW_VERSION < Version("3.1.0"):
+        raise RuntimeError(
+            "Cosmos AF3 plugin requires Airflow >= 3.1. External views are unavailable on earlier versions."
+        )
+
 
 API_BASE = conf.get("api", "base_url", fallback="")  # reads AIRFLOW__API__BASE_URL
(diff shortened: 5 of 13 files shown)


tatiana (commented):
Hi @anyapriya , thanks a lot for the contribution!

Since you marked this PR as a breaking change, could you give an example of DAG that would break?

Do you see a way of accomplishing the same goal without being a breaking change?

anyapriya:
@tatiana 

I think it would change functionality unexpectedly rather than explicitly break something.  Dags should continue to run, they will just behave differently.  Here's an example:

User wants to run JUST the source tests, but not models and model tests so they use Node conversion on type of model.
- Pre-PR behavior: has task-groups for source+source_tests, plus tasks for modified_model_task.  Does NOT run model_tests.  
- Post-PR behavior: has a task-group for source+source_tests, plus a task group for modified_model_task+model_tests.  DOES run model_tests.  

I think situations like this are edge-cases that could be gotten around with different select/exclude parameters, but it is different behavior.  But I really want your team's input on this since you're more familiar with the package and with the intention of node converters.  

The only way I can see of accomplishing the same goal without being a breaking change is to add a "TaskConverter" (which works the way I proposed in the PR) in addition to the current "NodeConverter" functionality.  But I don't think that addresses the bug I originally raised where NodeConverter breaks when passing in the type of Test, just provides a workaround.  If we went with the alternative path, then I'd really recommend a condition to raise an exception if someone attempts use NodeConverter on a resource type of test, which I think is a breaking change in a different way.

anyapriya:
Hi Tatiana, thanks for the feedback!  Understood, that does seem a little safer, especially if this is going out before a major version.  I'll do my best to implement it that way.

anyapriya:
@tatiana Apologies for the long delay on this, things have been crazy!  But I think this finally accomplishes your ask to do this new behavior by feature flag.

anyapriya:
@tatiana just fixed the pre-commit issue that was caused by merging in main

anyapriya:
@tatiana one test failed with the following exception:
```
airflow.exceptions.AirflowException: BigQuery job airflow_simple_dag_async_stg_payments_run_2025_10_28T15_41_35_635567_00_00_e0cd4e5d81c1560f09c6f358ea399345 failed: {'reason': 'notFound', 'message': 'Not found: Table astronomer-dag-authoring:release_17.raw_payments was not found in location US'}
```

I don't think that's related to my PR though I might be wrong?  Any thoughts?

tatiana:
> @tatiana one test failed with the following exception:
> 
> ```
> airflow.exceptions.AirflowException: BigQuery job airflow_simple_dag_async_stg_payments_run_2025_10_28T15_41_35_635567_00_00_e0cd4e5d81c1560f09c6f358ea399345 failed: {'reason': 'notFound', 'message': 'Not found: Table astronomer-dag-authoring:release_17.raw_payments was not found in location US'}
> ```
> 
> I don't think that's related to my PR though I might be wrong? Any thoughts?

@anyapriya it seems unrelated, just retriggered and if it fails again, we'll troubleshoot it

anyapriya:
Re-run worked!  🎉

cosmos/airflow/graph.py
```diff
@@ -111,7 +111,7 @@ def calculate_leaves(tasks_ids: list[str], nodes: dict[str, DbtNode]) -> list[st
 def exclude_detached_tests_if_needed(
     node: DbtNode,
     task_args: dict[str, str],
-    detached_from_parent: dict[str, DbtNode] | None = None,
+    detached_from_parent: dict[str, list[DbtNode]] | None = None,
```
tatiana, line 114: If we want this change to be non-breaking:
```suggestion
    detached_from_parent: dict[str, DbtNode | list[DbtNode]] | None = None,
```
And we can normalise within the function itself:
```
        if detached_from_parent is None:
            detached_from_parent = {}
        else:
            # Normalize so that every value is always a list
            detached_from_parent = {
                key: value if isinstance(value, list) else [value]
                for key, value in detached_from_parent.items()
            }
```

cosmos/airflow/graph.py
```diff
@@ -152,7 +152,7 @@ def create_test_task_metadata(
     on_warning_callback: Callable[..., Any] | None = None,
     node: DbtNode | None = None,
     render_config: RenderConfig | None = None,
-    detached_from_parent: dict[str, DbtNode] | None = None,
+    detached_from_parent: dict[str, list[DbtNode]] | None = None,
```
tatiana, line 155: Same feedback from https://github.com/astronomer/astronomer-cosmos/pull/1759/files#r2470359625 applies here

cosmos/airflow/graph.py
```diff
@@ -297,7 +297,7 @@ def create_task_metadata(
     test_behavior: TestBehavior = TestBehavior.AFTER_ALL,
     test_indirect_selection: TestIndirectSelection = TestIndirectSelection.EAGER,
     on_warning_callback: Callable[..., Any] | None = None,
-    detached_from_parent: dict[str, DbtNode] | None = None,
+    detached_from_parent: dict[str, list[DbtNode]] | None = None,
```
tatiana, line 300: Same feedback from https://github.com/astronomer/astronomer-cosmos/pull/1759/files#r2470359625 applies here

docs/configuration/render-config.rst
```diff
@@ -27,7 +27,7 @@ The ``RenderConfig`` class takes the following arguments:
 - ``normalize_task_display_name``: This function allows users to set a custom user-defined function to alter the display name independently of the model name. This way, the task_id can be preserved while the model display name is modified.
 - ``should_detach_multiple_parents_tests``: A boolean to control if tests that depend on multiple parents should be run as standalone tasks. See `Parsing Methods <testing-behavior.html>`_ for more information.
 - ``enable_owner_inheritance``: (introduced in 1.10.2) A boolean to control if dbt owners should be imported as part of the airflow DAG owners. Defaults to True.
-
+- ``node_conversion_by_task_group``: A boolean to control if node_converters are used at the task group level (ex. converting models with test_behavior=AFTER_EACH means the entire task group is converted including the run task and the test task), or the individual task level (gives more granularity for converting just the run tasks or just the test tasks).  Defaults to True.
```
tatiana, line 30: This is a great strategy to make the change non-breaking, @anyapriya. Thank you!

It may be worth moving this definition next to `node_converters` on line 18, since they are related.

cosmos/airflow/graph.py
```diff
@@ -435,6 +435,62 @@ def is_detached_test(node: DbtNode) -> bool:
...
+        )
+        logger.debug(f"Converting node <{node.unique_id}> task <{task_id}> using <{conversion_function.__name__}>")
+        task = conversion_function(  # type: ignore
+            dag=dag,
+            task_group=task_group,
+            dbt_project_name=dbt_project_name,
+            execution_mode=execution_mode,
+            task_args=task_args,
+            test_behavior=test_behavior,
+            source_rendering_behavior=source_rendering_behavior,
```
tatiana, line 479:
```suggestion
            source_rendering_behavior=source_rendering_behavior,
            source_pruning=source_pruning,
```

This property was introduced in https://github.com/astronomer/astronomer-cosmos/pull/1988 and will also be released as part of Cosmos 1.11

cosmos/airflow/graph.py
```diff
@@ -435,6 +435,62 @@ def is_detached_test(node: DbtNode) -> bool:
...
+            test_indirect_selection=test_indirect_selection,
+            on_warning_callback=on_warning_callback,
+            normalize_task_id=normalize_task_id,
```
tatiana, line 482:
```suggestion
            normalize_task_id=normalize_task_id,
            normalize_task_display_name=normalize_task_display_name,
```

This was introduced in https://github.com/astronomer/astronomer-cosmos/pull/1761 and will also be part of the 1.11.0 release

cosmos/airflow/graph.py
```diff
@@ -435,6 +435,62 @@ def is_detached_test(node: DbtNode) -> bool:
...
+            node=node,
+            task_id=task_id,
+            detached_from_parent=detached_from_parent,
```
tatiana, line 485:
```suggestion
            detached_from_parent=detached_from_parent,
            enable_owner_inheritance=enable_owner_inheritance,
```
This was introduced in #1787 and it was released as part of Cosmos 1.10.2

cosmos/airflow/graph.py
```diff
@@ -435,6 +435,62 @@ def is_detached_test(node: DbtNode) -> bool:
...
+        )
+        if task is not None:
+            logger.debug(f"Conversion of node <{node.unique_id}> task <{task_id}> was successful!")
+    else:
+        task = create_airflow_task(task_meta, dag, task_group)
+    return task
```
tatiana, line 491: Personal note while reviewing:

This code is very similar to 
https://github.com/astronomer/astronomer-cosmos/blob/9953f968d6c743bf431af96f1d6a62a90a6dff1c/cosmos/airflow/graph.py#L767-L794

The main differences are:
- This function focuses on converting an individual task instead of a task group
- Instead of defaulting to `generate_task_or_group`, the new method defaults to `create_airflow_task`
- There were a few missing properties that were introduced after this PR was first created, but that need to be added; otherwise, we'll have inconsistent behaviour between how we convert task groups and tasks

cosmos/airflow/graph.py
```diff
@@ -805,7 +908,24 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
...
+            resource_type=DbtResourceType.TEST,  # type: ignore
+            execution_mode=execution_mode,
+            task_args=task_args,
+            test_behavior=test_behavior,
+            source_rendering_behavior=source_rendering_behavior,
+            test_indirect_selection=test_indirect_selection,
+            on_warning_callback=on_warning_callback,
+            normalize_task_id=normalize_task_id,
+            detached_from_parent=detached_from_parent,
+            node_conversion_by_task_group=render_config.node_conversion_by_task_group,
```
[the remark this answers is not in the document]
anyapriya: Does not break because `**kwargs` but still worth cleaning up since it's unused.  Done.

cosmos/airflow/graph.py
```diff
@@ -823,7 +943,24 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
...
+                resource_type=node.resource_type,
+                execution_mode=execution_mode,
+                task_args=task_args,
+                test_behavior=test_behavior,
+                source_rendering_behavior=source_rendering_behavior,
+                test_indirect_selection=test_indirect_selection,
+                on_warning_callback=on_warning_callback,
+                normalize_task_id=normalize_task_id,
+                detached_from_parent=detached_from_parent,
+                node_conversion_by_task_group=render_config.node_conversion_by_task_group,
```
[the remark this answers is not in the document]
anyapriya: Does not break because `**kwargs` but still worth cleaning up since it's unused.  Done.

cosmos/airflow/graph.py
```diff
@@ -444,23 +500,28 @@ def generate_task_or_group(
...
     source_rendering_behavior: SourceRenderingBehavior,
     test_indirect_selection: TestIndirectSelection,
+    dbt_project_name: str | None = None,
+    node_converters: dict[DbtResourceType, Callable[..., Any]] = {},
     source_pruning: bool = False,
     on_warning_callback: Callable[..., Any] | None = None,
     normalize_task_id: Callable[..., Any] | None = None,
     normalize_task_display_name: Callable[..., Any] | None = None,
-    detached_from_parent: dict[str, DbtNode] | None = None,
+    detached_from_parent: dict[str, list[DbtNode]] | None = None,
```
tatiana, line 509: Same feedback from https://github.com/astronomer/astronomer-cosmos/pull/1759/files#r2470359625 applies here

cosmos/airflow/graph.py
```diff
@@ -484,9 +545,25 @@ def generate_task_or_group(
...
+                    node=node,
+                    resource_type=node.resource_type,
+                    execution_mode=execution_mode,
+                    task_args=task_args,
+                    test_behavior=test_behavior,
+                    source_rendering_behavior=source_rendering_behavior,
+                    test_indirect_selection=test_indirect_selection,
+                    on_warning_callback=on_warning_callback,
+                    normalize_task_id=normalize_task_id,
+                    detached_from_parent=detached_from_parent,
```
tatiana, line 565:
```suggestion
                    detached_from_parent=detached_from_parent,
                    source_pruning=source_pruning,
                    normalize_task_display_name=normalize_task_display_name,
                    enable_owner_inheritance=enable_owner_inheritance,
```

tatiana: All these properties were added to `RenderConfig` in the PRs #1761, #1787, and #1988, and we probably want them here as well.

cosmos/airflow/graph.py
```diff
@@ -497,11 +574,43 @@ def generate_task_or_group(
...
+                    node=node,
+                    resource_type=DbtResourceType.TEST,  # type: ignore
+                    execution_mode=execution_mode,
+                    task_args=task_args,
+                    test_behavior=test_behavior,
+                    source_rendering_behavior=source_rendering_behavior,
+                    test_indirect_selection=test_indirect_selection,
+                    on_warning_callback=on_warning_callback,
+                    normalize_task_id=normalize_task_id,
+                    detached_from_parent=detached_from_parent,
```
tatiana, line 592:
```suggestion
                    detached_from_parent=detached_from_parent,
                    normalize_task_display_name=normalize_task_display_name,
                    enable_owner_inheritance=enable_owner_inheritance,
```

Same explanation as in:
https://github.com/astronomer/astronomer-cosmos/pull/1759/files#r2470511753

Except that we probably do not need to give the source_prunning one to test nodes.

cosmos/airflow/graph.py
```diff
@@ -497,11 +574,43 @@ def generate_task_or_group(
...
+                node=node,
+                resource_type=node.resource_type,
+                execution_mode=execution_mode,
+                task_args=task_args,
+                test_behavior=test_behavior,
+                source_rendering_behavior=source_rendering_behavior,
+                test_indirect_selection=test_indirect_selection,
+                on_warning_callback=on_warning_callback,
+                normalize_task_id=normalize_task_id,
+                detached_from_parent=detached_from_parent,
```
tatiana, line 612: Needs the same change as: https://github.com/astronomer/astronomer-cosmos/pull/1759/files#r2470518583

Since we are declaring `generate_or_convert_task` multiple times with many parameters, we probably want to refactor to avoid repeating 14 lines each time. As the code is now, it is very error-prone

cosmos/airflow/graph.py
```diff
@@ -751,7 +860,7 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
     normalize_task_display_name = render_config.normalize_task_display_name
     enable_owner_inheritance = render_config.enable_owner_inheritance
     tasks_map: dict[str, Union[TaskGroup, BaseOperator]] = {}
-    task_or_group: TaskGroup | BaseOperator
+    task_or_group: TaskGroup | BaseOperator | None
```
tatiana, line 863: In which scenario do we want this to be `None`?

cosmos/airflow/graph.py
```diff
@@ -805,7 +908,23 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
...
+            node=node,
+            resource_type=DbtResourceType.TEST,  # type: ignore
+            execution_mode=execution_mode,
+            task_args=task_args,
+            test_behavior=test_behavior,
+            source_rendering_behavior=source_rendering_behavior,
+            test_indirect_selection=test_indirect_selection,
+            on_warning_callback=on_warning_callback,
+            normalize_task_id=normalize_task_id,
+            detached_from_parent=detached_from_parent,
```
tatiana, line 926: Same feedback as: https://github.com/astronomer/astronomer-cosmos/pull/1759/files#r2470518583

Since we are repeating this block 5 times, it feels like we should refactor and review the `generate_or_convert_task` interface to avoid so many arguments. To share the `render_config` may simplify the interface significantly

cosmos/airflow/graph.py
```diff
@@ -823,7 +942,23 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
...
+                resource_type=node.resource_type,
+                execution_mode=execution_mode,
+                task_args=task_args,
+                test_behavior=test_behavior,
+                source_rendering_behavior=source_rendering_behavior,
+                test_indirect_selection=test_indirect_selection,
+                on_warning_callback=on_warning_callback,
+                normalize_task_id=normalize_task_id,
+                detached_from_parent=detached_from_parent,
+            )
```
tatiana, line 961: Same feedback as: https://github.com/astronomer/astronomer-cosmos/pull/1759/files#r2470518583

Since we are repeating this block 5 times, it feels like we should refactor and review the `generate_or_convert_task` interface to avoid so many arguments. To share the `render_config` may simplify the interface significantly

cosmos/airflow/graph.py
anyapriya: I made the change because it looks like it's defined as a Dict of lists and it seemed like inconsistent (that was causing me issues as well on the pre-commit):
```

    detached_from_parent: dict[str, list[DbtNode]] = defaultdict(list)
    identify_detached_nodes(nodes, render_config, detached_nodes, detached_from_parent)
...

def identify_detached_nodes(
    nodes: dict[str, DbtNode],
    render_config: RenderConfig,
    detached_nodes: dict[str, DbtNode],
    detached_from_parent: dict[str, list[DbtNode]],
) -> None:
    """
    Given the nodes that represent a dbt project and the test_behavior, identify the detached test nodes
    (test nodes that have multiple dependencies and should run independently).

    Change in-place the dictionaries detached_nodes (detached node ID : node) and detached_from_parent (parent node ID that
    is upstream to this test and the test node).
    """
    if should_create_detached_nodes(render_config):
            if is_detached_test(node):
                detached_nodes[node_id] = node
                for parent_id in node.depends_on:
                    detached_from_parent[parent_id].append(node)


```

Even the function `exclude_detached_tests_if_needed`  has the parameter set as `dict[str, DbtNode]` but then assumes that it's a dict of lists of DbtNodes  and has a type: ignore on that line.  Where is it ever used as a dict[str, DbtNode] ?  I can change it but I would love to better understand before I do since it seems like a possible bug to me.

cosmos/airflow/graph.py
anyapriya: I'll need to think about the best way to do this.  Just to confirm, you want the refactor as part of this PR and not part of a follow-up?

cosmos/airflow/graph.py
anyapriya: Great question, I was making it consistent with the fact that that's the return type of generate_task_or_group.  I can remove the None from both if you think that's better.  It could also be possible that a user uses a conversion function on some tasks and returns None for others?

cosmos/airflow/graph.py
tatiana: I believe you're right, @anyapriya , closing this comment

cosmos/airflow/graph.py
tatiana: Addressed in https://github.com/astronomer/astronomer-cosmos/pull/2080

cosmos/airflow/graph.py
tatiana: Addressed in https://github.com/astronomer/astronomer-cosmos/pull/2080

cosmos/airflow/graph.py
tatiana: I think it is fine for now, @anyapriya , we can review this in future if it brings problems.

cosmos/airflow/graph.py
tatiana: I managed to clear it up a bit in https://github.com/astronomer/astronomer-cosmos/pull/2080

cosmos/airflow/graph.py
tatiana: Addressed in 50923d0f in https://github.com/astronomer/astronomer-cosmos/pull/2080

cosmos/airflow/graph.py
tatiana: Addressed in 50923d0f in https://github.com/astronomer/astronomer-cosmos/pull/2080

cosmos/airflow/graph.py
tatiana: I think I've addressed this in https://github.com/astronomer/astronomer-cosmos/pull/2080

tatiana:
@anyapriya I addressed what I believe was the last necessary feedback in a follow-up PR:
https://github.com/astronomer/astronomer-cosmos/pull/2080

I also wanted to confirm how much the changes introduced would impact DAG parsing performance in Cosmos, so we released Cosmos 1.12.0a1 & 1.12.0a2 based on the changes in this PR and https://github.com/astronomer/astronomer-cosmos/pull/2080.

Even though there is a performance degradation, it is in the order of 10ms for a dbt project with 239 nodes. It is irrelevant when compared to the time it takes to run dbt ls. More details on how I analysed this below.

I used our [benchmark project](https://github.com/astronomer/cosmos-benchmark/tree/main/benchmark) for this, with the following versions:
- astronomer-cosmos
  - 1.11.0
  - 1.12.0a1
- dbt=1.10.13
- bigquery=1.9.0

Cosmos 1.11.0, in average:
- It took 4.25s to parse the dbt project for DAG using LoadMode.DBT_LS
- It took 0.0377s to build the Airflow DAG.

Cosmos 1.12.0a1 & 1.12.0a2, in average:
- It took 4.26s to parse the dbt project for DAG using LoadMode.DBT_LS
- It took 0.0383s to build the Airflow DAG.

I also spoke to @corsettigyg, who uses the `node_converters` feature extensively at GetYourGuide - and he'll help us validate the latest alpha release to make sure we have not broken anything for their use case. If there are any problems, we can address them in a follow-up PR before the stable version of Cosmos with this feature is released.

For anyone interested in testing this feature, please, check out:
- Release notes: [v1.12.0a2](https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.12.0a2)
- PyPI package [here](https://pypi.org/project/astronomer-cosmos/1.12.0a2/)

corsettigyg:
I verified on my side and it is all good 😄  we do iterate a lot over cosmos nodes, but this PR did not break anything on our side, which was a positive surprise.

corsettigyg (approved):

tatiana:
Thank you a lot for reviewing, @corsettigyg! It's great to confirm in another real-world project.

Issue #1751, opened 2025-05-07
Task execution hanging on Airflow 2 when using dbt runner in DAG processing


pankajkoti:
We’ve received several user reports indicating that tasks are hanging after upgrading to Cosmos 1.9.0.

I attempted to reproduce the issue on Airflow 3 but was unable to reproduce it there. However, when testing on Airflow 2, I successfully reproduced the hang when upgrading from Cosmos 1.8 to 1.9.

Upon further investigation—thanks in part to our various 1.9 alpha releases—I traced the issue to version 1.9.0a3 (1.9.0a2 does not exhibit the problem). The hang appears to originate from the use of dbt Runner during DAG processing, specifically at [this line](https://github.com/astronomer/astronomer-cosmos/blob/21a2f13915ae60014b9a323137a181934ba5e948/cosmos/dbt/graph.py#L270).

When I switched to using a subprocess-based approach at that point, tasks progressed normally without getting stuck.

Although the exact reason why dbt Runner causes tasks to hang in this context is still unclear, we need to investigate how the dbt Runner library interacts with Airflow’s task execution system. Understanding this behaviour is key to resolving the issue and ensuring we can take advantage of the optimisations we’re targeting and aimed for in PR #1484 by using dbt Runner in that part of the code.

PR #1760, opened 2025-05-09
Disable dbt static parser during Airflow task execution using dbt runner

pankajkoti:
This PR adds support for conditionally applying the `--no-static-parser` dbt flag in Cosmos operators, ensuring it is included only when InvocationMode.DBT_RUNNER is used during task execution.

### Rationale
**Static Parser Issue**: User reports and investigation revealed that, starting with Cosmos 1.9.0 (see PR #1484), using dbtRunner for both DAG parsing and task execution in Airflow 2.x can cause task hangs. This is due to dbt's static parser interacting poorly with Cosmos's use of temporary project directories, especially when the temp paths differ between parsing and execution.
**Workaround**: Adding the `--no-static-parser` flag when invoking dbtRunner during task execution avoids these hangs and ensures reliable operation. This flag is not needed (and should not be added) when using the subprocess invocation mode.

closes: #1751 
related: #1750 

---------

### DIFF
--- a/cosmos/__init__.py
+++ b/cosmos/__init__.py
@@ -6,7 +6,7 @@ Astronomer Cosmos is a library for rendering dbt workflows in Airflow.
 Contains dags, task groups, and operators.
 """
 
-__version__ = "1.10.1a2"
+__version__ = "1.10.0"
 
 
 from cosmos.airflow.dag import DbtDag
--- a/cosmos/__init__.py
+++ b/cosmos/__init__.py
@@ -6,7 +6,7 @@ Astronomer Cosmos is a library for rendering dbt workflows in Airflow.
 Contains dags, task groups, and operators.
 """
 
-__version__ = "1.10.0"
+__version__ = "1.10.1a2"
 
 
 from cosmos.airflow.dag import DbtDag
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import inspect
 import json
-import logging
 import os
 import tempfile
 import time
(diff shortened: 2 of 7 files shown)


cosmos/__init__.py
```diff
@@ -6,7 +6,7 @@
 Contains dags, task groups, and operators.
 """
 
-__version__ = "1.10.0"
+__version__ = "1.10.1a2"
```
pankajkoti, line 9:
```suggestion
__version__ = "1.10.0"
```

cosmos/operators/local.py
```diff
@@ -2,7 +2,6 @@
 
 import inspect
 import json
-import logging
```
pankajkoti, line 5: This was removed by pre-commit and I confirmed that it's not getting used in the module.

tatiana (approved):
@pankajkoti its great you found this alternative than setting SUBPROCESS as the standard `InvocationMethod` during task execution. I'm glad we have a path to avoid the hanging behaviour some customers were facing.

Would it be worth for us to add a follow-up ticket for us to review in the future if we can find alternatives to this? According to dbt there can be significant performance improvements by enabling this setting, and we may be able to find alternative solutions in the future for this problem

pankajkoti:
> Would it be worth for us to add a follow-up ticket for us to review in the future if we can find alternatives to this? According to dbt there can be significant performance improvements by enabling this setting, and we may be able to find alternative solutions in the future for this problem

Yes, although I have not found any significant improvement with using static parsing, nor a degradation upon disabling it :), I have logged a follow-up ticket https://github.com/astronomer/astronomer-cosmos/issues/1771 to investigate that

PR #1761, opened 2025-05-09
Allow to set task display name as a user-defined function

corsettigyg:
## Description

re-opened #1756 after resetting the repository by mistake.

> Currently cosmos allows to override the task_id value with the param [normalize_task_id](https://astronomer.github.io/astronomer-cosmos/configuration/task-display-name.html), but there is no process in place to override the display_name . At my current company we have some logic in place using the task_id and corresponding suffixes, but we would like to still change how tasks are displayed in the airflow UI, plus from a user's perspective it is sometimes confusing to have the suffix being displayed in the airflow UI instead of just having the name of the model itself. It would be nice to have a way to keep the task_id as it is while overriding the display_name instead so if a user is already using the task_ids for some custom logic, it is possible to change how the tasks are displayed in the UI without causing issues.

## Breaking Change?

No
## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -193,6 +193,7 @@ def _get_task_id_and_args(
     args: dict[str, Any],
     use_task_group: bool,
     normalize_task_id: Callable[..., Any] | None,
+    normalize_task_display_name: Callable[..., Any] | None,
     resource_suffix: str,
     include_resource_type: bool = False,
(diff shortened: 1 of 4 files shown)


tatiana (commented):
Hi @corsettigyg ! Thanks a lot for the ongoing contributions.

I'd love to understand more about this use case. Does your company currently use the same or different approaches to create `display_name` and `task ID`?

PR #1768, opened 2025-05-19
Support custom `packages-install-path`

tatiana:
The feature introduced in #1670 (Support running `dbt deps` incrementally to pre-defined `dbt_packages` during task execution) did not work as expected if users had defined a custom path for `packages-install-path`. It only worked if the default (`dbt_packages` was being used. This PR aims to solve the issue.


pankajkoti (approved):
LGTM, nice fix 🙇🏽

Issue #1652, opened 2025-04-04
[Bug] Increased airflow memory usage after installing cosmos


rubanolha:
### Astronomer Cosmos Version

1.9.2

### dbt-core version

1.9.3

### Versions of dbt adapters

dbt-adapters==1.14.3
dbt-artifacts-parser==0.7.0
dbt-common==1.16.0
dbt-core==1.9.3
dbt-databricks==1.9.7
dbt-extractor==0.5.1
dbt-semantic-interfaces==0.7.4
dbt-spark==1.9.2
sqlfluff-templater-dbt==3.3.0

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

LOCAL

### InvocationMode

DBT_RUNNER

### airflow version

2.10.4

### Operating System

macOS 15.3.2, debian 12

### Deployment

Official Apache Airflow Helm Chart

### Deployment details

Tested on two different environments, and both can reproduce the issue:
- Locally on Mac with a local Airflow setup, using a virtual environment with Python 3.12.8.
- Airflow Helm Chart deployed via Argo CD, using Kubernetes Executor workers, with Python 3.12.8.

### What happened?

After adding cosmos to our Airflow project (as part of pyproject), we observed a noticeable CPU increase on each pod, which leads to OOM on lightweight InitContainers such as checkDb and waitForDbMigrations (which have a 256m CPU limit by default), and a ~100 MB memory increase on other pods and while running any Airflow command locally.

Investigation Findings:
- During startup, Airflow loads all plugin init files, including those provided by cosmos.
- Installing cosmos (via pip install astronomer-cosmos) before running `airflow db check` causes an additional memory usage of approximately 50–100 MB.
- Uninstalling cosmos (pip uninstall astronomer-cosmos) decreases memory usage by 50–100 MB.

So every k8s pod now uses approximately +100 MB more memory.

The memory was measured using [memory-profiler ](https://pypi.org/project/memory-profiler/) locally and with the kubernetes.memory.usage metric for the Kubernetes deployment.

### Relevant log output

```shell
Init:CrashLookBackOff
The container last terminated 2 minutes ago with exit code 137 because of OOMKilled. 
Db-migration: COntainer îs waiting because of PodInitializing. It is not started and not ready.
```

### How to reproduce

To test locally with local airflow [set up](https://airflow.apache.org/docs/apache-airflow/2.3.0/start/local.html)
1. pip install memory-profiler
2. pip install matplotlib
[...]

Issue #1471, opened 2025-01-16
Improve memory footprint of Cosmos `DbtDag`, `DbtTaskGroup` and operators


tatiana:
**Context**

Some users have reported high memory consumption per Airflow worker node with (and without) Cosmos, when using Airflow Celery Executors. Using Airflow Kubernetes Executor didn't help with the task-level memory consumption.

A customer shared numbers to run a trivial Airflow task is around 250 MB, without Cosmos. Using Cosmos increases the consumption to 300 MB. While running small DAGs with few parallel tasks may not be an issue, it can become a resource management issue for larger workflows.

An example of how the memory was tracked was to instantiate a sequence of `PythonOperator`s that would execute the following function:

```
def write_temp_files():
    process = psutil.Process()
    temp_dir = tempfile.mkdtemp()
    print(f"Temporary directory created: {temp_dir}")

    try:
        for i in range(10):  # Create 10 temporary files
            file_path = os.path.join(temp_dir, f"temp_file_{i}.txt")
            with open(file_path, "w") as temp_file:
                temp_file.write("X" * 10**6)  # Write 1 MB of data

            mem_info = process.memory_info()
            print(f"Step {i + 1}: Memory Usage: {mem_info.rss / (1024 * 1024):.2f} MB")

            time.sleep(1)  # Simulate delay between file writes

    finally:
        for file_name in os.listdir(temp_dir):
            os.remove(os.path.join(temp_dir, file_name))
        os.rmdir(temp_dir)
        print(f"Temporary directory deleted: {temp_dir}")
```

**What we can do**

There is a limit to what we can do at a Cosmos level on improving performance since the plain Airflow baseline is already high.

The goal of this ticket is to:
- [ ] confirm the memory consumption per task in a distributed environment, such as Astro
- [ ] spike making imports local inside `def` and not at top-level modules
- [ ] evaluate other possible improvements

josix:
Hello @tatiana, I'm interested in the performance enhancement topic. If the task isn't urgent, I'd be happy to start by investigating the current memory footprint and familiarizing myself with the codebase. Please feel free to assign it to me if possible, thanks!

tatiana:
@josix, we'd love help on this; thank you very much! I'm assigning the ticket to you - please let us know if you need any support.

sbaldassin:
@josix hey here's the customer that shared the numbers above :). please do let me know if there's anything you want me to test/try. cheers

Issue #1213, opened 2024-09-23
Stop importing/exposing everything that can be used by end-users in Cosmos __init__


tatiana:
The current approach is very error-prone:
https://github.com/astronomer/astronomer-cosmos/blob/aa26fc35b7caf5e456e1d582ebab4056c5bfb34f/cosmos/__init__.py

People who add new execution modes have to change multiple places, and the tendency is for this module to grow indefinitely. We should stop doing this and clean up what we want to expose from within the root path of Cosmos.

pankajkoti:
Should we create a label saying `breaking-change` and attach to such issues?

tatiana:
I think this is a great suggestion, @pankajkoti , could you do this, please?

pankajkoti:
Thanks @tatiana for the agreement. I have created a label and attached to the ones that I felt are breaking changes(most of them already had Cosmos 2.0.0 milestone set). Rethinking, I feel this could be duplicacy to setting the Milestone 2.0.0, but let's see if it is of any help to also have that label. If we think it's not needed, we can delete that label.

pankajkoti:
we also introduced #2531 that could help further until we address this

PR #1769, opened 2025-05-19
Optimise memory usage with optional explicit imports

pankajkoti:
This PR introduces a new configuration flag `enable_memory_optimised_imports` under the `cosmos` Airflow config section (environment variable `AIRFLOW__COSMOS__ENABLE_MEMORY_OPTIMISED_IMPORTS`) to optimise memory usage when Cosmos is installed but not actively used or when only certain modules of Cosmos need to be used (achieved by importing them explicitly with their full module names).

## Changes made to accommodate the above

- Introduce `enable_memory_optimised_imports` in `cosmos/settings.py` and guard eager imports in `__init__.py`.
- Extract provider info into `cosmos/provider_info.py` and update entry-points.

## Problem 

When Cosmos is installed, it eagerly imports many classes and modules (e.g., `DbtDag`, `operators`, etc) in `__init__.py`, leading to increased memory usage—observed to be approximately 200MB per task per worker node even if Cosmos isn’t actively used. 

## Proposed Solution

By default, `enable_memory_optimised_imports` is set to `False`, preserving the current behaviour and maintaining backward compatibility (i.e., all top-level exports remain available). When `explicit_imports` is set to `True`, top-level imports such as `DbtDag` are no longer automatically exposed via `cosmos.__init__.py`. This prevents the loading of large modules unless explicitly imported, resulting in reduced memory usage.

In Cosmos 2.0, this will become the default behaviour, and we'll remove the existing behaviour of allowing users to import everything from cosmos (`__init__.py`) as mentioned in #1213 

## Usage

To enable optimised imports:
```
export AIRFLOW__COSMOS__ENABLE_MEMORY_OPTIMISED_IMPORTS=True
```

### Non-Cosmos DAG
The following experiment was conducted on an Astro deployment that had only a single non-Cosmos DAG (DAG with 2 simple BashOperator tasks echoing outputs) running, with the `astronomer-cosmos` package installed in the deployment.

**Memory usage with default approach of `enable_memory_optimised_imports` config disabled ~900MB**

Screenshot 2025-05-20 at 1 20 07 AM

**Memory usage with `enable_memory_optimised_imports` config enabled ~700MB**

Screenshot 2025-05-20 at 1 20 22 AM

### Cosmos DAG
The following experiment was conducted on an Astro deployment that had the below Cosmos DAG running a jaffle-shop dbt project
DAG Code:
```
from datetime import datetime
from cosmos.airflow.dag import DbtDag
from cosmos.config import ProjectConfig, RenderConfig
from cosmos.constants import LoadMode, InvocationMode, TestBehavior
from include.profiles import snowflake_db
from include.constants import jaffle_shop_path, venv_execution_config

simple_dag = DbtDag(
    project_config=ProjectConfig(jaffle_shop_path),
    profile_config=snowflake_db,
    execution_config=venv_execution_config,
    render_config=RenderConfig(
        test_behavior=TestBehavior.NONE,
    ),
    schedule=None,
    catchup=False,
    dag_id="simple_dag",
    tags=["simple"],
    default_args={
        "retries": 2,
    },
)
```
where below are the values for imported constants in the above DAG
```
jaffle_shop_path = Path("/usr/local/airflow/dbt/jaffle_shop")
dbt_executable = Path("/usr/local/airflow/dbt_venv/bin/dbt")
venv_execution_config = ExecutionConfig(dbt_executable_path=str(dbt_executable))
```

**Memory usage with default approach of `enable_memory_optimised_imports` config disabled.**

It was observed that when **DAGs are running the memory usage peaks to 1.8-2.0GB and when no DAGs are running (idle worker), the memory usage hovered around ~990 MB**
Screenshot 2025-05-21 at 5 20 13 PM

**Memory usage with `enable_memory_optimised_imports` config enabled**

It was observed that for **the first DAG run the memory usage peaked upto 1.6 GB but for subsequent DAG runs the memory usage hovered around ~780 MB. This memory usage of ~780 MB remained consistent when DAGs were run (I gave about 5 subsequent DAG runs one after the other) or the worker was idle.**

Screenshot 2025-05-21 at 5 18 38 PM

This change thus provides users with more control over Cosmos’s memory footprint with leveraging the optional config.

closes: #1652 
related: #1213 
related: #1471 

---------


cosmos/telemetry.py
```diff
@@ -21,12 +22,12 @@ def should_emit() -> bool:
     return settings.enable_telemetry and not settings.do_not_track and not settings.no_analytics
 
 
-def collect_standard_usage_metrics() -> dict[str, object]:
+def collect_standard_usage_metrics() -> dict[str, Any]:
```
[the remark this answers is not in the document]
pankajkoti: mypy complained on that, hence keeping it as is.

tests/test_telemetry.py
```diff
@@ -45,7 +45,7 @@ class MockFailedResponse:
     text = "Non existent URL"
 
 
-@patch("cosmos.telemetry.httpx.get", return_value=MockFailedResponse())
+@patch.object(telemetry.httpx, "get", return_value=MockFailedResponse())
```
tatiana, line 48: Why did we introduce this and other telemetry-related changes as part of this PR?

cosmos/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "1.10.0"
```
tatiana, line 1: Since we changed the approach, does this change still make sense? What are the benefits of having the version here and not in `cosmos/__init__.py`?

tatiana (approved):
Hi @pankajkoti, this looks great. Thank you very much for working on this and finding a solution to the initial problem. I'm glad we're moving towards improving Cosmos' memory footprint.

Some minor feedback in-line, and some additional comments:
- It would be great if we could have an example of how to use explicit import paths, complementing the documentation you added
- It's probably worth stating that in Cosmos 2.0, this will become the default behaviour, and we'll remove the existing behaviour of allowing people to import everything from `cosmos`, referencing #1213

cosmos/version.py
pankajkoti: yes, nice catch. Moved it back to init module.

tests/test_telemetry.py
pankajkoti: The new tests that I wrote were affecting these tests for Python version 3.8-3.10, although these succeeded for Python 3.11, 3.12 without this change. I guess since I was manipulating the sys.modules to remove cosmos in the new tests they were affecting here as they references were getting altered. Not sure why it works for later version of Python. I have refactored my newly added tests to undo this change and ensure they are not affected.

DanMawdsleyBA:
I just tried this out but when I made the changes all of dags were not able to render with error below:
image

ImportError: cannot import name 'DbtTaskGroup' from 'cosmos' (/usr/local/airflow/.local/lib/python3.11/site-packages/cosmos/__init__.py)

using version 1.10.1

pankajkoti:
> I just tried this out but when I made the changes all of dags were not able to render with error below: 
> ImportError: cannot import name 'DbtTaskGroup' from 'cosmos' (/usr/local/airflow/.local/lib/python3.11/site-packages/cosmos/**init**.py)
> 
> using version 1.10.1

Hey @DanMawdsleyBA, when you enable this config, you need to use fully qualified imports for all Cosmos classes and methods. For example, check out this sample DAG:
https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/basic_cosmos_dag_full_module_path_imports.py

In your case, for the specific case of `DbtTaskGroup`, the import would need to be from `cosmos.airflow.task_group import DbtTaskGroup` instead


Please also refer to docs regarding this config: https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html#enable-memory-optimised-imports

PR #1773, opened 2025-05-20
Fix Airflow dependencies in the CI

tatiana:
Our CI tests have been quite unstable lately; this PR aims to fix the most recent issues.


pankajkoti (approved):
Approving so that we can rebase this in other PRs if needed.

Issue #1776, opened 2025-05-23
[Bug] AttributeError: '_PythonDecoratedOperator' object has no attribute 'env when using Docs operators


pankajkoti:
### Astronomer Cosmos Version

1.10.0

### What happened?

Cosmos 1.10 complain on using the following operator with a @task decorator

```
DbtDocsGCSOperator(
            dag=context["dag"],
            task_id="dbt_docs_generate",
            install_deps=True,
            dbt_cmd_flags=["--static"],
            project_dir=dbt_project_path,
            profile_config=cosmos_configs["profile_config"],
            should_store_compiled_sql=False,
            connection_id=gcp_conn_id,
            bucket_name=bucket_name,
            folder_dir=repository_name,
            # on_dbt_execute_callback=ingest_dbt_metadata_callable(dbt_metadata_path, profile_name, repository_name),
        ).execute(context=context)  # type: ignore
```

### Relevant log output

```shell
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 762, in _execute_task
    result = _execute_callable(context=context, **execute_callable_kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/taskinstance.py", line 733, in _execute_callable
    return ExecutionCallableRunner(
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/utils/operator_helpers.py", line 252, in run
    return self.func(*args, **kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/baseoperator.py", line 406, in wrapper
    return func(self, *args, **kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/decorators/base.py", line 266, in execute
    return_value = super().execute(context)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/models/baseoperator.py", line 406, in wrapper
    return func(self, *args, **kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 238, in execute
    return_value = self.execute_callable()
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/operators/python.py", line 256, in execute_callable
    return runner.run(*self.op_args, **self.op_kwargs)
  File "/home/airflow/.local/lib/python3.10/site-packages/airflow/utils/operator_helpers.py", line 252, in run
[...]

PR #1661, opened 2025-04-10
Refactor Rendered Task Instance Fields (RTIF) handling for Airflow 2.x and 3.x

pankajkoti:
## Description
This PR refactors how we handle Rendered Task Instance Fields (RTIF) across different Airflow versions, improving code organisation and version-specific behaviour handling.

## Changes
- Extracted RTIF handling logic into a dedicated `_override_rtif` method
- Split version-specific behavior:
  - Airflow 2.x: Maintains existing behavior of updating RTIFs in the database
  - Airflow 3.x: Sets a flag (`overwrite_rtif_after_execution`) for later processing. Airflow 3 no longer allows making direct database calls as we do we with Airflow 2.x for supporting the same. Airflow 3 nicely adds this `overwrite_rtif_after_execution` which magically handles this for us.
- Removed nested function definitions for better code organization
- Updated test structure to properly test both Airflow versions
- Simplified `store_compiled_sql` and `store_freshness_json` methods by removing nested functions
- Updated related test files to accommodate the new structure

## Impact
This change improves code maintainability and makes version-specific behaviors more explicit, while maintaining the same functionality for end users. No changes are required from users of the library.

Screenshot: RTIF compiled_sql populating and working in Airflow 3
Screenshot 2025-04-09 at 9 55 53 PM

As you can see, it's not prettily printing the SQL field, but I have filed an [issue](https://github.com/apache/airflow/issues/49064) requesting to natively support this in Airflow

## Related Issues
closes: #1636

PR #1777, opened 2025-05-23
Fix RTIF override logic when should_store_compiled_sql=False to restore pre-refactor behaviour

pankajkoti:
PR #1661 introduced a breaking change for standalone Cosmos operators when users set `should_store_compiled_sql=False`, as detailed in issue #1776.

The issue stems from a refactor in PR #1661. Previously, the RTIF refresh calls to the database were skipped because they occurred within the same function that checked the flag. After the refactor, these operations were moved to a separate `_override_rtif` function, but the `should_store_compiled_sql` condition was not checked there. As a result, the operator behaviour broke, as observed in the issue.

To restore the original behaviour, this change adds a condition in the override_rtif function to exit early when should_store_compiled_sql is False, ensuring consistency with the pre-refactor logic.

Closes: #1776


tatiana (approved):
Excellent work at supporting the community and fixing this so quickly, @pankajkoti ! Did @tuantran0910 have a chance to confirm it solved the original issue?

pankajkoti:
>Did @tuantran0910 have a chance to confirm it solved the original issue?

Yes, @tuantran0910 confirmed that the fix worked for them in https://apache-airflow.slack.com/archives/C059CC42E9W/p1748014917886709?thread_ts=1747031958.683949&cid=C059CC42E9W

Issue #1417, opened 2024-12-21
[Feature] Add ProfileConfig support for GCP Cloud Job Run Execution Mode


indraneel07:
### Description

Currently, ProfileConfigs are not supported with GCP Cloud Job Run Execution Mode. We can enable this similar to what was done in https://github.com/astronomer/astronomer-cosmos/pull/1347 for Docker Execution Mode.
![Image](https://github.com/user-attachments/assets/97002c53-8777-49fc-b2b0-76c5080cba6b)

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

pankajastro:
Thank you for submitting this feature request. I’ve added it to the 1.9.0 milestone for now.

PR #1783, opened 2025-05-27
fix: add support for `ProfileConfig` in gcp cloud run job execution mode

ramonvermeulen:
## Description

I am currently running into an issue with the cloud run job operator where the profiles and target is not picked-up, my config:

The Image contains a `profiles.yml` in the working dir, see [precedence connection profiles type 3](https://docs.getdbt.com/docs/core/connect-data-platform/connection-profiles#advanced-customizing-a-profile-directory).

```python
from airflow import DAG
from cosmos import DbtTaskGroup, ExecutionConfig, ProfileConfig, ProjectConfig
from cosmos.constants import ExecutionMode
from pendulum import datetime

GCP_PROJECT_ID = "XXX"
GCP_LOCATION = "XXX"
GCP_CLOUD_RUN_JOB_NAME = "XXX"

project_config = ProjectConfig(
    manifest_path="gs://XXXXXXXXXXX/manifest.json",
    manifest_conn_id="google_cloud_default",
    project_name="my_dbt",
)

    execution_mode=ExecutionMode.GCP_CLOUD_RUN_JOB,
    dbt_project_path="./",
)

    target_name="cosmos",
    profile_name="my_dbt_taskgroup",
    profiles_yml_filepath="./profiles.yml",
)

with DAG(
    dag_id="my_dbt_dag_cosmos",
    start_date=datetime(2025, 1, 1),
    schedule="0 0 * * *",
    catchup=False,
) as dag:
    pre_dbt_workflow = EmptyOperator(task_id="pre_dbt_workflow")

    my_dbt_taskgroup = DbtTaskGroup(
        group_id="my_dbt_taskgroup",
        project_config=project_config,
        execution_config=execution_config,
        operator_args={
            "project_id": GCP_PROJECT_ID,
            "region": GCP_LOCATION,
            "job_name": GCP_CLOUD_RUN_JOB_NAME,
        },
        default_args={"retries": 0},
        dag=dag,
    )

    post_dbt_workflow = EmptyOperator(task_id="post_dbt_workflow")

pre_dbt_workflow >> my_dbt_taskgroup >> post_dbt_workflow
```

Logs in Airflow:
```
[2025-05-27, 10:10:40 CEST] {gcp_cloud_run_job.py:127} INFO - Running command: ['dbt', 'run', '--models', 'stg_xxx']
```

The `--target` and `--profile` are not set while using a `profile_config` with the `ExecutionMode.GCP_CLOUD_RUN_JOB`.

Took the example from:
https://github.com/astronomer/astronomer-cosmos/blob/a2188226aecaacf6ad5d3c933fecb6b2f7938b3c/cosmos/operators/kubernetes.py#L132

For me it even feels like this maybe should be handled in the [base class](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/base.py#L30), but there is probably a reason why that is not the case.

https://github.com/astronomer/astronomer-cosmos/blob/a2188226aecaacf6ad5d3c933fecb6b2f7938b3c/cosmos/operators/base.py#L265

## Related Issue(s)

Closes https://github.com/astronomer/astronomer-cosmos/issues/1417

## Breaking Change?
My feeling says it should not be breaking, given it only adds functionality and doesn't modify and or removes functionality (maybe `--project-dir` does, given I had to change the test, good to double check this). 

However, I don't know the code base that good, so please verify when reviewing.

## Checklist

### DIFF
--- a/tests/operators/test_gcp_cloud_run_job.py
+++ b/tests/operators/test_gcp_cloud_run_job.py
@@ -50,6 +50,7 @@ PROFILE_CONFIG = ProfileConfig(
     profiles_yml_filepath="./profiles.yml",
 )
 
+
 def skip_on_empty_operator(test_func):
     """
     Skip the test if DbtGcpCloudRunJob operators couldn't be imported.
@@ -222,6 +223,7 @@ def test_dbt_gcp_cloud_run_job_build_command():
                 "--project-dir",
                 "my/dir",
             ]
+
 
 @skip_on_empty_operator
 def test_dbt_gcp_cloud_run_job_build_command_with_profile_config():


tatiana:
@ramonvermeulen thanks a lot for working on this. I agree with your analysis that ideally this would be the behaviour in the base class, perhaps that was not the case because originally there weren't so many execution modes. After we merge this fix, would you like to create a separate PR to try to generalise this solution, so other execution modes could leverage this?

ramonvermeulen:
> @ramonvermeulen thanks a lot for working on this. I agree with your analysis that ideally this would be the behaviour in the base class, perhaps that was not the case because originally there weren't so many execution modes. After we merge this fix, would you like to create a separate PR to try to generalise this solution, so other execution modes could leverage this?

Thanks, no problem, it is something I came across because we are trying out cosmos and have a lot of different targets on our `profiles.yml`.

Sure, I am fine doing a PR (probably later this week) to resolve this onto the base class. Want to spent some more time on that because I am not sure what the implications will be. Probably a lot of unit tets will fail, and want to be sure that it isn't going to be a breaking change.

I wrote unit tests for this PR and did the implementation, however, I did not yet test it on a real Airflow instance yet. Will update my comment once I confirmed also via "manual" testing, that my proposed change actually solves the issue.

ramonvermeulen:
When testing locally, at least I see the correct arguments are provided:

image

ramonvermeulen:
Small update, tested this with `v1.11.0a1` in a GCP Composer instance, and it seems to work fine now!

Issue #1728, opened 2025-04-30
[Bug] DAG owner becomes concatenation of model owners


linchun3:
### Astronomer Cosmos Version

1.9.2

### dbt-core version

1.8.1

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

2.9.3

### Operating System

macOS

### Deployment

Other

### What happened?

Since the changes introduced in PR #1082 (https://github.com/astronomer/astronomer-cosmos/pull/1082), the owner attribute of the generated Airflow DAG appears to be set to a concatenation (or possibly a list formatted as a string) of all unique owners specified in the meta.owner field of the dbt models processed by Cosmos.

This overrides the owner specified via `DEFAULT_ARGS` passed to the parent Airflow DAG when using `DbtTaskGroup`, or potentially passed directly during instantiation.

### Observed behavior
When Cosmos processes multiple dbt models within a DbtTaskGroup, each having a meta: `{owner: 'model_owner_x'}` tag, the resulting Airflow DAG's owner field becomes a combined string like `'model_owner_1, model_owner_2, model_owner_3'`. This happens even if the parent DAG has `default_args={"owner": "specific_dag_owner"}` defined.

Expected behavior
The DAG owner should be configurable independently of the individual model owners. Specifically:

    The owner defined in the DAG's default_args (e.g., DAG(..., `default_args={"owner": "my_dag_owner"}`)) should be respected and assigned to the DAG.
    If no owner is provided in the DAG's default_args, the behavior might default to Airflow's standard behavior (often airflow). The concatenation of model owners should ideally not be the default or only behavior without explicit configuration.

Impact
This change negatively impacts organizations where DAG ownership and model ownership are distinct roles. For example:

    A central platform/foundation team might own the Airflow DAG execution (owner: 'foundation_team').
    Many different analysts or teams might own the individual dbt models (owner: 'analyst_a', owner: 'team_b', etc.).

With the current behavior, the Airflow UI shows a potentially very long list of owners (e.g., hundreds) for the DAG, making it difficult to identify the true DAG operational owner and cluttering the UI.

### Relevant log output

```shell

```

### Steps to Reproduce

    Define multiple dbt models, each with a different value in meta: {owner: ...}.
        Example model 1: meta: {owner: 'team_alpha'}
        Example model 2: meta: {owner: 'team_beta'}
        Example model 3: meta: {owner: 'data_science'}

[...]

PR #1787, opened 2025-05-27
Allow to disable owner inheritance from dbt into airflow DAG owners

CorsettiS:
## Description

From slack
`Hey everyone! :wave: I'm working on an Airflow DAG that uses DbtTaskGroup (Cosmos v1.10.1). I've noticed that the Airflow tasks are showing both the DAG owner (set in default_args) and the owner defined in my dbt model's _doc_.yml. Is there a way to configure Cosmos to not pull the owner from the dbt model? I'm aiming to keep the DAG owner clean. Any insights would be greatly`

This PR introduces the flag `disable_owner_inheritance` which should make it possible

## Breaking Change?

No
## Checklist

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -66,6 +66,7 @@ class RenderConfig:
     :param airflow_vars_to_purge_dbt_ls_cache: Specify Airflow variables that will affect the LoadMode.DBT_LS cache.
     :param normalize_task_id: A callable that takes a dbt node as input and returns the task ID. This allows users to assign a custom node ID separate from the display name.
     :param should_detach_multiple_parents_tests: A boolean that allows users to decide whether to run tests with multiple parent dependencies in separate tasks.
+    :param disable_owner_inheritance: A boolean that allows users to disable the owner inheritance from dbt models to airflow tasks. Defaults to False.
     """
 
     emit_datasets: bool = True
@@ -87,6 +88,7 @@ class RenderConfig:
     airflow_vars_to_purge_dbt_ls_cache: list[str] = field(default_factory=list)
     normalize_task_id: Callable[..., Any] | None = None
     should_detach_multiple_parents_tests: bool = False
+    disable_owner_inheritance: bool | None = False
 
     def __post_init__(self, dbt_project_path: str | Path | None) -> None:
         if self.env_vars:
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -66,7 +66,7 @@ class RenderConfig:
     :param airflow_vars_to_purge_dbt_ls_cache: Specify Airflow variables that will affect the LoadMode.DBT_LS cache.
     :param normalize_task_id: A callable that takes a dbt node as input and returns the task ID. This allows users to assign a custom node ID separate from the display name.
     :param should_detach_multiple_parents_tests: A boolean that allows users to decide whether to run tests with multiple parent dependencies in separate tasks.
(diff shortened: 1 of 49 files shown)


corsettigyg:
- For reference, I created the `disable_owner_inheritance` to avoid making it into a breaking change, although I particularly believe that the default behaviour should be `enable_owner_inheritance` and this one should be set to False by default. To avoid headaches, I kept things as they are though.

- Since the owners are being propagated in many places and now it is possible to disable/enable this behaviour, creating the tests is no pleasant task since it is a very repetitive process if we want to hit the coverage on the changed lines for all situations. I ended up using AI to automate most of the test creation since it would be very time consuming to do it manually. On a quick look the tests looked fine to me, but as a traditional data engineer I might still take a better look later once I have more time 😄

docs/configuration/render-config.rst
```diff
@@ -24,6 +24,7 @@ The ``RenderConfig`` class takes the following arguments:
 - ``source_rendering_behavior``: Determines how source nodes are rendered when using cosmos default source node rendering (ALL, NONE, WITH_TESTS_OR_FRESHNESS). Defaults to "NONE" (since Cosmos 1.6). See `Source Nodes Rendering <./source-nodes-rendering.html>`_ for more information.
 - ``normalize_task_id``: A callable that takes a dbt node as input and returns the task ID. This function allows users to set a custom task_id independently of the model name, which can be specified as the task’s display_name. This way, task_id can be modified using a user-defined function, while the model name remains as the task’s display name. The display_name parameter is available in Airflow 2.9 and above. See `Task display name <./task-display-name.html>`_ for more information.
 - ``should_detach_multiple_parents_tests``: A boolean to control if tests that depend on multiple parents should be run as standalone tasks. See `Parsing Methods <testing-behavior.html>`_ for more information.
+- ``disable_owner_inheritance``: A boolean to control if dbt owners should be imported as part of the airflow DAG owners. Defaults to False.
```
pankajkoti, line 27:
```suggestion
- ``disable_owner_inheritance``: A boolean to control if dbt owners should not be imported as part of the airflow DAG owners. Defaults to False.
```

cosmos/config.py
```diff
@@ -87,6 +88,7 @@ class RenderConfig:
     airflow_vars_to_purge_dbt_ls_cache: list[str] = field(default_factory=list)
     normalize_task_id: Callable[..., Any] | None = None
     should_detach_multiple_parents_tests: bool = False
+    disable_owner_inheritance: bool | None = False
```
pankajkoti, line 91: wondering if naming it `disable_dbt_owner_inheritance` could make it clearer.

pankajkoti (approved):
LGTM. couple of minor suggestions in-line

cosmos/config.py
corsettigyg: hm since cosmos already refers to dbt only, Idk if adding `dbt` would be redundant or not.  I particularly like 3 words for a param name, but I have no strong preference for any option TBH

linchun3:
Thanks for this PR! It should help to close https://github.com/astronomer/astronomer-cosmos/issues/1728

Any idea on the timeline for the release of Cosmos 1.11?

tatiana (approved):

tatiana:
Thanks a lot, @corsettigyg ! @linchun3 We'll release this in 1.10.2 (although it can be perceived as a feature, it is also a sort of bug fix for some users). The estimate is to release it on 6 June (Friday)

Issue #1789, opened 2025-05-27
[Refactor] Refactor the `ProfilesConfig` handling for generating the dbt command to the base class


ramonvermeulen:
As described in [this comment](https://github.com/astronomer/astronomer-cosmos/pull/1783#issuecomment-2911947214) it would be nice if the `ProfilesConfig` handling can be moved to the base class. The `ProfilesConfig` is being used to set the `--target`, `--profiles-dir` and `--profile` dbt arguments.

I am creating this issue to be able to discuss the implementation and progress.

PR #1794, opened 2025-05-30
refactor: moved profile_config member to the `AbstractDbtBase` base class

ramonvermeulen:
## Description
The `ProfileConfig` is used to add the correct `--profiles-dir`, `--profile` and `--target` arguments to the build dbt argument. This is something I discussed with @tatiana in https://github.com/astronomer/astronomer-cosmos/pull/1783#issuecomment-2911947214, but I guess the more input/discussion the better, because for me it is hard to see/know all edge cases.

## Related Issue(s)

Closes https://github.com/astronomer/astronomer-cosmos/issues/1789
Related https://github.com/astronomer/astronomer-cosmos/pull/1783

## Breaking Change?
Might be, I don't know the code base well enough to know if this is a breaking change. At least I tried to keep the interface the same for anything inheriting from `AbstractDbtLocalBase` instead of `AbstractDbtBase`, so that the profile_config type stays `ProfileConfig` and not `ProfileConfig | None` as on the `AbstractDbtBase`.

## Checklist

### DIFF
--- a/cosmos/operators/base.py
+++ b/cosmos/operators/base.py
@@ -297,6 +297,7 @@ class AbstractDbtBase(metaclass=ABCMeta):
         env = self.get_env(context)
 
         from cosmos.operators.local import AbstractDbtLocalBase
+
         # AbstractDbtLocalBase has its own implementation for these flags via `_generate_dbt_flags`
         if not type(self) is AbstractDbtLocalBase:
             # Parse ProfileConfig and add additional arguments to the dbt_cmd


tatiana:
Thanks a lot for following up on this, @ramonvermeulen ! Once the checks are passing and we're happy with this refactor, I'll try to loop in users of the execution modes affected so we can collect feedback before releasing this

ramonvermeulen:
> Thanks a lot for following up on this, @ramonvermeulen ! Once the checks are passing and we're happy with this refactor, I'll try to loop in users of the execution modes affected so we can collect feedback before releasing this

Check, I think that is definetly a good idea because for me it is quite hard to grasp the downstream impact. Especially because I see some classes such as: https://github.com/astronomer/astronomer-cosmos/blob/d8056626336ba40f619db9f86fc26d0c445977dc/cosmos/operators/local.py#L146

Have their own implementation for generating and adding these profile, target, and project dir related flags, for instance:
https://github.com/astronomer/astronomer-cosmos/blob/d8056626336ba40f619db9f86fc26d0c445977dc/cosmos/operators/local.py#L491

Because it uses some temporary directory logic to clone the project in before execution:
https://github.com/astronomer/astronomer-cosmos/blob/d8056626336ba40f619db9f86fc26d0c445977dc/cosmos/operators/local.py#L600

And quite some classes inherit directly or indirectly from this `AbstractDbtLocalBase`. Maybe it is good to already loop someone into the discussion that has a good overview on the overall core code of cosmos to see if this refactor makes sense?

I am currently busy with a lot of things at the same time, but when I have some free time I will try to take another look at this.

PR #1796, opened 2025-06-03
Upgrade Python and Airflow used to run MyPy checks

tatiana:


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, mypy-fix]
```
```suggestion
    branches: [main]
```

PR #1801, opened 2025-06-09
Add support to base64 encoded pem in Snowflake profiles

### LINKED ISSUE #1798: [Bug] Snowflake Provider 6.3 breaking change broke `private_key_content`
### Astronomer Cosmos Version

Cosmos 1.10.1 (and previous versions)

### dbt-core version

Any

### Versions of dbt adapters

Any

### LoadMode

AUTOMATIC

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

Any

### Operating System

Any

_No response_

### Deployment

Other

### Deployment details

Irrelevant

### What happened?

Starting from `apache-airflow-providers-snowflake` [version 6.3](https://airflow.apache.org/docs/apache-airflow-providers-snowflake/stable/changelog.html#id2), there was a change in how `private_key_content` is handled (from plain text to base64 encode).

Users are currently hitting the following error when running with Cosmos:
```
Could not deserialize key data. The data may be in an incorrect format...
```
This is the exact error they've seen when the `private_key_content` is malformed.

Cosmos currently does not support the updated key format introduced in 6.3.

Until this is fixed, users should pin the Snowflake provider.

Since it was a security fix, I don't think this breaking change will be reverted:
https://github.com/apache/airflow/issues/47003

### Relevant log output

```shell
Could not deserialize key data. The data may be in an incorrect format...
```

### How to reproduce

Instantiate a profile that uses `private_key_content` as described in:
https://astronomer.github.io/astronomer-cosmos/profiles/SnowflakePrivateKeyPem.html

### Anything else :)?

No

### Contact Details

_No response_

brunocmartins:
## Description

Starting from `apache-airflow-providers-snowflake` version 6.3, the [provider expects the
`private_key_content` field to be base64 encoded](https://airflow.apache.org/docs/apache-airflow-providers-snowflake/stable/changelog.html#id2) instead of plain text PEM. This change
adds decoding of base64 encoded keys while preserving backwards compatibility with older
versions that use plain text PEM.

## Related Issue(s)

Closes #1798

## Breaking Change?

None

## Checklist

### DIFF
--- a/cosmos/profiles/snowflake/user_encrypted_privatekey_env_variable.py
+++ b/cosmos/profiles/snowflake/user_encrypted_privatekey_env_variable.py
@@ -68,6 +68,10 @@ class SnowflakeEncryptedPrivateKeyPemProfileMapping(SnowflakeBaseProfileMapping)
         if conn_dejson.get("extra__snowflake__account"):
             conn_dejson = {key.replace("extra__snowflake__", ""): value for key, value in conn_dejson.items()}
 
+        private_key_content = conn_dejson.get("private_key_content")
+        if private_key_content:
+            conn_dejson["private_key_content"] = self._decode_private_key_content(private_key_content)
+
         conn.extra = json.dumps(conn_dejson)
 
         return conn
--- a/cosmos/profiles/snowflake/base.py
+++ b/cosmos/profiles/snowflake/base.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import base64
+import binascii
 from typing import Any
 
 from cosmos.profiles.base import BaseProfileMapping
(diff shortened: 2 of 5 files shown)


cosmos/profiles/snowflake/user_encrypted_privatekey_env_variable.py
```diff
@@ -68,6 +68,10 @@ def conn(self) -> Connection:
         if conn_dejson.get("extra__snowflake__account"):
             conn_dejson = {key.replace("extra__snowflake__", ""): value for key, value in conn_dejson.items()}
 
+        private_key_content = conn_dejson.get("private_key_content")
+        if private_key_content:
+            conn_dejson["private_key_content"] = self._decode_private_key_content(private_key_content)
```
pankajastro, line 73: Do we know why changes were made on the Airflow provider side? Should we make our changes backwards-compatible if the old provider allows the key to work without encoding? That is, apply the new logic only if the user has installed the updated provider. WDYT?

brunocmartins: The change were made due to how Airflow handles sensitive fields as a one-liner field (https://github.com/apache/airflow/pull/49467). From provider 6.3 on, it will only support base64 encoded private key contents. I opted on adding backwards compatibility to avoid adding new provider requirements (`apache-airflow-providers-snowflake>=6.3`) in further Cosmos versions

PR #1821, opened 2025-06-20
Kubernetes Pod Operator convert `container_resources`

johnhoran:
## Description

Slightly odd and I'm not sure if this is the best solution, but this is the code base I have access to, so lets go with it.

I'm trying to specify resource limits in cosmos using https://astronomer.github.io/astronomer-cosmos/configuration/operator-args.html#overriding-operator-arguments-per-dbt-node-or-group-of-nodes, specifically I'm adding
```yaml
    meta:
      cosmos:
        operator_kwargs:
          container_resources:
            limits:
              memory: "2000Mi"
```
to a DBT project and when running locally using `astro dev start` this works correctly.  However when running in astronomer hosted, it doesn't apply.  Researching further if I have an operator with
```
KubernetesPodOperator(
        ...
        container_resources={
            "limits": {"cpu": "100m", "memory": "100Mi", "ephemeral-storage": "512Mi"},
            "requests": {
                "cpu": "100m",
                "memory": "100Mi",
                "ephemeral-storage": "512Mi",
            },
        },
    )
```
then the resources will be overridden with the defaults set on the deployment in astronomer hosted, though running locally it works correctly.  If I specify them as 
```
KubernetesPodOperator(
        ...
        container_resources=k8s.V1ResourceRequirements(
            limits={"cpu": "100m", "memory": "100Mi", "ephemeral-storage": "512Mi"},
            requests={"cpu": "100m", "memory": "100Mi", "ephemeral-storage": "512Mi"},
        ),
    )
```
then it all works correctly, but this precludes my ability to set them in cosmos as the argument will always be passed as a dictionary there.

It seems whatever code astronomer is using on the backend to apply the default container resources doesn't work with anything except `V1ResourceRequirements`.  Arguably this is correct behaviour since the signature in the kubernetes pod operator specifies the type to be `V1ResourceRequirements | None`, even though the code itself does work with dictionaries.

Anyway the proposed solution is to convert this argument to the correct type before it is passed to the kubernetes pod operator.  Thoughts?

### DIFF
--- a/tests/operators/test_kubernetes.py
+++ b/tests/operators/test_kubernetes.py
@@ -483,7 +483,7 @@ def test_kubernetes_pod_container_resources():
     b_operator = DbtRunOperationKubernetesOperator(
         task_id="run_macro_command",
         macro_name="macro",
-        project_dir = DBT_ROOT_PATH / "jaffle_shop",
+        project_dir=DBT_ROOT_PATH / "jaffle_shop",
         container_resources=k8s.V1ResourceRequirements(**resources),
     )
     assert isinstance(b_operator.container_resources, k8s.V1ResourceRequirements)
@@ -493,6 +493,6 @@ def test_kubernetes_pod_container_resources():
     c_operator = DbtRunOperationKubernetesOperator(
         task_id="run_macro_command",
         macro_name="macro",
-        project_dir = DBT_ROOT_PATH / "jaffle_shop",
+        project_dir=DBT_ROOT_PATH / "jaffle_shop",
     )
     assert c_operator.container_resources is None


johnhoran:
> Hi @johnhoran thank you very much for raising this and creating the PR so things work in Astro. Which type of executor are you using? Where is the k8s cluster you're instantiating these pods?

Hi @tatiana, We are running in astronomer 13 in a dedicated hosted AWS cluster using the kubernetes executor.

tatiana:
@johnhoran I discussed this matter with fellow Astronomers, and @kaxil raised this question:
> Did you (or someone from your team) set some defaults in Astro UI as defined in https://www.astronomer.io/docs/astro/deployment-resources/#configure-kubernetes-pod-resources ?

This may justify the behaviour you're seeing

johnhoran:
> @johnhoran I discussed this matter with fellow Astronomers, and @kaxil raised this question:
> 
> > Did you (or someone from your team) set some defaults in Astro UI as defined in https://www.astronomer.io/docs/astro/deployment-resources/#configure-kubernetes-pod-resources ?
> 
> This may justify the behaviour you're seeing

As far as I know it isn't possible to create a deployment without the defaults set.  I guess the real question is if this behaviour is intentional or not, like I said, it is arguably correct.

PR #1822, opened 2025-06-22
fix: dependencies installation with project level variables

AlexandrKhabarov:
## Description

Failed at runtime phase during dependencies installation if project level variables are used
```
dbt deps \
    --target $FLOPPA_TARGET \
    --profiles-dir $FLOPPA_FOLDER \
    --project-dir $FLOPPA_FOLDER
14:43:11  Running with dbt=1.8.9
14:43:11  Encountered an error:
Compilation Error
  Could not render ingestion_date = '{{ var("ingestion_date") }}': Required var 'ingestion_date' not found in config:
  Vars supplied to <Configuration> = {}
```

## Breaking Change?

No breaking changes

## Checklist

### DIFF
--- a/cosmos/operators/kubernetes.py
+++ b/cosmos/operators/kubernetes.py
@@ -98,6 +98,10 @@ class DbtKubernetesBaseOperator(AbstractDbtBase, KubernetesPodOperator):  # type
                     pass
 
         AbstractDbtBase.__init__(self, **base_kwargs)
+
+        container_resources = operator_kwargs.get("container_resources")
+        if isinstance(container_resources, dict):
+            operator_kwargs["container_resources"] = k8s.V1ResourceRequirements(**container_resources)
         KubernetesPodOperator.__init__(self, **operator_kwargs)
 
     def build_env_args(self, env: dict[str, str | bytes | PathLike[Any]]) -> None:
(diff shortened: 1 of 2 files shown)


AlexandrKhabarov:
@tatiana can you take a look, please?

AlexandrKhabarov:
Hello @tatiana 

I think that such test exists - https://github.com/astronomer/astronomer-cosmos/blob/main/tests/dbt/test_graph.py#L1837

I am using LoadMode.AUTOMATIC. DBT_LS is part of it

AlexandrKhabarov:
Also, I created a PR in a past to fix problem with parsing - https://github.com/astronomer/astronomer-cosmos/pull/1543

tatiana:
Thank you very much, @AlexandrKhabarov !

The test  https://github.com/astronomer/astronomer-cosmos/blob/main/tests/dbt/test_graph.py#L1837 covers the issue in the DAG parsing/rendering while converting during the `dbt ls` command execution - so apparently the problem was already solved there.

Since you fixed the behaviour in the local operator execution, it may be worth adapting or adding a test in https://github.com/astronomer/astronomer-cosmos/blob/main/tests/operators/test_local.py
To make sure we don't cause any regressions, and this fix remains there.

It could be a very similar test to what you did in the test `tests/operators/test_virtualenv.py` validates that `--vars` were not set. In the case of the local operator, we'd like to confirm that we're setting the correct arguments when trying to run deps from there.

PR #1860, opened 2025-07-04
docs(scheduling.rst): fix character name

ThePsyjo:
The example shows `/` which is a slash

## Description

The text says 'backslash' which is `\` and the example uses `/` which is a 'slash' for 'forward slash'

## Related Issue(s)

-

## Breaking Change?

-

## Checklist


tatiana (approved):
Thanks for the fix, @ThePsyjo !

PR #1864, opened 2025-07-09
[DRAFT FOR REFERENCE] Enable dbt show in Cosmos

iancmoritz:
## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -704,6 +704,27 @@ def build_airflow_graph(
             test_task = create_airflow_task(test_meta, dag, task_group=task_group)
             tasks_map[node_id] = test_task
 
+    import os
+
+    include_show_local_task = os.getenv("COSMOS__INCLUDE_SHOW_LOCAL_TASK", "false").lower() in ("true", "1", "t")
+    # Add DbtShowLocalOperator to the DAG
+    if include_show_local_task and execution_mode in (ExecutionMode.LOCAL, ExecutionMode.VIRTUALENV):
+        from cosmos.operators.local import DbtShowLocalOperator
+        from cosmos.io import log_to_xcom
+
+        show_task_id = f"{dbt_project_name}__show"
+        show_operator = DbtShowLocalOperator(
+            task_id=show_task_id,
+            project_dir=task_args.get("project_dir", ""),
+            profile_config=task_args.get("profile_config"),
+            callback=log_to_xcom,
+            inline="select * from stg_customers where first_name = 'Jesus'",
+            dag=dag,
+            task_group=task_group,
+            **{k: v for k, v in task_args.items() if k not in ["task_id", "profile_config", "project_dir"]},
+        )
+        tasks_map[show_task_id] = show_operator
+
     create_airflow_task_dependencies(nodes, tasks_map)
     if settings.enable_setup_async_task:
         _add_dbt_setup_async_task(
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -706,19 +706,19 @@ def build_airflow_graph(
 
     import os
 
-    include_show_local_task = os.getenv("COSMOS__INCLUDE_SHOW_LOCAL_TASK", "false").lower() in ("true", "1", "t")
+    include_show_local_task = os.getenv("COSMOS__RENDER_DBT_SHOW_LOCAL_TASK", "false").lower() in ("true", "1", "t")
     # Add DbtShowLocalOperator to the DAG
     if include_show_local_task and execution_mode in (ExecutionMode.LOCAL, ExecutionMode.VIRTUALENV):
         from cosmos.operators.local import DbtShowLocalOperator
         from cosmos.io import log_to_xcom
 
-        show_task_id = f"{dbt_project_name}__show"
+        show_task_id = f"cosmos__dbt_show_local"
         show_operator = DbtShowLocalOperator(
             task_id=show_task_id,
             project_dir=task_args.get("project_dir", ""),
             profile_config=task_args.get("profile_config"),
             callback=log_to_xcom,
-            inline="select * from stg_customers where first_name = 'Jesus'",
+            inline="{{ var.value.COSMOS__DBT_SHOW_LOCAL_INLINE_QUERY }}",
             dag=dag,
             task_group=task_group,
             **{k: v for k, v in task_args.items() if k not in ["task_id", "profile_config", "project_dir"]},
--- a/cosmos/io.py
+++ b/cosmos/io.py
@@ -87,6 +87,39 @@ def upload_to_gcp_gs(
             )
 
 
+def _extract_show_list(log_string: str) -> list:
+    """
+    Extracts the JSON list after "show" key from a log string.
+
+    TODO: This is not something we'll actually use, but it's useful for PoC purposes.
+    """
+    import re
+    import json
+
+    # Regex to find: "show": <whitespace> <JSON array starting with [ and ending with ]>
+    pattern = r'"show"\s*:\s*(\[[^\]]*\])'
+
+    # If you might have nested { } inside, you can make it lazy with:
+    # r'"show"\s*:\s*(\[[\s\S]*?\])'
+    # which is more greedy but more flexible for multiple lines
+
+    match = re.search(pattern, log_string, re.DOTALL)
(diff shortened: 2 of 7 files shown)


cosmos/airflow/graph.py
```diff
@@ -704,6 +704,27 @@ def build_airflow_graph(
             test_task = create_airflow_task(test_meta, dag, task_group=task_group)
             tasks_map[node_id] = test_task
 
+    import os
+
+    include_show_local_task = os.getenv("COSMOS__RENDER_DBT_SHOW_LOCAL_TASK", "false").lower() in ("true", "1", "t")
```
tatiana, line 709: We should leverage Airlfow config and keep this sort of definition inside the `settings.py` module, following similar configurations in Cosmos:
* https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/settings.py
* https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html

cosmos/airflow/graph.py
```diff
@@ -704,6 +704,27 @@ def build_airflow_graph(
...
+            project_dir=task_args.get("project_dir", ""),
+            profile_config=task_args.get("profile_config"),
+            callback=log_to_xcom,
+            inline="{{ var.value.COSMOS__DBT_SHOW_LOCAL_INLINE_QUERY }}",
+            dag=dag,
+            task_group=task_group,
+            **{k: v for k, v in task_args.items() if k not in ["task_id", "profile_config", "project_dir"]},
+        )
+        tasks_map[show_task_id] = show_operator
+
```
tatiana, line 727: It is unclear why we should change `DbtDag` and `DbtTaskGroup` to expose `DbtShowLocalOperator`. Users should control if and how many of these operators to instantiate directly in their Dags and TaskGroups.

If we decide to add `DbtShowLocalOperator` "automatically" to `DbtDag` and `DbtTaskGroup`, it likely makes more sense to have it by the end of the DAG, as a leaf node, since it would be used to query models and run inline SQL associated to the run transformations.

cosmos/io.py
```diff
@@ -68,31 +68,110 @@ def upload_to_gcp_gs(
...
-                f"/{context['run_id']}"
-                f"/{context['task_instance'].task_id}"
-                f"/{context['task_instance']._try_number}"
-                f"{dirpath.split(project_dir)[-1]}/{filename}"
-            )
+    # Get all files in target directory
+    for root, _, files in os.walk(target_dir):
+        for file in files:
+            file_path = os.path.join(root, file)
+            destination_file_path = os.path.relpath(file_path, os.path.join(project_dir, os.pardir))
```
tatiana, line 83: Nice refactoring! Since it is unrelated to the objective of this PR (enable dbt show in Cosmos), let's have it in a separate PR.

cosmos/io.py
```diff
@@ -68,31 +68,110 @@ def upload_to_gcp_gs(
...
+    # Get the full path to the log file
+    log_path = Path(project_dir) / log_relative_path
+
+    # Read the log file if it exists
+    log_content = ""
+    if log_path.exists():
+        try:
+            with open(log_path, encoding="utf-8") as file:
+                log_content = file.read()
+        except Exception as error:
```
tatiana, line 163: In the line above, it means that we will catch everything. Including exceptions that we cannot or should not recover from.

More on exception handling in Python: https://jerrynsh.com/stop-using-exceptions-like-this-in-python/.

cosmos/io.py
```diff
@@ -68,31 +68,110 @@ def upload_to_gcp_gs(
...
+
+
+def log_to_xcom(
+    project_dir: str,
+    log_relative_path: str = "logs/dbt.log",
+    xcom_key: str = "dbt_logs",
+    **kwargs: Any,
+) -> None:
+    """
+    Callback function that reads a log file and pushes its contents to XCom.
```
tatiana, line 131: Why are we parsing the log file instead of the dbt standard output?

dev/dags/user_defined_profile.py
```diff
@@ -9,7 +9,8 @@
 from airflow import DAG
```
tatiana, line 9: Probably https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/example_operators.py#L70-L77 would be a more suitable location for this example, as the show operator does not have any particular requirement for end users to give the profiles YAML.

cosmos/operators/base.py
```diff
@@ -482,6 +482,39 @@ class DbtCompileMixin:
...
+        **kwargs: Any,
+    ) -> None:
+        self.inline = inline
+        super().__init__(**kwargs)
+
+    def add_cmd_flags(self) -> list[str]:
+        flags = []
+
+        flags.extend(["--output", "json"])
+        # flags.extend(["--log-format-file", "json"])
```
tatiana, line 510: Why is this line commented?

cosmos/operators/base.py
```diff
@@ -482,6 +482,39 @@ class DbtCompileMixin:
...
+    Mixin for dbt show command.
+
+    :param inline: SQL query to execute
+    """
+
+    base_cmd = ["show"]
+    ui_color = "#6C8EBF"  # A muted blue color
+    ui_fgcolor = "#FFFFFF"
+
+    template_fields: Sequence[str] = ("inline",)
```
tatiana, line 496: Maybe we should also include the show [`--limit` ](https://docs.getdbt.com/reference/commands/show)flag and allow it to be templated

cosmos/io.py
```diff
@@ -68,31 +68,110 @@ def upload_to_gcp_gs(
...
+        except Exception as error:
+            log_content = f"Error reading log file {log_path}: {str(error)}"
+    else:
+        log_content = f"Log file not found: {log_path}"
+
+    context = get_current_context()
+
+    # Next, retrive the JSON from the plaintext log file. The JSON always is contained in a key {"show": [...]}.
+    json_content = _extract_show_list(log_content)
+    context["ti"].xcom_push(key=xcom_key, value=json_content)
```
tatiana, line 172: How big can this content be? It would be interesting have an idea of the sizes so we don't trash the Airflow metadata database.

dev/dags/user_defined_profile.py
```diff
@@ -45,6 +46,21 @@
...
+            target_name="dev",
+            profiles_yml_filepath=PROFILES_FILE_PATH,
+        ),
+        project_dir=DBT_ROOT_PATH / "jaffle_shop",
+        task_id="show_dbt",
+        callback=log_to_xcom,
+        inline="{{ var.value.COSMOS__DBT_SHOW_LOCAL_INLINE_QUERY }}",
+        install_deps=True,
+        append_env=True,
+        quiet=True,
```
tatiana, line 61: What would be the advantage of using `quiet=True` with the show operator? The logs are usually quite important, particularly for this command, when people are interacting with Airflow.

cosmos/operators/base.py
```diff
@@ -482,6 +482,39 @@ class DbtCompileMixin:
...
+
+    def __init__(
+        self,
+        inline: str | None = None,
```
tatiana, line 500: When we add tests, we should also make sure that `--select` works also

tatiana (commented):
Hey @iancmoritz , this is a very exciting feature! Thanks a lot for the proposal.

A few comments here - others follow inline:

1. What is the benefit of changing the interface of `DbtDag` and `DbtTaskGroup`? In the same way end-users can leverage Cosmos by instantiating operators directly (example: https://github.com/astronomer/astronomer-cosmos/blob/460e58c18bf5fc95cdb90a110193206fda7936cd/dev/dags/example_operators.py#L70-L77), they could instantiate the `DbtShowLocalOperator`. It feels like this could just be an operator in Cosmos, and end-users decide when and if to instantiate it. My first impression is that the change in `cosmos/airflow/graph.py` is unecessary.

2. How do you see this ported to other Cosmos execution modes? Currently Cosmos has [nine execution modes](https://astronomer.github.io/astronomer-cosmos/getting_started/execution-modes.html) in place. Do you see any problems or limitations in trying to implement this for others?

3. I understand this PR is in draft mode, but since you asked for feedback, are you planning to add tests and docs as well? This is a requirement so we can merge this feature

joppevos:
I'm curious when you'd want to use `dbt show` as a task on Airflow. What's the use-case?
I like `dbt show` locally to debug and iterate on models

PR #1807, opened 2025-06-18
Add  Airflow 2.11 to the test matrix

tatiana:

PR #1872, opened 2025-07-16
CI: Add back accidentally deleted python-version matrix for running unit tests

pankajkoti:
Looks like the CI is failing to run tests after having accidentally dropped the matrix entries for python-version in https://github.com/astronomer/astronomer-cosmos/pull/1807

This came to my attention while I'm working on another PR whose action run complained about this in https://github.com/astronomer/astronomer-cosmos/actions/runs/16320855168

We are getting a yml validation error for the missing matrix key. The PR fixes this issue by adding back the deleted matrix entry.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main,add-python-version-matrix-unit-tests]
+    branches: [main]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,add-python-version-matrix-unit-tests]
```
```suggestion
    branches: [main]
```

tatiana (approved):
Thanks a lot for fixing this, @pankajkoti !

Issue #1873, opened 2025-07-16
Fix failures in the CI


pankajkoti:
Refer Action run: https://github.com/astronomer/astronomer-cosmos/actions/runs/16322249943/ to fix few failures we've in the CI

pankajkoti:
We have merged the following PRs to fix the various issues in the CI:
1. https://github.com/astronomer/astronomer-cosmos/pull/1876
2. https://github.com/astronomer/astronomer-cosmos/pull/1877
3. https://github.com/astronomer/astronomer-cosmos/pull/1880


CI is now greee on commit https://github.com/astronomer/astronomer-cosmos/commit/782c5f174997306c716ad61df85005045080584a after merging PR https://github.com/astronomer/astronomer-cosmos/pull/1880

Issue #1802, opened 2025-06-10
Renable `source_rendering_dag` and `basic_cosmos_task_group_different_owners` DAGs test once Airflow 3 supports DatasetAlias with non-ASCII characters


tatiana:
Airflow 3.0.0, 3.0.1 and 3.02 do not support DatasetAlias with non-ASCII characters, as reported in:
https://github.com/apache/airflow/issues/51566

During the development of #1778, when we attempted to run

```
hatch run tests.py3.10-3.0-1.9:test-integration
```

In particular:
```
pytest -vv 'tests/test_example_dags.py::test_example_dag[source_rendering_dag]'
```

Failed with the following exception:
```                                                                     │
│ /Users/tati/Code/cosmos-fresh/astronomer-cosmos/cosmos/operators/local.py:793 in                 │
│ register_dataset                                                                                 │
│                                                                                                  │
│    790 │   │   │   │   from airflow.sdk.definitions.asset import AssetAlias                      │
│    791 │   │   │   │                                                                             │
│    792 │   │   │   │   # This line was necessary in Airflow 3.0.0, but this may become automati  │
│ ❱  793 │   │   │   │   self.outlets.append(AssetAlias(dataset_alias_name))  # type: ignore[attr  │
│    794 │   │   │   │   for outlet in new_outlets:                                                │
│    795 │   │   │   │   │   context["outlet_events"][AssetAlias(dataset_alias_name)].add(outlet)  │
│    796                                                                                           │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │            context = {                                                                       │ │
│ │                      │   'dag': <DbtDag                                                      │ │
│ │                      │   │   dag_id='source_rendering_dag'                                   │ │
│ │                      │   │   schedule='@daily'                                               │ │
│ │                      │   │   #tasks=20                                                       │ │
│ │                      │   >,                                                                  │ │
│ │                      │   'inlets': [],                                                       │ │
│ │                      │   'map_index_template': None,                                         │ │
│ │                      │   'outlets': [],                                                      │ │
│ │                      │   'run_id': 'manual__2025-06-10T09:37:09.344569+00:00',               │ │
│ │                      │   'task': <Task(DbtRunLocalOperator): ｍｕｌｔｉｂｙｔｅ_run>,        │ │
│ │                      │   'task_instance': <RuntimeTaskInstance                               │ │
[...]

PR #1877, opened 2025-07-17
Fix dbt Fusion tests: Harden DbtNode against null config/meta 

pankajkoti:
This PR fixes a crash in `DbtNode` when dbt emits `null` for `config` or `meta` and a bit of housekeeping around the demo projects and tests. The fixes are:
1.  Handle `null` `config`/`meta` in `DbtNode` (observed in case of using dbt Fusion)
dbt may output `"config": null` or `"meta": null` for nodes with no explicit config block.
to fix this, updated property accessors:
`meta` — safely coalesces to {}
`owner` — safely coalesces to ""
Parsers (`parse_dbt_ls_output`, `load_from_dbt_manifest`) now create every node with
```
  config=node_dict.get("config") or {}
```
ensuring the field is always a dict.
2. `basic_cosmos_task_group_different_owners` example DAG now targets the `altered_jaffle_shop` project
   - Point `basic_cosmos_task_group_different_owners.py` at the altered_jaffle_shop project (`dev/dags/dbt/altered_jaffle_shop`) instead of the `jaffle_shop` project.
   - In `altered_jaffle_shop/dbt_project.yml`, add a seed-level `meta` block to showcase per-resource profile overrides.
   - Remove the conflicting seed meta from the `jaffle_shop` project (it was using `postgres_profile`, which clashed with the Fusion tests’ `snowflake_profile`)
3. At the same time the PR disables running the `basic_cosmos_task_group_different_owners` example DAG on Airflow 3 due to a known issue #1802 on Airflow 3 wrt DatasetAlias with non-ascii characters (present in the `altered_jaffle_shop` project)

related: #1873 
related: #1802

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -307,7 +307,7 @@ def parse_dbt_ls_output(project_path: Path | None, ls_stdout: str) -> dict[str, 
                     # dbt-core defined the node path via "original_file_path", dbt fusion identifies it via "path"
                     file_path=base_path / (node_dict["original_file_path"] or node_dict.get("path")),
                     tags=node_dict.get("tags", []),
-                    config=node_dict.get("config", {}),
+                    config=node_dict.get("config") or {},
                     has_freshness=(
                         is_freshness_effective(node_dict.get("freshness"))
                         if DbtResourceType(node_dict["resource_type"]) == DbtResourceType.SOURCE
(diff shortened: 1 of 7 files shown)


tatiana (approved):
Thanks a lot for solving the CI issues, @pankajkoti , and restoring our tests.

PR #1878, opened 2025-07-17
Make implementation & docs consistent regarding `use_dataset_airflow3_uri_standard`

Anti0ff:
changed `enable_dataset_airflow3_uri` to `use_dataset_airflow3_uri_standard`

## Description

In [documentation](https://astronomer.github.io/astronomer-cosmos/configuration/scheduling.html) pointed
`export AIRFLOW__COSMOS__USE_DATASET_AIRFLOW3_URI_STANDARD=1`
but in settings used variable `enable_dataset_airflow3_uri` - not `use_dataset_airflow3_uri_standard`

Seems, **or code in settings.py isn't correct, or docs need to be updated** with `AIRFLOW__COSMOS__ENABLE_DATASET_AIRFLOW3_URI`

Otherwise in [local.py](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/local.py#L692C1-L725C27) the condition `if settings.use_dataset_airflow3_uri_standard` does not work (not used `airflow_3_uri` but logs warning)


cosmos/settings.py
```diff
@@ -19,7 +19,7 @@
 cache_dir = Path(conf.get("cosmos", "cache_dir", fallback=DEFAULT_CACHE_DIR) or DEFAULT_CACHE_DIR)
 enable_cache = conf.getboolean("cosmos", "enable_cache", fallback=True)
 enable_dataset_alias = conf.getboolean("cosmos", "enable_dataset_alias", fallback=True)
-use_dataset_airflow3_uri_standard = conf.getboolean("cosmos", "enable_dataset_airflow3_uri", fallback=False)
+use_dataset_airflow3_uri_standard = conf.getboolean("cosmos", "use_dataset_airflow3_uri_standard", fallback=False)
```
tatiana: Thanks for raising this issue and trying to fix it, @Anti0ff! Unfortunately, this would be a breaking change. Is there any chance we could support both, for now?

Anti0ff: > Is there any chance we could support both, for now?

May be something like
`conf.getboolean("cosmos", "enable_dataset_airflow3_uri", fallback=conf.getboolean("cosmos", "use_dataset_airflow3_uri_standard", fallback=False))`
(if `enable_dataset_airflow3_uri` is not set - try to read `use_dataset_airflow3_uri_standard` - and `False` if both are not set)

Or just change docs [there for configuration/scheduling](https://astronomer.github.io/astronomer-cosmos/configuration/scheduling.html) and [there for cosmos-conf](https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html?utm_cta=website-events-featured-summit#use-dataset-airflow3-uri-standard)

tatiana (approved):
Thanks a lot for making the implementation and docs consistent, @Anti0ff !

cosmos/settings.py
tatiana: It looks great, thank you very much!

Issue #1883, opened 2025-07-21
[Bug] Suggested MWAA startup script breaks MWAA environment


jaklan:
### Deployment

Amazon (AWS) MWAA

### What happened?

I'm really surprised it hasn't been noticed yet, but there's a serious issue with [the MWAA documentation](https://astronomer.github.io/astronomer-cosmos/getting_started/mwaa.html#create-a-startup-script) - that part is fundamentally wrong:

```bash
#!/bin/sh

export DBT_VENV_PATH="${AIRFLOW_HOME}/dbt_venv"
export PIP_USER=false

python3 -m venv "${DBT_VENV_PATH}"

${DBT_VENV_PATH}/bin/pip install <your-dbt-adapter>

export PIP_USER=true
```

The issue is related to `PIP_USER` environment variable - it should't be there. MWAA images (available [here](https://github.com/aws/amazon-mwaa-docker-images)) neither set up such envar nor use `--user` flag anywhere. Airflow and its dependencies are installed in `~/.local/` not because of that flag, but due to fallback mechanism of `pip` which installs packages in a home directory when it doesn't have write permissions to the system one (`/usr/local/` in this case).

But it's not only about `PIP_USER` being redundant - it's about its impact on `PythonVirtualenvOperator`. When you run such operator with `system_site_packages` parameter set to `True` - dependencies... won't be installed in a venv, but in the home directory! This way you can destroy the whole Airflow environment on a specific worker (been there, seen that - that's how we found out that issue). Here's the description what exactly happens:

<img width="1622" height="1120" alt="Image" src="https://github.com/user-attachments/assets/da701ef1-0ea1-4dac-b203-1c9e2c09647f" />

### How to reproduce

- `python -m venv .venv --system-site-packages`
- `source .venv/bin/activate`
- `PIP_USER=true pip install <whatever>`

and see how the world (== your home directory) burns.

### Anything else :)?

Solution is simple - just remove both lines related to `PIP_USER`.

jaklan:
Fun fact - this exact issue was noticed and fixed in an official Airflow Docker image... 4 years ago:

> In Airflow image we are setting PIP_USER variable to true, in order to
install all the packages by default with the --user flag. However
this is a problem if a virtualenv is created later which happens in
PythonVirtualenvOperator. We are unsetting this variable here, so that
it is not set when PIP is run by Airflow later on.

https://github.com/apache/airflow/pull/15774

[...]

PR #1884, opened 2025-07-22
Update suggested MWAA startup script

jaklan:
## Description

Issue described in https://github.com/astronomer/astronomer-cosmos/issues/1883.

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/1883

## Breaking Change?

No

## Checklist


pankajkoti (approved):
Thanks for reporting this issue @jaklan and updating the document with the fix so quickly!

PR #1888, opened 2025-07-25
Resolve Zizmor error alerts for unpinned image references; mark alert for pull_request_target ignored

pankajkoti:
scan results from this PR: https://github.com/astronomer/astronomer-cosmos/security/code-scanning?query=is%3Aopen+pr%3A1888

vs

on main: https://github.com/astronomer/astronomer-cosmos/security/code-scanning (this will change after merging this PR, but can be compared while the PR is still open)

related: https://github.com/astronomer/oss-integrations-private/issues/156

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -43,4 +43,4 @@ jobs:
 
       # More information in https://github.com/pypa/gh-action-pypi-publish
       - name: Publish package distributions to PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc  # v1.12.4
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,4 +16,4 @@ jobs:
       id-token: write
     steps:
       - id: deployment
-        uses: sphinx-notes/pages@v3
+        uses: sphinx-notes/pages@16601c3551fb1aff4d3f9cbffe7f8c86e123c917  # v3.4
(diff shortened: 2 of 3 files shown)


tatiana (commented):
@pankajkoti thanks a lot for doing further security improvements in the repo.

Two questions:
1. Why are we using the Git hash (`76f52bc884231f62b9a034ebfe128415bbaabdfc`) instead of the tag (e.g `v1.12.4`) here?

2. Currently we're using GH [Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot) to keep our GitHub Actions [dependencies](https://github.com/astronomer/astronomer-cosmos/blob/main/.github/dependabot.yml) up-to-date, including with the latest security fixes.  Will the approach of pinning a GH commit hash also work with this?

pankajkoti:
> Two questions:
> 
> 1. Why are we using the Git hash (`76f52bc884231f62b9a034ebfe128415bbaabdfc`) instead of the tag (e.g `v1.12.4`)?

Zizmor still complains about v1.12.4 as they(tags) are replaceable and mutable whereas commit SHAs are immutable and irreplacable making them secure.

> 2. Currently we're using GH [Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot) to keep our GitHub Actions [dependencies](https://github.com/astronomer/astronomer-cosmos/blob/main/.github/dependabot.yml) up-to-date, including with the latest security fixes.  Will the approach of pinning a GH commit hash also work with this?

I looked a bit into this. GitHub recommends having commit SHAs for 3rd party Actions. And as long as we mention the correspondign versions in comments against those SHAs, dependabot will be able to read and use those for it's functioning. They added this support in https://github.com/dependabot/dependabot-core/pull/5951

tatiana (approved):
Looks great, thanks for claryfing the Dependabot behaviour, @pankajkoti !

PR #1890, opened 2025-07-25
Resolve 'credential persistence through GitHub Actions artifacts' warnings from Zizmor

pankajkoti:
The persist-credentials option in the actions/checkout action determines whether the Git credentials used to fetch the repository are left in the Git configuration after the checkout step. Setting persist-credentials to true (the default) means the credentials will remain, allowing subsequent steps in the workflow to perform authenticated Git operations without re-authentication. Setting it to false removes the credentials, which is useful for security when you don't want later steps to have access to the credentials.

scan results from this PR: https://github.com/astronomer/astronomer-cosmos/security/code-scanning?query=is%3Aopen+pr%3A1890

vs

on main: https://github.com/astronomer/astronomer-cosmos/security/code-scanning (this will change after merging this PR, but can be compared while the PR is still open)

related: https://github.com/astronomer/oss-integrations-private/issues/156

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -44,6 +44,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
+      with:
+        persist-credentials: false
 
     # Add any setup steps before running the `github/codeql-action/init` action.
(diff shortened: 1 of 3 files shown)


tatiana (approved):
Excellent work, @pankajkoti ! Thanks for improving the security of our repo.

PR #1902, opened 2025-08-05
Fix broken CI test

tatiana:
Installing openlineage-airflow was leading to AF3 to be installed instead of AF2.8 when running

```bash
hatch run tests.py3.11-2.8-1.9:test-integration-setup
```

### DIFF
--- a/scripts/test/integration-setup.sh
+++ b/scripts/test/integration-setup.sh
@@ -10,9 +10,13 @@ NEXT_MINOR_VERSION=$(echo "$DBT_VERSION" | awk -F. '{print $1"."$2+1}')
 # we install using the following workaround to overcome installation conflicts, such as:
 # apache-airflow 2.3.0 and dbt-core [0.13.0 - 1.5.2] and jinja2>=3.0.0 because these package versions have conflicting dependencies
 pip uninstall -y 'dbt-bigquery' 'dbt-databricks' 'dbt-duckdb' 'dbt-postgres' 'dbt-vertica' 'dbt-core'
-rm -rf airflow.*
+
+rm $AIRFLOW_HOME/airflow.cfg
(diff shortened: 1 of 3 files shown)


scripts/test/pre-install-airflow.sh
```diff
@@ -1,16 +1,4 @@
...
-set -e
-
-AIRFLOW_VERSION="$1"
-PYTHON_VERSION="$2"
-
-# Use this to set the appropriate Python environment in Github Actions,
-# while also not assuming --system when running locally.
-if [ "$GITHUB_ACTIONS" = "true" ] && [ -z "${VIRTUAL_ENV}" ]; then
-  py_path=$(which python)
+ython)
```
tatiana:
```suggestion
#!/bin/bash

set -v
set -x
set -e

AIRFLOW_VERSION="$1"
PYTHON_VERSION="$2"

# Use this to set the appropriate Python environment in Github Actions,
# while also not assuming --system when running locally.
if [ "$GITHUB_ACTIONS" = "true" ] && [ -z "${VIRTUAL_ENV}" ]; then
  py_path=$(which python)
```

scripts/test/integration-setup.sh
```diff
@@ -27,8 +31,6 @@ fi
 
 uv pip install -U "dbt-core==$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica 'dbt-databricks' pyspark
```
tatiana:
```suggestion
uv pip install -U "dbt-core==$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica 'dbt-databricks' pyspark

pip install  -U openlineage-airflow
```

pankajkoti (approved):
Nice catch and fix! Thanks!

Issue #1488, opened 2025-01-24
Evaluate DAG versioning & bundles with a Airflow 3 pre-release


tatiana:
**What we want to do**

Evaluate how DAG Bundles and DAG versioning are behaving with Cosmos.

Try out Airflow 3.0 with Cosmos example DAGs:
```
COMMIT_SHA=573cd95 AIRFLOW_HOME=$(pwd) uvx \
  --python 3.11 \
  --from "git+https://github.com/apache/airflow@$COMMIT_SHA" \
  --with "git+https://github.com/apache/airflow@$COMMIT_SHA#egg=apache-airflow-task-sdk&subdirectory=task_sdk" \
  --with "git+https://github.com/apache/airflow@$COMMIT_SHA#egg=local-providers&subdirectory=providers" \
  bash -c "
    cd \$(python -c 'import airflow, os; print(os.path.dirname(airflow.__file__))') && \
    # Build the new UI (located in the 'ui' directory)
    cd ui && pnpm install && pnpm build && \
    # Build the legacy UI (located in the 'www' directory)
    cd ../www && yarn install --frozen-lockfile && yarn run build && \
    airflow standalone
  "
```
See the UI working in:
http://localhost:9091/webapp/
http://localhost:8080/

Focus on the work of these AIPs:
* [AIP-63: DAG Versioning](https://cwiki.apache.org/confluence/display/AIRFLOW/AIP-63%3A+DAG+Versioning)
* [AIP-66: DAG Bundles & Parsing](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=294816356)

**Acceptance criteria**

* [ ] Document which Airflow/Cosmos version was tested
* [ ] Raise any concerns and problems faced with @jedcunningham and team
* [ ] Have an action plan

tatiana:
@phanikumv suggested we try this with the nightly image. We can check with Rahul

tatiana:
Tables of interest:
- dag_run:
- dag_version:
- serialized_dag:
- dag_code:

1. Run Cosmos DAG - v X
2. Without changing Cosmos DAG, re-run v X
3. Change Cosmos python code, re-run v Y
4. Change dbt project, re-run, v Z

We need to use a version backend. We need to use the Git Repo

TODO:
- add diagram
- add reference on GitRepo

This is expected to work with local executors only.

This is likely not expected to run using DAG test.

tatiana:
[...]

PR #1907, opened 2025-08-06
Add dbt project's hash to dag docs to support dag versioning in Airflow 3

### LINKED ISSUE #1868: [Feature] Support DAG versioning in Airflow 3 with cosmos for dbt project change
### Description

@pankajkoti to update description

### Use case/motivation

_No response_

### Related issues

_No response_

pankajkoti:
This PR leverages the existing method for generating a hash of the dbt project contents and appends that hash to the DAG documentation. This enables Airflow 3’s automatic DAG versioning feature to track changes in dbt project content.

Airflow 3 introduces automatic DAG versioning by detecting changes in a DAG’s structure or metadata. By embedding the dbt content hash into the doc_md field (a part of the serialized DAG), this change allows Airflow to:
	•	Detect when dbt project files are added, modified, or deleted
	•	Automatically generate a new DAG version in response to those changes

This ensures accurate and automatic versioning of DAGs tied to evolving dbt projects, without requiring manual intervention.

closes: #1868

### DIFF
--- a/cosmos/cache.py
+++ b/cosmos/cache.py
@@ -42,6 +42,7 @@ from cosmos.settings import (
     remote_cache_dir_conn_id,
 )
 from cosmos.settings import remote_cache_dir as settings_remote_cache_dir
+from cosmos.versioning import _create_folder_version_hash
 
 logger = get_logger(__name__)
 VAR_KEY_CACHE_PREFIX = "cosmos_cache__"
@@ -52,7 +53,7 @@ def _configure_remote_cache_dir() -> Path | None:
     if not settings_remote_cache_dir:
         return None
 
-    _configured_cache_dir = None
+    _configured_cache_dir: Path | None = None
 
     cache_dir_str = str(settings_remote_cache_dir)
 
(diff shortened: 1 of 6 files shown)


cosmos/cache.py
```diff
@@ -267,37 +268,6 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
         shutil.copy(str(source_manifest_filepath), str(target_manifest_filepath))
 
 
-def _create_folder_version_hash(dir_path: Path) -> str:
```
pankajkoti, line 270: Moved this function as is to a new module `cosmos/versioning.py` since we now also need the same implementation to be called in the converter.py for DAG versioning support

tests/test_cache.py
```diff
@@ -123,38 +122,6 @@ def test__copy_partial_parse_to_project_msg_fails_msgpack(mock_unpack, tmp_path,
     assert "Unable to patch the partial_parse.msgpack file due to ValueError()" in caplog.text
 
 
-def test__create_folder_version_hash(tmp_path, caplog):
```
pankajkoti, line 126: moved this test as is to `tests/test_versioning.py` module corresponding to the source method movement

tatiana (approved):
Looks great, @pankajkoti , thank you! Very excited to see Cosmos DAGs being versioned in AF3!

tatiana:
BTW: as we discussed, please, let's check on a long-term solution with @jedcunningham and other Airflow commiters - to see if we can have a more robust solution that does not rely on the Airflow DAG docs.

Last time we discussed this, #1488, there wasn't a better way - but there may be one moving forward 🤞

pmuntianu:
Hi Team, this change is impacting parsing performance of cosmos because our IO is slow in k8s. And there is no way to turn this feature off while we are using airflow 2+. Do you have any suggestions?

pankajkoti:
> impacting parsing performance

hi @pmuntianu, I just logged an issue https://github.com/astronomer/astronomer-cosmos/issues/2386 to allow disabling this work done here in the PR. We will try to prioritise this, but would like to check with you that, before we get a chance to take a stab at it, is there a chance you could contribute a PR for this?

pmuntianu:
Hi @pankajkoti , thanks a ton. This is really pormpt response. I appreciate your dedication. For us, the solution is to not pass `dag` explicitly to the `DbtTaskGroup` and instead use `with` comprehension. Something like this:
```python
        with dag:
            DbtTaskGroup(
                # by commenting line below we prevent airflow dag caching in _add_dbt_project_hash_to_dag_docs method
                # this is important because our io is slow in k8s. We might negatively impact caching as well.
                # dag=dag,
                group_id=f"{self.workflow_settings.name}_group",
                project_config=self._parse_project(),
...
```

Thanks a lot one more time for taking this work

Issue #1875, opened 2025-07-16
[Bug] Caching behavior doesn't like . in DAG name


collinmcnulty:
### Astronomer Cosmos Version

1.10.1

### dbt-core version

unclear, maybe fusion

### LoadMode

AUTOMATIC

### ExecutionMode

AIRFLOW_ASYNC

### InvocationMode

None

### airflow version

2.11.0

### Operating System

debian bookworm

### Deployment

Astronomer

### What happened?

I think I found a bug in cosmos's caching behavior. It stores the result of dbt ls in a Variable based on the DAG name, but if the name has a . in it, that Variable name will be invalid when working with certain secret managers, including Google cloud's.

```
Invalid secret ID myhost-cosmos_cache__dag.moredag.yetmoredag
Only ASCII alphabets (a-Z), numbers (0-9), dashes (-), and underscores (_) are allowed in the secret ID.
```

### Relevant log output

```shell

```

### How to reproduce

1. Put a . in the dag name
2. Use Google Cloud secrets service
3. Use automatic parse method

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

tatiana:
Thanks a lot for logging this, @collinmcnulty ! Original issue, if we want to track: https://astronomer.zendesk.com/agent/tickets/78978

jtrobles-cdd:
Hello! I'm in charge of an (or the?) Astro Deployment that's affected by support ticket 78978.

Here are a few more details about our setup:

> ### dbt-core version

1.10.4

> ### Versions of dbt adapters

dbt-bigquery 1.10.0

> ### LoadMode

DBT_LS

> ### ExecutionMode

LOCAL

> ### InvocationMode

SUBPROCESS

tatiana:
Thank you very much for all the details, @collinmcnulty and @jtrobles-cdd. We just managed to add this ticket to our sprint, and we'll try to get a resolution sooner rather than later.

PR #1908, opened 2025-08-06
 Fix cache behaviour when DAG name contains "."

tatiana:
Cosmos stores the result of dbt ls as a Variable based on the DAG name, but if the name has a . in it, that Variable name will be invalid when working with certain secret managers, including Google Cloud's.
    
This is an example of an error that is raised:
  ```
  Invalid secret ID myhost-cosmos_cache__dag.moredag.yetmoredag
  Only ASCII alphabets (a-Z), numbers (0-9), dashes (-), and underscores (_) are allowed in the secret ID.
  ```
  
  In this PR, we replace those names to use underscores instead of ".".
  
  Closes: #1875

### DIFF
--- a/tests/dbt/test_graph.py
+++ b/tests/dbt/test_graph.py
@@ -1908,7 +1908,7 @@ def test_save_dbt_ls_cache(mock_variable_set, mock_datetime, tmp_dbt_project_dir
     assert hash_args == "d41d8cd98f00b204e9800998ecf8427e"
     if sys.platform == "darwin":
         # We faced inconsistent hashing versions depending on the version of MacOS/Linux - the following line aims to address these.
-        assert hash_dir in ("c2c47529eaec412281bdb243a479b734", "71bbf303ad4e06a7b1e2be20e0b73c0d")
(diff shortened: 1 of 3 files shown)


tests/dbt/test_graph.py
```diff
@@ -1908,7 +1908,7 @@ def test_save_dbt_ls_cache(mock_variable_set, mock_datetime, tmp_dbt_project_dir
     assert hash_args == "d41d8cd98f00b204e9800998ecf8427e"
     if sys.platform == "darwin":
         # We faced inconsistent hashing versions depending on the version of MacOS/Linux - the following line aims to address these.
-        assert hash_dir in ("c2c47529eaec412281bdb243a479b734", "71bbf303ad4e06a7b1e2be20e0b73c0d")
+        assert hash_dir in ("e16e6aef5f03f4e12d744b9412d90272", "71bbf303ad4e06a7b1e2be20e0b73c0d")
```
tatiana, line 1911: Fix the local tests dbt ls cache (outside of the CI)

pankajkoti (approved):
Nice fix!

PR #1767, opened 2025-05-19
Add dbt-core 1.10b to test matrix

tatiana:
Enable dbt 1.10 b3 in the test matrix so we can identify ahead of time if anything stopped working with the newest version of the dbt-core release.

PR #1909, opened 2025-08-07
Pin package dbt-databricks

pankajastro:
This PR remove the min version limitation of the package `dbt-databricks`.  The `dbt-databricks` version `1.10.8` has been yanked.

### DIFF
--- a/scripts/test/integration-setup.sh
+++ b/scripts/test/integration-setup.sh
@@ -31,7 +31,8 @@ else
 fi
 
 if [ "$DBT_VERSION" = "1.10" ]; then
-    uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica "dbt-databricks>=1.10.8" pyspark
+    # The package dbt-databricks version 1.10.8 has been yanked from PyPI: https://pypi.org/project/dbt-databricks/1.10.8/
+    uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica "dbt-databricks!=1.10.8" pyspark
 else
     uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica dbt-databricks pyspark
 fi
--- a/scripts/test/integration-setup.sh
+++ b/scripts/test/integration-setup.sh
@@ -30,9 +30,7 @@ else
     airflow db init
 fi
 
-if [ "$DBT_VERSION" = "1.10" ]; then
-    uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica dbt-databricks pyspark
-fi
+uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica dbt-databricks pyspark
 
 if python3 -c "import sys; print(sys.version_info >= (3, 9))" | grep -q 'True'; then
   pip install  'dbt-duckdb' "airflow-provider-duckdb>=0.2.0" "apache-airflow==$AIRFLOW_VERSION"
--- a/scripts/test/integration-setup.sh
+++ b/scripts/test/integration-setup.sh
@@ -31,9 +31,6 @@ else
 fi
 
 if [ "$DBT_VERSION" = "1.10" ]; then
-    # The package dbt-databricks version 1.10.8 has been yanked from PyPI: https://pypi.org/project/dbt-databricks/1.10.8/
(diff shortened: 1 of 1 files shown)


scripts/test/integration-setup.sh
```diff
@@ -31,7 +31,8 @@ else
 fi
 
 if [ "$DBT_VERSION" = "1.10" ]; then
-    uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica "dbt-databricks>=1.10.8" pyspark
+    # The package dbt-databricks version 1.10.8 has been yanked from PyPI: https://pypi.org/project/dbt-databricks/1.10.8/
+    uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica "dbt-databricks!=1.10.8" pyspark
```
pankajkoti:
```suggestion
    uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica dbt-databricks pyspark
```
wouldn't that more or less be the same? yanked versions won't be available to install anyway.

And then do we need the if else block, because both of them would be same.

I am curious if there was a reason we had to introduce the min version for dbt-databricks for dbt-core1.10 introduced in https://github.com/astronomer/astronomer-cosmos/pull/1767/files#diff-7fd89f34bb771040d82b88cfc3edcbdc536a0a5d2dbedc9dbf66635103edd6b4R33-R34. @tatiana do you remember why we added the min version pin for dbt-databricks for dbt1.10?

tatiana: dbt Core 1.10.x has conflicts with many dbt Databricks 1.10.x versions. These were raised in several of our integration tests, including `tests.py3.11-2.8-1.10`.

Some of the commits in [this PR](https://github.com/astronomer/astronomer-cosmos/pull/1767/commits) attempted to solve this issue:
* 8d1825afd9c4e551e9613fb569b9057c9de95e02
* 8938a48e55cb61278bef1a9f9e4944237aebb2f6

One of the many errors that were raised while I was troubleshooting this was (just by running  `dbt ls --target dev`) from the `jaffle_shop_python` directory:
```
  File "/Users/tatiana.alchueyr/Library/Application Support/hatch/env/virtual/astronomer-cosmos/4lAAXoPP/tests.py3.11-2.8-1.10/lib/python3.11/site-packages/dbt/adapters/databricks/catalogs/_unity.py", line 14, in __init__
    super().__init__(config)
  File "/Users/tatiana.alchueyr/Library/Application Support/hatch/env/virtual/astronomer-cosmos/4lAAXoPP/tests.py3.11-2.8-1.10/lib/python3.11/site-packages/dbt/adapters/catalogs/_integration.py", line 93, in __init__
    self.file_format: Optional[str] = config.file_format
                                      ^^^^^^^^^^^^^^^^^^
AttributeError: 'types.SimpleNamespace' object has no attribute 'file_format'
```

This is a bug or mismatch in the dbt-databricks adapter (version 1.10.2) - even though we had not explicitly enabled the  Unity Catalog integration. 

I am not sure of all the combinations that led to this error, but I would highly suggest we pin the minimal so we don't face this bug again.

scripts/test/integration-setup.sh
```diff
@@ -31,8 +31,6 @@ else
 fi
 
 if [ "$DBT_VERSION" = "1.10" ]; then
-    uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica "dbt-databricks>=1.10.8" pyspark
```
tatiana, line 34: I suggest we keep this and replace 1.10.8 by 1.10.7, since it wasn't yanked - yet 🙈 
https://pypi.org/project/dbt-databricks/1.10.7/#history

scripts/test/integration-setup.sh
```diff
@@ -31,8 +31,6 @@ else
 fi
 
 if [ "$DBT_VERSION" = "1.10" ]; then
```
tatiana: We need to handle "$DBT_VERSION" != "1.10" as well..! Otherwise the tests will fail, because those adaptors won't be available.

scripts/test/integration-setup.sh
pankajastro: I have removed the min req version, i.e >=1.10.8. I think this way we will know if something break or not work with future release but please let me know if you want to set it to >=1.10.7

tatiana: As long as our CI is happy, we're good! Probably the `~=` change I had introduced already solve the problem 🤞

PR #1915, opened 2025-08-11
Fix tags extraction

ms32035:
## Description

Looks like the hardening missed the tags extraction in one place

## Related Issue(s)

#1877 

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -931,7 +931,7 @@ class DbtGraph:
                     resource_type=DbtResourceType(node_dict["resource_type"]),
                     depends_on=node_dict.get("depends_on", {}).get("nodes", []),
                     file_path=self.execution_config.project_path / _normalize_path(node_dict["original_file_path"]),
-                    tags=node_dict["tags"],
+                    tags=node_dict.get("tags", []),
                     config=node_dict.get("config") or {},
                     has_freshness=(
                         is_freshness_effective(node_dict.get("freshness"))


tatiana (commented):
Thanks for fixing this, @ms32035 ! Is there any chance we could extend the test to also cover this change? We'll merge and cut a new alpha

tatiana (approved):
I'm happy with the change as long as we add a test to avoid regressions on this.

cosmos/dbt/graph.py
```diff
@@ -931,7 +931,7 @@ def load_from_dbt_manifest(self) -> None:
                     resource_type=DbtResourceType(node_dict["resource_type"]),
                     depends_on=node_dict.get("depends_on", {}).get("nodes", []),
                     file_path=self.execution_config.project_path / _normalize_path(node_dict["original_file_path"]),
-                    tags=node_dict["tags"],
+                    tags=node_dict.get("tags", []),
```
pankajkoti, line 934:
```suggestion
                    tags=node_dict.get("tags") or [],
```
This is better when we expect tags to be a list, and it can happen that `tags` key is present but set to `null` / `None`, and then it won't be set to `[]` with the current implementation `node_dict.get("tags". [])`

pankajkoti (commented):
Thanks for the contribution, @ms32035!

Quick question to scope this: did you run into a real manifest where `tags` was missing (dbt Core or dbt Fusion), or is this a defensive change in anticipation?

From what I've seen, both dbt Core and dbt Fusion always include a `tags` key on nodes (empty list when unset). PR #1877 was focused on dbt Fusion node defaults based on actual failures and didn’t touch `tags`. If there’s a concrete repro where `tags` is absent, could you share a snippet of the manifest and your dbt versions?

Could you please provide:
- dbt Core/Fusion versions used
- Minimal manifest snippet showing the node without tags
- Steps to produce that manifest (command + config)

ms32035:
@pankajkoti we ran into this when we were testing conversion of one of our projects to dbt fusion. Didn't really look into the details of what exactly is going on in the manifest, as the project is quite large, and the manifest specification is not public. It might equally well be a bug in how dbt fusion generates the manifest, but the project runs otherwise with the CLI

pankajkoti:
> @pankajkoti we ran into this when we were testing conversion of one of our projects to dbt fusion. Didn't really look into the details of what exactly is going on in the manifest, as the project is quite large, and the manifest specification is not public. It might equally well be a bug in how dbt fusion generates the manifest, but the project runs otherwise with the CLI

Thanks @ms32035 for the quick response. No objections to the change proposed here apart from the suggestion in https://github.com/astronomer/astronomer-cosmos/pull/1915#discussion_r2272307372.

By any chance, do you have a stack trace of the error? Or can you confirm that this fix alone solves your issue? 

I am worried if there is something more to fix & if yes, we can log that issue for investigation too.

ms32035:
```
  File "/home/airflow/.local/lib/python3.12/site-packages/cosmos/dbt/graph.py", line 562, in load
    load_method[method]()
  File "/home/airflow/.local/lib/python3.12/site-packages/cosmos/dbt/graph.py", line 934, in load_from_dbt_manifest
    tags=node_dict["tags"],
         ~~~~~~~~~^^^^^^^^
KeyError: 'tags'
```
@pankajkoti the fix resolved the issue, and the dag loaded. You're absolutely right that there might be something else. We only tried to convert one of several dbt projects we have to dbt fusion, and I'm quite sure that once we get down to others, there will be more issues, but these are impossible to predict at the moment.

cosmos/dbt/graph.py
pankajkoti: I will address this in PR #1918

pankajkoti:
Thanks @ms32035, for the clarification. I am following up on this PR in PR #1918 to add a test to ensure we do not have regressions. And I'm proceeding ahead with merging the PR. Thanks for your contributions!

PR #1918, opened 2025-08-13
Test hardening of dbt node having tags as unset or missing

pankajkoti:
Follow-up PR for #1915 adding tests to ensure cosmos handles well when dbt tags are missing or set to None on a DbtNode.

related: #1915


pankajkoti:
We observed test failures here before rebasing main (i.e. before PR #1915 was merged): https://github.com/astronomer/astronomer-cosmos/actions/runs/16935353240/job/47990492166?pr=1918

pankajkoti:
After rebasing PR #1915, we still have the test failing: https://github.com/astronomer/astronomer-cosmos/actions/runs/16935513205/job/47991044286?pr=1918#step:6:1083, which is when the tags key is present but set to None. To fix this, I have pushed a commit: https://github.com/astronomer/astronomer-cosmos/pull/1918/commits/89ae41751e549f5e932449e72134e4ab81055c4c

tatiana (approved):
Thanks a lot, @pankajkoti , looks great!

PR #1920, opened 2025-08-13
Use pepy.tech downloads badge in README

pankajkoti:
Since https://img.shields.io badges are no longer working due to [pypistats.org being down](https://github.com/crflynn/pypistats.org/issues/82), use the alternative badge 
provider pepy.tech to display download counts in the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,8 +4,8 @@
 .. |ossrank| image:: https://img.shields.io/endpoint?url=https://ossrank.com/shield/2121
     :target: https://ossrank.com/p/2121-astronomer-cosmos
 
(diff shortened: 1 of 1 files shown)


tatiana (approved):

tatiana:
Looks great, thanks, @pankajkoti ! I noticed there are some red tests - unrelated to the change, but still..! Could you check if re-running would make them pass - and, if not, log a ticket for us to address this in a separate PR? Thanks a lot!

pankajkoti:
Thanks, @tatiana. Re-running those went fine. Probably just flaky—if I see them again, I’ll log an issue.

PR #1934, opened 2025-08-22
Use XCom to store sql when using ExecutionMode.AIRFLOW_ASYNC

### LINKED ISSUE #1891: Use XCom to exchange data when using `ExecutionMode.AIRFLOW_ASYNC`
As of Cosmos 1.10, the pre-compiled SQL files are exchanged by being uploaded to a remote object store and being downloaded by each correspondent task.

This task proposes using XCom instead, which would result in less configuration and potentially less overhead.

We should implement a PoC and compare the performance of both approaches.

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1891

Feature: Use XCom to store SQL queries when ExecutionMode.AIRFLOW_ASYNC is enabled.

This change introduces an option to store the SQL query in Airflow’s XCom instead of uploading to and downloading from remote storage (GCS/SC) during AIRFLOW_ASYNC execution mode. This approach significantly reduces the network overhead associated with file transfers in the setup task.

Implementation:

- Added a new flag upload_sql_to_xcom to enable or disable this behaviour.
- The flag is disabled by default to maintain backwards compatibility.

Benefit:

- Reduced API call network latency.
- Improved overall task execution time when running in async mode.

Observation: A dbt project containing 129 models takes approximately 500 seconds to complete when uploading and downloading SQL files to/from a bucket via network calls. In contrast, using XCom, the same takes approximately 2 seconds. Tested on PR: https://github.com/astronomer/cosmos-benchmark/pull/9

Alpha: https://pypi.org/project/astronomer-cosmos/1.11.0a3/

### DIFF
--- a/cosmos/operators/_asynchronous/bigquery.py
+++ b/cosmos/operators/_asynchronous/bigquery.py
@@ -132,7 +132,7 @@ class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtL
     def base_cmd(self) -> list[str]:
         return ["run"]
 
-    def get_sql_from_xocm(self, context: Context) -> str:
+    def get_sql_from_xcom(self, context: Context) -> str:
         start_time = time.time()
         file_path = self.async_context["dbt_node_config"]["file_path"]
         project_dir_parent = str(Path(self.project_dir).parent)
@@ -181,7 +181,7 @@ class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtL
         if settings.enable_setup_async_task:
 
             if settings.upload_sql_to_xocm:
-                sql_query = self.get_sql_from_xocm(context)
+                sql_query = self.get_sql_from_xcom(context)
             else:
                 sql_query = self.get_remote_sql()
 
@@ -201,7 +201,7 @@ class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtL
             self.log.info("SQL cannot be made available, skipping registration of compiled_sql template field")
             return
         if settings.upload_sql_to_xocm:
-            sql = self.get_sql_from_xocm(context)
+            sql = self.get_sql_from_xcom(context)
         else:
             sql = self.get_remote_sql().strip()
         self.log.debug("Executed SQL is: %s", sql)
--- a/cosmos/operators/_asynchronous/bigquery.py
+++ b/cosmos/operators/_asynchronous/bigquery.py
@@ -1,10 +1,14 @@
 from __future__ import annotations
 
+import base64
 import time
+import zlib
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Sequence
 
 import airflow
+
+from cosmos.operators.base import _sanitize_xcom_key
 
 try:
     from airflow.providers.google.cloud.operators.bigquery import BigQueryInsertJobOperator
@@ -132,11 +136,10 @@ class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtL
         start_time = time.time()
         file_path = self.async_context["dbt_node_config"]["file_path"]
         project_dir_parent = str(Path(self.project_dir).parent)
-        relative_file_path = str(file_path).replace(project_dir_parent, "").lstrip("/")
-        dbt_dag_task_group_identifier = self.async_context["dbt_dag_task_group_identifier"]
-        run_id = self.async_context["run_id"]
-        xcom_model_path = f"{dbt_dag_task_group_identifier}/{run_id}/run/{relative_file_path}"
-        sql_query = context["ti"].xcom_pull(task_ids="dbt_setup_async", key=xcom_model_path)
+        sql_model_path = str(file_path).replace(project_dir_parent, "").lstrip("/")
+        compressed_b64_sql = context["ti"].xcom_pull(task_ids="dbt_setup_async", key=_sanitize_xcom_key(sql_model_path))
+        compressed_b64_sql = base64.b64decode(compressed_b64_sql)
+        sql_query = zlib.decompress(compressed_b64_sql).decode("utf-8")
 
         elapsed_time = time.time() - start_time
         self.log.info("SQL file download completed in %.2f seconds.", elapsed_time)
--- a/cosmos/operators/_asynchronous/bigquery.py
+++ b/cosmos/operators/_asynchronous/bigquery.py
@@ -14,7 +14,7 @@ except ImportError:
         "with with `pip install apache-airflow-providers-google`."
     )
 
-from airflow.utils.context import Context
+from airflow.utils.context import Context  # type: ignore
 from packaging.version import Version
 
 from cosmos import settings
@@ -128,6 +128,20 @@ class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtL
     def base_cmd(self) -> list[str]:
         return ["run"]
 
+    def get_sql_from_xocm(self, context: Context) -> str:
+        start_time = time.time()
+        file_path = self.async_context["dbt_node_config"]["file_path"]
+        project_dir_parent = str(Path(self.project_dir).parent)
+        relative_file_path = str(file_path).replace(project_dir_parent, "").lstrip("/")
+        dbt_dag_task_group_identifier = self.async_context["dbt_dag_task_group_identifier"]
+        run_id = self.async_context["run_id"]
+        xcom_model_path = f"{dbt_dag_task_group_identifier}/{run_id}/run/{relative_file_path}"
+        sql_query = context["ti"].xcom_pull(task_ids="dbt_setup_async", key=xcom_model_path)
+
+        elapsed_time = time.time() - start_time
+        self.log.info("SQL file download completed in %.2f seconds.", elapsed_time)
+        return sql_query  # type: ignore
+
     def get_remote_sql(self) -> str:
         start_time = time.time()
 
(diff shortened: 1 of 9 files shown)


cosmos/operators/_asynchronous/bigquery.py
```diff
@@ -128,6 +132,19 @@ def __init__(
     def base_cmd(self) -> list[str]:
         return ["run"]
 
+    def get_sql_from_xocm(self, context: Context) -> str:
```
tatiana:
```suggestion
    def get_sql_from_xcom(self, context: Context) -> str:
```

docs/getting_started/async-execution-mode.rst
```diff
@@ -7,14 +7,14 @@ Airflow Async Execution Mode
...
 
 Advantages of Airflow Async Mode
 ++++++++++++++++++++++++++++++++
 
 - **Improved Task Throughput:** Async tasks free up Airflow workers by leveraging the Airflow Trigger framework. While long-running SQL transformations are executing in the data warehouse, the worker is released and can handle other tasks, increasing overall task throughput.
-- **Faster Task Execution:** With Cosmos ``SetupAsyncOperator``, the SQL transformations are precompiled and uploaded to a remote location. Instead of invoking a full dbt run during each dbt model task, the SQL files are downloaded from this remote path and executed directly. This eliminates unnecessary overhead from running the full dbt command, resulting in faster and more efficient task execution.
+- **Faster Task Execution:** With Cosmos ``SetupAsyncOperator``, the SQL transformations are precompiled and uploaded to a remote location or xcom. Instead of invoking a full dbt run during each dbt model task, the SQL files are downloaded from this remote path and executed directly. This eliminates unnecessary overhead from running the full dbt command, resulting in faster and more efficient task execution.
```
tatiana: We should state the difference we observed in time, detailing what was the size of the DAG tested.

tatiana (commented):
Hi @pankajastro, thank you very much!

1. Performance gain

The PR description states:
> Observation: A dbt project containing 129 models takes approximately 500 seconds to complete when uploading and downloading SQL files to/from a bucket via network calls. In contrast, using XCom, the same takes approximately 2 seconds.

Please share more details so that others can reproduce it. We'll be changing the `cosmos_benchmark` repo to public, so it is fine to reference it.

2. Default behaviour and possibility of introducing a breaking change

We strive to be as compatible with Semantic Versioning as possible.

However, given the little adoption of this feature, I would be supportive of making this a breaking change and having XCom as the default behaviour of Cosmos async. It makes it easier to adopt the feature, and the performance times speak for themselves.

If we decide to follow this path, we can have a flag to opt-out of this behaviour (and revert to the original implementation, if users want to).

@pankajastro @pankajkoti WDYT?

3. We should update the documentation either way:
https://astronomer.github.io/astronomer-cosmos/getting_started/async-execution-mode.html

This could be done in a follow-up PR

pankajastro:
> Please share more details so that others can reproduce it. We'll be changing the cosmos_benchmark repo to public, so it is fine to reference it.

I have added the PR link; please let me know if you would like me to include any specific details. The  running steps should be different from how we normally run any benchmark airflow+dbt test

> Default behaviour and possibility of introducing a breaking change

I'm unsure because it's still a breaking change, and we should void it in a minor version, but open to hearing others.

docs/getting_started/async-execution-mode.rst
pankajastro: Added the performance observation

pankajastro:
I gave some more thought to making this the default behaviour.

Currently, by default, we store the SQL in the `target/` folder, and only upload it to a remote location when `enable_setup_async_task` is set to True.

Making this new approach the default would require removing the current default behaviour (i.e., storing SQL in the target/ folder). The `enable_setup_async_task` configuration will not have any effect. Additionally, some effort would be needed for cleanup and testing.

Given other priorities, I suggest we either:
- **Put this on hold**, or
- **Merge it in its current form**, and revisit this cleanup once we make more progress on the watcher tasks.

That said, please let me know if you'd prefer to proceed with the cleanup now.

**Impact and Proposed Changes:**
- Remove enable_setup_async_task config
- Remove enable_teardown_async_task config
- Always render/run setup and teardown tasks in async execution mode
- Stop storing SQL in the target/ folder
- If remote_target_path and remote_target_path_conn_id are set, store SQL in the remote bucket
- By default, store SQL in Airflow XCom

tatiana:
@pankajastro  My request for storing the compiled SQL in Airflow XCom by default remains.

I firmly believe we should make the flag `upload_sql_to_xcom=True` by default as part of this PR.

Given that we reduce the time to share compiled SQL files via XCom by 250x - and the little adoption this execution mode has had so far - we should introduce this breaking change and document it.

Regarding your suggestions, I agree with addressing them in the future, in separate PRs.

pankajastro:
> @pankajastro My request for storing the compiled SQL in Airflow XCom by default remains.
> 
> I firmly believe we should make the flag `upload_sql_to_xcom=True` by default as part of this PR.
> 
> Given that we reduce the time to share compiled SQL files via XCom by 250x - and the little adoption this execution mode has had so far - we should introduce this breaking change and document it.
> 
> Regarding your suggestions, I agree with addressing them in the future, in separate PRs.

Updated the default value for upload_sql_to_xcom to true

docs/getting_started/async-execution-mode.rst
```diff
@@ -7,14 +7,14 @@ Airflow Async Execution Mode
 
 The Airflow async execution mode in Cosmos is designed to improve pipeline performance. This execution mode could be preferred when you’ve long running resources and you want to run them asynchronously by leveraging Airflow’s `deferrable operators <https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/deferring.html>`__. In this mode, additional operators—``SetupAsyncOperator`` and ``TeardownAsyncOperator``—are added to your workflow.
 
-- **SetupAsyncOperator:** This task runs a mocked ``dbt run`` command on your dbt project, which outputs compiled SQL files to the project’s target directory. These compiled SQLs are then uploaded to a remote location specified by the ``remote_target_path`` configuration.
+- **SetupAsyncOperator:** This task runs a mocked ``dbt run`` command on your dbt project, which outputs compiled SQL files to the project’s target director. If ``upload_sql_xcom`` is enabled, the compiled SQL files will be uploaded to Airflow XCom. Otherwise, they will be uploaded to the remote location specified by the ``remote_target_path`` configuration.
```
tatiana:
```suggestion
- **SetupAsyncOperator:** This task runs a mocked ``dbt run`` command on your dbt project, which outputs compiled SQL files to the project’s target director. If ``upload_sql_xcom`` is enabled (default behaviour), the compiled SQL files will be uploaded to Airflow XCom. Otherwise, they will be uploaded to the remote location specified by the ``remote_target_path`` configuration.
```

docs/getting_started/async-execution-mode.rst
```diff
@@ -7,14 +7,14 @@ Airflow Async Execution Mode
...
-- **Faster Task Execution:** With Cosmos ``SetupAsyncOperator``, the SQL transformations are precompiled and uploaded to a remote location. Instead of invoking a full dbt run during each dbt model task, the SQL files are downloaded from this remote path and executed directly. This eliminates unnecessary overhead from running the full dbt command, resulting in faster and more efficient task execution.
+- **Faster Task Execution:** With Cosmos ``SetupAsyncOperator``, the SQL transformations are precompiled and uploaded to a remote location or xcom. Instead of invoking a full dbt run during each dbt model task, the SQL files are downloaded from this remote path and executed directly. This eliminates unnecessary overhead from running the full dbt command, resulting in faster and more efficient task execution. We have observed a dbt project with 129 models takes ~500 seconds with remote SQL file upload/download, but only ~2 seconds using xcom.
```
tatiana:
```suggestion
- **Faster Task Execution:** With Cosmos ``SetupAsyncOperator``, the SQL transformations are precompiled and uploaded to xcom (default behaviour) or a remote location. Instead of invoking a full dbt run during each dbt model task, the SQL files are downloaded from this remote path and executed directly. This eliminates unnecessary overhead from running the full dbt command, resulting in faster and more efficient task execution. We have observed a dbt project with 129 models takes ~500 seconds with remote SQL file upload/download, but only ~2 seconds using xcom.
```

cosmos/settings.py
```diff
@@ -43,6 +43,7 @@
 remote_cache_dir_conn_id = conf.get("cosmos", "remote_cache_dir_conn_id", fallback=None)
 remote_target_path = conf.get("cosmos", "remote_target_path", fallback=None)
 remote_target_path_conn_id = conf.get("cosmos", "remote_target_path_conn_id", fallback=None)
+upload_sql_to_xocm = conf.getboolean("cosmos", "upload_sql_to_xcom", fallback=True)
```
tatiana: Need to fix this here, and in the other places using it
```suggestion
upload_sql_to_xcom = conf.getboolean("cosmos", "upload_sql_to_xcom", fallback=True)
```

tatiana (approved):
Happy for the changes to be merged once the checks pass, @pankajastro !

Very impressive performance improvement, I didn't realise using a remote object store would add so much overhead. 250x reduction in time to exchange SQL files between Airflow tasks is very significant!

vincent-pauwels-al:
On the documentation, it states "BigQuery support only: This mode only supports BigQuery as the target database. If a different target is specified, Cosmos will throw an error indicating the target database is unsupported in this mode."

Why does Airflow async only supports Bigquery?
I'm currently working on dbt-athena

pankajastro:
> Why does Airflow async only supports Bigquery?

The async implementation in Cosmos isn’t database-agnostic because we patch the dbt adapter at runtime. See https://github.com/astronomer/astronomer-cosmos/blob/a9fca6a74d287d4e62f6146ed55ab371d6189d08/cosmos/operators/_asynchronous/bigquery.py#L34

You’ll probably need to do something similar for Athena.

PR #1939, opened 2025-08-26
Reuse the virtual env for AIRFLOW_ASYNC setup task

### LINKED ISSUE #1936: Refactor AIRFLOW_ASYNC setuptask so it doesn't install the virtualenv
TODO: @pankajastro

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1936

Screenshot 2025-08-26 at 6 39 09 PM

DAG
```python

# [START airflow_async_execution_mode_example]
simple_dag_async = DbtDag(
    project_config=ProjectConfig("/usr/local/airflow/dbt/jaffle_shop"),
        async_py_requirements=[f"dbt-bigquery=={DBT_ADAPTER_VERSION}"],
    ),

    render_config=RenderConfig(test_behavior=TestBehavior.NONE),
    # normal dag parameters
    schedule=None,
    catchup=False,
    dag_id="simple_dag_async",
    tags=["simple"],
    operator_args={
        "location": "US",
        "install_deps": True,
        "full_refresh": True,
        "virtualenv_dir": "/usr/local/airflow/include/dbt_venv2",
    },
)
```
Alpha: https://pypi.org/project/astronomer-cosmos/1.11.0a4/

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -477,6 +477,7 @@ def _add_dbt_setup_async_task(
     task_group: TaskGroup | None,
     render_config: RenderConfig | None = None,
     async_py_requirements: list[str] | None = None,
+    virtualenv_dir: str | None = None,
 ) -> None:
     if execution_mode != ExecutionMode.AIRFLOW_ASYNC:
         return
@@ -489,6 +490,8 @@ def _add_dbt_setup_async_task(
         task_args["selector"] = render_config.selector
         task_args["exclude"] = render_config.exclude
         task_args["py_requirements"] = async_py_requirements
+
+    task_args["virtualenv_dir"] = virtualenv_dir
 
     setup_task_metadata = TaskMetadata(
         id=DBT_SETUP_ASYNC_TASK_ID,
@@ -643,6 +646,8 @@ def build_airflow_graph(
     detached_from_parent: dict[str, list[DbtNode]] = defaultdict(list)
     identify_detached_nodes(nodes, render_config, detached_nodes, detached_from_parent)
 
+    virtualenv_dir = task_args.pop("virtualenv_dir", None)
+
     for node_id, node in nodes.items():
         conversion_function = node_converters.get(node.resource_type, generate_task_or_group)
         if conversion_function != generate_task_or_group:
@@ -714,6 +719,7 @@ def build_airflow_graph(
             task_group,
             render_config=render_config,
             async_py_requirements=async_py_requirements,
+            virtualenv_dir=virtualenv_dir,
         )
     if settings.enable_teardown_async_task:
         _add_teardown_task(
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -477,7 +477,6 @@ def _add_dbt_setup_async_task(
     task_group: TaskGroup | None,
     render_config: RenderConfig | None = None,
     async_py_requirements: list[str] | None = None,
-    virtualenv_dir: str | None = None,
 ) -> None:
     if execution_mode != ExecutionMode.AIRFLOW_ASYNC:
         return
@@ -490,8 +489,6 @@ def _add_dbt_setup_async_task(
         task_args["selector"] = render_config.selector
         task_args["exclude"] = render_config.exclude
         task_args["py_requirements"] = async_py_requirements
-
-    task_args["virtualenv_dir"] = virtualenv_dir
 
     setup_task_metadata = TaskMetadata(
         id=DBT_SETUP_ASYNC_TASK_ID,
@@ -714,12 +711,11 @@ def build_airflow_graph(
         _add_dbt_setup_async_task(
             dag,
             execution_mode,
-            task_args,
+            {**task_args, "virtualenv_dir": virtualenv_dir},
             tasks_map,
             task_group,
             render_config=render_config,
             async_py_requirements=async_py_requirements,
-            virtualenv_dir=virtualenv_dir,
         )
     if settings.enable_teardown_async_task:
         _add_teardown_task(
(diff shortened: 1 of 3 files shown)


cosmos/airflow/graph.py
```diff
@@ -643,6 +643,8 @@ def build_airflow_graph(
     detached_from_parent: dict[str, list[DbtNode]] = defaultdict(list)
     identify_detached_nodes(nodes, render_config, detached_nodes, detached_from_parent)
 
+    virtualenv_dir = task_args.pop("virtualenv_dir", None)
```
pankajastro: Would this cause any issues for virtual env execution mode?

tatiana: Why are we popping this here..?

pankajastro: If we don't pop, other tasks will also receive the `virtualenv_dir` kwarg, which is invalid. 

```
Broken DAG: [/usr/local/airflow/dags/simple_dag_async.py]
  File "/usr/local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 508, in apply_defaults
    result = func(self, **kwargs, default_args=default_args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 959, in __init__
    raise AirflowException(
airflow.exceptions.AirflowException: Invalid arguments were passed to DbtRunAirflowAsyncOperator (task_id: customers_run). Invalid arguments were:
**kwargs: {'virtualenv_dir': '/usr/local/airflow/cosmos-env'}
```

pankajastro: Added a check before pop operation

tatiana (commented):
@pankajastro Thanks a lot for implementing this.

Some points:

1. Possibility of switching from `DbtRunVirtualenvOperator` to `DbtRunLocalOperator` 

We discussed about this in a call, but could you confirm why we are not changing the implementation to use `DbtRunLocalOperator` instead of `DbtRunVirtualenvOperator` https://github.com/astronomer/astronomer-cosmos/blob/426fbee0b39c078aaa48b3fe7586d128131b441b/cosmos/operators/_asynchronous/__init__.py#L21

Ideally, we'd remove the management of the virtualenv outside the async operator, and let the user manage this (being careful not to use the same between the async implementation and non-async DAGs/operators).

My concern with continuing to use the `DbtRunVirtualenvOperator` is that Cosmos will still create, potentially, the Python venv per Airflow worker node (which in the case of the Airflow `KubernetesExecutor` is very inefficient. Even considering `CeleryExecutors` were used, it will continue always trying to update the dependencies every time the setup task is run, even when users do not need to pay this cost.

2.  Possibility using this with pre-created venv

Does this approach work if users pre-create
`/usr/local/airflow/include/dbt_venv2` in their Dockerfile? It still remains with the issue of reaching out to PyPI and updating the dependencies.

3. Should we create some sort of test for this feature, to confirm that the venv is not recreated unnecessary?

4. We should update the docs explaining how users can benefit from this improvement

For instance, in the https://astronomer.github.io/astronomer-cosmos/getting_started/async-execution-mode.html?

pankajastro:
Thanks for feedback @tatiana 

> Possibility of switching from DbtRunVirtualenvOperator to DbtRunLocalOperator

We currently patch the dbt CLI in SetupAsyncOperator.
- This patching is inherently risky, and using a separate virtual environment helps minimise the risk surface.
- Switching to `DbtRunLocalOperator` would require us to manage the unpatching process, which is error-prone and increases the risk of corrupting the dbt-cli installation.
- There may also be custom dbt-related operators that rely on the local dbt CLI. Switching to `DbtRunLocalOperator` could introduce issues with these custom operators.

Therefore, switching to `DbtRunLocalOperator` does not appear to be the right choice IMHO.

> Possibility of using this with a pre-created virtual environment

This is supported by default and works as expected.

tatiana:
@pankajastro 
> We currently patch the dbt CLI in SetupAsyncOperator.
> This patching is inherently risky, and using a separate virtual environment helps minimise the risk surface.
Switching to `DbtRunLocalOperator` would require us to manage the unpatching process, which is error-prone and increases the risk of corrupting the dbt-cli installation.
There may also be custom dbt-related operators that rely on the local dbt CLI. Switching to DbtRunLocalOperator could introduce issues with these custom operators.
Therefore, switching to DbtRunLocalOperator does not appear to be the right choice IMHO.

I agree it would be very risky to patch and unpatch and use a single venv for both sync and async operators. My thought was different from this proposal. The idea is that users could have two venvs:
- one that is a patched venv, for async operators
- the other that is a non-patched venv, for non-async operators

If we decided to follow this approach, to make it even safer, we could show how to patch the venv in the `Dockerfile`.

That said, this is a further optimisation, and given the amount of time we're saving with the approach, I'm happy for this PR to be merged after (3) and (4) are addressed - and time will tell us if we need any further optimisations in this execution mode.

tatiana (approved):

pankajastro:
Added tests and created PR: https://github.com/astronomer/astronomer-cosmos/pull/1969 to add docs

pankajastro:
> I agree it would be very risky to patch and unpatch and use a single venv for both sync and async operators. My thought was different from this proposal. The idea is that users could have two venvs:
>
> - one that is a patched venv, for async operators
> - the other that is a non-patched venv, for non-async operators\

I'm sorry, I'm a bit confused now.
What do you mean by "async" and "non-async" here?
This PR is trying to address the SetupAsyncOperator only, which is a non-async operator. I'm trying to reuse the virtual env in this particular operator. For the rest of the operator/task, users keep using the way they are using now.

tatiana:
@pankajastro we discussed this further:
> I'm sorry, I'm a bit confused now.
What do you mean by "async" and "non-async" here?
This PR is trying to address the SetupAsyncOperator only, which is a non-async operator. I'm trying to reuse the virtual env in this particular operator. For the rest of the operator/task, users keep using the way they are using now.

In future, if we decide to implement this, we'd need to review the interfaces, so users of ASYNC would set two different venvs - one for the setup task, and the other functions that fall back into the local execution mode.

We all agreed that this is a potential further improvement that is out of scope in the current ticket.

Issue #1929, opened 2025-08-21
dbt Fusion 2.0.0-preview.3 breaking tests in CI


pankajkoti:
Check action: https://github.com/astronomer/astronomer-cosmos/actions/runs/17119471927/job/48556946882?pr=1928#step:7:628

Error traceback
```
  def run_command_with_subprocess(command: list[str], tmp_dir: Path, env_vars: dict[str, str]) -> str:
        """Run a command in a subprocess, returning the stdout."""
        process = Popen(
            command,
            stdout=PIPE,
            stderr=PIPE,
            cwd=tmp_dir,
            universal_newlines=True,
            env=env_vars,
        )
        stdout, stderr = process.communicate()
        returncode = process.returncode
    
        if 'Run "dbt deps" to install package dependencies' in stdout and command[1] == "ls":
            raise CosmosLoadDbtException(
                "Unable to run dbt ls command due to missing dbt_packages. Set RenderConfig.dbt_deps=True."
            )
    
        if returncode or "Error" in stdout.replace("WarnErrorOptions", ""):
            details = f"stderr: {stderr}\nstdout: {stdout}"
>           raise CosmosLoadDbtException(f"Unable to run {command} due to the error:\n{details}")
E           cosmos.dbt.graph.CosmosLoadDbtException: Unable to run ['/home/runner/.local/bin/dbt', 'ls', '--output', 'json', '--output-keys', 'name', 'unique_id', 'resource_type', 'depends_on', 'original_file_path', 'tags', 'config', 'freshness', '--project-dir', '/tmp/tmp5viijrkb', '--profiles-dir', '/home/runner/work/astronomer-cosmos/astronomer-cosmos/dev/dags/dbt/jaffle_shop', '--profile', 'snowflake_profile', '--target', 'dev'] due to the error:
E           stderr: error: dbt1060: Ignored unexpected key 'materialized'. Try '+materialized' instead. YAML path: 'jaffle_shop.materialized'.
E             --> dbt_project.yml:24:21
E           error: dbt1060: Ignored unexpected key 'materialized'. Try '+materialized' instead. YAML path: 'jaffle_shop.staging.materialized'.
E             --> dbt_project.yml:26:23
E           error: dbt0102: Deprecated test arguments: ["values"] at top-level detected. Please migrate to the new format under the 'arguments' field: https://docs.getdbt.com/reference/deprecations#missingargumentspropertyingenerictestdeprecation.
E             --> models/schema.yml:58:23
[...]

PR #1948, opened 2025-09-05
Fix running tests with dbt Fusion 2.0.0 preview versions

tatiana:
The dbt Fusion tests in Cosmos' main branch were failing with errors like:
```
>           raise CosmosLoadDbtException(f"Unable to run {command} due to the error:\n{details}")
E           cosmos.dbt.graph.CosmosLoadDbtException: Unable to run ['/home/runner/.local/bin/dbt', 'ls', '--output', 'json', '--output-keys', 'name', 'unique_id', 'resource_type', 'depends_on', 'original_file_path', 'tags', 'config', 'freshness', '--project-dir', '/tmp/tmp5viijrkb', '--profiles-dir', '/home/runner/work/astronomer-cosmos/astronomer-cosmos/dev/dags/dbt/jaffle_shop', '--profile', 'snowflake_profile', '--target', 'dev'] due to the error:
E           stderr: error: dbt1060: Ignored unexpected key 'materialized'. Try '+materialized' instead. YAML path: 'jaffle_shop.materialized'.
E             --> dbt_project.yml:24:21
E           error: dbt1060: Ignored unexpected key 'materialized'. Try '+materialized' instead. YAML path: 'jaffle_shop.staging.materialized'.
E             --> dbt_project.yml:26:23
E           error: dbt0102: Deprecated test arguments: ["values"] at top-level detected. Please migrate to the new format under the 'arguments' field: https://docs.getdbt.com/reference/deprecations#missingargumentspropertyingenerictestdeprecation.
E             --> models/schema.yml:58:23
E           error: dbt0102: Deprecated test arguments: ["values"] at top-level detected. Please migrate to the new format under the 'arguments' field: https://docs.getdbt.com/reference/deprecations#missingargumentspropertyingenerictestdeprecation.
E             --> models/staging/schema.yml:20:23
E           error: dbt0102: Deprecated test arguments: ["values"] at top-level detected. Please migrate to the new format under the 'arguments' field: https://docs.getdbt.com/reference/deprecations#missingargumentspropertyingenerictestdeprecation.
E             --> models/staging/schema.yml:31:23
E           
E           stdout: dbt-fusion 2.0.0-preview.3
E              Loading ../../home/runner/work/astronomer-cosmos/astronomer-cosmos/dev/dags/dbt/jaffle_shop/profiles.yml
E              Loading packages.yml
E           suggestion: Run 'dbt deps' to see the latest fusion compatible packages. For compatibility errors, try the autofix script: https://github.com/dbt-labs/dbt-autofix
E             Finished 'ls' target 'dev' with 5 errors in 91ms
```

Example:
https://github.com/astronomer/astronomer-cosmos/actions/runs/17485309398/job/49663095838?pr=1947

This issue is due to breaking changes introduced in the dbt schema and dbt project syntax as part of the dbt Fusion promotion (and only back-ported to dbt Core 1.10). The latest versions of dbt Fusion no longer accept the previous syntax, which was supported by dbt versions 1.9 and earlier.

This PR:
- Changes the `jaffle_shop` so that it is compatible with the new syntax. This has fixed the dbt Fusion tests.
- For tests that need to run with dbt Core versions before 1.10, modify the DAGs to use `altered_jaffle_shop`, allowing us to maintain both project syntaxes and validate against both versions. We must support dbt Core 1.9, as it has not yet reached the end of life (https://docs.getdbt.com/docs/dbt-versions/core).

There were a few changes to the CI, and a successful test run against these new configurations can be seen at:
https://github.com/astronomer/astronomer-cosmos/actions/runs/17733688014/job/50390249064

Tests worked with dbt Fusion 2.0.0 preview 13.

Closes: #1929

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, fix-1929]
+    branches: [main]
   # Also run on pull requests originating from forks. Although this is insecure by default, we need it to run
   # integration tests on forked PRs. As a guardrail, we’ve added an Authorize step to each job, which requires manually
   # approving the workflow run for each pushed commit. Approval only happens after a careful code review of the changes.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -381,7 +381,7 @@ jobs:
       matrix:
         python-version: [ "3.11" ]
         airflow-version: [ "2.11", "3.0" ]
-        dbt-version: ["1.5", "1.6", "1.7", "1.8", "1.9", "1.10"]
+        dbt-version: ["1.6", "1.7", "1.8", "1.9", "1.10"]
     services:
       postgres:
         image: postgres@sha256:4cd697181d4bd3ddc41a09012f339fa8cb5a8cd3d8b30130ea8378c176b6c494  # 14.18
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -434,6 +434,7 @@ jobs:
       AIRFLOW_HOME: /home/runner/work/astronomer-cosmos/astronomer-cosmos/
       PYTHONPATH: /home/runner/work/astronomer-cosmos/astronomer-cosmos/:$PYTHONPATH
 
+      AIRFLOW__COSMOS__ENABLE_DATASET_ALIAS: 0
       AIRFLOW_CONN_AWS_S3_CONN: ${{ secrets.AIRFLOW_CONN_AWS_S3_CONN }}
       AIRFLOW_CONN_GCP_GS_CONN: ${{ secrets.AIRFLOW_CONN_GCP_GS_CONN }}
       AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT: 90.0
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -421,16 +421,6 @@ jobs:
       - name: Test Cosmos against Airflow ${{ matrix.airflow-version }}, Python ${{ matrix.python-version }} and dbt ${{ matrix.dbt-version }}
         run: |
           hatch run tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }}-${{ matrix.dbt-version }}:test-integration-dbt-async
-        env:
-          AIRFLOW_HOME: /home/runner/work/astronomer-cosmos/astronomer-cosmos/
-          AIRFLOW_CONN_AWS_S3_CONN: ${{ secrets.AIRFLOW_CONN_AWS_S3_CONN }}
-          AIRFLOW_CONN_GCP_GS_CONN: ${{ secrets.AIRFLOW_CONN_GCP_GS_CONN }}
-          AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT: 90.0
-          PYTHONPATH: /home/runner/work/astronomer-cosmos/astronomer-cosmos/:$PYTHONPATH
-          AIRFLOW__COSMOS__ENABLE_CACHE: 0
-          AIRFLOW__COSMOS__REMOTE_TARGET_PATH: "s3://cosmos-remote-cache/target_compiled/"
-          AIRFLOW__COSMOS__REMOTE_TARGET_PATH_CONN_ID: aws_s3_conn
-          DBT_ADAPTER_VERSION: ${{ matrix.dbt-version }}
 
       - name: Upload coverage to Github
         uses: actions/upload-artifact@v4
@@ -441,10 +431,18 @@ jobs:
 
     env:
       AIRFLOW_HOME: /home/runner/work/astronomer-cosmos/astronomer-cosmos/
-      AIRFLOW_CONN_EXAMPLE_CONN: postgres://postgres:postgres@0.0.0.0:5432/postgres
       PYTHONPATH: /home/runner/work/astronomer-cosmos/astronomer-cosmos/:$PYTHONPATH
 
+      AIRFLOW_CONN_AWS_S3_CONN: ${{ secrets.AIRFLOW_CONN_AWS_S3_CONN }}
+      AIRFLOW_CONN_GCP_GS_CONN: ${{ secrets.AIRFLOW_CONN_GCP_GS_CONN }}
+      AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT: 90.0
+      AIRFLOW__COSMOS__ENABLE_CACHE: 0
+      AIRFLOW__COSMOS__REMOTE_TARGET_PATH: "s3://cosmos-remote-cache/target_compiled/"
+      AIRFLOW__COSMOS__REMOTE_TARGET_PATH_CONN_ID: aws_s3_conn
+      DBT_ADAPTER_VERSION: ${{ matrix.dbt-version }}
+
       # The following are only required because the profiles.yml file references them. They are not used by the tests selected by this job.
+      AIRFLOW_CONN_EXAMPLE_CONN: postgres://postgres:postgres@0.0.0.0:5432/postgres
       POSTGRES_HOST: localhost
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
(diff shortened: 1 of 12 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, fix-1929]
```
```suggestion
    branches: [main]
```

.github/workflows/test.yml
```diff
@@ -380,7 +382,7 @@ jobs:
       matrix:
         python-version: [ "3.11" ]
         airflow-version: [ "2.11", "3.0" ]
-        dbt-version: ["1.5", "1.6", "1.7", "1.8", "1.9", "1.10"]
+        dbt-version: ["1.6", "1.7", "1.8", "1.9", "1.10"]
```
pankajastro, line 385: Why are you removing the dbt-1.5? Do we need to update some docs, too?

.github/workflows/test.yml
```diff
@@ -614,7 +624,7 @@ jobs:
       matrix:
         python-version: [ "3.12" ]
         airflow-version: [ "2.10", "3.0" ]
-        dbt-version: [ "1.9" ]
+        dbt-version: [ "1.10" ]
```
pankajastro, line 627: same question as above for changing 1.9 -> 1.10

pankajastro (approved):

.github/workflows/test.yml
```diff
@@ -369,6 +369,8 @@ jobs:
           POSTGRES_PORT: 5432
           AIRFLOW__COSMOS__REMOTE_TARGET_PATH: "s3://cosmos-remote-cache/target_compiled/"
           AIRFLOW__COSMOS__REMOTE_TARGET_PATH_CONN_ID: aws_s3_conn
+          DBT_PROJECT_NAME: "altered_jaffle_shop"  # The syntax of jaffle_shop is not supported in dbt 1.5.4, only in dbt >= 1.10
+          TEST_SINGLE_DAG: "basic_cosmos_task_group.py"
```
pankajkoti, line 373: Is there some motivation behind running just a single task for dbt 1.5.4? Can we please add a comment for that too here, similar to the above comment for the `DBT_PROJECT_NAME`?

.github/workflows/test.yml
pankajkoti: +1, curious on this.

tatiana: @pankajastro We still have other tests running against dbt 1.5 - I just removed this one to simplify the already complex usage of `altered_jaffle_shop` and `jaffle_shop` depending on the dbt version.

tests/test_example_dags.py
```diff
@@ -72,7 +72,6 @@ def get_dag_bag() -> DagBag:  # noqa: C901
             file.writelines([f"{dagfile}\n"])
 
         if DBT_VERSION < Version("1.6.0"):
-            file.writelines(["simple_dag_async.py\n"])
             file.writelines(["example_model_version.py\n"])
             file.writelines(["example_operators.py\n"])
```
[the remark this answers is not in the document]
tatiana: Tests remain working

.github/workflows/test.yml
tatiana: @pankajastro We are still running tests against dbt 1.9. Unfortunately, we do not have a dbt project that is compatible with both dbt versions 1.9 and 1.10. For this reason, I changed the matrices. All the previous versions continue to be tested in one way or another.

In this particular case, there was no compelling reason to test against dbt 1.9 - it was simply the latest version of dbt when we introduced this job.

My preference was to change the minimum amount of DAGs using `altered_jaffle_shop` as possible, without losing specific tests that were needed for specific versions.

.github/workflows/test.yml
```diff
@@ -441,9 +433,27 @@ jobs:
...
+      DBT_ADAPTER_VERSION: ${{ matrix.dbt-version }}
+
+      # The following are only required because the profiles.yml file references them. They are not used by the tests selected by this job.
+      AIRFLOW_CONN_EXAMPLE_CONN: postgres://postgres:postgres@0.0.0.0:5432/postgres
+      POSTGRES_HOST: localhost
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+      POSTGRES_DB: postgres
+      POSTGRES_SCHEMA: public
+      POSTGRES_PORT: 5432
```
pankajkoti, line 455: Is there a reason we need to move those environment variables here? Our other jobs have most of the environment variables localised to the step that runs the tests.

.github/workflows/test.yml
```diff
@@ -441,9 +433,27 @@ jobs:
 
     env:
       AIRFLOW_HOME: /home/runner/work/astronomer-cosmos/astronomer-cosmos/
-      AIRFLOW_CONN_EXAMPLE_CONN: postgres://postgres:postgres@0.0.0.0:5432/postgres
       PYTHONPATH: /home/runner/work/astronomer-cosmos/astronomer-cosmos/:$PYTHONPATH
 
+      DBT_PROJECT_NAME: "altered_jaffle_shop" # The syntax of jaffle_shop is not supported in dbt 1.5.4, only in dbt >= 1.10
+      AIRFLOW__COSMOS__ENABLE_DATASET_ALIAS: 0
```
pankajkoti, line 439: Should we add a comment for this environment variable on why we disable the dataset aliases for this job?

scripts/test/integration.sh
```diff
@@ -22,4 +22,4 @@ pytest -vv \
     --ignore=tests/perf \
     --ignore=tests/test_async_example_dag.py \
     --ignore=tests/test_example_k8s_dags.py \
-    -k 'not ( example_cosmos_python_models or example_virtualenv or jaffle_shop_kubernetes)'
+    -k 'not (simple_dag_async or example_cosmos_python_models or example_virtualenv or jaffle_shop_kubernetes)'
```
pankajkoti, line 25: Are we ignoring the `simple_dag_async` because it is acting flaky, or some other reason?

pankajkoti: I now guess, because we have the `Run-Integration-Tests-DBT-Async` job?

.github/workflows/test.yml
tatiana: Yes..! Previously, even when we wanted to run a single DAG in specific tests, we had to parse most of the DAG Bag. We didn't have a way to specify only parsing the DAG of interest, even though that is what the CI job needed.

This was particularly problematic when trying to run specific DAGs that use `altered_jaffle_shop` and have a dbt syntax compatible only with dbt versions 1.9 or earlier, or DAGs that use `jaffle_shop` and are only compatible with dbt versions 1.10 or later.

.github/workflows/test.yml
tatiana: It is because we're now using the `altered_jaffle_shop` which has non-ASCII models and this raises an exception in Airflow versions that try to create the correspondent Dataset alias - since they do not support non-ASCII values.

This is further explained in:
https://github.com/astronomer/astronomer-cosmos/issues/1802

scripts/test/integration.sh
tatiana: Yes, I removed this redundancy because we we already had the `Run-Integration-Tests-DBT-Async` job covering this test-case

pankajkoti: understood, thanks

.github/workflows/test.yml
pankajkoti: understood. Would be nice if we could link this issue in the comment so that we can remove it once the issue is resolved.

.github/workflows/test.yml
tatiana: I didn't realise we were trying to keep the envvars closer to the step, sorry.  We can revert this change in a future change to `test.yml`! Please, feel free to create a PR with this change

.github/workflows/test.yml
tatiana: Added comment in https://github.com/astronomer/astronomer-cosmos/pull/1963

.github/workflows/test.yml
tatiana: Added comment in https://github.com/astronomer/astronomer-cosmos/pull/1963

Issue #1950, opened 2025-09-10
[Feature] Introduce a new execution mode with a dbt runner task and subsequent watcher tasks


tatiana:
### Description

**Context**

As of Cosmos 1.10, each "run" task that maps to a dbt model invokes dbt (either as a binary or as a library).

We noticed that the cost to run the same pipeline with plain dbt core varies significantly by running:
- the whole dbt command using a single command
- running one dbt command per model

AS an example, for the https://github.com/google/fhir-dbt-analytics project, these numbers were, on average, 5min30 versus 32min.

**Proposed solution**

* Use existing Cosmos DAG rendering techniques
* Have a single Airlfow task to run "all the pipeline" (selected by the user)
* Try to use dbt-core callbacks https://docs.getdbt.com/reference/programmatic-invocations#registering-callbacks to track how the model's execution is progressing and update different Xcoms (one Xcom per model)
* All the other tasks, by default, should watch their designated Xcom
* Have a strategy for retrying the pipeline

This proposal follows up a successful internal PoC (https://github.com/astronomer/oss-integrations-private/issues/185), available in the branch https://github.com/astronomer/astronomer-cosmos/tree/single-run-execution-mode.

### Overview of the plan

https://docs.google.com/document/d/1lQkbmurAZvnisVUybEY8ZkvuX-00M5wGySnhGEHZEfw/edit?tab=t.0

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

PR #1952, opened 2025-09-11
Add sample dbt events for the dbt watcher execution mode

pankajkoti:
related: #1950


tatiana (approved):
Thanks for adding the messages that we'll use for the interfaces of the operators, @pankajkoti !
Happy for this PR to be merged once the tests are green.

PR #1963, opened 2025-09-15
Improve comments after feedback on #1948

tatiana:
Addresses feedback that was posted after merging the PR:
https://github.com/astronomer/astronomer-cosmos/pull/1948/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -370,7 +370,7 @@ jobs:
           AIRFLOW__COSMOS__REMOTE_TARGET_PATH: "s3://cosmos-remote-cache/target_compiled/"
           AIRFLOW__COSMOS__REMOTE_TARGET_PATH_CONN_ID: aws_s3_conn
           DBT_PROJECT_NAME: "altered_jaffle_shop"  # The syntax of jaffle_shop is not supported in dbt 1.5.4, only in dbt >= 1.10
-          TEST_SINGLE_DAG: "basic_cosmos_task_group.py"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -370,7 +370,7 @@ jobs:
           AIRFLOW__COSMOS__REMOTE_TARGET_PATH: "s3://cosmos-remote-cache/target_compiled/"
           AIRFLOW__COSMOS__REMOTE_TARGET_PATH_CONN_ID: aws_s3_conn
           DBT_PROJECT_NAME: "altered_jaffle_shop"  # The syntax of jaffle_shop is not supported in dbt 1.5.4, only in dbt >= 1.10
-          TEST_SINGLE_DAG: "basic_cosmos_task_group.py"
+          TEST_SINGLE_DAG: "basic_cosmos_task_group.py"  # There are circumstances when we only need to run a single DAG, but we were parsing all the DAGs. This change is to avoid parsing all the DAGs.
```
pankajkoti, line 373: It would be nice to check if we can drop support for dbt 1.5

Currently, we're testing only one DAG for dbt 1.5 for a known case we want to test, but I guess it makes it hard to remember the exact reasons/changes we wanted to test across versions.

pankajkoti (approved):

pankajkoti:
LGTM, thanks for quickly addressing the follow-up comments. Apologies, I could not complete my review earlier on #1948!

.github/workflows/test.yml
tatiana: @pankajkoti I'm happy for us to drop support for dbt 1.5 in a follow-up PR. Would you like to do this?
I'm focused on our P1s now, and slightly apprehensive of add more issues to our on-growing backlog

PR #1969, opened 2025-09-15
Add docs note for param virtualenv_dir for async execution mode

pankajastro:
Add docs for PR: https://github.com/astronomer/astronomer-cosmos/pull/1939

Screenshot 2025-09-15 at 9 57 21 PM

### DIFF
--- a/docs/getting_started/async-execution-mode.rst
+++ b/docs/getting_started/async-execution-mode.rst
@@ -200,7 +200,7 @@ The ``run`` tasks will run asynchronously via the deferrable operator, freeing u
 
    1. The deferrable operator is currently supported for dbt models only when using BigQuery. Adding support for other adapters is on the roadmap.
 
-   2. By default, the ``SetupAsyncOperator`` creates and executes within a new isolated virtual environment for each run. To reuse an existing virtual environment, use the ``virtualenv_dir`` parameter within the ``operator_args`` of the ``DbtDag``.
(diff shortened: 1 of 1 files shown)


pankajkoti (approved):

docs/getting_started/async-execution-mode.rst
```diff
@@ -197,4 +197,13 @@ Create an Airflow connection with following configurations
 The ``run`` tasks will run asynchronously via the deferrable operator, freeing up worker slots while waiting on I/O or long-running tasks.
 
 .. note::
-    The deferrable operator is currently supported for dbt models only when using BigQuery. Adding support for other adapters is on the roadmap.
+
+   1. The deferrable operator is currently supported for dbt models only when using BigQuery. Adding support for other adapters is on the roadmap.
+
+   2. By default, the ``SetupAsyncOperator`` creates and executes within a new isolated virtual environment for each run. To reuse an existing virtual environment, use the ``virtualenv_dir`` parameter within the ``operator_args`` of the ``DbtDag``.
```
tatiana:
```suggestion
   2. By default, the ``SetupAsyncOperator`` creates and executes within a new isolated virtual environment for each task run, which can cause performance issues. To reuse an existing virtual environment, use the ``virtualenv_dir`` parameter within the ``operator_args`` of the ``DbtDag``.
```

tatiana: @pankajastro, could you add numbers on how much the performance can be improved? This was set up. Would it make sense for us to have a default virtual environment (venv) for the async setup task?

pankajastro: If we enable it by default, it may cause some permission issues, depending on the user's file system. Also, we will have to provide a default path. The current way I feel is consistent with the virtual env execution mode

PR #1981, opened 2025-09-17
Add Scarf metric to collect the execution mode uses

### LINKED ISSUE #1975: Track adoption for watcher execution mode using Scarf
Implement tracking of the execution_mode usage in Cosmos using Scarf
This will help us gain insights into the adoption and usage patterns of the execution mode feature across deployments.

Tracking execution mode usage will:
- Provide visibility into how widely the feature is being adopted.
- Inform future development and prioritization.

Related Work
- Initial Scarf integration PR: https://github.com/astronomer/astronomer-cosmos/pull/1397

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1975

This PR introduces tracking of the execution mode in `DbtDag` and `DbtTaskGroup` to help understand usage patterns and support better decision-making during development.

- Collecting execution modes used by `DbtDag` and `DbtTaskGroup`.
- Collect multiple execution modes per DbtTaskGroup (e.g., when using multiple DbtTaskGroups with different execution settings). Concatenating modes using `__ ` as a delimiter (e.g., kubernetes__local).
- Introducing a new versioned metrics endpoint to collect execution mode data while preserving support for the legacy format.
variable-downloads-execution_modes-astronomer-2025-09-18T21_35_51 537Z

### DIFF
--- a/cosmos/listeners/dag_run_listener.py
+++ b/cosmos/listeners/dag_run_listener.py
@@ -56,6 +56,9 @@ def get_execution_modes(dag: DAG) -> str:
         if (getattr(task, "_task_module", None) or task.__class__.__module__).startswith("cosmos.")
     }
 
+    if not modes:
+        return "none"
+
     # Sorted to ensure consistent and predictable output
     return "__".join(sorted(modes))
 
--- a/cosmos/listeners/dag_run_listener.py
+++ b/cosmos/listeners/dag_run_listener.py
@@ -56,9 +56,6 @@ def get_execution_modes(dag: DAG) -> str:
         if (getattr(task, "_task_module", None) or task.__class__.__module__).startswith("cosmos.")
     }
 
-    if not modes:
-        return "none"
-
     # Sorted to ensure consistent and predictable output
     return "__".join(sorted(modes))
 
--- a/cosmos/listeners/dag_run_listener.py
+++ b/cosmos/listeners/dag_run_listener.py
@@ -1,11 +1,15 @@
 from __future__ import annotations
 
 import hashlib
+from typing import TYPE_CHECKING
 
 from airflow import __version__ as airflow_version
 from airflow.listeners import hookimpl
-from airflow.models.dag import DAG
-from airflow.models.dagrun import DagRun
+
+if TYPE_CHECKING:
+    from airflow.models.dag import DAG
+    from airflow.models.dagrun import DagRun
+
 from packaging import version
 
 from cosmos import telemetry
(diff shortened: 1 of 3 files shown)


cosmos/listeners/dag_run_listener.py
```diff
@@ -44,7 +48,21 @@ def total_cosmos_tasks(dag: DAG) -> int:
...
+        (getattr(task, "_task_module", None) or task.__class__.__module__).split(".")[2]
+        for task in dag.task_dict.values()
+        if (getattr(task, "_task_module", None) or task.__class__.__module__).startswith("cosmos.")
+    }
+
+    if not modes:
+        return "none"
+
+    # Sorted to ensure consistent and predictable output
+    return "__".join(sorted(modes))
```
tatiana, line 60: Have you considered other alternatives to this approach?

tatiana (approved):
Hi @pankajastro,

Looks good, thanks for adding this!

Using the URL path to represent multi-valued fields may not be the ideal approach.

An alternative is using query strings. If we used a query string, we could log how many operators of each type are being used. However, since the Scarf UI doesn't display them, it would probably be worse for our analysis.

Have you considered other strategies?

Overall, we need to revisit our telemetry tooling - but I believe that, for now, this solves our immediate problem.

Happy for this to be merged once:
- the checks are green
- we have the expected coverage

cosmos/listeners/dag_run_listener.py
pankajastro: The string concatenation is here for a special case, only when we have more than one DbtTaskGroup in a DAG, and they have different execution modes.

I believe we will have small metrics for this case, and I feel this approach is reasonable, but please let me know if you have some other suggestions for this.

pankajastro:
> Using the URL path to represent multi-valued fields may not be the ideal approach.
> 
> An alternative is using query strings. If we used a query string, we could log how many operators of each type are being > used. However, since the Scarf UI doesn't display them, it would probably be worse for our analysis.

I did not find much documentation for the query param approach, and we have been using url path for other variables, but we can explore the query param approach. I have created an issue: https://github.com/astronomer/astronomer-cosmos/issues/1991 to track it.

cosmos/listeners/dag_run_listener.py
```diff
@@ -44,7 +48,21 @@ def total_cosmos_tasks(dag: DAG) -> int:
...
+    """Determine the execution mode(s) based on task modules in the DAG."""
+    modes = {
+        (getattr(task, "_task_module", None) or task.__class__.__module__).split(".")[2]
+        for task in dag.task_dict.values()
+        if (getattr(task, "_task_module", None) or task.__class__.__module__).startswith("cosmos.")
+    }
+
+    if not modes:
+        return "none"
+
```
pankajastro:
```suggestion
```

PR #1982, opened 2025-09-18
Add `DbtProducerWatcherOperator` for the proposed `ExecutionMode.WATCHER`

### LINKED ISSUE #1958: Add root dbt build task with event callbacks and XCOM Support [ExecutionMode.WATCHER]
Develop the root dbt build task, which is responsible for triggering the dbt build command and registering callback functions in the dbtRunner.invoke command to capture model events. The task acts as the root of the new execution mode, pushing relevant XComs for downstream watcher tasks.

related: #1950 

Acceptance Criteria:
- [ ] Implementation for the root task 
- [ ] Unit tests covering the change

pankajkoti:
This PR introduces the `DbtProducerWatcherOperator ` in 
`cosmos/operators/watcher.py` for use with the proposed `ExecutionMode.WATCHER`. 
The operator triggers a single dbt build and streams real-time per-model 
run statuses via dbtRunner events, pushing keys such as `nodefinished_<uid>`, 
aggregated `dbt_startup_events`, and falling back to pushing `run_results` 
from the target directory to XCom when dbtRunner is unavailable. 

Additionally, local execution has been updated with a 
`_push_run_results_to_xcom` helper and a `push_run_results_to_xcom` flag, 
enabling gzip+base64–compressed run-results to be stored in XCom for 
fallback support.

closes: #1958
closes: https://github.com/astronomer/oss-integrations-private/issues/238

### DIFF
--- a/cosmos/dbt/runner.py
+++ b/cosmos/dbt/runner.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import sys
 from functools import lru_cache
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Callable
 
 from cosmos.dbt.project import change_working_directory, environ
 from cosmos.exceptions import CosmosDbtRunError
@@ -39,7 +39,7 @@ def is_available() -> bool:
 
 
 @cache
-def get_runner() -> dbtRunner:
+def _get_cached_dbt_runner() -> dbtRunner:
     """
     Retrieves a dbtRunner instance.
     """
@@ -48,7 +48,21 @@ def get_runner() -> dbtRunner:
     return dbtRunner()
 
 
-def run_command(command: list[str], env: dict[str, str], cwd: str) -> dbtRunnerResult:
+def get_runner(callbacks: list[Callable] | None = None) -> dbtRunner:  # type: ignore[type-arg]
+    """
+    Retrieves a dbtRunner instance.
+    """
+    if callbacks and isinstance(callbacks, list):
+        from dbt.cli.main import dbtRunner
+
+        return dbtRunner(callbacks=callbacks)
+
+    return _get_cached_dbt_runner()
+
+
+def run_command(
+    command: list[str], env: dict[str, str], cwd: str, callbacks: list[Callable] | None = None  # type: ignore[type-arg]
+) -> dbtRunnerResult:
     """
     Invokes the dbt command programmatically.
     """
@@ -58,7 +72,7 @@ def run_command(command: list[str], env: dict[str, str], cwd: str) -> dbtRunnerR
     cli_args = command[1:]
     with change_working_directory(cwd), environ(env):
         logger.info("Trying to run dbtRunner with:\n %s\n in %s", cli_args, cwd)
-        runner = get_runner()
+        runner = get_runner(callbacks=callbacks)
         result = runner.invoke(cli_args)
     return result
 
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -579,7 +579,6 @@ class AbstractDbtLocalBase(AbstractDbtBase):
         self.log.info("Pushed run results to XCom")
 
     def _handle_post_execution(self, tmp_project_dir: str, context: Context, push_run_results_to_xcom: bool) -> None:
-        self.log.info("Handling post execution with push_run_results_to_xcom: %s", push_run_results_to_xcom)
         self.store_freshness_json(tmp_project_dir, context)
         self.store_compiled_sql(tmp_project_dir, context)
         self._override_rtif(context)
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -580,7 +580,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
 
     def _push_run_results_to_xcom(self, tmp_project_dir: str, context: Context) -> None:
         run_results_path = Path(tmp_project_dir) / "target" / "run_results.json"
-        if not run_results_path.is_file():
+        if not run_results_path.is_file():  # pragma: no cover
             raise AirflowException(f"run_results.json not found at {run_results_path}")
 
         try:
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import base64
+import gzip
 import inspect
 import json
 import os
@@ -555,13 +557,39 @@ class AbstractDbtLocalBase(AbstractDbtBase):
         if partial_parse_file.exists():
             cache._update_partial_parse_cache(partial_parse_file, self.cache_dir)
 
-    def _handle_post_execution(self, tmp_project_dir: str, context: Context) -> None:
+    def _push_run_results_to_xcom(self, tmp_project_dir: str, context: Context) -> None:
+        run_results_path = Path(tmp_project_dir) / "target" / "run_results.json"
+        if not run_results_path.is_file():
+            raise AirflowException(f"run_results.json not found at {run_results_path}")
+
+        try:
+            with run_results_path.open() as fp:
+                raw = json.load(fp)
+        except json.JSONDecodeError as exc:  # pragma: no cover
+            raise AirflowException("Invalid JSON in run_results.json") from exc
+        self.log.debug("Loaded run results from %s", run_results_path)
+
+        # results_mapping = {result["unique_id"]: result for result in raw.get("results", [])}
+        # compressed = base64.b64encode(gzip.compress(json.dumps(results_mapping).encode())).decode()
+        # self.log.debug("Parsed %d entries out of run_results.json", len(results_mapping))
+
+        compressed = base64.b64encode(gzip.compress(json.dumps(raw).encode())).decode()
+        context["ti"].xcom_push(key="run_results", value=compressed)
+
+        self.log.info("Pushed run results to XCom")
+
+    def _handle_post_execution(self, tmp_project_dir: str, context: Context, push_run_results_to_xcom: bool) -> None:
+        self.log.info("Handling post execution with push_run_results_to_xcom: %s", push_run_results_to_xcom)
         self.store_freshness_json(tmp_project_dir, context)
         self.store_compiled_sql(tmp_project_dir, context)
         self._override_rtif(context)
 
         if self.should_upload_compiled_sql:
             self._upload_sql_files(tmp_project_dir, "compiled")
+
+        if push_run_results_to_xcom:
+            self._push_run_results_to_xcom(tmp_project_dir, context)
+
         if self.callback:
             self.callback_args.update({"context": context})
             if isinstance(self.callback, list):
@@ -589,6 +617,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
         context: Context,
         run_as_async: bool = False,
         async_context: dict[str, Any] | None = None,
+        push_run_results_to_xcom: bool = False,
     ) -> FullOutputSubprocessResult | dbtRunnerResult | str:
         """
         Copies the dbt project to a temporary directory and runs the command.
@@ -642,7 +671,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
                 if self.partial_parse:
                     self._update_partial_parse_cache(tmp_dir_path)
 
-                self._handle_post_execution(tmp_project_dir, context)
+                self._handle_post_execution(tmp_project_dir, context, push_run_results_to_xcom)
                 self.handle_exception(result)
 
                 if run_as_async and async_context:
@@ -836,6 +865,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
         cmd_flags: list[str] | None = None,
         run_as_async: bool = False,
         async_context: dict[str, Any] | None = None,
+        **kwargs: Any,
     ) -> FullOutputSubprocessResult | dbtRunnerResult:
         # If this is an async run and we're using the setup task, make sure to include the full_refresh flag if set
         if run_as_async and settings.enable_setup_async_task and getattr(self, "full_refresh", False):
@@ -847,7 +877,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
         dbt_cmd, env = self.build_cmd(context=context, cmd_flags=cmd_flags)
         dbt_cmd = dbt_cmd or []
         result = self.run_command(
-            cmd=dbt_cmd, env=env, context=context, run_as_async=run_as_async, async_context=async_context
+            cmd=dbt_cmd, env=env, context=context, run_as_async=run_as_async, async_context=async_context, **kwargs
         )
         return result
 
@@ -948,10 +978,11 @@ class DbtBuildLocalOperator(DbtBuildMixin, DbtLocalBaseOperator):
 
         self.on_warning_callback and self.on_warning_callback(warning_context)
 
-    def execute(self, context: Context, **kwargs: Any) -> None:
-        result = self.build_and_run_cmd(context=context, cmd_flags=self.add_cmd_flags())
+    def execute(self, context: Context, **kwargs: Any) -> FullOutputSubprocessResult | dbtRunnerResult:
+        result = self.build_and_run_cmd(context=context, cmd_flags=self.add_cmd_flags(), **kwargs)
         if self.on_warning_callback:
             self._handle_warnings(result, context)
+        return result
 
 
 class DbtLSLocalOperator(DbtLSMixin, DbtLocalBaseOperator):
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -578,7 +578,9 @@ class AbstractDbtLocalBase(AbstractDbtBase):
 
         self.log.info("Pushed run results to XCom")
 
-    def _handle_post_execution(self, tmp_project_dir: str, context: Context, push_run_results_to_xcom: bool) -> None:
+    def _handle_post_execution(
+        self, tmp_project_dir: str, context: Context, push_run_results_to_xcom: bool = False
+    ) -> None:
         self.store_freshness_json(tmp_project_dir, context)
         self.store_compiled_sql(tmp_project_dir, context)
         self._override_rtif(context)
@@ -977,11 +979,10 @@ class DbtBuildLocalOperator(DbtBuildMixin, DbtLocalBaseOperator):
 
         self.on_warning_callback and self.on_warning_callback(warning_context)
 
-    def execute(self, context: Context, **kwargs: Any) -> FullOutputSubprocessResult | dbtRunnerResult:
+    def execute(self, context: Context, **kwargs: Any) -> None:
         result = self.build_and_run_cmd(context=context, cmd_flags=self.add_cmd_flags(), **kwargs)
         if self.on_warning_callback:
             self._handle_warnings(result, context)
-        return result
 
 
 class DbtLSLocalOperator(DbtLSMixin, DbtLocalBaseOperator):
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -581,13 +581,13 @@ class AbstractDbtLocalBase(AbstractDbtBase):
 
     def _push_run_results_to_xcom(self, tmp_project_dir: str, context: Context) -> None:
         run_results_path = Path(tmp_project_dir) / "target" / "run_results.json"
-        if not run_results_path.is_file():  # pragma: no cover
+        if not run_results_path.is_file():
             raise AirflowException(f"run_results.json not found at {run_results_path}")
 
         try:
             with run_results_path.open() as fp:
                 raw = json.load(fp)
-        except json.JSONDecodeError as exc:  # pragma: no cover
+        except json.JSONDecodeError as exc:
             raise AirflowException("Invalid JSON in run_results.json") from exc
         self.log.debug("Loaded run results from %s", run_results_path)
 
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import base64
-import gzip
 import inspect
 import json
 import os
@@ -592,11 +591,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
             raise AirflowException("Invalid JSON in run_results.json") from exc
         self.log.debug("Loaded run results from %s", run_results_path)
 
-        # results_mapping = {result["unique_id"]: result for result in raw.get("results", [])}
-        # compressed = base64.b64encode(gzip.compress(json.dumps(results_mapping).encode())).decode()
-        # self.log.debug("Parsed %d entries out of run_results.json", len(results_mapping))
-
-        compressed = base64.b64encode(gzip.compress(json.dumps(raw).encode())).decode()
+        compressed = base64.b64encode(zlib.compress(json.dumps(raw).encode())).decode()
         context["ti"].xcom_push(key="run_results", value=compressed)
 
         self.log.info("Pushed run results to XCom")
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -187,6 +187,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
         should_store_compiled_sql: bool = True,
         should_upload_compiled_sql: bool = False,
         append_env: bool = True,
+        dbt_runner_callbacks: list[Callable] | None = None,  # type: ignore[type-arg]
         **kwargs: Any,
     ) -> None:
         self.task_id = task_id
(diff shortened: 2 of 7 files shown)


cosmos/operators/local.py
```diff
@@ -555,13 +557,39 @@ def _update_partial_parse_cache(self, tmp_dir_path: Path) -> None:
...
+        # compressed = base64.b64encode(gzip.compress(json.dumps(results_mapping).encode())).decode()
+        # self.log.debug("Parsed %d entries out of run_results.json", len(results_mapping))
+
+        compressed = base64.b64encode(gzip.compress(json.dumps(raw).encode())).decode()
+        context["ti"].xcom_push(key="run_results", value=compressed)
+
+        self.log.info("Pushed run results to XCom")
+
+    def _handle_post_execution(self, tmp_project_dir: str, context: Context, push_run_results_to_xcom: bool) -> None:
+        self.log.info("Handling post execution with push_run_results_to_xcom: %s", push_run_results_to_xcom)
```
pankajkoti:
```suggestion
```

cosmos/operators/local.py
```diff
@@ -555,13 +557,38 @@ def _update_partial_parse_cache(self, tmp_dir_path: Path) -> None:
...
+                raw = json.load(fp)
+        except json.JSONDecodeError as exc:  # pragma: no cover
+            raise AirflowException("Invalid JSON in run_results.json") from exc
+        self.log.debug("Loaded run results from %s", run_results_path)
+
+        # results_mapping = {result["unique_id"]: result for result in raw.get("results", [])}
+        # compressed = base64.b64encode(gzip.compress(json.dumps(results_mapping).encode())).decode()
+        # self.log.debug("Parsed %d entries out of run_results.json", len(results_mapping))
+
+        compressed = base64.b64encode(gzip.compress(json.dumps(raw).encode())).decode()
```
pankajkoti: TODO: Decide if we want to process and push the results or push as is.

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
         "data": {"adapter_name": "bigquery", "adapter_version": "=1.9.0"},
     }
 }
+
+
+class DbtBuildCoordinatorOperator(DbtBuildLocalOperator):
```
tatiana: In the case of the `ExecutionMode.AIRFLOW_ASYNC`, we named the operator that pre-compiled SQL `SetupAsyncOperator`. We have a separate ticket to analyse whether we can make this task a setup task for the watcher ones, so it is unclear if it will ever become a setup task as per the Airlfow definition. Would it be worth making it consistent by calling it `SetupWatcherOperator`?

If not, I suggest we are consistent with other Cosmos operators (e.g. `DbtTestAirflowAsyncOperator`), where we have `Dbt<Feature><ExecutionMode>Operator`

One possibility:
- `DbtProducerWatcherOperator`

And we could then name the base consumer task implementation as:
- `DbtConsumerWatcherOperator`

I believe we should not have `Build` in the name because:
- We may want to use `dbt run` in some circumstances (e.g. depending on if and how we decide to implement `TestBehavior.AFTER_EACH`).
- It feels more like an implementation detail instead of a command we want to expose in our UI explicitly

That said, I know naming is one of the hardest things in computing - so I'm happy to hear other thoughts

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
...
+    """Run dbt build and coordinate model run statuses via XCom for *WATCHER* execution mode .
```
```suggestion
    """Run dbt build and update XCom with the progress of each model, as part of the *WATCHER* execution mode.
```

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
...
+          ``dbt_startup_events``;
+        – pushes each ``NodeFinished`` event immediately to XCom under
+          ``nodefinished_<unique_id>`` (gzipped+base64 JSON) so downstream
+          sensors can react with near-zero latency.
+
+    - **When ``dbtRunner`` is *not* available** (older dbt or
+      ``InvocationMode=SUBPROCESS``) we fallback to delayed strategy: after
+      dbt exits we read ``target/run_results.json`` and push the whole mapping
+      once under key ``run_results`` to XCom.  Sensors can poll this key but will not
+      get per-model updates until the build completes.
```
```suggestion
      get per-model updates until the build completes - by the end of the execution of all dbt nodes.
```

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
...
+    def __init__(
+        self,
+        *,
+        render_config: RenderConfig | None = None,
+        **kwargs: Any,
+    ) -> None:
+        # Store so we can honour select/exclude when building flags
+        self.render_config: RenderConfig | None = render_config
+
+        task_id = kwargs.pop("task_id", "dbt_build_coordinator")
```
tatiana: This should be renamed based on the conclusion we reach in https://github.com/astronomer/astronomer-cosmos/pull/1982/files#r2367925322

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
...
+        if self.render_config is not None and self.render_config.exclude:
+            flags.extend(["--exclude", *self.render_config.exclude])
+        if self.render_config is not None and self.render_config.select:
+            flags.extend(["--select", *self.render_config.select])
+        return flags
+
+    @staticmethod
+    def _serialize_event(ev: EventMsg) -> dict[str, Any]:
+        """Convert structured dbt EventMsg to plain dict."""
+        from google.protobuf.json_format import MessageToDict
```
tatiana, line 155: Why are we using a Google library here and a method from the protobuf subpackage?

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
         "data": {"adapter_name": "bigquery", "adapter_version": "=1.9.0"},
     }
 }
+
+
```
tatiana, line 120: A separate discussion from the naming: what are the pros and cons of this task inheriting from `DbtBuildLocalOperator`?

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
...
+      get per-model updates until the build completes.
+
+    This keeps the heavy dbt work centralised while providing near real-time
+    feedback and granular task-level observability downstream.
+    """
+
+    def __init__(
+        self,
+        *,
+        render_config: RenderConfig | None = None,
```
tatiana: We were trying to keep the `render_config` as much as possible separate from the DAG rendering itself, and avoid mixing things up.

While the `ExecutionMode.AIRFLOW_ASYNC` also uses render_config, the interface was slightly different, and `render_config` was not exposed as an argument of the operator, but as part of `task_args`:
https://github.com/astronomer/astronomer-cosmos/blob/869ebb887bb4a9da01d6fdc4d85b6361880bf3ce/cosmos/airflow/graph.py#L493-L498

Would it be worth to follow the same approach?

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
...
+    ) -> None:
+        self.log.debug("DbtBuildCoordinatorOperator: handling node finished event: %s", ev)
+        ti = context["ti"]
+        uid = ev.data.node_info.unique_id
+        ev_dict = self._serialize_event(ev)
+        payload = base64.b64encode(gzip.compress(json.dumps(ev_dict).encode())).decode()
+        ti.xcom_push(key=f"nodefinished_{uid.replace('.', '__')}", value=payload)
+
+    @contextmanager
+    def _patch_runner(self, callback: Callable[[EventMsg], None]) -> Any:
```
tatiana: Why do we need to patch the runner? A docstring could be very useful

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
...
+
+        def _patched_get_runner() -> Any:
+            from dbt.cli.main import dbtRunner
+
+            return dbtRunner(callbacks=[callback])
+
+        _dbt_runner_mod.get_runner = _patched_get_runner  # type: ignore[assignment]
+        if hasattr(original, "cache_clear"):
+            # Clear the cache of the original get_runner function to ensure that
+            # no stale cached runners are used after monkey-patching. This prevents
```
tatiana: Do we really need monkey-patching in this execution mode when using `dbtRunner`?

tatiana (commented):
Really excited to see this work advancing, @pankajkoti , thank you!

I left several questions in-line

cosmos/operators/watcher.py
pankajkoti: I’m not inclined to call it a Setup operator because what we’re doing isn’t just setting up — we’re also executing a command.

On the other hand, Producer–Consumer also doesn’t seem like the best fit. The “producer” here isn’t really generating something for later heavy-lift processing — the  “consumer” is mostly just sensing XCOM rather than performing substantial downstream work.

I do like the consistency argument with other Cosmos operators. How about `DbtCommandWatcherOperator `?

@pankajastro, would have more ideas on the naming of the operator? 

In the absence of more ideas, happy to settle on `DbtProducerWatcherOperator` or `DbtCommandWatcherOperator` or something similar.

cosmos/operators/watcher.py
pankajkoti: the base command that comes with it and all the functions and flow that come with `DbtLocalBaseOperator`. Also, happy to extend with `DbtLocalBaseOperator` and override the `base_cmd` method here.

cosmos/operators/watcher.py
pankajkoti: That appears to be the most reliable method for converting the dbt event to a dictionary. Other simple approaches using the json library were not reliable/usable. The library `protobuf` comes with the dbt-core [library](https://github.com/dbt-labs/dbt-core/blob/faeee357b1e5d366160e6de7d112ac6ebbbbcf3d/core/setup.py#L61).  I used the reference for this from the dbt-core [test](https://github.com/dbt-labs/dbt-core/blob/faeee357b1e5d366160e6de7d112ac6ebbbbcf3d/tests/unit/events/test_types.py#L2)

cosmos/operators/watcher.py
pankajkoti: Yes, all we need are the `select` and `exclude` filters to be passed to the DBT command to run. Happy to use them based on how we're planning to pass them while constructing the tasks.

cosmos/operators/watcher.py
pankajkoti: yes, we're registering the callback function by patching the dbtRunner that we use in the base class (local.py) https://github.com/astronomer/astronomer-cosmos/blob/869ebb887bb4a9da01d6fdc4d85b6361880bf3ce/cosmos/operators/local.py#L91

cosmos/operators/watcher.py
pankajkoti: yes, will add a docstring. reasoning is to register the callback functions with dbtRunner

cosmos/operators/watcher.py
```diff
@@ -90,3 +118,132 @@
...
+        startup_events: list[dict[str, Any]] = []
+
+        if use_events:
+
+            def _callback(ev: EventMsg) -> None:
+                name = ev.info.name
+                if name in {"MainReportVersion", "AdapterRegistered"}:
+                    self._handle_startup_event(ev, startup_events)
+                elif name == "NodeFinished":
+                    self._handle_node_finished(ev, context)
```
pankajastro, line 199: ~I'm not sure if I'm missing something here, but why do we have an else block here? Don't we need both the startup and finished events in XCOM?~

cosmos/operators/local.py
```diff
@@ -577,13 +578,40 @@ def _update_partial_parse_cache(self, tmp_dir_path: Path) -> None:
         if partial_parse_file.exists():
             cache._update_partial_parse_cache(partial_parse_file, self.cache_dir)
 
-    def _handle_post_execution(self, tmp_project_dir: str, context: Context) -> None:
+    def _push_run_results_to_xcom(self, tmp_project_dir: str, context: Context) -> None:
+        run_results_path = Path(tmp_project_dir) / "target" / "run_results.json"
+        if not run_results_path.is_file():  # pragma: no cover
+            raise AirflowException(f"run_results.json not found at {run_results_path}")
```
pankajastro, line 585: Do we need to raise this error? I think the user might intentionally disable it: https://docs.getdbt.com/reference/global-configs/json-artifacts

cosmos/operators/watcher.py
pankajastro: This operator executes a dbt command and produces metadata for downstream tasks.

I agree that we may not always use `dbt build` in future, and I also agree that this operator doesn't perform significant setup like the async execution setup task does.

While this particular operator doesn't watch but rather executes a command, we've agreed on using the "watcher" name for the execution mode. Given that, I'm happy with the name `DbtProducerWatcherOperator` or `DbtCommandWatcherOperator`.

tatiana: > I’m not inclined to call it a Setup operator because what we’re doing isn’t just setting up — we’re also executing a command.

We were also executing a command in the setup `ExecutionMode.AIRFLOW_ASYNC` - the main difference being that we were changing the behaviour of that command, instead of running the original one. In both cases, we share artefacts that are used by "subscriber" tasks. The downstream tasks enable end-users to view the lineage and retry.

> On the other hand, Producer–Consumer also doesn’t seem like the best fit. The “producer” here isn’t really generating something for later heavy-lift processing — the “consumer” is mostly just sensing XCOM rather than performing substantial downstream work.

The [Producer-Consumer](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern) design pattern does not define how much work each part should perform; it only describes the interface of communication. The fact that the consumer only updates the status of an Airflow task and exposes the SQL does not make it a lesser consumer than one that would do more work.

> I do like the consistency argument with other Cosmos operators. How about DbtCommandWatcherOperator ?

I find `DbtCommandWatcherOperator` ambiguous, because a command can represent both an instruction received via an external interface or an authoritative order.

> While this particular operator doesn't watch but rather executes a command, we've agreed on using the "watcher" name for the execution mode. Given that, I'm happy with the name DbtProducerWatcherOperator or DbtCommandWatcherOperator.

Happy with `DbtProducerWatcherOperator` or alternatives such as
- `DbtLeadWatcherOperator`
- `DbtInputWatcherOperator`
- `DbtPreWatcherOperator`
- `DbtStartWatcherOperator`
- `DbtControllerWatcherOperator`

> While this particular operator doesn't watch but rather executes a command, we've agreed on using the "watcher" name for the execution mode.

If the 'watcher' execution name is misleading, we could also discuss and rename this execution mode to 'PUBSUB' or something else, so it represents both operations that are involved in the execution.

cosmos/operators/watcher.py
tatiana: > That appears to be the most reliable method for converting the dbt event to a dictionary. Other simple approaches using the json library were not reliable/usable.

Why was using the JSON library not reliable/usable?

> The library protobuf comes with the dbt-core [library](https://github.com/dbt-labs/dbt-core/blob/faeee357b1e5d366160e6de7d112ac6ebbbbcf3d/core/setup.py#L61). I used the reference for this from the dbt-core [test](https://github.com/dbt-labs/dbt-core/blob/faeee357b1e5d366160e6de7d112ac6ebbbbcf3d/tests/unit/events/test_types.py#L2)

It feels risky to have our implementation dependent on users having dbt Core installed in the same Python virtualenv as Cosmos. This has not been a requirement so far, particularly due to dependency conflicts, and there will be scenarios (e.g. when we fallback to `run_results.json`) when it will not be available.

Protocol buffers can introduce an additional level of complexity - and if we decide for this path, we need to make sure we introduce this additional dependency to Cosmos itself.

cosmos/operators/watcher.py
tatiana: What would we lose by using `DbtLocalBaseOperator` instead of `DbtBuildLocalOperator`?
The benefit is that we could switch between using `run` and `build`, depending on the test behaviour, in future tasks.
BTW: just double-checking, we're handling `dbt deps` in one of those parent classes, right?

cosmos/operators/watcher.py
tatiana: We'll likely also need `selector`, in case end-users are using YAML-based selectors.

cosmos/operators/watcher.py
tatiana: My understanding from the [dbt docs example](https://docs.getdbt.com/reference/programmatic-invocations#registering-callbacks) is that we do not need to patch, that we could set `dbtRunner` callbacks while instantiating the class:
```
from dbt.cli.main import dbtRunner
from dbt_common.events.base_types import EventMsg

def print_version_callback(event: EventMsg):
    if event.info.name == "MainReportVersion":
        print(f"We are thrilled to be running dbt{event.data.version}")

dbt = dbtRunner(callbacks=[print_version_callback])
dbt.invoke(["list"])
```

Would the implementation be cleaner if we changed the interface of our `get_runner`/`
` functions, to accept callbacks:
https://github.com/astronomer/astronomer-cosmos/blob/8627d63b40b4ba50c139995e08901c1352a53707/cosmos/dbt/runner.py#L42-L48

And adapted the local operator to use it?
https://github.com/astronomer/astronomer-cosmos/blob/8627d63b40b4ba50c139995e08901c1352a53707/cosmos/operators/local.py#L473

cosmos/operators/watcher.py
tatiana: Monkey patching alters the behaviour of objects at runtime. Even though it can be a handy feature (as we observed in the `ExecutionMode.AIRFLOW_ASYNC` implementation), it can also make our code much more complicated to understand and debug. I suggest we avoid this pattern unless strict necessary.

cosmos/operators/watcher.py
```diff
@@ -1,4 +1,32 @@
-status_model_fhir_dbt_analytics_active_encounters_daily_nodefinished = {
+from __future__ import annotations
+
+import base64
+import gzip
```
pankajastro: Shall we use `zlib`? 
https://github.com/astronomer/astronomer-cosmos/blob/8627d63b40b4ba50c139995e08901c1352a53707/cosmos/dbt/graph.py#L12

cosmos/operators/watcher.py
pankajastro: Shall we keep this class private?

cosmos/operators/local.py
pankajkoti: yes, but if user disables it then our sensors won't be able to consume and sense the status. So, for this execution mode to use and when they also don't have dbtRunner available (hence cannot use dbt Events and need to use this fallback mode), run_results need to be available. WDYT?

cosmos/operators/watcher.py
pankajkoti: yes, using DbtLocalBaseOperator and dbt deps is handled in it.

cosmos/operators/watcher.py
pankajkoti: Decided to use `DbtProducerWatcherOperator` for now and later change the execution mode name upon consensus with the team.

pankajkoti: I am in favour of keeping this as public. So that we could also show in examples how to use this operator in individual mode if need be. It allows users to extend and customise the opreator too based on their use cases.

cosmos/operators/watcher.py
pankajkoti: yes, removed render_config, and initialising base/parent  class `DbtLocalBaseOperator` with (*args, **kwargs) so that the upstream implementation applies for those selectors.

cosmos/operators/watcher.py
pankajkoti: like discussed over call how json serde is unusable due to the output produced, and we only have this when dbt events path is used (meaning dbtRunner is available, hence dbt Core is available and protobuf comes with it), we're keeping it as is restricted to minimum scope of dbt events path and the import path for the library is local to the method(no top-level import necessiating installation of the protobuf dependency).

cosmos/operators/watcher.py
pankajkoti: removed the patching and added capability to pass callbacks to base get_runner() method

cosmos/operators/watcher.py
pankajkoti: yes removed this monkeypatching and added capability to base get_runner to accept callbacks

cosmos/dbt/runner.py
```diff
@@ -48,7 +48,21 @@ def get_runner() -> dbtRunner:
     return dbtRunner()
 
 
-def run_command(command: list[str], env: dict[str, str], cwd: str) -> dbtRunnerResult:
+def get_runner(callbacks: list[Callable] | None = None) -> dbtRunner:  # type: ignore[type-arg]
```
pankajkoti, line 51: adding a separate method here and moving the existing @cache interface to `_get_cached_dbt_runner` because having the `@cache` decorator complains that `callbacks` param is not hashable.

pankajkoti: traceback when adding `callbacks` param and updating the existing `get_runner(callbacks)` with `@cache` decorator:

```
ERROR - Task failed with exception: source="task"
TypeError: unhashable type: 'list'
File "/usr/local/lib/python3.12/site-packages/airflow/sdk/execution_time/task_runner.py", line 867 in run
File "/usr/local/lib/python3.12/site-packages/airflow/sdk/execution_time/task_runner.py", line 1159 in _execute_task
File "/usr/local/lib/python3.12/site-packages/airflow/sdk/bases/operator.py", line 397 in wrapper
File "/usr/local/astronomer-cosmos/cosmos/operators/watcher.py", line 224 in execute
File "/usr/local/astronomer-cosmos/cosmos/operators/base.py", line 316 in execute
File "/usr/local/astronomer-cosmos/cosmos/operators/local.py", line 906 in build_and_run_cmd
File "/usr/local/astronomer-cosmos/cosmos/operators/local.py", line 675 in run_command
File "/usr/local/astronomer-cosmos/cosmos/operators/local.py", line 555 in _install_dependencies
File "/usr/local/astronomer-cosmos/cosmos/operators/local.py", line 476 in run_dbt_runner
File "/usr/local/astronomer-cosmos/cosmos/dbt/runner.py", line 64 in run_command
```

cosmos/operators/local.py
pankajkoti: I will leave this as is in the PR and let @pankajastro decide in his PR on how we wants to produce and consume the `run_results` XCOM

cosmos/operators/watcher.py
pankajkoti: yes, nice suggestion. using zlib now.

cosmos/dbt/runner.py
tatiana: Thanks, @pankajkoti , looks good!

cosmos/operators/local.py
tatiana: Perhaps in a follow-up PR, we could have some validation to confirm users are not passing the `--no-write-json` flag - if they do, we should error. @pankajkoti, could you log the task? Happy for this comment to be resolved once we have the ticket link here

cosmos/operators/local.py
```diff
@@ -577,13 +579,36 @@ def _update_partial_parse_cache(self, tmp_dir_path: Path) -> None:
...
+
+        try:
+            with run_results_path.open() as fp:
+                raw = json.load(fp)
+        except json.JSONDecodeError as exc:  # pragma: no cover
```
tatiana: Why aren't we covering this?

tatiana (approved):
Looks great, @pankajkoti , thanks a lot for the great work and addressing the feedback.
I'm happy for us to merge this PR once the follow up ticket is logged.

cosmos/operators/local.py
pankajkoti: okay, thanks. Logged a ticket: https://github.com/astronomer/astronomer-cosmos/issues/1994

cosmos/operators/local.py
pankajkoti: Initially I considered it less significant to cover, but have added tests now and removed the pragmas in commit: https://github.com/astronomer/astronomer-cosmos/pull/1982/commits/1a8056b12e584b296208b0fcada7e915b74c7718

cosmos/operators/watcher.py
```diff
@@ -90,3 +116,94 @@
...
+        raw_ts = getattr(info, "ts", None)
+        ts_val = raw_ts.ToJsonString() if hasattr(raw_ts, "ToJsonString") else str(raw_ts)  # type: ignore[union-attr]
+        startup_events.append({"name": info.name, "msg": info.msg, "ts": ts_val})
+
+    def _handle_node_finished(
+        self,
+        ev: EventMsg,
+        context: Context,
+    ) -> None:
+        self.log.debug("DbtBuildCoordinatorOperator: handling node finished event: %s", ev)
```
```suggestion
        self.log.debug("DbtProducerWatcherOperator: handling node finished event: %s", ev)
```

cosmos/operators/watcher.py
```diff
@@ -90,3 +116,94 @@
...
+        # Only push startup events; per-model statuses are available via individual nodefinished_<uid> entries.
+        if startup_events:
+            ti.xcom_push(key="dbt_startup_events", value=startup_events)
+
+    def execute(self, context: Context, **kwargs: Any) -> Any:
+        if not self.invocation_mode:
+            self._discover_invocation_mode()
+
+        use_events = self.invocation_mode == InvocationMode.DBT_RUNNER and EventMsg is not None
+        self.log.debug("DbtBuildCoordinatorOperator: use_events=%s", use_events)
```
```suggestion
        self.log.debug("DbtProducerWatcherOperator: use_events=%s", use_events)
```

PR #1986, opened 2025-09-19
Use official postgres image from Docker hub for kubernetes setup

pankajkoti:
## Checklist

### DIFF
--- a/scripts/test/postgres-deployment.yaml
+++ b/scripts/test/postgres-deployment.yaml
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: postgres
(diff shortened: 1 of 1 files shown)


pankajkoti:
Other test failures are due to another issue with `hatch,` which is being resolved in #1987

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,change-postgres-image-k8s-tests]
```
```suggestion
    branches: [main]
```

PR #1988, opened 2025-09-19
Support to prune sources without downstream references in dbt projects

corsettigyg:
## Description

In cosmos, once the sources are rendered into our dag, an issue that might arise is that if we are selecting only a portion of our project or if we have dead-code in the sources.yml file, tasks for these unchecked sources will still be created, which will essentially pollute the UI. 
Screenshot 2025-09-19 at 12 56 21

For large projects with hundreds of sources, users often select a fraction of the main project per DbtDag object, but all sources being rendered at once essentially means that the project is unreadable. The idea is to allow users to pass a `source_prunning` flag to RenderConfig, which will essentially remove source tasks without downstream dependencies

Screenshot 2025-09-19 at 13 00 58

## Breaking Change?

Should not be
## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,9 +40,7 @@ jobs:
           python-version: "3.10"
           architecture: "x64"
 
-      # Remove the `click` installation(which currently has an upper bound 8.3.0) once the
-      # issue https://github.com/pypa/hatch/issues/2050 is resolved.
-      - run: pip3 install hatch "click<8.3.0"
+      - run: pip3 install "hatch>=1.14.2"
       - run: hatch run tests.py3.10-2.10-1.9:type-check
 
   Run-Unit-Tests:
@@ -98,11 +96,9 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install packages and dependencies
-        # Remove the `click` installation(which currently has an upper bound 8.3.0) once the
-        # issue https://github.com/pypa/hatch/issues/2050 is resolved.
         run: |
           python -m pip install uv
-          uv pip install --system hatch "click<8.3.0"
+          uv pip install --system "hatch>=1.14.2"
           hatch -e tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }}-${{ matrix.dbt-version }} run pip freeze
 
       - name: Test Cosmos against Airflow ${{ matrix.airflow-version }}, Python ${{ matrix.python-version }} and dbt ${{ matrix.dbt-version }}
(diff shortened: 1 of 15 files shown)


corsettigyg:
At Getyourguide we have an internal logic  of `source_pruning` that loops over the airflow tasks to fetch the upstream tasks of our source tasks and remove the source tasks that have none

## Initial approach
I tried implementing a `get_downstream` method that would allow a node to fetch all of its downstream references. it ended up being harder than i thought due to the fact it ends up being a O(n) operation and even by caching, the code got dirty very fast

## Current approach
I am trying to leverage the filtered_nodes param. lets see if it works

@tatiana @pankajastro Once you have some time, can you take a look? I will also iterate over it later, but early thoughts are appreciated 😄

tatiana:
HI @corsettigyg, thanks a lot for contributing this, it seems a valuable addition to Cosmos.

I know it is in draft mode - and I'll avoid adding any further comments right now - but I'd love it if we could get some early thoughts from @arojasb3 , who initially implemented the source-node support.

corsettigyg:
@tatiana ready for review now 💯 tested it out in many different scenarios and it is working fine. my only small concern would be with performance since it is yet a loop for the sources, which can end up being kinda expensive.
Screenshot 2025-09-23 at 14 33 55

tests/dbt/test_graph.py
```diff
@@ -654,7 +654,7 @@ def test_load_via_dbt_ls_with_exclude(postgres_profile_config):
 @pytest.mark.integration
 @pytest.mark.parametrize(
     "project_dir,node_count",
-    [(DBT_PROJECTS_ROOT_DIR / ALTERED_DBT_PROJECT_NAME, 39), (DBT_PROJECTS_ROOT_DIR / "jaffle_shop_python", 28)],
+    [(DBT_PROJECTS_ROOT_DIR / ALTERED_DBT_PROJECT_NAME, 40), (DBT_PROJECTS_ROOT_DIR / "jaffle_shop_python", 28)],
```
corsettigyg, line 657: For reference, this is because I added a new source

dev/dags/example_source_pruning.py
```diff
@@ -0,0 +1,54 @@
...
+    operator_args={
+        "install_deps": True,  # install any necessary dependencies before running any dbt command
+        "full_refresh": True,  # used only in dbt commands that support this flag
+    },
+    render_config=RenderConfig(
+        source_rendering_behavior=SourceRenderingBehavior.ALL,
+        source_pruning=True,
+        exclude=[
+            "ｍｕｌｔｉｂｙｔｅ"
+        ],  # Exclude multibyte model to avoid Airflow 3.0 AssetAlias ASCII validation issues
```
corsettigyg, line 44: had to exclude this model since the sources are defined in the altered_jaffle_shop and the `m u l t i b y t e` source actually does not work with airflow 3 :(

tatiana: This is quite annoying, but you found a good workaround for now.
We have a ticket related to the underlining issue:
https://github.com/astronomer/astronomer-cosmos/issues/1802

pankajastro:
Overall, this looks good to me, and I'm happy for it to be merged once @tatiana's feedback is addressed.
A couple of questions out of curiosity:
- Is it common for dbt projects to leave dead code in `sources.yml`?
- Does the current source rendering respect the `--select` flag?

corsettigyg:
@tatiana 
1. it is already disabled by default 💯   python short-circuits the check if source_pruning is False (default)
2. no strong opinions here. I see the advantage to have it as a global config particularly speaking so i can implement it as such 😄 

@pankajastro 
1. Yes, especially when you have a big dbt project and you select per tags. You end up with a LOT of not used sources
2. yes

tatiana:
@corsettigyg just saw you're on holidays until the 8th - I'll merge this PR, you'll have lots of opportunity to do the same in the future!

Issue #1959, opened 2025-09-15
Guarantee the build task is run before sensors [ExecutionMode.WATCHER]


pankajkoti:
Ensure that the dbt build root task always runs before watcher/sensor tasks begin consuming XCom events. This guarantees that watcher tasks don’t start waiting on events from an uninitialized run.

Explore the following potential ideas to achieve this:
1. Airflow priority weight is set on the build task.
2. Trigger rules (always?) configured to enforce execution order.

related: #1950 

Acceptance Criteria:
- [ ] The root dbt build task always runs before sensor tasks

Issue #1972, opened 2025-09-16
Support overriding operator_args for build_operator


pankajastro:
Extend support for [operator_args](https://astronomer.github.io/astronomer-cosmos/configuration/operator-args.html) to build tasks. This will allow users to customise operator-level parameters, such as timeout and other dbt parameters, for execution settings specific to build task behaviour.

PR #1989, opened 2025-09-19
Add configurable `watcher_build_coordinator_priority_weight` making run build operator with priority

pankajkoti:
The PR introduces `[cosmos] watcher_build_coordinator_priority_weight` Airflow setting (default 9999) in `cosmos/settings.py`.
`DbtBuildCoordinatorOperator` now reads this value and passes it to `BaseOperator(priority_weight=…)`, ensuring the build-coordinator task runs before others.

closes: #1959 
closes: https://github.com/astronomer/oss-integrations-private/issues/237

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -579,7 +579,6 @@ class AbstractDbtLocalBase(AbstractDbtBase):
         self.log.info("Pushed run results to XCom")
 
     def _handle_post_execution(self, tmp_project_dir: str, context: Context, push_run_results_to_xcom: bool) -> None:
-        self.log.info("Handling post execution with push_run_results_to_xcom: %s", push_run_results_to_xcom)
         self.store_freshness_json(tmp_project_dir, context)
         self.store_compiled_sql(tmp_project_dir, context)
         self._override_rtif(context)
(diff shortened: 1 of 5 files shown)


pankajkoti:
> Hi @pankajkoti !
> 
> Thanks for adding this.
> 
> Cosmos users are already overwhelmed with configurations. The implementation in [c48cd85](https://github.com/astronomer/astronomer-cosmos/commit/c48cd850172ba1ae15935627b4a81d8e0b27b449) gives users yet another configuration.
> 
> Let's decide on a priority for the runner task - which should be higher than the priority we set for the watcher tasks - and we can just hard-code them.
> 
> Ultimatelly, the problem we're trying to solve is to make sure the watchers do not start before the runner task - and hard-coding values would be good enough.

In this case, the configuration comes with a default value, so most users won’t ever need to touch it. By contrast, if we hard-code it, we leave no room for advanced users who actively use task priorities to adapt things to their deployment setup. This knob is really aimed at that group, not the average user.

Airflow itself exposes a wide surface of configurations, and as long as we provide sensible defaults and good documentation, I don’t think the presence of this option adds friction. Users can safely ignore it unless they have a reason not to.

That said, if there’s a strong consensus that it should be hard-coded, I’m okay to remove it as a configurable parameter — I just wanted to highlight the tradeoff for those advanced cases.

tatiana:
Hi @pankajkoti , thanks for sharing the benefits of exposing this as a configuration. 
One of our following goals will be to simplify Cosmos and reduce the number of configurations, thereby improving adoption and allowing us to focus on areas that deliver the most value. We can see from support calls that several users feel lost due to the numerous configurations available.
I remain in favour of hard-coding this value for now.
Users will be able to override this value as part of #1972, without the need for a custom Airflow configuration.

pankajkoti:
Closing in favour of https://github.com/astronomer/astronomer-cosmos/pull/1995

PR #1995, opened 2025-09-25
Add default priority_weight for DbtProducerWatcherOperator

### LINKED ISSUE #1959: Guarantee the build task is run before sensors [ExecutionMode.WATCHER]

pankajkoti:
Add a high-value `priority_weight` (9999) to the `DbtProducerWatcherOperator`
so that we guarantee this task runs before any sensor tasks in the DAG that
would have the default `priority_weight=1` 

Refer: https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/priority-weight.html

closes: #1959 
closes: https://github.com/astronomer/oss-integrations-private/issues/237

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -23,6 +23,8 @@ except ImportError:  # pragma: no cover
 logger = logging.getLogger(__name__)
 
 
+PRODUCER_OPERATOR_DEFAULT_PRIORITY_WEIGHT = 9999
+
 # Example dbt event JSON dictionaries (kept for reference)
 nodefinished_model__fhir_dbt_utils__fhir_table_list = {
     "info": {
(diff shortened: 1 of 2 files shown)


cosmos/operators/watcher.py
```diff
@@ -147,6 +147,7 @@ class DbtProducerWatcherOperator(DbtLocalBaseOperator):
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         task_id = kwargs.pop("task_id", "dbt_producer_watcher_operator")
+        kwargs["priority_weight"] = kwargs.get("priority_weight", 9999)
```
tatiana: @pankajkoti let's add the value 9999 as a constant on the top of the file

tatiana (approved):
A minor suggestion - but looks good, thanks, @pankajkoti !

cosmos/operators/watcher.py
pankajkoti: okay moved to the top of the file as a constant

PR #1997, opened 2025-09-25
Cleanup example dbt event JSON dictionaries kept for XCOM reference

pankajkoti:
Now that PR #1982 is merged and we have the XCOM getting populated, we can clean up the example dbt event payload dictionaries kept in the watcher.py module. If needed to refer again, we can refer to the previous PR #1952 that added these examples.

related: #1952 
related: #1950


tatiana (approved):
Thanks for the clean-up, @pankajkoti !

PR #1998, opened 2025-09-25
Add `DbtConsumerWatcherSensor` for the proposed `ExecutionMode.WATCHER`

### LINKED ISSUE #1971: Add Watcher task for Execution Mode Watcher to monitor models status
Implement a sensor to monitor the status of DBT models triggered via the `dbt build` command on the root dbt build task/node in a DAG. This will allow workflows to wait for DBT execution to complete and respond accordingly.

Related Issues
- https://github.com/astronomer/astronomer-cosmos/issues/1964
- https://github.com/astronomer/astronomer-cosmos/issues/1950

Proof of Concept
- https://github.com/astronomer/astronomer-cosmos/compare/single-run-execution-mode?expand=1

Acceptance Criteria:
- [x] Implementation for the sensor 
- [x] Unit tests covering the change

pankajastro:
This PR add `DbtConsumerWatcherSensor` for the proposed ExecutionMode.WATCHER.

Introduced a new `DbtConsumerWatcherSensor` class that extends BaseSensorOperator and reuses `DbtRunLocalOperator` functionality to monitor and handle the status of a dbt model execution triggered by a coordinating task.

- Added DbtConsumerWatcherSensor:
- Pulls XComs from the upstream master_task_id to check dbt model status.
- Handles task retries by reconstructing and re-running dbt commands with sanitized flags.
- Implemented _filter_flags helper to exclude incompatible flags like --select or --exclude.
- Decodes and decompresses dbt event messages from XComs to inspect model status.
- Logs detailed information for debugging and visibility during execution and retries.
- Raises AirflowException if the model finishes with a non-success status.

Airflow task to test with jaffle-shop
```python
sensor = DbtConsumerWatcherSensor(
        task_id="jaffle_shop__customers",
        producer_task_id="dbt_build",
        poke_interval=1,
        profile_config=bigquery_db,
        project_dir=DBT_PROFILES_DIR,
        extra_context={"dbt_node_config": {"unique_id": "model.jaffle_shop.customers"}}
    )
```

closes: https://github.com/astronomer/astronomer-cosmos/issues/1971

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -132,7 +132,6 @@ class DbtNodeStatusSensor(BaseSensorOperator, DbtRunLocalOperator):  # type: ign
     def __init__(
         self,
         *,
-        model_unique_id: str,
         profile_config: ProfileConfig | None = None,
         project_dir: str | None = None,
         profiles_dir: str | None = None,
@@ -141,6 +140,7 @@ class DbtNodeStatusSensor(BaseSensorOperator, DbtRunLocalOperator):  # type: ign
         timeout: int = 60 * 60,  # 1 h safety valve
         **kwargs: Any,
     ) -> None:
+        extra_context = kwargs.pop("extra_context") if "extra_context" in kwargs else {}
         super().__init__(
             poke_interval=poke_interval,
             timeout=timeout,
@@ -149,7 +149,7 @@ class DbtNodeStatusSensor(BaseSensorOperator, DbtRunLocalOperator):  # type: ign
             profiles_dir=profiles_dir,
             **kwargs,
         )
-        self.model_unique_id = model_unique_id
+        self.model_unique_id = extra_context.get("dbt_node_config", {}).get("unique_id")
         self.master_task_id = master_task_id
 
     @staticmethod
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import ast
 import base64
 import json
 import logging
@@ -211,7 +210,7 @@ class DbtConsumerWatcherSensor(BaseSensorOperator, DbtRunLocalOperator):  # type
 
         compressed_bytes = base64.b64decode(compressed_b64_event_msg)
         event_json_str = zlib.decompress(compressed_bytes).decode("utf-8")
-        event_json = ast.literal_eval(event_json_str)
+        event_json = json.loads(event_json_str)
 
         self.log.info("Node Info: %s", event_json_str)
 
@@ -225,7 +224,7 @@ class DbtConsumerWatcherSensor(BaseSensorOperator, DbtRunLocalOperator):  # type
 
         compressed_bytes = base64.b64decode(compressed_b64_run_results)
         run_results_str = zlib.decompress(compressed_bytes).decode("utf-8")
-        run_results_json = ast.literal_eval(run_results_str)
+        run_results_json = json.loads(run_results_str)
 
         self.log.debug("Run results: %s", run_results_json)
 
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import ast
 import base64
 import json
 import logging
@@ -12,8 +13,15 @@ if TYPE_CHECKING:  # pragma: no cover
     except ImportError:
         from airflow.utils.context import Context  # type: ignore[attr-defined]
 
+try:
+    from airflow.sdk.bases.sensor import BaseSensorOperator
+except ImportError:
+    from airflow.sensors.base import BaseSensorOperator
+from airflow.exceptions import AirflowException
+
+from cosmos.config import ProfileConfig
 from cosmos.constants import InvocationMode
-from cosmos.operators.local import DbtLocalBaseOperator
+from cosmos.operators.local import DbtLocalBaseOperator, DbtRunLocalOperator
 
 try:
     from dbt_common.events.base_types import EventMsg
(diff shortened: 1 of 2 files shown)


cosmos/operators/watcher.py
```diff
@@ -210,3 +218,150 @@ def _callback(ev: EventMsg) -> None:
...
         return super().execute(context=context, **kwargs)
+
+
+class DbtNodeStatusSensor(BaseSensorOperator, DbtRunLocalOperator):  # type: ignore[misc]
+    template_fields = ("model_unique_id",)
+
+    def __init__(
+        self,
+        *,
+        model_unique_id: str,
```
tatiana: As part of `kwargs`, Cosmos tasks receive the key `'extra_context'`, that contains {'dbt_node_config': {'unique_id': 'model.jaffle_shop.customers'}

Can we leverage this information instead of having a custom field?

pankajastro: Ok, how will you pass the `master_task_id` in `extra_context` dict?

cosmos/operators/watcher.py
```diff
@@ -210,3 +218,150 @@ def _callback(ev: EventMsg) -> None:
...
+        )
+
+        if not self.invocation_mode:
+            self._discover_invocation_mode()
+        use_events = self.invocation_mode == InvocationMode.DBT_RUNNER and EventMsg is not None
+
+        if use_events:
+            status = self._get_status_from_events(ti)
+        else:
+            status = self._get_status_from_run_results(ti)
```
pankajkoti, line 266: I wanted to discuss whether we should rely on dbt Runner being available when making that check, or if we should add fallbacks in XCOM. Specifically, the fallback would first check the dedicated node_finished XCOM for the model, and if that isn’t present, then check the run_result XCOM.

I’m not sure if this situation could arise, but if users customise and deploy different nodes with different images—for example, the Producer image has dbt Runner installed while the Consumer image does not—then relying solely on this approach could be misleading.

pankajastro: are you saying the user can run the sensor with the DBT runner and build with subprocess, or vice versa? I'm curious why they can't run both using either subprocess or the runner.

pankajkoti: Yes, what I mean to say is the sensor is not doing any dbt related work. So it does not need to have dbt installed in the same environment as Airflow unlike the producer that may have dbt installed to leverage dbtRunner to produce events (I am thinking of a scenario where the build task uses Node A with image having dbt installed, and the sensors running on Node B that may not have dbt installed because they don't need it)

pankajastro: In case of sensor retry we use dbt, no?

pankajkoti: Oh, right, it will need for retry, but not in usual scenarios for just sensing the XCOMs. Even for retries, they could use a subprocess, as that involves running only a single model, so dbt Runner might not be needed. I am not sure if this is how users would use it, but just bringing it up for discussion that should we assume the invocation mode is the same across producer and consumers or not? Given that we're planning to work on https://github.com/astronomer/oss-integrations-private/issues/225, users might be able to override the invocation mode across those (say dbtRunner for producer and subprocess for sensors)

pankajastro:
PR: https://github.com/astronomer/astronomer-cosmos/pull/2003 to fix type check

cosmos/operators/watcher.py
pankajastro: @tatiana, any input on this

cosmos/operators/watcher.py
pankajastro: Fixed the model_unique_id

cosmos/operators/watcher.py
```diff
@@ -210,3 +218,151 @@ def _callback(ev: EventMsg) -> None:
...
+class DbtNodeStatusSensor(BaseSensorOperator, DbtRunLocalOperator):  # type: ignore[misc]
+    template_fields = ("model_unique_id",)
+
+    def __init__(
+        self,
+        *,
+        profile_config: ProfileConfig | None = None,
+        project_dir: str | None = None,
+        profiles_dir: str | None = None,
+        master_task_id: str = "dbt_producer_watcher",
```
tatiana: We should replace any references from `master` to `producer`
```suggestion
        producer_task_id: str = "dbt_producer_watcher",
```

cosmos/operators/watcher.py
```diff
@@ -210,3 +218,150 @@ def _callback(ev: EventMsg) -> None:
...
+
+        self.log.info(
+            "Pulling status from task_id '%s' for model '%s'",
+            self.master_task_id,
+            self.model_unique_id,
+        )
+
+        if not self.invocation_mode:
+            self._discover_invocation_mode()
+        use_events = self.invocation_mode == InvocationMode.DBT_RUNNER and EventMsg is not None
```
tatiana, line 261: It's probably worth adding a comment that we are deciding on which XCom to use indirectly based on the `InvocationMode.DBT_RUNNER` - we may want to revisit this in the future if we see this is not always the case, in order to decouple further the producer from the consumers

cosmos/operators/watcher.py
tatiana: We discussed this and we decided that, for now, we'll tackle  this via: https://github.com/astronomer/astronomer-cosmos/pull/1998/files/1cb3649d2ed27c7036f58f75d5bdc35b25d5ef93#r2381640154

cosmos/operators/watcher.py
```diff
@@ -210,3 +218,151 @@ def _callback(ev: EventMsg) -> None:
         # Fallback – push run_results.json via base class helper
         kwargs["push_run_results_to_xcom"] = True
         return super().execute(context=context, **kwargs)
+
+
+class DbtNodeStatusSensor(BaseSensorOperator, DbtRunLocalOperator):  # type: ignore[misc]
```
```suggestion
class DbtConsumerWatcherSensor(BaseSensorOperator, DbtRunLocalOperator):  # type: ignore[misc]
```

cosmos/operators/watcher.py
pankajastro: Added a comment about assumption

cosmos/operators/watcher.py
pankajastro: Added a comment about the assumption

cosmos/operators/watcher.py
```diff
@@ -116,3 +124,151 @@ def _callback(ev: EventMsg) -> None:
...
+
+        node_finished_key = f"nodefinished_{self.model_unique_id.replace('.', '__')}"
+        compressed_b64_event_msg = ti.xcom_pull(task_ids=self.producer_task_id, key=node_finished_key)
+
+        if not compressed_b64_event_msg:
+            return None
+
+        compressed_bytes = base64.b64decode(compressed_b64_event_msg)
+        event_json_str = zlib.decompress(compressed_bytes).decode("utf-8")
+        event_json = ast.literal_eval(event_json_str)
```
tatiana: @pankajastro why we are using `ast.literal_eval` instead of something like `json.loads`?

tatiana (approved):
Excellent work, @pankajastro , one last question - and happy for the PR to be merged once that's discussed!

cosmos/operators/watcher.py
pankajastro: vibe coding!! yes json should work

PR #1993, opened 2025-09-23
Add `DbtModelStatusSensor` for the proposed `ExecutionMode.WATCHER`

pankajastro:
This PR add `DbtModelStatusSensor` for the proposed ExecutionMode.WATCHER.

Introduced a new `DbtModelStatusSensor` class that extends BaseSensorOperator and reuses `DbtRunLocalOperator` functionality to monitor and handle the status of a dbt model execution triggered by a coordinating task.

- Added DbtModelStatusSensor:
- Pulls XComs from the upstream master_task_id to check dbt model status.
- Handles task retries by reconstructing and re-running dbt commands with sanitized flags.
- Implemented _filter_flags helper to exclude incompatible flags like --select or --exclude.
- Decodes and decompresses dbt event messages from XComs to inspect model status.
- Logs detailed information for debugging and visibility during execution and retries.
- Raises AirflowException if the model finishes with a non-success status.

Airflow task to test with jaffle-shop
```python
sensor = DbtModelStatusSensor(
        task_id="jaffle_shop__customers",
        model_unique_id="model.jaffle_shop.customers",
        master_task_id="dbt_build",
        poke_interval=1,
        profile_config=bigquery_db,
        project_dir=DBT_PROFILES_DIR,

    )
```

closes: https://github.com/astronomer/astronomer-cosmos/issues/1971

Issue #1973, opened 2025-09-16
Support operator_args for watcher tasks


pankajastro:
Extend support for [operator_args](https://astronomer.github.io/astronomer-cosmos/configuration/operator-args.html) to watcher tasks in Cosmos. This will allow users to customise operator-level parameters such as timeout, poke_interval, and other execution settings specific to watcher behaviour.

why?
- Longer or shorter sensor timeouts.
- Custom poke_interval values.
- Other Airflow operator-level overrides as needed.

tatiana:
As part of Cosmos 1.12.0, we'll have:
- `operator_args` being sent to both watcher sensor and producer tasks
- `setup_operator_args` being setin to producer tasks

This allows users to customise behaviour for these different node types. If we need any further adjustment, we can make it, but this fulfills the initial goal with this ticket.

Issue #245, opened 2023-04-18
Build base images for running dbt in Docker


chrishronek:
As it exists today, we expect customers to bring their own images, we should try to abstract this where possible.

PR #1999, opened 2025-09-25
Introduce `ExecutionMode.WATCHER` to reduce DAG run time by 1/5

### LINKED ISSUE #1959: Guarantee the build task is run before sensors [ExecutionMode.WATCHER]

### LINKED ISSUE #1964: Introduce `ExecutionMode.WATCHER`
Expose the `ExecutionMode.WATCHER` that will instantiate the operators created in:
- https://github.com/astronomer/astronomer-cosmos/issues/1958
- https://github.com/astronomer/astronomer-cosmos/issues/1971

Following the proposal described in #1950.

tatiana:
Introduce a new high-performance execution mode, named `ExecutionMode.WATCHER`, following the implementation of the producer and consumer operators in PRs #1982, #1993 and #1998.

Initial performance analysis indicates that this mode will reduce the total DAG Run time to execute some dbt pipelines in Airflow to 1/5 of the original time. For example, if a Cosmos `DbtDag` takes 5 minutes to run with the default `ExecutionMode.LOCAL`, it will now run in 1 minute with the new `ExecutionMode.WATCHER`. The performance gain will be comparable to running the dbt CLI locally.

In the near future, there will also be benefits related to CPU and memory utilisation, as users will be able to run the producer task on a more powerful node with increased CPU and memory resources. In comparison, the consumer nodes can have less CPU and memory. Further development (#1972 and #1973), testing, and analysis are needed to evaluate this.

# Context

As of Cosmos 1.10, when users leverage the default `ExecutionMode.LOCAL`, each dbt model becomes an Airflow run task, and dbt is invoked in each of those tasks.

We noticed that the cost to run the same pipeline with plain dbt core varies significantly by running:
- the whole dbt command using a single command
- running one dbt command per model

For example, for the https://github.com/google/fhir-dbt-analytics project, these numbers were, on average, 5 minutes and 30 seconds (by running a single `dbt run` for the whole pipeline) versus 32 minutes (when using 184 `dbt run` commands as illustrated in https://gist.github.com/tatiana/c7831173ab09bf05d88839fb0b557920).

Similar to the [`ExecutionMode.AIRFLOW_ASYNC`](https://astronomer.github.io/astronomer-cosmos/getting_started/async-execution-mode.html), this mode aims to reduce the number of times the dbt command is invoked, while still allowing users to have observability of the dbt workflow via the Airflow UI and being able to retry individual tasks.

# Overall solution

* Use existing Cosmos DAG rendering techniques - implemented in this PR
* Have a single Airlfow task to run "all the pipeline" (selected by the user) - implemented in #1982
* Use dbt Core callbacks https://docs.getdbt.com/reference/programmatic-invocations#registering-callbacks to track how the model's execution is progressing and update different Xcoms (one Xcom per model) - implemented in #1982
* All the other tasks, by default, should watch their designated Xcom - implemented in #1998 and used in this PR

This proposal follows up on a successful internal PoC (https://github.com/astronomer/oss-integrations-private/issues/185), available in the branch https://github.com/astronomer/astronomer-cosmos/tree/single-run-execution-mode.

# Benefits

An initial performance analysis by @pankajkoti showed promising results:

| Experiment                                                    | Number of threads | Execution time (s) |
|---------------------------------------------------------------|-------------------|--------------------|
| dbt build                                                     | 4                 | 6 - 7              | 
| dbt run for each of model locally                             |                   | 30                 | 
| Cosmos default ExecutionMode.LOCAL in Astro CLI locally       |                   | 10 - 15            |
| Cosmos proposed ExecutionMode.WATCHER in Astro CLI locally    | 1                 | 26                 |
|                                                               | 2                 | 14                 | 
|                                                               | 4                 | 7                  | 
|                                                               | 8                 | 4                  | 
|                                                               | 16                | 2                  |
| The ExecutionMode.WATCHER in Airflow with an Astro deployment | 8                 | 5                  | 

# Example of usage

Example of DAG topology, with the producer task preceding the others.
Screenshot 2025-10-06 at 17 34 53

The dbt root nodes are set with `trigger_rule` `always`, so they start sensing once the producer begins.
Screenshot 2025-10-06 at 17 44 32

Producer task runs dbt Core, as shown on the logs:
Screenshot 2025-10-06 at 17 47 59

Consumer task senses XCom, waiting for producer to finish running dbt:
Screenshot 2025-10-06 at 17 48 12

Evidence that producer is running concurrently to the dbt root nodes sensing:
Screenshot 2025-10-06 at 17 46 55

# Related tickets

Closes #1964
Closes #1959 (*)

(*) I ended up implementing this while trying to enforce the producer task to run before the consumer tasks when running  `airflow dags test`.

Outside of the scope of this PR:
- Documentation (this will be added as part of #245)
- We are not implementing support for the following operators in the `ExecutionMode.WATCHER` mode:
  - LS
  - Run operation
  - Docs
  - Clone
Since it does not make sense to have them, we can review them later.

There are many other tasks related to this execution mode that can be tracked by searching issues using `label:execution:watcher`:
https://github.com/astronomer/astronomer-cosmos/issues?q=is%3Aissue%20state%3Aopen%20label%3Aexecution%3Awatcher

# Credits

The idea for this approach appeared in a discussion with @ashb.

The implementation of this feature is the result of teamwork with @pankajastro and @pankajkoti, both directly and indirectly involvement via PoC and previous PRs:

- Co-authored-by: Pankaj Koti <pankaj.koti@astronomer.io>
- Co-authored-by: Pankaj Singh <pankaj.singh@astronomer.io>


cosmos/airflow/graph.py
```diff
@@ -709,6 +753,15 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
             tasks_map[node_id] = test_task
 
     create_airflow_task_dependencies(nodes, tasks_map)
+
+    if execution_mode == ExecutionMode.WATCHER:
+        _add_producer(
+            dag,
+            task_args,
+            tasks_map,
+            task_group,
```
pankajkoti: Same question as above. Do we need to pass the `render_config`, here?

cosmos/airflow/graph.py
```diff
@@ -505,6 +518,37 @@ def _add_dbt_setup_async_task(
     tasks_map[DBT_SETUP_ASYNC_TASK_ID] = setup_airflow_task
 
 
+def _add_producer(
+    dag: DAG,
+    task_args: dict[str, Any],
+    tasks_map: dict[str, Any],
+    task_group: TaskGroup | None,
+    render_config: RenderConfig | None = None,
```
pankajkoti, line 549: Do we need `render_config` here or not? I was guessing that we would get the required fields from the `task_args`. Also asking because I do not see `render_config` getting passed while making the call to this function.

cosmos/operators/watcher.py
```diff
@@ -116,3 +232,192 @@ def _callback(ev: EventMsg) -> None:
...
+
+class DbtSnapshotWatcherOperator(DbtSnapshotMixin, DbtModelStatusSensor):
+    """
+    Watches for the progress of dbt snapshot execution, run by the producer task (DbtProducerWatcherOperator).
+    """
+
+    template_fields: Sequence[str] = DbtModelStatusSensor.template_fields  # type: ignore[operator]
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
```
pankajkoti: I like the strategy here of extending the Sensor!

pankajkoti (commented):
It's looking great to me. I don't have any additional comments at this point, apart from the `render_config` question for the producer task. Hope we can merge this soon! 🤞🏽

cosmos/airflow/graph.py
tatiana: We need to pass the `--select`, `--selector`, and `--exclude` options to the producer operator, so it does not run against the entire dbt project.

At the moment, this configuration lives in `render_config` - and with `ExecutionMode`s like `AIRFLOW_ASYNC` and `WATCHER`, the "setup" task has to know those values. This approach is consistent with our implementat of async.

cosmos/airflow/graph.py
tatiana: Fixed the missing `render_config`

pankajastro (approved):
Looks great, @tatiana!!

yetudada:
I'm so excited to see this PR! Well done @tatiana @pankajkoti @pankajastro!

Issue #1960, opened 2025-09-15
Fail sensor tasks immediately if the root dbt build task fails [ExecutionMode.WATCHER]


pankajkoti:
Ensure that watcher/sensor tasks automatically fail when the root dbt build task fails. This prevents DAGs from hanging indefinitely and makes failure conditions explicit.

Proposal:
- Explore if an Airflow task (sensor) can check the status of another Airflow task (root dbt build task)
- Alternatively, check if the root dbt build command should push an XCOM with `on_failure_callback` that the sensor tasks should first pull before polling for the status of the task

Acceptance Criteria:
- [ ] Watcher tasks fail immediately if the root task fails.
- [ ] Unit tests confirm failure propagation.

related: #1950

PR #2000, opened 2025-09-25
Push producer's task completion status to XCOM

pankajkoti:
This PR pushes a `task_status` XCOM for the `DbtProducerWatcherOperator` task after its completion -- be it failed or succeeded.

The idea is that sensors will pull this XCOM while checking for the result of the model. If a sensor sees that the `DbtProducerWatcherOperator` task has completed (either succeeded or failed), but the `DbtProducerWatcherOperator` has not pushed the model status to XCOM—whether via the `node_finished...` XCOM or the `run_results` XCOM—it means the operator has already finished processing and will not push any further status updates.

In that case, the sensor should immediately mark the model as failed, since there is no chance of receiving a run status for it in the future, and the sensor task itself should also fail.

related: https://github.com/astronomer/oss-integrations-private/issues/239
related: https://github.com/astronomer/astronomer-cosmos/issues/1960

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -121,6 +121,6 @@ class DbtProducerWatcherOperator(DbtLocalBaseOperator):
             context["ti"].xcom_push(key="task_status", value="completed")
             return return_value
 
-        except Exception as e:
+        except Exception:
             context["ti"].xcom_push(key="task_status", value="completed")
             raise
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Looks great, thank you very much @pankajkoti . This is one of the few scenarios I think it justifies for us to use the broad exception.

PR #2003, opened 2025-09-26
Pin Airflow version in type check CI job

pankajastro:
Pin the Airflow version in pre-commit hook for mypy


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, use_af3_in_typecheck]
```
```suggestion
    branches: [main]
```

pankajkoti (approved):

Issue #1865, opened 2025-07-10
[Bug] Excluding unit tests from DbtTaskGroup doesn't work


Pawel-Drabczyk:
### Astronomer Cosmos Version

1.10.1

### dbt-core version

1.9.0

### Versions of dbt adapters

dbt-bigquery==1.6.6
dbt-core==1.6.5
# via dbt-bigquery
dbt-extractor==0.4.1
dbt-semantic-interfaces==0.2.2

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

KUBERNETES

### InvocationMode

None

### airflow version

2.10.4

### Operating System

For local development I am using Ubuntu24.04

### Deployment

Official Apache Airflow Helm Chart

### What happened?

I am running the DbtTaskGroup operator with the `test_behavior=TestBehavior.AFTER_EACH` , `source_rendering_behavior=SourceRenderingBehavior.WITH_TESTS_OR_FRESHNESS`, `load_mode = LoadMode.DBT_MANIFEST` and `execution_config=ExecutionConfig(execution_mode=ExecutionMode.KUBERNETES)`. I want to exclude the DBT unit tests from being run during the DbtTaskGroup  , I want to run just the data tests. 

I tried to exclude the unit tests by tagging them and adding parameter `exclude = ["tag:silenced_test", "tag:unit_testing"]`, but they are still executed (the tests were tagged). I also tried `exclude = ["test_type:unit"]`, but the effect was the same.

### Relevant log output

```shell
Logs when using `exclude = ["test_type:unit"]`:

local-airflow-dag-import-test  | /opt/airflow/dags/vinaudit_elt/dag.py  | Traceback (most recent call last):     
local-airflow-dag-import-test  |                                        |   File                                 
local-airflow-dag-import-test  |                                        | "/home/airflow/.local/lib/python3.11/si
local-airflow-dag-import-test  |                                        | te-packages/cosmos/dbt/selector.py",   
local-airflow-dag-import-test  |                                        | line 704, in select_nodes              
local-airflow-dag-import-test  |                                        |     validate_filters(exclude, select)  
local-airflow-dag-import-test  |                                        |   File                                 
local-airflow-dag-import-test  |                                        | "/home/airflow/.local/lib/python3.11/si
local-airflow-dag-import-test  |                                        | te-packages/cosmos/dbt/selector.py",   
local-airflow-dag-import-test  |                                        | line 757, in validate_filters          
[...]

Issue #1763, opened 2025-05-12
[Bug] exclude=["resource_type:unit_test"] not being passed


kandharvishnu:
### Astronomer Cosmos Version

1.9.2

### dbt-core version

1.9.2

### LoadMode

AUTOMATIC

### ExecutionMode

AIRFLOW_ASYNC

### InvocationMode

None

### airflow version

2.10.5

### Operating System

debian

### Deployment

Astronomer

### What happened?

When both `data_tests` and `unit_tests` are defined for a model, attempting to exclude the unit tests using:

```
render_config = RenderConfig(exclude=["resource_type:model"])
```

does not prevent the unit tests from running.

However, when using operator_args with the following configuration:

```
operator_args = {
    "dbt_cmd_flags": ["--include", "resource_type:unit_test"],
}
```
the unit tests are correctly excluded as expected.

### Relevant log output

```shell
when render_config=RenderConfig(exclude=["resource_type:unit_test"]) is set, totally 4 test cases ran, which includes unit_test order_items::test_supply_costs_sum_correctly

code:

exclude_unit_test = DbtTaskGroup(
group_id="exclude_unit_test",
project_config=ProjectConfig(DBT_PROJECT_PATH),
profile_config=profile_config,
execution_config=execution_config,
default_args={"retries": 2},
render_config=RenderConfig(exclude=["resource_type:unit_test"]),
)

Logs:

[2025-04-29, 07:10:21 UTC] {logging_mixin.py:190} INFO - 07:10:21 1 of 4 START test not_null_order_items_order_item_id ........................... [RUN]
[2025-04-29, 07:10:21 UTC] {logging_mixin.py:190} INFO - 07:10:21 1 of 4 PASS not_null_order_items_order_item_id ................................. [PASS in 0.07s]
[2025-04-29, 07:10:21 UTC] {logging_mixin.py:190} INFO - 07:10:21 2 of 4 START test relationships_order_items_order_id__order_id__ref_orders_ .... [RUN]
[2025-04-29, 07:10:21 UTC] {logging_mixin.py:190} INFO - 07:10:21 2 of 4 PASS relationships_order_items_order_id__order_id__ref_orders_ .......... [PASS in 0.08s]
[2025-04-29, 07:10:21 UTC] {logging_mixin.py:190} INFO - 07:10:21 3 of 4 START test unique_order_items_order_item_id ............................. [RUN]
[...]

Issue #1530, opened 2025-02-10
[Feature] Add support to `--exclude-resource-type` flag


tatiana:
### Description

Add support to dbt `--exclude-resource-type` (https://docs.getdbt.com/reference/global-configs/resource-type, available since dbt-core 1.8) in `DbtDag` and `DbtTaskGroup`type when using `LoadMode.MANIFEST` and `LoadMode.CUSTOM`.

This complements the `resoure_type:` select statement already available since Cosmos 1.9 (#1510).

### Use case/motivation

Support newly introduced dbt-core arguments.

### Related issues

Related: #1510

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

PR #2005, opened 2025-09-26
Removing exclude_resource_type functionality since it's broken

### LINKED ISSUE #1733: [Bug] exclude_resource_type doesn't work in select / exclude with dbt ls in cosmos
### Astronomer Cosmos Version

1.10.0

### dbt-core version

1.8.4

### Versions of dbt adapters

_No response_

### LoadMode

DBT_LS

### ExecutionMode

LOCAL

### InvocationMode

DBT_RUNNER

### airflow version

2.10.5

### Operating System

Debian GNU/Linux 12

_No response_

### Deployment

Astronomer

### Deployment details

_No response_

### What happened?

The new "exclude_resource_type:" feature in select / exclude seems broken with dbt ls.  Additional details:

- DBT allows you to select or exclude certain resource types with the [flags](https://docs.getdbt.com/reference/global-configs/resource-type) "--resource-type x" or "--exclude-resource-type x" 
- [DBT node selector](https://docs.getdbt.com/reference/node-selection/methods#resource_type) works with "resource_type:x" in the "--select" or "--exclude" tags (ex. "--select resource_type:x") 
- Cosmos allows you to use "--select" or "--exclude" flags, but NOT the "--resource-type x" or "--exclude-resource-type x"
- New cosmos feature adds "exclude_resource_type:<resource>" similarly to "resource_type:x" in either the select or exclude tags, but this breaks dbt ls with the error `'exclude_resource_type' is not a valid method name`

### Relevant log output

```shell
cosmos.dbt.graph.CosmosLoadDbtException: Unable to run ['/usr/local/bin/dbt', 'ls', '--output', 'json', '--project-dir', '/tmp/tmpexno75tu', '--profiles-dir', '/tmp/cosmos/profile/147a36f81d4697b947dc02cdd963615081da6f695e3abfa1274e4e10164418b7', '--profile', 'spoileralert', '--target', 'local', '--select', 'exclude_resource_type:unit_test', '--no-partial-parse'] due to the error:
stderr: Runtime Error
  'exclude_resource_type' is not a valid method name
```

### How to reproduce

Add the following two options to the render config:

```
    select=["exclude_resource_type:unit_test"],
    load_method=LoadMode.DBT_LS,
```

### Anything else :)?

I don't think this is urgent since I believe that using "resource_type:x" in the "exclude" tag should work pretty much the same, but still wanted to raise in case this is unexpected.  The double negative was also weird to me: if you were to put "exclude_resource_type:x" in the "exclude" tag, then how is that expected to work?  Does that actually select the resource by excluding the exclusion?  

### Contact Details

_No response_

anyapriya:
## Description

Essentially undoes [this PR](https://github.com/astronomer/astronomer-cosmos/pull/1580).  

While dbt did the ability to use --exclude-resource-type flag (which matches --resource-type that already exists), it is meant as a separate flag.  The way this was implemented was mixing it up with the "resource_type:" tag that can be used as part of the "--select" flag - there is no corresponding "exclude_resource_type:" tag in dbt.  As a result, it wasn't working the way it was intended as explained in [this issue](https://github.com/astronomer/astronomer-cosmos/issues/1733).  

It's also unclear what putting a tag of exclude_resource_type into the "exclude" parameter is meant to do.  

In order to add --exclude-resource-type correctly, it would need to be properly implemented as a flag instead of a tag which I think is a much bigger change.   

Additionally, I can't think of any situations where this would be needed instead of using the "resource_type:" tag in the exclude parameter of the render config.  

## Related Issue(s)

Closes #1733 

## Breaking Change?

Sort of : removes functionality, but the functionality was broken.  

## Checklist


pankajastro:
@anyapriya, are you considering these changes as a bug fix? Could you please share an example to help me better understand the issue you're trying to address?

anyapriya:
> @anyapriya, are you considering these changes as a bug fix? Could you please share an example to help me better understand the issue you're trying to address?

@pankajastro As mentioned in [the linked issue](https://github.com/astronomer/astronomer-cosmos/issues/1733), if you use the following settings:
```
    select=["exclude_resource_type:unit_test"],
    load_method=LoadMode.DBT_LS,
```
Then it results in this error:
```
cosmos.dbt.graph.CosmosLoadDbtException: Unable to run ['/usr/local/bin/dbt', 'ls', '--output', 'json', '--project-dir', '/tmp/tmpexno75tu', '--profiles-dir', '/tmp/cosmos/profile/147a36f81d4697b947dc02cdd963615081da6f695e3abfa1274e4e10164418b7', '--profile', 'spoileralert', '--target', 'local', '--select', 'exclude_resource_type:unit_test', '--no-partial-parse'] due to the error:
stderr: Runtime Error
  'exclude_resource_type' is not a valid method name
```

I think it should be removed for the following reasons: 
1. It's broken when using dbt_ls 
2. `--select exclude_resource_type:unit_test` and `--exclude exclude_resource_type:unit_test` are not valid ways to exclude a given resource type in dbt.  The fact that this differs from base dbt logic selection logic is confusing.  The actual ways that dbt allows are the following: 
  a. `--select resource_type:unit_test` ([docs](https://docs.getdbt.com/reference/node-selection/methods)) - possible in cosmos
  b. `--exclude resource_type:unit_test` ([docs](https://docs.getdbt.com/reference/node-selection/methods)) - possible in cosmos
  c. `--resource-type unit_test` ([docs](https://docs.getdbt.com/reference/global-configs/resource-type)) - not possible in cosmos
  d. `--exclude-resource-type unit_test` ([docs](https://docs.getdbt.com/reference/global-configs/resource-type)) - not possible in cosmos
3. It's confusing what putting `exclude_resource_type:unit_test` in the exclude parameter is meant to do.  It looks like it select the resource type because of a double negative?  But that does not seem clear.  DBT doesn't run into this issue because they don't allow it.  
4. I don't think it's necessary.  Not only can you achieve the intended functionality of putting `exclude_resource_type:unit_test` in the select, by instead putting `resouce_type:unit_test` in the exclude, but DBT [docs](https://docs.getdbt.com/reference/global-configs/resource-type) say the following: "Note that the list of dbt resource types is mutually exclusive and collectively exhaustive (MECE). This means that any --resource-type selection can also be achieved by excluding the other resource types using --exclude-resource-type, and vice versa." 
5. Selecting and excluding specific tests/resource types have a different bug where they're not passed into the test tasks as mentioned in issue #1763 and issue #1865.  I have a separate PR (#2006) to fix this for exclusions but fixing it for `select` is a lot harder.  As a result, I think only allowing for putting `resource_type:unit_test` in the exclude is a lot safer than putting `exclude_resource_type:unit_test` in the select.  

As mentioned in the issue, I'm open to alternative options after input from the cosmos team, but went this route after not hearing any response, and because I ultimately think it's the cleanest approach for the reasons stated above.  

I think if there really is an interest in adding exclude resource type functionality, then it should be added as its own flag as a separate parameter instead of as part of the select / exclude parameters, along with the resource type flag for consistency, but that seems like a decent amount of work to implement with little value beyond what already exists.

anyapriya:
I'm seeing that 5 tests are failing.  It seems like the failing tests are due to the following:
- 4 failing integration tests due to "Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential." see [here](https://bigquery.googleapis.com/bigquery/v2/projects/astronomer-dag-authoring/datasets?maxResults=10000&prettyPrint=false:) 
- 1 failing Type-Check test - this is failing in other PRs too like [this one](https://github.com/astronomer/astronomer-cosmos/pull/2000) , looks like it's planned to be fixed [here](https://github.com/astronomer/astronomer-cosmos/pull/2012)

Neither seems related to this PR, would love recommendation from the cosmos team on how to handle this

pankajastro:
@tatiana, the issue https://github.com/astronomer/astronomer-cosmos/issues/1530 mentioned to support `--exclude-resource-type` for `LoadMode.MANIFEST` and `LoadMode.CUSTOM`. Why not the `DBT_LS`?

tatiana:
@anyapriya I agree with your analysis. It was a mistake to accept that contribution. I'm talking to the original contributor of this feature next week to confirm the possibility of them changing how they accomplish this, before we go ahead to remove the feature.

PR #2009, opened 2025-09-30
Add Airflow 3 plugin for dbt docs with multiple dbt projects support

### LINKED ISSUE #862: Support multiple projects' dbt docs
Discussion here: https://github.com/astronomer/astronomer-cosmos/pull/737#issuecomment-1964349542

Better to get this done now before 1.4 ships than after, I think.

tldr: The main worry I have is that the API for supporting multiple projects' docs can get very convoluted. Ini file format / env vars for config do not natively support arrays of objects, so you need to do something that isn't pretty or is complicated (e.g. read from a yaml file) basically no matter what.

I'm leaving this as an open ended issue to discuss solutions.

### LINKED ISSUE #1587: Evaluate how to support dbt docs in Airflow 3

pankajkoti:
This PR introduces Airflow 3 plugin support for hosting dbt documentation with multiple project configurations. The implementation replaces the disabled Airflow 2.x Flask-based plugin with a new FastAPI-based plugin compatible with Airflow 3.x.

### Key Changes
- New FastAPI-based plugin implementation (af3_plugin_impl.py) that serves dbt docs for Airflow 3 via external views
- Support for multiple dbt projects through JSON configuration or legacy single-project settings

**Screenshot showing multiple projects listed in "Browse" tab icon**

Screenshot 2025-10-27 at 9 20 08 PM

**Screenshot showing dbt Docs in Airflow 3.1  "external views"** 

Screenshot 2025-10-27 at 9 20 46 PM

closes: https://github.com/astronomer/astronomer-cosmos/issues/1587
closes: https://github.com/astronomer/astronomer-cosmos/issues/862
closes: https://github.com/astronomer/oss-integrations-private/issues/205
closes: https://github.com/astronomer/oss-integrations-private/issues/206

### DIFF
--- a/cosmos/plugin/af3_plugin_impl.py
+++ b/cosmos/plugin/af3_plugin_impl.py
@@ -171,7 +171,7 @@ def create_cosmos_fastapi_app() -> FastAPI:  # noqa: C901
             index_local = cfg_local.get("index") or "index.html"
             if not docs_dir_local:
                 return HTMLResponse(
-                    content=(f"<pre>Cosmos dbt docs error: slug={slug_alias} not configured (missing dir)</pre>"),
+                    content=(f"<pre>Cosmos dbt docs error: slug={html.escape(slug_alias, quote=True)} not configured (missing dir)</pre>"),
                     status_code=404,
                 )
             try:
@@ -179,17 +179,17 @@ def create_cosmos_fastapi_app() -> FastAPI:  # noqa: C901
             except FileNotFoundError:
                 return HTMLResponse(
                     content=(
-                        f"<pre>Cosmos dbt docs error: index not found for slug={slug_alias}\n"
-                        f"path={op.join(docs_dir_local, index_local)} conn_id={conn_id_local or ''}</pre>"
+                        f"<pre>Cosmos dbt docs error: index not found for slug={html.escape(slug_alias, quote=True)}\n"
+                        f"path={html.escape(op.join(docs_dir_local, index_local), quote=True)} conn_id={html.escape(conn_id_local or '', quote=True)}</pre>"
                     ),
                     status_code=404,
                 )
             except Exception as e:
                 return HTMLResponse(
                     content=(
-                        f"<pre>Cosmos dbt docs error: index read failed for slug={slug_alias}\n"
-                        f"path={op.join(docs_dir_local, index_local)} conn_id={conn_id_local or ''}\n"
-                        f"exception={type(e).__name__}: {e}</pre>"
+                        f"<pre>Cosmos dbt docs error: index read failed for slug={html.escape(slug_alias, quote=True)}\n"
+                        f"path={html.escape(op.join(docs_dir_local, index_local), quote=True)} conn_id={html.escape(conn_id_local or '', quote=True)}\n"
+                        f"exception={html.escape(type(e).__name__, quote=True)}: {html.escape(str(e), quote=True)}</pre>"
                     ),
                     status_code=500,
                 )
--- a/cosmos/plugin/af3_plugin_impl.py
+++ b/cosmos/plugin/af3_plugin_impl.py
@@ -213,13 +213,15 @@ def create_cosmos_fastapi_app() -> FastAPI:  # noqa: C901
                     status_code=404,
                 )
             except Exception as e:
+                logging.exception(
+                    f"Error reading manifest for slug '{slug_alias}', path '{op.join(docs_dir_local, 'manifest.json')}', conn_id '{conn_id_local}': {e}"
+                )
                 return JSONResponse(
                     content={
                         "error": "manifest read failed",
                         "slug": slug_alias,
                         "path": op.join(docs_dir_local, "manifest.json"),
                         "conn_id": conn_id_local,
-                        "exception": f"{type(e).__name__}: {e}",
                     },
                     status_code=500,
                 )
--- a/cosmos/plugin/af3_plugin_impl.py
+++ b/cosmos/plugin/af3_plugin_impl.py
@@ -2,14 +2,20 @@ from __future__ import annotations
 
 import json
 import mimetypes
+import os
 import os.path as op
-from typing import Any, Optional, Tuple
+from contextlib import contextmanager
+from typing import Any, Generator, Optional, Tuple, TypeVar
+from unittest.mock import patch
 
 from airflow.configuration import conf
 from airflow.plugins_manager import AirflowPlugin
+from airflow.sdk import ObjectStoragePath
 from fastapi import FastAPI
 from fastapi.responses import HTMLResponse, JSONResponse, Response
 from fastapi.staticfiles import StaticFiles
+
+S = TypeVar("S")
 
 
 def bucket_and_key(path: str) -> Tuple[str, str]:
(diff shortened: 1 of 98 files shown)


cosmos/plugin/af3_plugin_impl.py
```diff
@@ -0,0 +1,419 @@
...
+def bucket_and_key(path: str) -> Tuple[str, str]:
+    from urllib.parse import urlsplit
+
+    parsed_url = urlsplit(path)
+    bucket = parsed_url.netloc
+    key = parsed_url.path.lstrip("/")
+    return bucket, key
+
+
+def open_s3_file(path: str, conn_id: Optional[str]) -> str:
```
tatiana: While talking to @dwreeves he suggested we may want to use https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/objectstorage.html instead of having specific functions per object storage

tatiana:
@pankajkoti based on [internal discussions](https://astronomer.slack.com/archives/C01UJJEN0P3/p1760359042377239?thread_ts=1759165841.026969&cid=C01UJJEN0P3), to overcome the error:
```
cannot import name 'SUPERVISOR_COMMS' from 'airflow.sdk.execution_time.task_runner' (/usr/local/lib/python3.12/site-packages/airflow/sdk/execution_time/task_runner.py) [airflow.task.hooks.airflow.providers.amazon.aws.hooks.s3.S3Hook] loc=base_aws.py:631
```

We can try to use:
```
from airflow.models.connection import Connection

conn = Connection.get_connection_from_secrets(conn_id)
```

To retrieve the connection in the current implementation and see if it works. If this is successful, we can further look to see if `ObjectStorage` would work - and potentially refactor the code to use it, if end-users are using AF 2.8 or higher.

One of the recommendations was to get connection using above method and store it as env variable temporarily where the hooks need access, until there is a better way to implement this plugin.

tatiana:
Recommendation from @kaxil for now is to do something like:

```py
@contextmanager
def connection_env(conn_id: str) -> Generator[None, None, None]:
    """
    Temporarily expose a connection as AIRFLOW_CONN_{CONN_ID} in the environment.

    This allows hooks and SDK code resolving connections via the environment
    variables backend to find the connection during the scope of the context.
    """
    from airflow.models.connection import Connection as ORMConnection

    conn = ORMConnection.get_connection_from_secrets(conn_id)
    env_name = f"AIRFLOW_CONN_{conn_id.upper()}"
    env_value = conn.get_uri()
    with patch_environ({env_name: env_value}):
        yield

def set_connection_as_env(conn_id: str):
    """
    Decorator that sets AIRFLOW_CONN_{CONN_ID} for the duration of the call.

    Example:
        @set_connection_as_env("my_conn")
        def my_func(...):
            ...
    """

    def _decorator(func: Callable[..., S]) -> Callable[..., S]:
        def _wrapper(*args, **kwargs):
            with connection_env(conn_id):
                return func(*args, **kwargs)

        return _wrapper

    return _decorator 
```

In the meantime, @pierrejeambrun logged a ticket in the Apache Airflow repo so Airflow itself can improve this type of use case:
https://github.com/apache/airflow/issues/56583

docs/airflow3_compatibility/index.rst
```diff
@@ -58,8 +58,32 @@ investigation.
...
+
+.. code-block:: ini
+
+   [cosmos]
+   # Map of slug -> {dir, conn_id, index, name}
+   dbt_docs_projects = {
+     "core": {"dir": "/path/to/core/target", "index": "index.html", "name": "dbt Docs (Core)"},
+     "mart": {"dir": "s3://bucket/path/to/mart/target", "conn_id": "aws_default", "name": "dbt Docs (Mart)"}
+   }
+
```
tatiana, line 63: Could you also give an example of doing the same by exporting the envvar? We should also update https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html - in a follow up PR

cosmos/plugin/airflow3.py
```diff
@@ -0,0 +1,293 @@
+from __future__ import annotations
```
tatiana, line 1: Could we rename this Python file to `airflow3.py` or `af3.py`? It is already in the plugin package, so I don't think it needs `plugin` in the name. And it feels like `impl` doesn't give much value.

We should probably rename the Airflow 2 Implementation as well, so we are consistent.

cosmos/plugin/airflow3.py
```diff
@@ -0,0 +1,293 @@
...
+from airflow.sdk import ObjectStoragePath
+from fastapi import FastAPI
+from fastapi.responses import HTMLResponse, JSONResponse, Response
+
+API_BASE = conf.get("api", "base_url", fallback="")  # reads AIRFLOW__API__BASE_URL
+API_BASE_PATH = urlsplit(API_BASE).path.rstrip("/")
+
+
+@contextmanager
+def connection_env(conn_id: str | None = None) -> Generator[None, None, None]:  # pragma: no cover
```
tatiana, line 31: This is a workaround for a limitation in Airflow 3.1.0. Do we have any tickets in the Airflow repo that we could link from here?

cosmos/plugin/af3_plugin_impl.py
```diff
@@ -0,0 +1,293 @@
...
+        return content
+
+
+def open_file(path: str, conn_id: str | None = None) -> Any:
+    """
+    Retrieve a file from http, https, gs, s3, or wasb.
+
+    Raise a (base Python) FileNotFoundError if the file is not found.
+    """
+    if path.strip().startswith(("s3://", "gs://", "gcs://", "wasb://", "abfs://", "az://", "http://", "https://")):
```
tatiana: Would this list be available somewhere else via constants, so we don't have to hardcode the suffix protocols—and change them depending on new protocols that may be supported by Airflow's ObjectStorage in the future?

cosmos/plugin/airflow3.py
```diff
@@ -0,0 +1,419 @@
...
+        # If no conn_id is provided, rely on boto3 environment/instance role credentials
+        return open_s3_file(path, conn_id=conn_id if conn_id else None)
+    elif path.strip().startswith("gs://"):
+        return open_gcs_file(path, conn_id=conn_id)
+    elif path.strip().startswith("wasb://"):
+        return open_azure_file(path, conn_id=conn_id)
+    elif path.strip().startswith("http://") or path.strip().startswith("https://"):
+        return open_http_file(path, conn_id=conn_id)
+    else:
+        with open(path) as f:
```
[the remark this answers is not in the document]
tatiana: Yes, this is part of the feature.

cosmos/plugin/af3_plugin_impl.py
```diff
@@ -0,0 +1,293 @@
...
+    let msgData = event.data;
+    if (typeof msgData === 'string' && msgData.startsWith('#!')) {
+      let updateUrl = new URL(window.location);
+      updateUrl.hash = msgData;
+      hashChangeLock = true;
+      history.replaceState(null, null, updateUrl);
+    }
+  });
+</script>
+"""
```
tatiana: How different is this iFrame from the one implemented for airflow2?

cosmos/plugin/af3_plugin_impl.py
```diff
@@ -0,0 +1,293 @@
...
+                for key, value in parsed.items():
+                    if not isinstance(value, dict):  # pragma: no cover
+                        continue
+                    projects[str(key)] = {
+                        "dir": value.get("dir"),
+                        "conn_id": value.get("conn_id"),
+                        "index": value.get("index", "index.html"),
+                        "name": value.get("name"),
+                    }
+        except Exception:
```
tatiana: Could you add a more specific exception here? It is usually better to minimise code within the try/except block and to capture a specific exception.

cosmos/plugin/af3_plugin_impl.py
```diff
@@ -0,0 +1,293 @@
...
+            # Ignore malformed config; fall back to legacy
```
tatiana: If customers attempt to use the new `dbt_docs_projects` and its configuration has issues, I believe we should not fallback to the old configuration. We should be explicit with users about the error and provide clear instructions so they can fix it. If they want to fallback to the old way, they can do this explicitly.

cosmos/plugin/airflow3.py
```diff
@@ -0,0 +1,419 @@
...
+                f'Index URL: <a href="{iframe_src}">{iframe_src}</a>. '
+                f'Try <a href="/cosmos/{slug_alias}/manifest.json">manifest.json</a> and '
+                f'<a href="/cosmos/{slug_alias}/catalog.json">catalog.json</a>.'
+                f"</div>"
+            )
+            return (
+                '<div style="height:100%;display:flex;flex-direction:column;">'
+                f"{debug_panel}"
+                f'<iframe src="{iframe_src}" style="border:0;flex:1 1 auto;"></iframe>'
+                "</div>"
```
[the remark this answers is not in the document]
tatiana: It's an inherent part of how the feature works. @pankajkoti, could you document these security concerns in the follow-up docs PR?

cosmos/plugin/airflow3.py
```diff
@@ -0,0 +1,335 @@
...
+        @app.get(f"/{slug}/dbt_docs", response_class=HTMLResponse)
+        def dbt_docs_view(slug_alias: str = slug) -> str:  # type: ignore[no-redef]
+            cfg_local = projects.get(slug_alias, {})
+            if not cfg_local.get("dir"):
+                return "<div>dbt Docs are not configured.</div>"
+            iframe_src = f"/cosmos/{slug_alias}/dbt_docs_index.html"
+            return (
+                '<div style="height:100%;display:flex;flex-direction:column;">'
+                f'<iframe src="{iframe_src}" style="border:0;flex:1 1 auto;"></iframe>'
+                "</div>"
```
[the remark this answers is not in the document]
tatiana: Same here https://github.com/astronomer/astronomer-cosmos/pull/2009/files#r2460141065

cosmos/plugin/af3_plugin_impl.py
```diff
@@ -0,0 +1,293 @@
...
+                return JSONResponse(
+                    content={
+                        "error": "catalog not found",
+                        "slug": slug_alias,
+                        "path": op.join(docs_dir_local, "catalog.json"),
+                        "conn_id": conn_id_local,
+                    },
+                    status_code=404,
+                )
+            except Exception as e:
```
tatiana: Why are we capturing `Exception` here?

cosmos/plugin/airflow3.py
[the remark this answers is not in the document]
pankajkoti: The suggestion was already applied, so the warning was and is no longer relevant. Hence, I had marked the conversation as resolved.

cosmos/plugin/airflow3.py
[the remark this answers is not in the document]
pankajkoti: This part of the code was already removed, so I marked the conversation as resolved. It was complaining about the debug panel code that we had added during development earlier. The warning was and is no longer relevant.

docs/airflow3_compatibility/index.rst
pankajkoti: yes, will do today. I have created a tracking issue for the docs PR now https://github.com/astronomer/oss-integrations-private/issues/252 and also to address this comment.

pankajkoti: added the env var approach in this section in the current PR. will follow-up with the docs PR to update cosmos-conf.html

cosmos/plugin/airflow3.py
pankajkoti: yes, makes sense. renamed to `airflow3.py` and `airflow2.py`

cosmos/plugin/airflow3.py
pankajkoti: added a comment and also linked the PR for the upcoming fix.

cosmos/plugin/af3_plugin_impl.py
pankajkoti: moved it to cosmos/constants.py

cosmos/plugin/af3_plugin_impl.py
pankajkoti: It is the same script. Moved and centralised it to `cosmos/plugin/snippets.py`. I am happy to rename the module if you have any suggestion.

cosmos/plugin/af3_plugin_impl.py
pankajkoti: yes added specific JSONDecodeError

cosmos/plugin/af3_plugin_impl.py
pankajkoti: okay, makes sense. removed the fallback. And we already have the invalid json exception logged above.

cosmos/plugin/af3_plugin_impl.py
pankajkoti: One exception we caught above was FileNotFoundError. I kept a blanket Exception not knowing what all exceptions could come on the backend and returning  JSONResponse with status code 500. I have now narrowed the caught exceptions to `(OSError, ValueError, RuntimeError, TimeoutError, PermissionError)`

docs/airflow3_compatibility/index.rst
```diff
@@ -46,6 +46,41 @@ functionality with Airflow 3 by extending our CI infrastructure:
...
+.. code-block:: bash
+
+   # Shell (note the single quotes to preserve JSON)
+   export AIRFLOW__COSMOS__DBT_DOCS_PROJECTS='{"core":{"dir":"/path/to/core/target","index":"index.html","name":"dbt Docs (Core)"},"mart":{"dir":"s3://bucket/path/to/mart/target","conn_id":"aws_default","name":"dbt Docs (Mart)"}}'
+
+.. code-block:: dockerfile
+
+   # Dockerfile (ENV does not require outer quotes)
+   ENV AIRFLOW__COSMOS__DBT_DOCS_PROJECTS={"core":{"dir":"/path/to/core/target","index":"index.html","name":"dbt Docs (Core)"},"mart":{"dir":"s3://bucket/path/to/mart/target","conn_id":"aws_default","name":"dbt Docs (Mart)"}}
```
pankajkoti:
```suggestion
   # Dockerfile
   ENV AIRFLOW__COSMOS__DBT_DOCS_PROJECTS='{"core":{"dir":"/path/to/core/target","index":"index.html","name":"dbt Docs (Core)"},"mart":{"dir":"s3://bucket/path/to/mart/target","conn_id":"aws_default","name":"dbt Docs (Mart)"}}'
```

cosmos/plugin/airflow3.py
```diff
@@ -0,0 +1,263 @@
...
+from cosmos.plugin.snippets import IFRAME_SCRIPT
+
+API_BASE = conf.get("api", "base_url", fallback="")  # reads AIRFLOW__API__BASE_URL
+API_BASE_PATH = urlsplit(API_BASE).path.rstrip("/")
+
+
+# TODO: This context manager and its usage in the method "_read_content_via_object_storage" has been added due
+# to the current limitation in Airflow 3.1.0 where plugins are not able to resolve connections via the API server.
+# Once this is fixed, potentially in PR https://github.com/apache/airflow/pull/56602 planned to be released in
+# Airflow 3.1.1, test the fix and remove this context manager and its usage.
```
tatiana, line 29: @pankajkoti It seems https://pypi.org/project/apache-airflow/3.1.1/ is out, do we have a follow up ticket for this?

tatiana (approved):
This looks great, @pankajkoti, thanks a lot for getting this to work.
Happy for this PR to be merged once the checks pass.

Tiny detail to be done before merging:
- Create a follow-up ticket for 3.1.1 or update comment
- Add a screenshot of the feature in the ticket description

Things that can be done on follow-up ticket:
- Update other parts of the Cosmos documentation with the changes, including
- Update other docs pages, included:
    -  https://astronomer.github.io/astronomer-cosmos/configuration/cosmos-conf.html
    - https://astronomer.github.io/astronomer-cosmos/configuration/hosting-docs.html

cosmos/plugin/airflow3.py
pankajkoti: Created a ticket https://github.com/astronomer/astronomer-cosmos/issues/2061

PR #2012, opened 2025-10-01
Fix mypy issues related to `ObjectStoragePath` in main branch

tatiana:
The solution utilises forward references (string literals) for the `ObjectStoragePath` type, allowing mypy to understand the types without requiring the actual class to be available at runtime. This is the standard approach for handling optional dependencies in type annotations.

Cosmos' main branch is red due to [mypy](https://github.com/python/mypy) checks failing, as observed in this sample GitHub Actions job:
https://github.com/astronomer/astronomer-cosmos/actions/runs/18122118077/job/51706664314

```
mypy-python..............................................................Failed
- hook id: mypy
- exit code: 1

cosmos/cache.py:76: error: Incompatible types in assignment (expression has type "ObjectStoragePath", variable has type "Path | None")  [assignment]
cosmos/cache.py:78: error: Item "None" of "Path | None" has no attribute "exists"  [union-attr]
cosmos/cache.py:78: note: Error code "union-attr" not covered by "type: ignore" comment
cosmos/cache.py:80: error: Item "None" of "Path | None" has no attribute "mkdir"  [union-attr]
cosmos/config.py:233: error: Incompatible types in assignment (expression has type "ObjectStoragePath", variable has type "Path | None")  [assignment]
cosmos/operators/local.py:325: error: Incompatible return value type (got "tuple[ObjectStoragePath, str]", expected "tuple[Path, str] | tuple[None, None]")  [return-value]
Found 5 errors in 3 files (checked 35 source files)
```

The same job was working without failures two days ago:
https://github.com/astronomer/astronomer-cosmos/actions/runs/18111649257/job/51539065565

**Changes introduced**

- Added `TYPE_CHECKING` imports: Added conditional imports for `ObjectStoragePath` in all affected files using the `TYPE_CHECKING` pattern to avoid runtime import errors when the class is not available.
- Updated type annotations: Used string literals (forward references) for `ObjectStoragePath` in type annotations to avoid mypy errors:
```
  Path | "ObjectStoragePath" | None instead of Path | ObjectStoragePath | None
```

Fixed specific issues:
- cache.py: Updated `_configured_cache_dir` variable and `_configure_remote_cache_dir()` function return type
- config.py: Updated `manifest_path` field and `mandatory_paths` dictionary type annotations
- operators/local.py: Updated `_configure_remote_target_path()` return `type and _construct_dest_file_path()` parameter type
- dbt/graph.py: Updated `_get_dbt_ls_remote_cache()` parameter type

**Verification**

I used the same command as the CI to confirm that the introduced changes fix the original issue:
```
hatch run tests.py3.10-2.10-1.9:type-check
mypy-python..............................................................Passed
```

And I also confirmed in the CI:
Screenshot 2025-10-01 at 16 29 55

https://github.com/astronomer/astronomer-cosmos/actions/runs/18167212785/job/51712529044?pr=2012

### DIFF
--- a/cosmos/cache.py
+++ b/cosmos/cache.py
@@ -26,7 +26,12 @@ from cosmos import settings
 
 if TYPE_CHECKING:
     try:
-        from airflow.io.path import ObjectStoragePath
+        from airflow.sdk import ObjectStoragePath
+    except ImportError:
+        try:
+            from airflow.io.path import ObjectStoragePath
+        except ImportError:
+            pass
     except ImportError:
         pass
 from cosmos.constants import (


cosmos/cache.py
```diff
@@ -22,6 +23,12 @@
 from sqlalchemy.orm import Session
 
 from cosmos import settings
+
+if TYPE_CHECKING:
+    try:
+        from airflow.io.path import ObjectStoragePath
```
corsettigyg: since airflow 3, the ObjectStoragePath import changed a little bit 

`from airflow.sdk import ObjectStoragePath`

One alternative is to do 

```
try:
    from airflow.sdk import ObjectStoragePath
except ImportError:
    try:
        from airflow.io.path import ObjectStoragePath
    except ImportError:
        pass
```
but it does feel a little bit too nested

cosmos/config.py
```diff
@@ -8,12 +8,18 @@
...
+from typing import TYPE_CHECKING, Any, Callable, Iterator
 
 import yaml
 from airflow.version import version as airflow_version
 
 from cosmos import settings
+
+if TYPE_CHECKING:
+    try:
+        from airflow.io.path import ObjectStoragePath
```
corsettigyg, line 20: same here

cosmos/cache.py
tatiana:
```suggestion
    try:
        from airflow.sdk import ObjectStoragePath
    except ImportError:
        try:
            from airflow.io.path import ObjectStoragePath
        except ImportError:
            pass
```

tatiana: Thanks a lot, @corsettigyg , just addressed your feedback! I believe it is a change worth making.

corsettigyg (approved):

PR #2030, opened 2025-10-13
Improve dataset/asset experience in Cosmos

tatiana:
While validating Cosmos Airflow Asset support in Airflow 3.1, I noticed that we could improve the overall experience.

Previously, Cosmos required both `openlineage-integration-common` and `openlineage-airflow` or `apache-airflow-providers-openlineage` to be installed to emit Airflow Datasets or Airflow Assets. However, the only necessary dependency is  `openlineage-integration-common`.

Users will need either `openlineage-airflow` or `apache-airflow-providers-openlineage`  for emitting OpenLineage events - but not Airflow Datasets and Aliasses.

This PR improves this behaviour by adding  `openlineage-integration-common` as a dependency of the Cosmos standard installation - and not failing to emit datasets if `openlineage-airflow` or `apache-airflow-providers-openlineage`   are not installed.

It was validated with Airflow 3.1.0 and validates with the following three DAGs:

# Producer Cosmos `DbtDag`

Existing `basic_cosmos_dag` example
```
import os
from datetime import datetime
from pathlib import Path

from cosmos import DbtDag, ProfileConfig, ProjectConfig

DEFAULT_DBT_ROOT_PATH = Path(__file__).parent / "dbt"
DBT_ROOT_PATH = Path(os.getenv("DBT_ROOT_PATH", DEFAULT_DBT_ROOT_PATH))
DBT_PROJECT_NAME = os.getenv("DBT_PROJECT_NAME", "jaffle_shop")
DBT_PROJECT_PATH = DBT_ROOT_PATH / DBT_PROJECT_NAME


    profile_name="default",
    target_name="dev",
        conn_id="example_conn",
        disable_event_tracking=True,
    ),
)

basic_cosmos_dag = DbtDag(
    project_config=ProjectConfig(DBT_PROJECT_PATH),
    operator_args={
        "full_refresh": True,  # used only in dbt commands that support this flag
    },
    # normal dag parameters
    schedule="@daily",
    catchup=False,
    dag_id="basic_cosmos_dag",
    default_args={"retries": 0},
)
```

# Consumer DAG scheduled based on an Asset

```
from datetime import datetime
from airflow import DAG
from airflow.sdk.definitions.asset import Asset


with DAG(
    "dataset_triggered_dag",
    description="A DAG that should be triggered via Dataset",
    start_date=datetime(2024, 9, 1),
    schedule=[Asset(uri="postgres://0.0.0.0:5432/postgres/public/orders")],
) as dag:
    t1 = EmptyOperator(
        task_id="task_1",
    )
    t2 = EmptyOperator(
        task_id="task_2",
    )

    t1 >> t2
```

# Consumer DAG scheduled based on a AssetAlias

```
from datetime import datetime
from airflow import DAG
from airflow.sdk.definitions.asset import AssetAlias


with DAG(
    "datasetalias_triggered_dag",
    description="A DAG that should be triggered via Dataset alias",
    start_date=datetime(2024, 9, 1),
    schedule=[AssetAlias(name="basic_cosmos_dag__orders__run")],
) as dag:

    t3 = EmptyOperator(
        task_id="task_3",
    )

    t3

```

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -116,8 +116,7 @@ from cosmos.operators.base import (
 )
 
 AIRFLOW_VERSION = Version(airflow.__version__)
-AIRFLOW_VERSION = Version(airflow.__version__)
-AIRFLOW_VERSION = Version(airflow.__version__)
+
 logger = get_logger(__name__)
 
 # The following is related to the ability of Cosmos parsing dbt artifacts and generating OpenLineage URIs


tatiana:
BTW: The lines not covered by tests were already not covered by tests before.

dev/dags/performance_dag.py
```diff
@@ -27,9 +27,7 @@
         DBT_ROOT_PATH / "perf",
     ),
     profile_config=profile_config,
-    render_config=RenderConfig(
-        dbt_deps=False,
-    ),
+    render_config=RenderConfig(dbt_deps=False, emit_datasets=False),
```
pankajastro, line 30: Just curious, why are we changing this?

cosmos/operators/local.py
```diff
@@ -151,6 +153,9 @@ class OperatorLineage:  # type: ignore
             job_facets: dict[str, str] = dict()
 
 
+AIRFLOW_VERSION = Version(airflow.__version__)
```
pankajastro: Why you have moved it here from top?

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "packaging>=22.0",
     "pydantic>=1.10.0",
     "virtualenv",
+    "openlineage-integration-common"  # Without this dependency, Airflow datasets/assets don't work
```
pankajastro, line 39: Shall we add this dependency in Cosmos extra `openlineage` instead of here https://github.com/astronomer/astronomer-cosmos/blob/f370999512127307e43dd1f15aeb3411f096b07f/pyproject.toml#L79

dev/dags/performance_dag.py
tatiana: I think we need to review these performance tests altogether...!

Due to changes in Apache Airflow itself, they started raising:
```
  File "/home/runner/.local/share/hatch/env/virtual/astronomer-cosmos/Za_bFbg4/tests.py3.11-2.10-1.9/lib/python3.11/site-packages/sqlalchemy/orm/dependency.py", line 1252, in _synchronize
    self._verify_canload(child)
  File "/home/runner/.local/share/hatch/env/virtual/astronomer-cosmos/Za_bFbg4/tests.py3.11-2.10-1.9/lib/python3.11/site-packages/sqlalchemy/orm/dependency.py", line 257, in _verify_canload
    raise exc.FlushError(
sqlalchemy.orm.exc.FlushError: Can't flush None value found in collection DatasetModel.aliases
```

As reported in:
https://github.com/apache/airflow/issues/42495

This is an example of error:
https://github.com/astronomer/astronomer-cosmos/actions/runs/18469435103/job/52619384083

Given this upstream issue, I thought it was easier to just disable this for this particular DAG.

cosmos/operators/local.py
```diff
@@ -124,10 +115,22 @@
     _sanitize_xcom_key,
 )
```
tatiana, line 117:
```suggestion

AIRFLOW_VERSION = Version(airflow.__version__)
```

cosmos/operators/local.py
tatiana:
```suggestion
```

pyproject.toml
tatiana: No. And the reason is that we are using `openlineage-integration-common` for computing Airflow Dataset/Alias URI name. Cosmos should emit those by default, instead of only working if users add optional dependencies.

cosmos/operators/local.py
tatiana: Just a mistake, fixed

dev/dags/performance_dag.py
pankajastro: Would it make sense to add this issue here, maybe as a comment?

tatiana: The issue is that it is an Airflow problem, not Cosmos. Although it works for Airflow 3, it is not clear if it will work for Airflow 2, so there may be nothing we could do while we still support Airflow 2. That said, if you see work that we could do in this direction, please go ahead and log a ticket!

PR #2035, opened 2025-10-15
Drop Python 3.9 Support

pankajastro:
Drop the Python 3.9 Support from Cosmos 

- Python 3.9 will reach EOL at the end of this month: https://www.python.org/downloads/
- Airflow has already dropped Python 3.9 in the latest release

Success CI: https://github.com/astronomer/astronomer-cosmos/actions/runs/18538281007/job/52839009081?pr=2035

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, drop_py3.9]
+    branches: [main]
   # Also run on pull requests originating from forks. Although this is insecure by default, we need it to run
   # integration tests on forked PRs. As a guardrail, we’ve added an Authorize step to each job, which requires manually
   # approving the workflow run for each pushed commit. Approval only happens after a careful code review of the changes.


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, drop_py3.9]
```
```suggestion
    branches: [main]
```

tatiana (commented):
@pankajastro, as we discussed during the call, I do not believe we should restrict the version of Python at this point for Cosmos end-users. Let's close and reopen this PR once we agree that the time has come.

From a code-complexity perspective, for now, supporting this Python version does not give us additional overhead. Python 3.9 has not been responsible for our CI issues, and Cosmos code base is not more complex because of Python 3.9.

Although Airflow 3 has stopped supporting Python 3.9, Airflow 2.11, which will be supported and maintained for one year ahead, still supports Python 3.9.

pankajastro:
> @pankajastro, as we discussed during the call, I do not believe we should restrict the version of Python at this point for Cosmos end-users. Let's close and reopen this PR once we agree that the time has come.
> 
> From a code-complexity perspective, for now, supporting this Python version does not give us additional overhead. Python 3.9 has not been responsible for our CI issues, and Cosmos code base is not more complex because of Python 3.9.
> 
> Although Airflow 3 has stopped supporting Python 3.9, Airflow 2.11, which will be supported and maintained for one year ahead, still supports Python 3.9.


As I mentioned during our discussion, I’m happy to wait for some time or even close this PR.

The intention behind this PR is not to reduce code complexity—as we can see, the changes only affect the CI—and it’s not meant to directly solve any CI issue either.

Airflow doesn’t backport Python version changes, so once a release is published with a specific Python version, it remains fixed. Airflow also follows a straightforward policy regarding Python versions: they drop support for a Python version in upcoming releases once it reaches EOL. I believe this helps maintainers reduce potential security risks and other related issues.

The main reasons for this PR are:
- Python 3.9 is reaching its end of life this month, so this drop is intended to proactively avoid potential issues.
- To keep the CI clean, making it easier to introduce testing for newer Python versions without adding unnecessary overhead.

**Regarding users on older Airflow versions:**
My understanding is that the oldest version we support is Airflow 2.4, which fortunately supports Python 3.10. So users still on older setups should be able to use Airflow 2.4 with Python 3.10. Ideally, it would be great if they could upgrade their Python version as well to stay current and secure.

pankajkoti:
I agree with @pankajastro's thoughts in [comment](https://github.com/astronomer/astronomer-cosmos/pull/2035#issuecomment-3410379578). My opinion is that we should be aggressive on dropping dbt and Python versions that have reached their EOL.

PR #2040, opened 2025-10-20
Fail sensor tasks immediately if the `ExecutionMode.WATCHER` producer task fails

pankajastro:
relates to: https://github.com/astronomer/astronomer-cosmos/issues/1960

This PR adds a default `on_failure_callback` to the producer task. The callback records the producer task’s state and ensures that the sensor fails immediately if the producer task has failed, improving error visibility and reducing unnecessary wait time.

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -88,11 +88,11 @@ class DbtProducerWatcherOperator(DbtLocalBaseOperator):
         task_id = kwargs.pop("task_id", "dbt_producer_watcher_operator")
         kwargs.setdefault("priority_weight", PRODUCER_OPERATOR_DEFAULT_PRIORITY_WEIGHT)
         kwargs.setdefault("weight_rule", WEIGHT_RULE)
-        on_failure_callback = self._set_on_failure_callback(kwargs)
+        on_failure_callback = self._set_on_failure_callback(kwargs.pop("on_failure_callback", None))
         super().__init__(task_id=task_id, *args, on_failure_callback=on_failure_callback, **kwargs)
 
     def _set_on_failure_callback(
-        self, kwargs: Any
+        self, user_callback: Any
     ) -> Union[Callable[[Context], None], List[Callable[[Context], None]]]:
         default_callback = self._store_producer_task_state
 
@@ -100,9 +100,6 @@ class DbtProducerWatcherOperator(DbtLocalBaseOperator):
             # Older versions only support a single callable
             return default_callback
         else:
-            # Airflow >= 2.6.0 supports a list of callbacks
-            user_callback = kwargs.get("on_failure_callback")
-
             if user_callback is None:
                 # No callback provided — use default in a list
                 return [default_callback]
@@ -320,13 +317,6 @@ class DbtConsumerWatcherSensor(BaseSensorOperator, DbtRunLocalOperator):  # type
             self.model_unique_id,
         )
 
-        producer_task_state = ti.xcom_pull(task_ids=self.producer_task_id, key="state")
-
-        if producer_task_state == "failed":
-            raise AirflowException(
-                f"""The dbt build command failed in producer task. Please check the log of task {self.producer_task_id} for details."""
-            )
-
         # We have assumption here that both the build producer and the sensor task will have same invocation mode
         if not self.invocation_mode:
             self._discover_invocation_mode()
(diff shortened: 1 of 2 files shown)


cosmos/operators/watcher.py
```diff
@@ -282,6 +315,13 @@ def poke(self, context: Context) -> bool:
             self.model_unique_id,
         )
 
+        producer_task_state = ti.xcom_pull(task_ids=self.producer_task_id, key="state")
+
+        if producer_task_state == "failed":
+            raise AirflowException(
+                f"""The dbt build command failed in producer task. Please check the log of task {self.producer_task_id} for details."""
+            )
```
pankajkoti: It may happen that the producer tasks reported failure because only few of the models/resources failed running but the rest of the models succeeded. So, my strong suggestion is that we check for this xcom only once the corresponsing model's status is checked first. And if that xcom is not available then fail the sensor.

pankajastro:
I'll create a follow-up PR to extend the documentation introduced in [PR #2046](https://github.com/astronomer/astronomer-cosmos/pull/2046)

tatiana (commented):
Thanks a lot for working on this, @pankajastro !

I have a behavioural question. In the case of Jaffle Shop, let's say there was an issue applying a seed. The producer task fails. Will all watcher tasks be marked as failed, even if they had not been scheduled and did not start sensing?

Please, could you increase the coverage?

pankajastro:
> I have a behavioural question. In the case of Jaffle Shop, let's say there was an issue applying a seed. The producer task fails. Will all watcher tasks be marked as failed, even if they had not been scheduled and did not start sensing?

If not scheduled, it should go to the `upstream_failed` state

cosmos/operators/watcher.py
pankajastro: I have moved it down to check after model's status is checked

tatiana:
> If not scheduled, it should go to the upstream_failed state

@pankajastro, Did you run a few DAGs to confirm this behaviour and ensure that tasks skipped because upstream tasks failed would not enter sensor mode when they attempt to run for the "first time"?

pankajastro:
> @pankajastro, Did you run a few DAGs to confirm this behaviour and ensure that tasks skipped because upstream tasks failed would not enter sensor mode when they attempt to run for the "first time"?

I'm not sure if I understand it fully, but why should it be skipped? 

Screenshot 2025-10-23 at 8 22 21 AM

tatiana:
@pankajastro In the case of Jaffle Shop, let's say the following happens:

The `dbt_producer_watcher` fails while trying to apply the seed `raw_customers`.
The task `raw_customers`, which was actively watching, will fail (it becomes red). It will have logs for a first try that was unsuccessful. If it is retried, that would be its second try.
However, the task `stg_customers` will have never been started; it would never have been initialised as a sensor. It will likely be marked as `upstream_failed` (I'm asking you to confirm this). If this hypothesis is correct, it will likely never be attempted, and therefore, no logs will be generated for this nonexistent first attempt.

So, my concern is—and this is what I'm asking you to test and confirm—what happens when someone clears the status of the seed `raw_customers` and its downstream tasks? Let's say it succeeds; the task `stg_customers` will then proceed to its second attempt, as expected, and execute `dbt run`. However, I'm worried that the try number on the `customers` task would be 1 (afterall, it is the first time it is actually being executed), and it would enter sensor mode instead of running the `dbt run`, and it would wait "forever" (until timeout happens) for an update from the dbt producer that will never come.

tatiana:
> @pankajastro In the case of Jaffle Shop, let's say the following happens:
> 
> The `dbt_producer_watcher` fails while trying to apply the seed `raw_customers`. The task `raw_customers`, which was actively watching, will fail (it becomes red). It will have logs for a first try that was unsuccessful. If it is retried, that would be its second try. However, the task `stg_customers` will have never been started; it would never have been initialised as a sensor. It will likely be marked as `upstream_failed` (I'm asking you to confirm this). If this hypothesis is correct, it will likely never be attempted, and therefore, no logs will be generated for this nonexistent first attempt.
> 
> So, my concern is—and this is what I'm asking you to test and confirm—what happens when someone clears the status of the seed `raw_customers` and its downstream tasks? Let's say it succeeds; the task `stg_customers` will then proceed to its second attempt, as expected, and execute `dbt run`. However, I'm worried that the try number on the `customers` task would be 1 (afterall, it is the first time it is actually being executed), and it would enter sensor mode instead of running the `dbt run`, and it would wait "forever" (until timeout happens) for an update from the dbt producer that will never come.

I implemented the approach that was discussed with @pankajastro @pankajkoti in #2062

tatiana (approved):
@pankajastro @pankajkoti  and I discussed a few scenarios, and we are extending the original implementation in this PR in a separate PR #2062. We still believe more work will be needed in this area. It seems that Airflow does not expose whether a task has a `State.UPSTREAM_FAILED` - which affects how we'll handle scenarios.

PR #2044, opened 2025-10-21
Fix `ExecutionMode.WATCHER` behaviour with `DbtTaskGroup`

### LINKED ISSUE #1961: Validate Watcher execution mode works with TaskGroups [ExecutionMode.WATCHER]
Validate that the Watcher execution mode works correctly when dbt build and watcher tasks are grouped inside Airflow TaskGroups and Cosmos DbtTaskGroups. Ensure dependencies, XCom usage, and retries continue to work inside nested DAG structures.

related: #1950

tatiana:
While validating `TaskGroup` in a session with an [end-user](https://us-35700.app.gong.io/call?id=1548801879239330851&email_type=call-ready-notification&xtid=45fp79m8frjifyiyqp0) with Cosmos [v1.11.0a8](https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.11.0a8), we identified that it was not working as expected.

There were two main issues:
1. We were not setting the `producer_task_id` argument correctly in the sensor tasks
2. The producer task was always being picked up after the sensor tasks, even though it had a higher priority weight and was an "upstream" task of the dbt root nodes (with a trigger rule always, so they could start sensing).

The first issue was a bug, which is now fixed.

The second issue occurs because the sensor tasks have a `trigger_rule` of "always". This setting works well in the case of `DbtDag`, as the only upstream task is the producer task, and the producer task and root dbt project nodes can be scheduled at the same time, with the task weight of the producer giving its precedence. However, in the case of a `DbtTaskGroup` with one or multiple upstream tasks, the producer task must wait for the upstream task, whereas the dbt project root nodes ("top sensors") do not have to wait.  This leads to sensors occupying the available lots before the producer is scheduled to arrive.

At some point, we suspected this was an issue with Airflow itself, but after a troubleshooting session with Ash, we realised it was not a bug in Airflow: https://github.com/apache/airflow/issues/56723.

The side-effect of the implementation in this PR is that `DbtTaskGroup` and `DbtDag` are behaving differently:
- In `DbtDag`, the producer is upstream of the dbt root nodes
- In `DbtTaskGroup`, the producer is not an upstream root node

The reason we are keeping this behaviour in `DbtDag` is that it is not leading to any issues. It is topologically "correct", making it easier for users to see the dependency between the tasks & potentially view the task list topologically sorted.

At this moment, we were unable to find a way to offer this same behaviour with `DbtTaskGroup`. @ashb , with whom I troubleshooted the original "problem", pointed out Airflow could potentially have a different "state" that wouldn't always be, but something like "start when my upstream task started".  However, by examining [Airflow's current implementation](https://github.com/apache/airflow/blob/c160ab70a00faa41ef5996c4ba4a6580a64194ad/airflow-core/src/airflow/ti_deps/deps/trigger_rule_dep.py), any changes to this area could potentially incur an additional cost to the scheduler and compromise Airflow's performance.

If we were able to "force" in the Airflow UI to have the producer task on the "top" both in the graph view and in the list of functions view, we could make `DbtDag`and `DbtTaskGroup` more consistent (by not having the producer task as upstream and dbt root nodes with trigger_rule always). Currently, in Airflow 3.0.x and Airflow 3.1.0, the producer task has been showing randomly, unfortunately, but we will work with the Airflow team to improve this and add follow-up tickets as needed.

Closes: https://github.com/astronomer/astronomer-cosmos/issues/1961
Closes: https://github.com/astronomer/oss-integrations-private/issues/240


dev/dags/example_watcher.py
```diff
@@ -57,3 +63,28 @@
...
+        execution_config=ExecutionConfig(
+            execution_mode=ExecutionMode.WATCHER,
+        ),
+        project_config=ProjectConfig(DBT_PROJECT_PATH),
+        render_config=RenderConfig(test_behavior=TestBehavior.NONE),
+        profile_config=profile_config,
+        operator_args=operator_args,
+    )
+
+    pre_dbt >> dbt_task_group
```
tatiana:
```suggestion
```

cosmos/airflow/graph.py
```diff
@@ -567,11 +567,28 @@ def _add_producer_watcher(
...
+        # the following only works with DbtDag. It does not work with DbtTaskGroup due to an Airflow bug
+        if "DbtDag" in dag.__class__.__name__:
+            # First, we tackle dbt graph nodes that are root nodes
+            if nodes and node_id in nodes and not nodes[node_id].depends_on:
+                producer_airflow_task >> task_or_taskgroup
+            for root_task in node_tasks:
+                if hasattr(root_task, "trigger_rule"):
+                    root_task.trigger_rule = task_args.get("trigger_rule", "always")
+
+        # We also need to set the producer task id too all consumer tasks, regardless if they are root or not
```
```suggestion
        # We also need to set the producer task ID to all consumer tasks, regardless of whether they are root or not
```

dev/dags/example_watcher.py
```diff
@@ -6,13 +6,16 @@
...
-
-# [END cosmos_init_imports]
 from cosmos.profiles import PostgresUserPasswordProfileMapping
 
+try:  # Airflow 3+
+    from airflow.providers.standard.operators.empty import EmptyOperator
+except ImportError:  # Airflow 2
+    pass
+
+
```
tatiana:
```suggestion
```

cosmos/airflow/graph.py
```diff
@@ -567,11 +567,28 @@ def _add_producer_watcher(
...
-            producer_airflow_task >> task
-            task.trigger_rule = task_args.get("trigger_rule", "always")
+        node_tasks = (
+            list(task_or_taskgroup.children.values())
+            if isinstance(task_or_taskgroup, TaskGroup)
+            else [task_or_taskgroup]
+        )
+
+        # the following only works with DbtDag. It does not work with DbtTaskGroup due to an Airflow bug
+        if "DbtDag" in dag.__class__.__name__:
```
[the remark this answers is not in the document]
tatiana: We cannot use `isinstance` because it would lead to a circular dependency.

cosmos/airflow/graph.py
```diff
@@ -567,11 +567,28 @@ def _add_producer_watcher(
         arguments=producer_task_args,
     )
     producer_airflow_task = create_airflow_task(producer_task_metadata, dag, task_group=task_group)
-    for task_id, task in tasks_map.items():
+
+    # Consumer tasks will need to be updated to use the producer task as a dependency
+    for node_id, task_or_taskgroup in tasks_map.items():
         # we want to make the producer task to be the parent of the root dbt nodes, without blocking them from sensing XCom
```
pankajkoti: Can we ensure these comments are valid for `DbtTaskGroup` too?

cosmos/airflow/graph.py
```diff
@@ -567,11 +567,28 @@ def _add_producer_watcher(
...
+            # First, we tackle dbt graph nodes that are root nodes
+            if nodes and node_id in nodes and not nodes[node_id].depends_on:
+                producer_airflow_task >> task_or_taskgroup
+                root_task = nodes[node_id]
+                if hasattr(root_task, "trigger_rule"):
+                    root_task.trigger_rule = task_args.get("trigger_rule", "always")
+
+        # We also need to set the producer task ID to all consumer tasks, regardless of whether they are root or not
+        for task in node_tasks:
+            task.producer_task_id = producer_airflow_task.task_id  # type: ignore[attr-defined]
```
pankajkoti: Does https://github.com/astronomer/astronomer-cosmos/blob/a9fca6a74d287d4e62f6146ed55ab371d6189d08/cosmos/operators/watcher.py#L183 have significance now?

pankajkoti (approved):
Happy to merge if we've validated the behaviour. Don't have time myself today to validate the issue and the fix.

cosmos/airflow/graph.py
tatiana: Yes! Previously the producer taks was being instantiated before, and the logic wouldn't work

cosmos/airflow/graph.py
tatiana: Absolutely, I'll update the comments so they are up-to-date, @pankajkoti ! Thanks for the feedback.

Unfortunately, the behaviour in `DbtDag` and `DbTaskGroup` is different, as described in the updated PR description. Until we have a solution in the Airflow UI, we should not be able to keep both behaving the same way.

tatiana: No, as per PR description, but I'll make the comments in-code more informative on the why

Issue #1974, opened 2025-09-16
Handle test behaviour for the execution mode Watcher


pankajastro:
Since the execution mode watcher runs the dbt build command, we are limited in the test behaviours we can support for this mode.

✅ Supported Test Behaviours:
- build
- none
- Any other test behaviour should raise an error — ideally during DAG parsing

pankajastro:
Created PR for `TestBehavior.NONE` https://github.com/astronomer/astronomer-cosmos/pull/2047.
For `TestBehavior.NONE`, we already wrap the `EmptyOperator` in Watcher execution mode https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/watcher.py#L355

PR #2047, opened 2025-10-23
Add support for `TestBehavior.NONE` with `ExecutionMode.WATCHER`

pankajastro:
Related to: https://github.com/astronomer/astronomer-cosmos/issues/1974

Screenshot 2025-10-23 at 10 58 33 AM

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -546,14 +546,14 @@ def _add_dbt_setup_async_task(
     tasks_map[DBT_SETUP_ASYNC_TASK_ID] = setup_airflow_task
 
 
-def _add_producer_watcher(
+def _add_producer_watcher_and_dependencies(
     dag: DAG,
     task_args: dict[str, Any],
     tasks_map: dict[str, Any],
     task_group: TaskGroup | None,
     render_config: RenderConfig | None = None,
+    nodes: dict[str, DbtNode] | None = None,
 ) -> str:
-
     producer_task_args = task_args.copy()
 
     if render_config is not None:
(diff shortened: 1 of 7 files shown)


cosmos/operators/watcher.py
```diff
@@ -116,6 +119,9 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
             ti.xcom_push(key="dbt_startup_events", value=startup_events)
 
     def execute(self, context: Context, **kwargs: Any) -> Any:
+        if self.test_behavior == TestBehavior.NONE:
+            dbt_flag_no_test = ["--exclude", "resource_type:test"]
```
anyapriya: When test_behavior is NONE, should this also exclude unit tests in addition to data tests?

pankajastro: Good point. The unit test was introduced in [dbt-1.8.0](https://github.com/dbt-labs/dbt-core/blob/1.8.latest/CHANGELOG.md#dbt-core-180---may-09-2024), so it may be tricky to check the dbt version if dbt is not installed in the same environment as Airflow and add this selective flag.

pankajastro: @tatiana / @pankajkoti any input here?

cosmos/airflow/graph.py
```diff
@@ -552,6 +552,7 @@ def _add_producer_watcher(
     tasks_map: dict[str, Any],
     task_group: TaskGroup | None,
     render_config: RenderConfig | None = None,
+    test_behavior: TestBehavior = TestBehavior.AFTER_EACH,
```
tatiana: The `test_behavior` is part of the `render_config`; we should not introduce a new argument here

cosmos/operators/watcher.py
tatiana: Yes, I agree with @anyapriya ,  it may be better to do `--exclude test_type:*`

cosmos/airflow/graph.py
```diff
@@ -561,6 +562,8 @@ def _add_producer_watcher(
         producer_task_args["selector"] = render_config.selector
         producer_task_args["exclude"] = render_config.exclude
 
+    producer_task_args["test_behavior"] = test_behavior
```
tatiana: @pankajastro It feels we changed several parts of the code interface that did not need to be changed. This PR:

- Exposed the concept of `test_behavior` to the producer task
- Modified the base operator to handle a new `kwarg`, `dbt_flag_no_test`

Did you consider simply extending the producer kwargs "exclude" list to contain  `"test_type:*`"? This could be done in `_add_producer_watcher`, which already receives `render_config`.

tests/airflow/test_graph.py
```diff
@@ -1100,6 +1102,42 @@ def test_owner(dbt_extra_config, expected_owner):
     assert output.leaves[0].owner == expected_owner
 
 
+@pytest.mark.parametrize("test_behavior", [TestBehavior.NONE, TestBehavior.AFTER_EACH])
+def test_test_behavior_for_watcher_mode(test_behavior):
```
tatiana, line 1106: Based on https://github.com/astronomer/astronomer-cosmos/pull/2047/files#r2459297168, my impression is that we do not need to have a complex test validating `build_airflow_graph` - which is a very sophisticated method. We could exclusively focus on the implementation and tests in `_add_producer_watcher` - simplyfing the code base and therefore its maintainance.

cosmos/airflow/graph.py
pankajastro: thanks, removed it

cosmos/airflow/graph.py
pankajastro: Yes, it worked. thanks

tests/airflow/test_graph.py
pankajastro: I liked it because it checks the task count in the DAG, which changes based on test behaviour.

tatiana: Sounds good, thank you!

cosmos/airflow/graph.py
tatiana: That's awesome, thanks a lot!

cosmos/airflow/graph.py
```diff
@@ -561,6 +561,9 @@ def _add_producer_watcher_and_dependencies(
         producer_task_args["selector"] = render_config.selector
         producer_task_args["exclude"] = render_config.exclude
 
+        if render_config.test_behavior == TestBehavior.NONE:
+            producer_task_args["exclude"] = producer_task_args["exclude"] + ["--exclude", "resource_type:test"]
```
tatiana: We do not need the `"--exclude"` here, Cosmos handles this when building the dbt command.

```suggestion
            producer_task_args["exclude"] = producer_task_args["exclude"] +  ["resource_type:test", "resource_type:unit_test"]
```

This also addresses @anyapriya 's [feedback](https://github.com/astronomer/astronomer-cosmos/pull/2047#discussion_r2454739019) on also covering `unit_test`. The way the selection works, in this case, is by union.

tatiana (approved):
@pankajastro, thanks for fixing this, and @anyapriya, for the prompt and spot-on feedback!

I'm happy for this PR to be merged once the [last change](https://github.com/astronomer/astronomer-cosmos/pull/2047#discussion_r2461024649) I requested has been applied and the checks pass.

PR #2062, opened 2025-10-27
Support running `State.UPSTREAM_FAILED` tasks when WATCHER consumer upstream tasks fail

tatiana:
This PR extends the implementation of `ExecutionMode.WATCHER` failure handling in #2040 to tackle running `State.UPSTREAM_FAILED` tasks.

And aims to offer a solution for the scenario described in:
https://github.com/astronomer/astronomer-cosmos/pull/2040#issuecomment-3436342796

> @pankajastro In the case of Jaffle Shop, let's say the following happens:
>
> The `dbt_producer_watcher` fails while trying to apply the seed `raw_customers`.
The task `raw_customers`, which was actively watching, will fail (it becomes red). It will have logs for a first try that was unsuccessful. If it is retried, that would be its second try.
However, the task `stg_customers` will have never been started; it would never have been initialised as a sensor. It will likely be marked as `upstream_failed` (I'm asking you to confirm this). If this hypothesis is correct, it will likely never be attempted, and therefore, no logs will be generated for this nonexistent first attempt.
>
> So, my concern is—and this is what I'm asking you to test and confirm—what happens when someone clears the status of the seed `raw_customers` and its downstream tasks? Let's say it succeeds; the task `stg_customers` will then proceed to its second attempt, as expected, and execute `dbt run`. However, I'm worried that the try number on the `customers` task would be 1 (afterall, it is the first time it is actually being executed), and it would enter sensor mode instead of running the `dbt run`, and it would wait "forever" (until timeout happens) for an update from the dbt producer that will never come.

Previously - in #2040 - if the status of the seed `raw_customers` was cleared and that task was run sucessfully falling back to `ExecutionMode.LOCAL`, the downstream tasks would fail. This happened because it was their first try, and they'd see the producer task had failed, and they would fail, giving a poor user-experience.

With the current PR, we attempt to understand if a task is a `State.UPSTREAM_FAILED` task based on the newly introduced property `poke_retry_number`.

Relates to: #1960


pankajkoti (approved):
The conditionals and approach look good to me now. Happy to merge once CI is green.

cosmos/operators/watcher.py
```diff
@@ -184,6 +184,7 @@ def _callback(ev: EventMsg) -> None:
 
 class DbtConsumerWatcherSensor(BaseSensorOperator, DbtRunLocalOperator):  # type: ignore[misc]
     template_fields = ("model_unique_id",)  # type: ignore[operator]
+    poke_retry_number: int = 0
```
pankajastro, line 187: Does poke_retry_number reset to 0 after a worker restart, or does it retain its previous value?

pankajastro (approved):

cosmos/operators/watcher.py
tatiana: When a worker starts, a new Python process is begun and this value is reset

PR #2076, opened 2025-11-03
Fix broken CI due to fastapi incompatibility with cadwyn for Airflow 3

pankajkoti:
Our CI started failing after fastapi [0.121.0 ](https://pypi.org/project/fastapi/0.121.0/) was released yesterday. It's observed that cadwyn(that airflow uses for its migrations) is [incompatible](https://github.com/zmievsa/cadwyn/issues/305) with the latest the release of fastapi. Hence, putting an upper bound for fastapi version which can be removed once cadwyn becomes compatible.

Succesful run: https://github.com/astronomer/astronomer-cosmos/actions/runs/19061976838

closes: #2079

PR #2077, opened 2025-11-03
refactor: remove duplicated _construct_dest_file_path

### LINKED ISSUE #1746: Refactor: remove code duplicate
Currently (Cosmos 1.10), we define the method `_construct_dest_file_path both` in two places:
* [cosmos.io](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/io.py)
* [cosmos.operators.local.AbstractDbtLocalBase._construct_dest_file_path](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/local.py)

They are slightly different, but we could try to unify and reuse them - to avoid potential future issues.

jx2lee:
## Description

Refactor destination path construction for remote object storage uploads in the Astronomer Cosmos library.

- a new helper function `construct_dest_file_path()` in `cosmos/io.py`, responsible for normalizing and building destination paths for artifact upload.
- Remove duplicated `_construct_dest_file_path` logic from the local dbt operator class; instead reuse the shared helper to keep both upload and local operator path logic
- Add a unit test (`tests/test_io.py`) covering `_construct_dest_file_path()` 

## Related Issue(s)

closes #1746

## Breaking Change?

No. This refactor preserves existing destination path semantics while changing internal implementation. Consumers invoking upload methods or operators will see no behavior change.

## Checklist

### DIFF
--- a/tests/operators/test_local.py
+++ b/tests/operators/test_local.py
@@ -32,7 +32,6 @@ from cosmos.dbt.parser.output import (
 )
 from cosmos.exceptions import CosmosDbtRunError, CosmosValueError
 from cosmos.hooks.subprocess import FullOutputSubprocessResult
-from cosmos.io import _construct_dest_file_path
 from cosmos.operators.local import (
     AbstractDbtLocalBase,
     DbtBuildLocalOperator,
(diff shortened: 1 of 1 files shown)


tests/test_io.py
```diff
@@ -75,16 +76,24 @@ def test_configure_remote_target_path_no_remote_target():
     assert _configure_remote_target_path() == (None, None)
 
 
-def test_construct_dest_file_path(dummy_kwargs):
-    """Test _construct_dest_file_path."""
+def testconstruct_dest_file_path(dummy_kwargs):
```
tatiana:
```suggestion
def test_construct_dest_file_path(dummy_kwargs):
```

cosmos/io.py
```diff
@@ -174,28 +174,35 @@ def _configure_remote_target_path() -> tuple[Path, str] | tuple[None, None]:
     return _configured_target_path, remote_conn_id
 
 
-def _construct_dest_file_path(
-    dest_target_dir: Path,
-    file_path: str,
-    source_target_dir: Path,
-    source_subpath: str,
-    **kwargs: Any,
+def construct_dest_file_path(
```
tatiana: It's probablu worth to keep the `_` so end-users don't use this as part of the Cosmos library, so if we need to change in future, it is not considered a breaking change:

```suggestion
def _construct_dest_file_path(
```

tatiana (commented):
Thanks a lot for doing this refactor, @jx2lee. Changes like this really help the Cosmos code base be easier to maintain.

I provided minor feedback inline, and I'm happy for us to merge this change once the tests pass.

Once we merge #2076, please, rebase based on the `main` branch

jx2lee:
> Thanks a lot for doing this refactor, @jx2lee. Changes like this really help the Cosmos code base be easier to maintain.
> 
> I provided minor feedback inline, and I'm happy for us to merge this change once the tests pass.
> 
> Please, rebase based on the `main` branch

@tatiana Thanks for waiting ! I was on a trip(🇯🇵), so update took a bit longer. Rebase is all done, too!

PR #2080, opened 2025-11-04
Refactor `airflow/graph.py` to simplify code-base

tatiana:
The goal with this PR is to refactor the `airflow/graph.py`  to make the code more manageable.

The PR #1759 should be merged before the current changes are applied.

The complexity of building DAGs has increased significantly with #1759. As part of this PR, we ended up duplicating configurations and argument passing in multiple parts of the code, adding more complexity to the `airflow/graph.py` module.

This PR also aims to add documentation changes that will help users of `node_converters` and `node_conversion_by_task_group`.

**Breaking changes**

The following functions are expected to be used internally only to Cosmos, so we hope these won't impact end-users, but we are documenting the changes just in case:
  - ``generate_task_or_group`` receives ``render_config`` instead of its individual configurations, such as ``test_behavior``, ``source_rendering_behavior`` and ``enable_owner_inheritance``
  - ``create_task_metadata`` receives ``render_config`` instead of its individual configurations, such as ``test_behavior``, ``source_rendering_behavior`` and ``enable_owner_inheritance``
  - ``create_task_metadata`` now expects the ``node_converters`` argument

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -102,7 +102,6 @@ class RenderConfig:
     normalize_task_display_name: Callable[..., Any] | None = None
     should_detach_multiple_parents_tests: bool = False
     enable_owner_inheritance: bool | None = True
-    node_conversion_by_task_group: bool | None = True
 
     def __post_init__(self, dbt_project_path: str | Path | None) -> None:
         if self.env_vars:


CHANGELOG.rst
```diff
@@ -1,6 +1,30 @@
 Changelog
 =========
 
+1.12.0 (2025-11-04)
+-------------------
+
+Breaking changes
+
+* The following functions are expected to be used internally only to Cosmos, so we hope these won't impact end-users, but we are documenting the changes just in case:
```
tatiana:
```suggestion
* The following functions are expected to be used internally only by Cosmos, so we hope these won't impact end-users, but we are documenting the changes just in case:
```

cosmos/config.py
tatiana, line 66:
```suggestion
    :param node_converters: a dictionary mapping a ``DbtResourceType`` into a callable. Users can control how to render dbt nodes in Airflow. Only supported when using ``load_method=LoadMode.DBT_MANIFEST`` or ``LoadMode.DBT_LS``.
    :param node_conversion_by_task_group: A boolean that allows users to do node conversion at the task group level instead of task level.  Defaults to True.
```

tatiana:
```suggestion
```

docs/configuration/render-config.rst
tatiana, line 31:
```suggestion
```

pankajastro (approved):

PR #2084, opened 2025-11-05
Add deferrable support for Watcher Sensor

### LINKED ISSUE #2059: [Enhancement] Support using deferrable sensors on `ExecutionMode.WATCHER`
### Description

As of Cosmos 1.11.0, the `ExecutionMode.WATCHER` consumer nodes are sensing for changes in the producer node synchronously. We should change the pokes to happen on the `triggerer`, to further optimise resource utilisation in Cosmos with this execution mode.

### Use case/motivation

_No response_

### Related issues

_No response_

pankajastro:
This PR adds async/deferrable support to the Watcher sensor

**Benefits**
- Reduces resource usage for long-running watcher tasks.
- Improves task scheduling and throughput.

**Limitation**
- Supports deferrable execution for model or run commands only.
- Retry execution remains synchronous, and deferrable happens only for try no 0.

**Future**
- Improve poke_interval and timeout handling to address performance issue: https://github.com/astronomer/astronomer-cosmos/issues/1973

closes: https://github.com/astronomer/astronomer-cosmos/issues/2059

Screenshot 2025-11-06 at 12 28 27 AM

Screenshot 2025-11-06 at 2 44 31 PM


cosmos/_triggers/watcher.py
```diff
@@ -0,0 +1,133 @@
...
+            if node_status == "success":
+                self.log.info("Model '%s' succeeded", self.model_unique_id)
+                yield TriggerEvent({"status": "success"})  # type: ignore[no-untyped-call]
+                return
+            elif node_status == "failed":
+                self.log.warning("Model '%s' failed", self.model_unique_id)
+                yield TriggerEvent({"status": "failed"})  # type: ignore[no-untyped-call]
+                return
+            elif dr_state == "failed":
+                yield TriggerEvent({"status": "failed"})  # type: ignore[no-untyped-call]
```
pankajkoti, line 121: can we make the messaging consistent with poke method here to include in the error logs

"The dbt build command failed in producer task. Please check the log of task {self.producer_task_id} for details."

cosmos/_triggers/watcher.py
```diff
@@ -0,0 +1,133 @@
...
+            (r for r in results if r.get("unique_id") == self.model_unique_id),
+            {},
+        )
+        return node_result.get("status")
+
+    async def run(self) -> AsyncIterator[TriggerEvent]:
+        self.log.info("Starting WatcherTrigger for model: %s", self.model_unique_id)
+
+        while True:
+            dr_state = await self.get_xcom_val("state")
```
pankajkoti:
```suggestion
            producer_task_state = await self.get_xcom_val("state")
```
Isn't it the producer task state?

pankajkoti (approved):
It's looking great to me. Have a couple of comments inline. 

Also, looks like a few tests are failing in the CI.

cosmos/_triggers/watcher.py
pankajastro: Yes, that's correct. fixed

cosmos/_triggers/watcher.py
pankajastro: Added this log message in `execute_complete` method

tatiana (commented):
@pankajastro, this is really exciting, thanks for finding a solution for it!
Do you see any downsides or risks with always using the deferrable implementation? What happens if the end-user does not have a triggerer in their Airflow deployment?
Should we also support people opting out of using the deferrable mode?
Finally, are there any performance improvements when switching to deferrable, when running `ExecutionMode.WATCHER` with our [benchmark project](https://github.com/astronomer/cosmos-benchmark)?

pankajastro:
> Do you see any downsides or risks with always using the deferrable implementation? What happens if the end-user does not have a triggerer in their Airflow deployment?

I’m a bit concerned about the number of database accesses from the trigger.

> Should we also support people opting out of using the deferrable mode?

I’ve thought about it, and I think we should expose a parameter at the sensor level to conditionally defer. However, it would be better to do this once https://github.com/astronomer/astronomer-cosmos/issues/1973 is solved.

> Finally, are there any performance improvements when switching to deferrable, when running ExecutionMode.WATCHER with our [benchmark project](https://github.com/astronomer/cosmos-benchmark)?

I have not done any performance testing.

tatiana (approved):
Looks great, thank you very much, @pankajastro !

PR #2087, opened 2025-11-06
Fix `ExecutionMode.WATCHER` deadlock in Airflow 3.0 & 3.1

### LINKED ISSUE #2057: [Bug] dbt Command hanging in the producer task with `ExecutionMode.WATCHER`
### Astronomer Cosmos Version

1.11.0a12

### dbt-core version

1.11.0-b3

### Versions of dbt adapters

dbt-bigquery=1.10.2

### LoadMode

AUTOMATIC

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

Astronomer Runtime 3.1-2 based on Airflow 3.1.0+astro.2

### Operating System

Astro CLI (Linux) on MacOS

_No response_

### Deployment

Other

### Deployment details

Local Astro CLI using Dockerfile
```
FROM astrocrpublic.azurecr.io/runtime:3.1-2
```

And requirements:
```
astronomer-cosmos==1.11.0a13

apache-airflow-providers-amazon[s3fs]
apache-airflow-providers-microsoft-azure
apache-airflow-providers-google
apache-airflow-providers-snowflake
apache-airflow-providers-databricks
```

### What happened?

When trying out `ExecutionMode.WATCHER` with the DAG:
```
from datetime import datetime, timedelta
from pathlib import Path

from airflow import DAG

from cosmos.airflow.dag import DbtDag
from cosmos.airflow.task_group import DbtTaskGroup
from cosmos.config import ProfileConfig, ProjectConfig, RenderConfig, ExecutionConfig
from cosmos.constants import LoadMode, ExecutionMode, InvocationMode, TestBehavior


PROFILE_PATH = Path("/usr/local/airflow/dbt/fhir-dbt-analytics/profiles.yml")
PROJECT_PATH = Path("/usr/local/airflow/dbt/jaffle_shop")

    execution_mode=ExecutionMode.WATCHER
)
#execution_config = ExecutionConfig()

    profile_name="fhir_dbt_analytics",
    target_name="dev",
    profiles_yml_filepath=PROFILE_PATH,
)


cosmos_jaffle_shop = DbtDag(
    dag_id="cosmos_jaffle_shop",
    project_config=ProjectConfig(PROJECT_PATH),
    execution_config=execution_config,
    tags=["watcher"],
)

operator_args = {
    "execution_timeout": timedelta(seconds=120),
}


with DAG(
    dag_id="cosmos_jaffle_shop_taskgroup",
    schedule="@daily",
    catchup=False,
):
    """
    The simplest example of using Cosmos to render a dbt project as a TaskGroup.
    """
    pre_dbt = EmptyOperator(task_id="pre_dbt")

    first_dbt_task_group = DbtTaskGroup(
        group_id="first_dbt_task_group",
        execution_config=execution_config,
        project_config=ProjectConfig(PROJECT_PATH),
        operator_args=operator_args,
    )

    pre_dbt >> first_dbt_task_group

```

The producer task hanged running the dbt command for a very long time, without completing, until I cancelled it. I haven't experienced this issue before.

Within 12 attempts to run this DAG, I saw this happen twice, in different parts of the dbt pipeline. The producer task should take less than 30s to run, and in both scenarios it took more than this - and I manually cancelled them:

Image

Image

### Relevant log output

```shell
Logs displayed in the screenshot - the dbt command just hang in different stages.
```

### How to reproduce

I had to re-run this DAG multiple times to get a similar behaviour.

### Anything else :)?

There are a few hypotheses why this happened:
1. There was a delay running the query against BQ (the database we were using in this demo)
2. There is an issue using the dbt Core callback feature, which sometimes gets lost (leading to it hanging indefinitely) - this may be specific to this dbt-core beta version
3. There is an issue with Airflow 3.1.0 when running a subprocess when it sometimes hangs

For now, we'll track the behaviour and see how often this happens. We should find the root cause and fix it.

### Contact Details

_No response_

tatiana:
Airflow 3.0 and 3.1 are currently not thread-safe when running `xcom_push`, which leads to deadlocks.

The problem description and troubleshooting are detailed in ticket #2057.

Unfortunately, since this issue is not deterministic, I was unable to create an automated test to reproduce and prevent regressions. That said, given that the producer node hangs indefinitely until the task times out, and given that we were no longer able to reproduce the issue after the current changes, I firmly believe we should proceed with this change.

Closes: #2057

Co-authored-by: Ash Berlin-Taylor <ash@astronomer.io>

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -6,6 +6,7 @@ import logging
 import zlib
 from datetime import timedelta
 from pathlib import Path
+from threading import Lock
 from typing import TYPE_CHECKING, Any, Callable, List, Union
 
 import airflow
@@ -14,7 +15,9 @@ from packaging.version import Version
 if TYPE_CHECKING:  # pragma: no cover
     try:
         from airflow.sdk.definitions.context import Context
+        from airflow.sdk.types import RuntimeTaskInstanceProtocol as TaskInstance
     except ImportError:
+        from airflow.models.taskinstance import TaskInstance  # type: ignore[assignment]
         from airflow.utils.context import Context  # type: ignore[attr-defined]
 
 try:
@@ -50,11 +53,23 @@ except ImportError:  # pragma: no cover
     EventMsg = None
 
 logger = logging.getLogger(__name__)
-
+xcom_set_lock = Lock()
 
 CONSUMER_OPERATOR_DEFAULT_PRIORITY_WEIGHT = 10
 PRODUCER_OPERATOR_DEFAULT_PRIORITY_WEIGHT = 9999
 WEIGHT_RULE = "absolute"  # the default "downstream" does not work with dag.test()
+
+
+def safe_xcom_push(task_instance: TaskInstance, key: str, value: Any) -> None:
+    """
+    Safely set an XCom value in a thread-safe manner in Airflow 3.0 and 3.1.
+    We noticed that the combination of using dbt (multi-threaded) and Airflow 3.0 and 3.1 to set Xcom lead to race conditions.
+    This leads the producer task to get stuck while running the dbt build command.
+    Unfortunately, since this is non-deterministic, and happens once every five runs, we were not able to have a proper test.
+    However, we applied this fix and run over 20 times a pipeline that would fail every 5 runs and this allowed us to no longer face the issue.
+    """
+    with xcom_set_lock:
+        task_instance.xcom_push(key=key, value=value)
 
 
 class DbtProducerWatcherOperator(DbtLocalBaseOperator):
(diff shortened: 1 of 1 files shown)


cosmos/operators/watcher.py
```diff
@@ -50,13 +53,25 @@
...
 
 CONSUMER_OPERATOR_DEFAULT_PRIORITY_WEIGHT = 10
 PRODUCER_OPERATOR_DEFAULT_PRIORITY_WEIGHT = 9999
 WEIGHT_RULE = "absolute"  # the default "downstream" does not work with dag.test()
 
 
+def safe_xcom_push(task_instance: TaskInstance, key: str, value: Any) -> None:
+    """
+    Safely set an XCom value in a thread-safe manner in Airflow 3.0 and 3.1.
+    We noticed that the combination of using dbt (multi-threaded) and Airflow 3.0 and 3.1 to set Xcom lead to race conditions.
```
```suggestion
    We noticed that the combination of using dbt (multi-threaded) and Airflow 3.0 and 3.1 to set XCom lead to race conditions.
```

cosmos/operators/watcher.py
```diff
@@ -50,13 +53,25 @@
...
+    This leads the producer task to get stuck while running the dbt build command.
+    Unfortunately, since this is non-deterministic, and happens once every five runs, we were not able to have a proper test.
+    However, we applied this fix and run over 20 times a pipeline that would fail every 5 runs and this allowed us to no longer face the issue.
+    """
```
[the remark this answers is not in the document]
tatiana: Addressed in b8e9de1e9a97892db60adc46a87085d235eee7cb

cosmos/operators/watcher.py
```diff
@@ -50,13 +53,25 @@
     EventMsg = None
 
 logger = logging.getLogger(__name__)
-
+xcom_set_lock = Lock()
```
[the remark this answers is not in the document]
tatiana: Module-level lock is right. It needs to be global per process.

pankajastro (approved):
Looks good! It might be nice to add a regression test to make sure this issue doesn’t come back in the future.

cosmos/operators/watcher.py
```diff
@@ -50,13 +53,25 @@
...
 
 CONSUMER_OPERATOR_DEFAULT_PRIORITY_WEIGHT = 10
 PRODUCER_OPERATOR_DEFAULT_PRIORITY_WEIGHT = 9999
 WEIGHT_RULE = "absolute"  # the default "downstream" does not work with dag.test()
 
 
+def safe_xcom_push(task_instance: TaskInstance, key: str, value: Any) -> None:
+    """
+    Safely set an XCom value in a thread-safe manner in Airflow 3.0 and 3.1.
```
pankajkoti, line 65: Do we need to mention specific Airflow versions here? Because we are not altering the behavior here based on Airflow versions here.

tatiana:
> Looks good! It would be beneficial to add a regression test to ensure this issue doesn’t recur in the future.

I fully agree, @pankajastro. I discussed this with @ashb, but since it would happen only from time to time, and I couldn't reproduce it with Airflow `dag.test()` or `airflow dags test`, I couldn't find an easy way to accomplish this. If we test against Astro in the future, we could ensure that we run multiple tests and have a method for avoiding this from regressing.

cosmos/operators/watcher.py
tatiana: I think we should mention we're using this because of these versions bug, but it is also harmless to use for others. It simplifies our code-base to not have to handle this differently based on Airflow version. If Airflow fixes this, we can then do the version check

PR #2089, opened 2025-11-07
Unify airflow version handling into `constants.py`

tatiana:
As part of supporting Airflow 3.0 and 3.1, we ended up adding very similar code in multiple places - this PR aims to simplify and clean it up.

Cosmos code base had logic similar to the following repeated in over 15 modules.
```
import airflow
from packaging.version import Version

AIRFLOW_VERSION = Version(airflow.__version__)
```

This PR refactors the code to define the `AIRFLOW_VERSION` in `constants.py`, to avoid duplication.

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -318,9 +318,9 @@ class AbstractDbtLocalBase(AbstractDbtBase):
             return None, None
 
         if not settings.AIRFLOW_IO_AVAILABLE:
-raise CosmosValueError(
-f"You're trying to specify remote target path {target_path_str}, but the required "
-f"Object Storage feature is unavailable in Airflow version {AIRFLOW_VERSION}. Please upgrade to "
+            raise CosmosValueError(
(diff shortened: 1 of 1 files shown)


cosmos/operators/local.py
```diff
@@ -321,6 +318,8 @@ def _configure_remote_target_path() -> tuple[Path | ObjectStoragePath, str] | tu
             return None, None
 
         if not settings.AIRFLOW_IO_AVAILABLE:
+            from airflow.version import version as airflow_version
+
             raise CosmosValueError(
                 f"You're trying to specify remote target path {target_path_str}, but the required "
                 f"Object Storage feature is unavailable in Airflow version {airflow_version}. Please upgrade to "
```
pankajkoti:
```suggestion
raise CosmosValueError(
f"You're trying to specify remote target path {target_path_str}, but the required "
f"Object Storage feature is unavailable in Airflow version {AIRFLOW_VERSION}. Please upgrade to "
```

Would like to check if we can do this here?

PR #2092, opened 2025-11-08
docs: Fix default values in documentation

dnskr:
## Description

The PR fixes default values in the documentation for `models_relative_path`, `seeds_relative_path` and `snapshots_relative_path`.

Current values are taken from:
https://github.com/astronomer/astronomer-cosmos/blob/975ebae2b4c6ef75dd442678fb538771bc717829/cosmos/config.py#L166-L168

Before changes:
image

After changes:
image

## Breaking Change?

No

## Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -43,7 +43,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v6.0.0
+      uses: actions/checkout@v6.0.1
       with:
         persist-credentials: false
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -12,7 +12,7 @@ jobs:
       contents: read
 
     steps:
-      - uses: actions/checkout@v6.0.0
+      - uses: actions/checkout@v6.0.1
         with:
           persist-credentials: false
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@v6.0.0
+      - uses: actions/checkout@v6.0.1
         with:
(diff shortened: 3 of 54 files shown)


docs/configuration/project-config.rst
```diff
@@ -40,7 +37,7 @@ Project Config Example
     config = ProjectConfig(
         dbt_project_path="/path/to/dbt/project",
         models_relative_path="models",
-        seeds_relative_path="data",
```
tatiana: This path was intentionally different than the default, we shouldn't change it.

docs/configuration/project-config.rst
```diff
@@ -40,7 +37,7 @@ Project Config Example
...
+        seeds_relative_path="seeds",
```
tatiana: This path was intentionally different than the default seeds, we shouldn't revert to the default, we could add something like `custom_seeds_folder` if you think it would be more explicit.

tatiana (commented):
Well spot, @dnskr , it makes sense to change those values based on the dbt documentation:
* https://docs.getdbt.com/reference/project-configs/snapshot-paths
* https://docs.getdbt.com/reference/project-configs/model-paths
* https://docs.getdbt.com/reference/project-configs/seed-paths

Since we implemented this feature, dbt now supports a list of paths instead of a single path.

Since this would be a breaking change, I've logged a ticket (#2100) for Cosmos 2.0 so we support a list of paths instead of single path.

docs/configuration/project-config.rst
dnskr: The value corresponds to an incorrect default used for `seeds_relative_path`, so I assumed it is also incorrect. If the difference is intentional, should I also change `models_relative_path` and `snapshots_relative_path` to non-default values here or remove them from the example?

tatiana: That would be great, thanks, @dnskr !

dnskr: The custom values have been changed as suggested in the comment.

PR #2093, opened 2025-11-08
Fix minor documentation typo

dnskr:
## Description

Fix minor typo in the documentation.

## Breaking Change?

No

## Checklist

### DIFF
--- a/docs/configuration/generating-docs.rst
+++ b/docs/configuration/generating-docs.rst
@@ -80,7 +80,7 @@ You can use the :class:`~cosmos.operators.DbtDocsGCSOperator` to generate and up
     from cosmos.operators import DbtDocsGCSOperator
 
     # then, in your DAG code:
-    generate_dbt_docs_aws = DbtDocsGCSOperator(
+    generate_dbt_docs_gcs = DbtDocsGCSOperator(
         task_id="generate_dbt_docs_gcs",
         project_dir="path/to/jaffle_shop",
         profile_config=profile_config,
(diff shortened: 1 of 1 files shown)


docs/configuration/generating-docs.rst
```diff
@@ -80,7 +80,7 @@ You can use the :class:`~cosmos.operators.DbtDocsGCSOperator` to generate and up
     from cosmos.operators import DbtDocsGCSOperator
 
     # then, in your DAG code:
-    generate_dbt_docs_aws = DbtDocsGCSOperator(
```
tatiana, line 83: Thanks for fixing this, @dnskr.

I strongly suggest that instead of hard-coding this here, that we reference an example code that is being run as part of Cosmos' CI:
https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/cosmos_manifest_example.py#L76-L88

Using the same pattern as:
https://github.com/astronomer/astronomer-cosmos/blob/d76bae0f71fed8102fce1168e391fb4888b5a3e2/docs/getting_started/watcher-execution-mode.rst?plain=1#L104-L107

dnskr: I think the following DAG and code should be used as a docs generation example:
https://github.com/astronomer/astronomer-cosmos/blob/db76233da894d185b074c86dbbd9f47c67d18bd7/dev/dags/dbt_docs.py#L64-L71

Also, the changes should be applied to the AWS and Azure examples for consistency.

However, the DAG code does not fully match existing example code. Would you like me to make the changes or should I leave it as is?

pankajastro: I think this PR fixes the typo only. I agree we should not hardcode it if possible, but maybe we can address this different PR since this example is hardcoded before the changes in this Pr.

Issue #2086, opened 2025-11-06
Sensors tasks hanging when producer tasks fail with OOM


pankajkoti:
One of our customers mentioned that they observed using the Watcher Execution mode that in one of the runs, the Producer task failed due to OOM but the dbt nodes sensor tasks were stuck in running state and that they needed to fail the DAG manually.

We already have a on_failure_callback implementation https://github.com/astronomer/astronomer-cosmos/blob/a95c81b86a863959b4bc0976a313fe75c54da5bb/cosmos/operators/watcher.py#L98C33-L98C59 that stores the producer task as failed and the sensor tasks consume this to fail the sensor tasks when the producer task has failed. However, looks like it may not be getting called when the producer tasks fail due to OOM. Reproduce and find a solution for this issue so that sensor tasks can be failed early.

pankajkoti:
I see that the process itself is exiting and we do not get a chance to catch the exception or push to XCOM the failure status

<img width="4928" height="1682" alt="Image" src="https://github.com/user-attachments/assets/f2e55f07-f1bb-4340-bf53-ecbde06eb5a7" />

pankajkoti:
We added PRs #2126, #2144, and #2145 to address this issue. Hence, closing the ticket now

PR #2101, opened 2025-11-10
Guard watcher callback exceptions to avoid hanging producer tasks

### LINKED ISSUE #2102: DbtProducerWatcherOperator callback exceptions hang WATCHER producer tasks
### Summary

When ExecutionMode.WATCHER runs via `DbtProducerWatcherOperator`, the dbtRunner callback can raise exceptions (e.g., while processing startup or node-finished events). dbt does not bubble up that exception and the Airflow task never finishes, leaving the DAG stuck without any clear failure signal.

### Reproduction
1. 
2. Run a WATCHER-mode DAG (InvocationMode.DBT_RUNNER).
3. Force `_handle_startup_event` or `_handle_node_finished` to raise (for example by monkeypatching them to throw).
4. Observe that the producer task stops making progress and never completes; no error is surfaced in Airflow.
### Expected

Callback errors should be captured and surfaced so the producer task can either fail or finish cleanly, allowing downstream sensors to proceed and operators to see useful logs.

### Actual

The exception vanishes (not bubbled up) inside dbtRunner, leaving the Airflow task hanging indefinitely with no diagnostic output.

pankajkoti:
The PR adds the following changes to address #2102:
- wrap the dbtRunner callback in DbtProducerWatcherOperator with a try/except so any failure during event handling is logged instead of silently hanging the airflow task
- ensure `_handle_startup_event`/`_handle_node_finished` exceptions surface via `logger.exception` while letting execution continue
- add a regression test covering the callback error path (`test_execute_callback_exception_is_logged`)
 
closes: #2102

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -191,11 +191,18 @@ class DbtProducerWatcherOperator(DbtLocalBaseOperator):
             if use_events:
 
                 def _callback(event_message: EventMsg) -> None:
-                    name = event_message.info.name
-                    if name in {"MainReportVersion", "AdapterRegistered"}:
-                        self._handle_startup_event(event_message, startup_events)
-                    elif name == "NodeFinished":
-                        self._handle_node_finished(event_message, context)
+                    try:
+                        name = event_message.info.name
+                        if name in {"MainReportVersion", "AdapterRegistered"}:
+                            self._handle_startup_event(event_message, startup_events)
+                        elif name == "NodeFinished":
+                            self._handle_node_finished(event_message, context)
+                    except Exception:
+                        event_name = getattr(getattr(event_message, "info", None), "name", "unknown")
+                        logger.exception(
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Looks great, @pankajkoti ! Thanks for improving this.

We still need to come up with a solution for the SIGKILL due to [OOM](https://airflow.apache.org/docs/apache-airflow/stable/troubleshooting.html#out-of-memory-error-oom) that some users are facing - but this is a part of a bigger task along the Airflow team.

pankajkoti:
> We still need to come up with a solution for the SIGKILL due to [OOM](https://airflow.apache.org/docs/apache-airflow/stable/troubleshooting.html#out-of-memory-error-oom) that some users are facing - but this is a part of a bigger task along the Airflow team.

Yes, we have a ticket for this: https://github.com/astronomer/astronomer-cosmos/issues/2086

PR #2114, opened 2025-11-11
Force watcher producer retries to zero

### LINKED ISSUE #2105: Set `ExecutionMode.WATCHER` producer retry to 0
- Update the code
- Update the documentation

pankajkoti:
The PR makes the following changes:
- hard-set `DbtProducerWatcherOperator` `retries` (and `default_args["retries"]`) to zero so WATCHER never auto-retries the full dbt build
-  Log at INFO that producer retries are disabled, and fail fast if Airflow attempts a retry (try number > 1), improving user visibility into the behavior.
- explain the rationale inline (sensor retry path handles per-model reruns)
- document behavior in the WATCHER guide and add tests covering the forced retry value

closes: #2105

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -183,6 +183,25 @@ class DbtProducerWatcherOperator(DbtLocalBaseOperator):
         safe_xcom_push(task_instance=context["ti"], key="state", value="failed")
 
     def execute(self, context: Context, **kwargs: Any) -> Any:
+        task_instance = context.get("ti")
+        if task_instance is None:
+            raise AirflowException("DbtProducerWatcherOperator expects a task instance in the execution context")
+
+        try_number = getattr(task_instance, "try_number", 1)
+
+        if try_number > 1:
+            retry_message = (
+                "Dbt WATCHER producer task does not support Airflow retries. "
+                f"Detected attempt #{try_number}; failing fast to avoid running a second dbt build."
+            )
+            self.log.error(retry_message)
+            raise AirflowException(retry_message)
+
+        self.log.info(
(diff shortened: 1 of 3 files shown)


cosmos/operators/watcher.py
```diff
@@ -102,6 +102,14 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         task_id = kwargs.pop("task_id", "dbt_producer_watcher_operator")
         kwargs.setdefault("priority_weight", PRODUCER_OPERATOR_DEFAULT_PRIORITY_WEIGHT)
         kwargs.setdefault("weight_rule", WEIGHT_RULE)
+        # Consumer watcher retry logic handles model-level reruns using the LOCAL execution mode; rerunning the producer
+        # would repeat the full dbt build and duplicate watcher callbacks which may not be processed by the consumers if
+        # they have already processed output XCOMs from the first run of the producer, so we disable retries.
+        default_args = dict(kwargs.get("default_args", {}) or {})
+        default_args["retries"] = 0
+        kwargs["default_args"] = default_args
+        kwargs["retries"] = 0
```
tatiana, line 111: It is extremely important that we have an info-level log explaining to the end-user that currently the producer task does not allow retries - and for this reason we're setting it to 0.

We should also probably check during the `execute` method, if the try is higher than the first run, we should fail and log why.

pankajkoti: Addressed both in https://github.com/astronomer/astronomer-cosmos/pull/2114/commits/aa2ce1ad7477876ab083818f74d6c7daa6e5e09b

tatiana (approved):
Great work, @pankajkoti , thank you!

PR #2117, opened 2025-11-12
Remove unused Python3.9 uses from Github action CI

pankajastro:
This PR clean unused Python3.9 uses from Github action CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, clean3.9]
+    branches: [main]
   # Also run on pull requests originating from forks. Although this is insecure by default, we need it to run
   # integration tests on forked PRs. As a guardrail, we’ve added an Authorize step to each job, which requires manually
   # approving the workflow run for each pushed commit. Approval only happens after a careful code review of the changes.


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, clean3.9]
```
```suggestion
    branches: [main]
```

tatiana (commented):
When did we remove Python 3.9 support..? My understanding is that we agreed to not do this change for now: #2035. I discussed with @yetudada and she was happy for us to continue support Python 3.9 for now as well. The Python 3.9 are the fastest tests to run in our CI and they have not caused mantainance overhead, so I'd like to see them being re-added.

tatiana:
I just saw it was removed by mistake in this PR:
https://github.com/astronomer/astronomer-cosmos/commit/a9fca6a74d287d4e62f6146ed55ab371d6189d08#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88

We should re-add it, and skip it only for the latest versions of Airflow that don't support it.

pankajkoti:
On the contrary, I agree with @pankajastro's thoughts in [comment](https://github.com/astronomer/astronomer-cosmos/pull/2035#issuecomment-3410379578). My opinion is that we should be aggressive on dropping dbt and Python versions that have reached their EOL.

tatiana:
I just checked the two utilisation sources we have, Astro Cloud customers, and Scarf. And I agree for us to remove Python 3.9 support.

The main reasoning:
- We currently have 0.4% of Astro customers who have Cosmos installed using Python 3.9.
- Within the Scarf data, we see the top 10 `python_version` utilised, and all of them are Python 3.10 or higher

More details in the [slack discussion](https://astronomer.slack.com/archives/C096EPSPUQ2/p1762955887669939).

tatiana (approved):

pankajastro:
> @pankajastro I believe it's important that we highlight in our user facing docs and Changelog that starting Cosmos 1.12.0 we're dropping support for Python 3.9. Would be nice if you could add a PR for the same when possible.

yes, created followup PR: https://github.com/astronomer/astronomer-cosmos/pull/2118

PR #2118, opened 2025-11-12
Drop Python 3.9 support

pankajastro:
Based on the discussion in [PR #2117](https://github.com/astronomer/astronomer-cosmos/pull/2117), we are officially dropping support for Python 3.9 from the main branch. Python 3.9 support will be removed starting with Cosmos v1.12.0.


pankajkoti (commented):
Is it possible to add a highlight in the Changelog for 1.12.0 that we're dropping support for Python 3.9?

tatiana (approved):
Thanks, @pankajastro ! I agree with @pankajkoti , if you could add this to the changelog, it would be great.

We already breaking changes section.

Issue #2072, opened 2025-10-29
Improve visibility of dbt command logs in `ExecutionMode.WATCHER` consumer tasks


tatiana:
Improve visibility of dbt command logs in `ExecutionMode.WATCHER` consumer tasks

pankajkoti:
Also address: https://github.com/astronomer/astronomer-cosmos/pull/2124#pullrequestreview-3459746264 to show model failure details in logs in a pretty/readable format for both deferrable and non-deferrable mode of the sensors.

tatiana:
Requested by customer: https://astronomer.slack.com/archives/C0607HRA2SC/p1765221813498099?thread_ts=1764677139.702229&cid=C0607HRA2SC

pankajastro:
In this ticket, we aim to:

- [x] Track all possible node statuses, not just terminal states
- [x] Track node start and finish timestamps
- [x] Capture error logs
- [x] Record the dbt-core and adapter versions

pankajastro:
https://github.com/astronomer/astronomer-cosmos/pull/2392 to track all possible node statuses, not just terminal states

pankajastro:
https://github.com/astronomer/astronomer-cosmos/pull/2412 to record the dbt-core and adapter versions

pankajastro:
draft PR to logs error message: https://github.com/astronomer/astronomer-cosmos/pull/2414

pankajastro:
Created https://github.com/astronomer/astronomer-cosmos/pull/2431 to capture error log, start-time, finish-time and non-terminal state

PR #2124, opened 2025-11-13
Clarify watcher deferrable failure messaging

### LINKED ISSUE #2123: Failure state error messaging is inaccurate in deferrable sensors for Watcher mode
For Watcher mode deferrable sensors, in the case of failure events which can happen either due to the model failing or the producer task failing before running the model, we're always raising in execute_complete saying that the producer task failed as can be seen at https://github.com/astronomer/astronomer-cosmos/blob/4994474b308e14e496bee97e3f44c2f0a8a37150/cosmos/operators/watcher.py#L398.

We already have separate error handling for those events ([model failed](https://github.com/astronomer/astronomer-cosmos/blob/4994474b308e14e496bee97e3f44c2f0a8a37150/cosmos/_triggers/watcher.py#L116), [producer failed](https://github.com/astronomer/astronomer-cosmos/blob/4994474b308e14e496bee97e3f44c2f0a8a37150/cosmos/_triggers/watcher.py#L120))  but looks like we don't propogate the differentiating reason from the TriggerEvents to the worker. 

We should raise exception with appropriate error messaging suggesting the end users on whether the model failed or the producer task failed when using deferrable sensors

pankajkoti:
The PR makes the following changes:
 - Enrich `WatcherTrigger` failure events with explicit `reason` codes (model_failed vs producer_failed) and log producer task crashes so deferrable sensors can react accurately.
  - Update `DbtConsumerWatcherSensor.execute_complete` to raise targeted messages for model and producer failures and otherwise exit cleanly.
  - Align the watcher operator/trigger tests with the new payload shape and add coverage for the reason-specific messaging paths.
 
closes: #2123

### DIFF
--- a/cosmos/_triggers/watcher.py
+++ b/cosmos/_triggers/watcher.py
@@ -115,10 +115,15 @@ class WatcherTrigger(BaseTrigger):
                 return
             elif node_status == "failed":
                 self.log.warning("Model '%s' failed", self.model_unique_id)
-                yield TriggerEvent({"status": "failed"})  # type: ignore[no-untyped-call]
+                yield TriggerEvent({"status": "failed", "reason": "model_failed"})  # type: ignore[no-untyped-call]
                 return
             elif producer_task_state == "failed":
-                yield TriggerEvent({"status": "failed"})  # type: ignore[no-untyped-call]
+                self.log.error(
+                    "Watcher producer task '%s' failed before delivering results for model '%s'",
+                    self.producer_task_id,
+                    self.model_unique_id,
+                )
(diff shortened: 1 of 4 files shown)


tatiana (approved):
This looks great, thanks a lot, @pankajkoti !
Moving forward (in a follow up PR), if the model failed, would there be a possibility to give more details on the failure?

pankajkoti:
> Moving forward (in a follow up PR), if the model failed, would there be a possibility to give more details on the failure?

Yes, I guess it should be done and could be tacked as part of ticket https://github.com/astronomer/astronomer-cosmos/issues/2072. I also added a [comment](https://github.com/astronomer/astronomer-cosmos/issues/2072#issuecomment-3527901384) about the same in the ticket

PR #2140, opened 2025-11-18
Improve example DAG `jaffle_shop_kubernetes.py`

tatiana:
Previously, the `DbtSeedKubernetesOperator` had a misleading `profile_config` argument that used a `ProfileMapping` class, which was not being used. 

Cosmos 1.11 does not support `ProfileMapping` during the execution of tasks created with `ExecutionMode.KUBERNETES`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,8 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
+default_language_version:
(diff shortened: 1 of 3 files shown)


pankajkoti:
CI seems to be failing with the change

tatiana:
> CI seems to be failing with the change
But it worked on my machine..! 😂 
I'll fix it before merging, thanks for the speedy review, @pankajkoti!

PR #2144, opened 2025-11-20
Watcher: fetch async producer status from Airflow runtime

pankajkoti:
With this PR, the Watcher deferrable trigger now mirrors the synchronous sensor (as implemented in PR #2126) by querying the producer task state through Airflow instead of the state XCom. The PR adds `_get_producer_task_status` to `WatcherTrigger` (database lookup in Airflow 2, `RuntimeTaskInstance` in Airflow 3) and wires the `run()` method to utilise it, so async sensors raise `producer_failed` events when the producer dies.
  
related: #2126 
related: #2086

### DIFF
--- a/cosmos/_triggers/watcher.py
+++ b/cosmos/_triggers/watcher.py
@@ -11,6 +11,8 @@ from airflow.triggers.base import BaseTrigger, TriggerEvent
 from asgiref.sync import sync_to_async
 from packaging.version import Version
 
+from cosmos._utils.watcher_state import build_producer_state_fetcher
+
 AIRFLOW_VERSION = Version(airflow.__version__)
 
 
@@ -72,7 +74,7 @@ class WatcherTrigger(BaseTrigger):
                         task_id=self.producer_task_id,
                         run_id=self.run_id,
                     )
-                    .first()
+                    .one_or_none()
                 )
                 if ti is None:
                     return None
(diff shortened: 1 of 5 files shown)


cosmos/_triggers/watcher.py
```diff
@@ -103,11 +105,58 @@ async def _parse_node_status(self) -> str | None:
...
+                    )
+                    if ti is not None:
+                        return str(ti.state)
+                    return None
+
+            return await sync_to_async(_fetch_state)()
+
+        try:
+            from airflow.sdk.execution_time.task_runner import RuntimeTaskInstance
+        except (ImportError, NameError) as exc:
```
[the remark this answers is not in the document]
pankajkoti: It does give NameError in cases when run via dags.test() command. Hence, need it here.

tests/_triggers/test_watcher.py
```diff
@@ -6,6 +8,15 @@
...
 
+def _import_side_effect(name: str, *args, **kwargs):
+    if name == "airflow.sdk.execution_time.task_runner":
+        raise ModuleNotFoundError("missing runtime")
+    return _real_import(name, *args, **kwargs)
+
+
+_real_import = __import__
+
+
```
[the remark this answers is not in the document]
pankajkoti: ignoring for now since no real impact. don't want to run the CI again for this suggestion in the tests, but I will remember it for next time.

cosmos/_triggers/watcher.py
```diff
@@ -72,8 +72,10 @@ def _get_xcom_val() -> Any | None:
                         task_id=self.producer_task_id,
                         run_id=self.run_id,
                     )
-                    .one()
+                    .first()
```
tatiana: Why are we getting the `.first()`? Would it make sense to capture the last?

cosmos/_triggers/watcher.py
```diff
@@ -103,11 +105,58 @@ async def _parse_node_status(self) -> str | None:
...
+            def _fetch_state() -> str | None:
+                with create_session() as session:
+                    ti = (
+                        session.query(TaskInstance)
+                        .filter_by(
+                            dag_id=self.dag_id,
+                            task_id=self.producer_task_id,
+                            run_id=self.run_id,
+                        )
+                        .first()
```
tatiana: Same question here: https://github.com/astronomer/astronomer-cosmos/pull/2144/files#r2545505614

cosmos/_triggers/watcher.py
```diff
@@ -103,11 +105,58 @@ async def _parse_node_status(self) -> str | None:
         )
         return node_result.get("status")
 
+    async def _get_producer_task_status(self) -> str | None:
+        """Retrieve the producer task state for both Airflow 2 and Airflow 3."""
+
+        if AIRFLOW_VERSION < Version("3.0.0"):
```
tatiana: Currently, it feels like we're duplicating, to an extent, the logic in https://github.com/astronomer/astronomer-cosmos/pull/2126. Would it make sense to refactor the code so part of the logic could be reused both in the sync and async functions

tests/_triggers/test_watcher.py
```diff
@@ -6,6 +8,15 @@
 from cosmos._triggers.watcher import AIRFLOW_VERSION, WatcherTrigger
 
 
+def _import_side_effect(name: str, *args, **kwargs):
```
tatiana: Why are we artificially causing this to happen? Shouldn't this be the natural path if we're using an older version of Airflow, and this case would be covered by the CI without the need fo mocking the import path?

tatiana (commented):
Thanks a lot for addressing this, @pankajkoti, and splitting the changes so we can apply them incrementally in both Cosmos 1.11.2 and 1.12.0.

Please, find questions in-line

cosmos/_triggers/watcher.py
pankajkoti: I checked the Airflow codebase and see that the standard approach seems to be using one_or_none() https://github.com/apache/airflow/blob/b34e63a246daa1ad6a2c27ac782faad5775c7476/airflow-core/src/airflow/api_fastapi/core_api/routes/ui/dags.py#L290

Hence, using that now. Also appears that .one() / .first() could raise errors and we're expecting it to return None when records are not found, so one_or_none() helps in that case.

cosmos/_triggers/watcher.py
pankajkoti: same as https://github.com/astronomer/astronomer-cosmos/pull/2144#discussion_r2547142589

cosmos/_triggers/watcher.py
pankajkoti: yes, refactored to move the common logic in _utils/watcher_state.py and using that across `operators/watcher.py` and `_triggers/watcher.py`

tests/_triggers/test_watcher.py
pankajkoti: It's actually for Airflow 3's RunTimeTaskInstance import. The import is available when running tests in Airflow 3, but we have a path to handle the import failure to `return None` that was added after observing failures in integration tests when using dags.test(). Hence, adding this unit test mock to simulate an import failure in the case of Airflow 3 and cover the None return. I have localised this method to the test now.

cosmos/_triggers/watcher.py
tatiana: Thanks a lot, @pankajkoti ! Are there any circumstances in which we'd get multiple values?

pankajkoti: I would not expect duplicates for the self.producer_task_id for a specific dag run id, unless we have a bug in calculating in populating the `self.producer_task_id`, or there exists an Airflow bug.

tatiana (approved):
Looks great, thank you very much!

PR #2152, opened 2025-11-22
Support real-time consumer updates when using `ExecutionMode.WATCHER` and `InvocationMode.SUBPROCESS`

### LINKED ISSUE #2060: [Enhancement] Consume from dbt struct log instead of callback in `ExecutionMode.WATCHER`
For giving end-users a "real time" update of the pipeline progress, the `ExecutionMode.WATCHER` implementation in Cosmos 1.11.0 uses dbt Core [callback](https://docs.getdbt.com/reference/programmatic-invocations#registering-callbacks) feature.

However, this only works if dbt and Airflow were installed in the same Python virtual env.

An alternative approach would be to leverage:
https://docs.getdbt.com/reference/events-logging#structured-logging

This would allow us to give a better experience when not having Airflow and dbt co-existing in the same environment, while also being a step towards having dbt Fusion support and K8s pod operator support.

For this to work, we'd need to change:
https://github.com/astronomer/astronomer-cosmos/blob/994c719cf69a6c8fe0b405c64a15e8ec956bb67b/cosmos/hooks/subprocess.py#L75-L94

To something like:
```
            self.sub_process = Popen(
                command,
                stdout=PIPE,
                stderr=STDOUT,
                cwd=cwd,
                env=env if env or env == {} else os.environ,
                preexec_fn=pre_exec,
                bufsize=1,               # Line-buffered
                universal_newlines=True  # Decode text directly
            )

            if self.sub_process is None:
                raise RuntimeError("The subprocess should be created here and is None!")

            print("Command output (streaming):")

            line = ""
            if self.sub_process.stdout is not None:
                for line in self.sub_process.stdout:
                    line = line.rstrip()
                    log_lines.append(line)

                    # Print to console
                    print(line)

                    # Send to external callback if provided
                    if on_output:
                        on_output(line)

            # Wait for process to finish
            exit_code = self.sub_process.wait()
```

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/2060
This PR modifies `ExecutionMode.WATCHER` so that, when `InvocationMode.SUBPROCESS` is used, sensor tasks can still be updated in real time based on dbt status information from structured JSON logs, rather than parsing `run_results.json` after the full dbt pipeline is executed. The change enables real-time status tracking for subprocess-based dbt invocations.

Key Changes:

- Added ``--log-format json`` flag to dbt build commands in watcher mode
- Implemented real-time JSON log parsing in subprocess hook to extract and store node status to XCom
- Refactored XCom utility functions to a new common module for better code organization

We can see here that the producer task is running but some node has already succeeded for WATCHER mode with SUBPROCESS invocation
Screenshot 2025-11-22 at 7 29 35 PM


cosmos/hooks/subprocess.py
```diff
@@ -28,15 +31,33 @@ class FullOutputSubprocessHook(BaseHook):  # type: ignore[misc]
     """Hook for running processes with the ``subprocess`` module."""
 
     def __init__(self) -> None:
-        self.sub_process: Popen[bytes] | None = None
+        self.sub_process: Popen[str] | None = None
         super().__init__()  # type: ignore[no-untyped-call]
 
+    def _store_dbt_resource_status_from_log(self, line: str, **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
pankajastro: Added docstring

cosmos/_utils/common.py
```diff
@@ -0,0 +1,26 @@
...
+    This leads the producer task to get stuck while running the dbt build command.
+    Unfortunately, since this is non-deterministic, and happens once every five runs, we were not able to have a proper test.
+    However, we applied this fix and run over 20 times a pipeline that would fail every 5 runs and this allowed us to no longer face the issue.
+    """
+    with xcom_set_lock:
+        task_instance.xcom_push(key=key, value=value)
+
+
+def get_xcom_val(task_instance: TaskInstance, task_ids: str | list[str], key: str) -> Any:
+    return task_instance.xcom_pull(task_ids, key=key)
```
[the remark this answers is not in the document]
pankajastro: Added a commet

cosmos/hooks/subprocess.py
```diff
@@ -28,15 +31,33 @@ class FullOutputSubprocessHook(BaseHook):  # type: ignore[misc]
...
+        try:
+            log_line = json.loads(line)
+            node_status = log_line.get("data", {}).get("node_info", {}).get("node_status")
+            unique_id = log_line.get("data", {}).get("node_info", {}).get("unique_id")
+
+            self.log.debug("Model: %s is in %s state", unique_id, node_status)
+
+            if node_status in ["success" or "failed"]:
+                context = kwargs.get("context")
+                assert context is not None  # Make MyPy happy
```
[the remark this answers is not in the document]
pankajastro: I want this fail when context is none

cosmos/operators/base.py
```diff
@@ -325,8 +325,9 @@ class DbtBuildMixin:
 
     template_fields: Sequence[str] = ("full_refresh",)
 
-    def __init__(self, full_refresh: bool | str = False, **kwargs: Any) -> None:
+    def __init__(self, full_refresh: bool | str = False, log_format: str | None = None, **kwargs: Any) -> None:
         self.full_refresh = full_refresh
+        self.log_format = log_format
         super().__init__(**kwargs)
```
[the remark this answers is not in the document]
pankajastro: Added docstring

cosmos/hooks/subprocess.py
```diff
@@ -28,15 +31,33 @@ class FullOutputSubprocessHook(BaseHook):  # type: ignore[misc]
     """Hook for running processes with the ``subprocess`` module."""
 
     def __init__(self) -> None:
-        self.sub_process: Popen[bytes] | None = None
+        self.sub_process: Popen[str] | None = None
```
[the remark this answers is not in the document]
pankajastro: This requires as we need to pass `text=True` or `universal_newlines=True` in `self.sub_process = Popen(...)`

cosmos/_utils/common.py
```diff
@@ -0,0 +1,27 @@
+from threading import Lock
```
tatiana: @pankajastro let's avoid a generic module name like `common.py`. Let's call this `xcom.py`

cosmos/_triggers/watcher.py
```diff
@@ -87,23 +87,21 @@ async def get_xcom_val(self, key: str) -> Any | None:
...
+
+            data_json = _parse_compressed_xcom(compressed_xcom_val)
             return data_json.get("data", {}).get("run_result", {}).get("status")  # type: ignore[no-any-return]
 
-        results = data_json.get("results", [])
-        node_result: dict[str, Any] = next(
-            (r for r in results if r.get("unique_id") == self.model_unique_id),
-            {},
-        )
-        return node_result.get("status")
```
tatiana, line 106: Why did we remove this part of the code?

cosmos/hooks/subprocess.py
```diff
@@ -28,15 +31,41 @@ class FullOutputSubprocessHook(BaseHook):  # type: ignore[misc]
...
+    def _store_dbt_resource_status_from_log(self, line: str, **kwargs: Any) -> None:
+        """
+        Parses a single line from dbt JSON logs and stores node status to Airflow XCom.
+
+        This method parses each log line from dbt when --log-format json is used,
+        extracts node status information, and pushes it to XCom for consumption
+        by downstream watcher sensors.
+        """
+        try:
+            log_line = json.loads(line)
```
tatiana, line 46: Could we wrap the `json.JSONDecodeError` exception handling only in this line?

cosmos/hooks/subprocess.py
```diff
@@ -28,15 +31,41 @@ class FullOutputSubprocessHook(BaseHook):  # type: ignore[misc]
...
+            node_status = log_line.get("data", {}).get("node_info", {}).get("node_status")
+            unique_id = log_line.get("data", {}).get("node_info", {}).get("unique_id")
+
+            self.log.debug("Model: %s is in %s state", unique_id, node_status)
+
+            # TODO: Handle and store all possible node statuses, not just the current success and failed
+            if node_status in ["success", "failed"]:
```
tatiana: What are the other possible status? Could we add a link to the documentation?

cosmos/hooks/subprocess.py
```diff
@@ -79,26 +108,32 @@ def pre_exec() -> None:
                 cwd=cwd,
                 env=env if env or env == {} else os.environ,
                 preexec_fn=pre_exec,
+                bufsize=1,  # line-buffered (works only in text mode)
+                text=True,
+                encoding=output_encoding,
+                errors="backslashreplace",
```
tatiana, line 114: Are there any downsides to using this configuration?

tatiana (approved):
This looks very good, @pankajastro. Thanks a lot for the work.
I left minor feedback, and I'm happy for the changes to be merged.

cosmos/_utils/common.py
pankajastro: Move this to `_utils/watcher_state.py` for now

cosmos/hooks/subprocess.py
pankajastro: Ok, reduced the scope for exception  handling

cosmos/hooks/subprocess.py
pankajastro: I think other state is "started"

cosmos/_triggers/watcher.py
pankajastro: This parsing was performed to retrieve the status of a dbt resource from the run_result. Now, dbt resource state is available in xcom, so we don't need parsing

cosmos/hooks/subprocess.py
pankajastro: If it won't be able to decode a byte then replace it with backslash escapes which may not be user friendly

cosmos/hooks/subprocess.py
pankajastro: Created issue: https://github.com/astronomer/astronomer-cosmos/issues/2181 to track it

Issue #2069, opened 2025-10-28
[DX] Move a few logs to the debug channel


matthieuauger:
### Description

Hello,

Currently my task output is getting filled with a lot of cosmos-related information I don't really need.

<img width="1870" height="343" alt="Image" src="https://github.com/user-attachments/assets/3837cd54-c291-4d9f-955f-dc44266bdf45" />

I tried to remove them but the only way I found is to reduce my whole log level and I miss useful INFO log if I do that.

Would you open for a Pull Request to move these lines of logs to the DEBUG channel?

I think they are mainly useful for DEBUGGING purposes but do not need to be visible everytime.

Thanks

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

matthieuauger:
Otherwise, if you have a solution to quiet them it would be really useful.

tatiana:
Hi @matthieuauger Thank you for proposing it. Since you offered to help in this area, I just assigned this ticket to you.

Are these logs from a single task?

Regarding the solution, I believe the main problem is the "duplicate" (4x) logs. If we were able to address this, then we should be good on having them as info. Would you be up to try to solve this using this approach?

matthieuauger:
Hello @tatiana thanks for the anwser et I'm happy to work on it.

I agree that the duplicate logs could be an issue but I also think that INFO logs should contain informations that is useful _most of the time_.

Here the parsing method, the node count, information about the DagBag looks like information useful when something is not working, for debugging purpose. Personally except when I changed my astro configuration (for example to parse a Manifest), I don't read these information at all on a daily basis. Any chances that we could lower the logging level?

Maybe another possibility is to have an option for the astronomer logs level but it would add a bit of complexity.

Regarding the duplicate logs do you have any insights on this? Could that be a configuration problem on my side? I see that I have the same bunch of entries for my different DAGs I don't know if other people have the same thing

Best

pankajkoti:
Closing since PR: https://github.com/astronomer/astronomer-cosmos/pull/2235 has improved the readability by grouping the parsing logs, which get collapsed by default. This was released with Cosmos 1.13.0. Requesting you try it out. I am closing this issue, but please re-open the ticket or file a new one if you feel there is something more we could do here.

PR #2163, opened 2025-11-25
move debug logs to debug channel

matthieuauger:
## Description

These logs are displayed on every task multiple times. They hide meaningful information logged in the tasks.
There are more useful when trying to debug performance issues.

Here is an example of what is shown in my tasks, this makes the logs less readable

image

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/2069

## Breaking Change?

This could be a breaking change is anyone is using these logs programatically but I doubt that is the case

## Checklist

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -576,8 +576,8 @@ class DbtGraph:
 
         self.update_node_dependency()
 
-        logger.info("Total nodes: %i", len(self.nodes))
-        logger.info("Total filtered nodes: %i", len(self.filtered_nodes))
+        logger.debug("Total nodes: %i", len(self.nodes))
+        logger.debug("Total filtered nodes: %i", len(self.filtered_nodes))
 
     def run_dbt_ls(
         self, dbt_cmd: str, project_path: Path, tmp_dir: Path, env_vars: dict[str, str]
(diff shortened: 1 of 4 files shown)


matthieuauger:
hello @pankajastro tests should be good now, just need your approval to execute tests again on CI

matthieuauger:
@pankajastro is this normal that my CI is blocked? Thanks

pankajastro:
> @pankajastro is this normal that my CI is blocked? Thanks

Yes, for security reasons and to avoid potential misuse, CI requires approval.

pankajastro:
I know the duplicated log lines aren’t ideal, but I think it’s important to understand why they’re happening.

pankajastro:
> I know the duplicated log lines aren’t ideal, but I think it’s important to understand why they’re happening.

I tried running a few DAGs and couldn’t see duplicate logs. We may need a broader discussion about changing the log level and deciding what we should and shouldn’t log.

pankajastro:
These logs are helpful for many users—they make it easier to understand what’s happening under the hood and also give a rough idea of performance.

PR https://github.com/astronomer/astronomer-cosmos/pull/2235 adds ``Airflow ::group::`` to group logs. I think the changes in the PR: https://github.com/astronomer/astronomer-cosmos/pull/2235 will significantly improve readability, and we may not need to change the log level to debug.

cosmos/dbt/graph.py
```diff
@@ -576,8 +576,8 @@ def load(
 
         self.update_node_dependency()
 
-        logger.info("Total nodes: %i", len(self.nodes))
-        logger.info("Total filtered nodes: %i", len(self.filtered_nodes))
+        logger.debug("Total nodes: %i", len(self.nodes))
+        logger.debug("Total filtered nodes: %i", len(self.filtered_nodes))
```
pankajastro, line 580: Maybe we can combine both and keep keep as info

pankajastro:
Closing since PR: https://github.com/astronomer/astronomer-cosmos/pull/2235 has improved the readability, and we will cut alpha soon for testing

Issue #1589, opened 2025-03-04
[Bug] mock profile for DatabricksOauthProfileMapping does not include `auth_type: 'oauth'`, which breaks `dbt ls` parsing when trying to load DAG


charlielor:
### Astronomer Cosmos Version

1.9.0

### dbt-core version

1.9.2

### Versions of dbt adapters

dbt-databricks==1.9.5

### LoadMode

DBT_LS

### ExecutionMode

LOCAL

### InvocationMode

DBT_RUNNER

### airflow version

2.10.3

### Operating System

Amazon Linux 2023

### Deployment

Amazon (AWS) MWAA

### What happened?

When Cosmos tries to parse a new Airflow DAG using `DatabricksOauthProfileMapping` as profile mapping, it is using mock values for the profile. Because the dbt-databricks adapter requires `auth_type: 'oauth'` when not using `token` (and using `client_id` and `client_secret`), dbt throws an error saying:

```
The config `auth_type: oauth` is required when not using access token`. 
```

and parsing of the DAG fails.

To get around this, I had to set RenderConfig's `enable_mock_profile` to `False`. 

### Relevant log output

```shell

```

### How to reproduce

1. Enable Airflow DAG processing log and set it log level INFO
2. Create an Airflow connection called `dbt-databricks` that uses `client_id`/`client_secret` as login/password, with `extra.http_path` as the path to SQL warehouse
3. Upload example DAG:
```
import logging

from datetime import datetime
from cosmos import DbtDag, ProjectConfig, ProfileConfig, ExecutionConfig, RenderConfig
from cosmos.profiles import DatabricksOauthProfileMapping
from cosmos.constants import TestBehavior, LoadMode

logger = logging.getLogger(__name__)

dag = DbtDag(
  project_config=ProjectConfig(dbt_project_path="/usr/local/airflow/dags/dbt-databricks-example"),
  profile_config=ProfileConfig(
    profile_name="dbt_databricks_example",
    target_name="prod",
    profile_mapping=DatabricksOauthProfileMapping(
      conn_id='dbt-databricks',
      profile_args=dict(
        catalog='dbt_databricks_example',
        schema='mart',
      ),
    ),
  ),
  operator_args=dict(full_refresh=False, cancel_query_on_kill=True, install_deps=True),
  render_config=RenderConfig(
    load_method=LoadMode.DBT_LS,
    select=['path:models/intermediate'],
    exclude=[],
    test_behavior=TestBehavior.NONE,
[...]

PR #2164, opened 2025-11-25
Add databricks oauth mock profile

fjmacagno:
## Description

Adds an override for mock_profile to DatabricksOauthProfileMapping so that the `dbt ls` command doesn't fail with mocked credentials.

## Related Issue(s)

Fixes https://github.com/astronomer/astronomer-cosmos/issues/1589 

## Checklist


fjmacagno:
Due to https://github.com/databricks/dbt-databricks/issues/1274 it would be more effective to make this not output client_id and client_secret and auth_type in the mocked profile, but that wouldn't be consistent behavior with the other profile classes.

Issue #2131, opened 2025-11-14
Improve DAG versioning test in Airflow 3


tatiana:
Scenarios:

1) no changes, no new version
    - have a Cosmos-powered DAG
    - airflow dags reserialise
    - check dag_version in the table
    - airflow dags reserialise
    - check dag_version is the same

2) changes that should not lead to DAG change, no new version
    - have a Cosmos-powered DAG
    - airflow dags reserialise
    - make a change that should not affect the DAG version
    - check dag_version in the table
    - airflow dags reserialise
    - check dag_version is the same

3) change the dbt project, confirm there is a new version
    - have a Cosmos-powered DAG
    - airflow dags reserialise
    - make a change that should not affect the DAG version
    - check dag_version in the table
    - airflow dags reserialise
    - check dag_version changed

tatiana:
Just chatted with @michal-mrazek and he's picking this up. For some reason, GH didn't allow me to assign it to him:
<img width="1348" height="343" alt="Image" src="https://github.com/user-attachments/assets/9a79e339-62f0-4ba1-8997-9d32a08b0e71" />

michal-mrazek:
Thank @tatiana, weird, I am sure I was assigned to some of the former issues, thanks though!

PR #2177, opened 2025-12-02
test: add Airflow 3 DAG versioning tests for Cosmos

michal-mrazek:
## Description

This PR adds a test for Airflow 3 versioning. We verify that adding a model to DbtDag will change a dag version.

## Related Issue(s)

https://github.com/astronomer/astronomer-cosmos/issues/2131

## Checklist

### DIFF
--- a/dev/dags/cosmos_manifest_selectors_example.py
+++ b/dev/dags/cosmos_manifest_selectors_example.py
@@ -19,7 +19,7 @@ from cosmos.profiles import DbtProfileConfigVars, PostgresUserPasswordProfileMap
 DEFAULT_DBT_ROOT_PATH = Path(__file__).parent / "dbt"
 DBT_ROOT_PATH = Path(os.getenv("DBT_ROOT_PATH", DEFAULT_DBT_ROOT_PATH))
 
-execution_config = ExecutionConfig(dbt_project_path=DBT_ROOT_PATH / "jaffle_shop")
+execution_config = ExecutionConfig(dbt_project_path=DBT_ROOT_PATH / "altered_jaffle_shop")
 
 profile_config = ProfileConfig(
     profile_name="default",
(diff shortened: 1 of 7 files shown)


pankajastro:
Thank you so much for the work on this PR — I really appreciate the effort! 🙌

One quick note: Cosmos handles DAG versioning a bit differently. For context, Cosmos added DAG-versioning support in this PR: https://github.com/astronomer/astronomer-cosmos/pull/1907

It might be more aligned with the Cosmos approach to structure the test like:

- Create a DAG object
- Run the DAG using the helper here: https://github.com/astronomer/astronomer-cosmos/blob/main/tests/utils.py#L29
and check the version with `DagVersion.get_latest_version(...)`
- Update the dbt project
- Run again and assert the version changes

Curious to hear what you think — happy to collaborate on the approach!

michal-mrazek:
> Thank you so much for the work on this PR — I really appreciate the effort! 🙌
> 
> One quick note: Cosmos handles DAG versioning a bit differently. For context, Cosmos added DAG-versioning support in this PR: #1907
> 
> It might be more aligned with the Cosmos approach to structure the test like:
> 
> * Create a DAG object
> * Run the DAG using the helper here: https://github.com/astronomer/astronomer-cosmos/blob/main/tests/utils.py#L29
>   and check the version with `DagVersion.get_latest_version(...)`
> * Update the dbt project
> * Run again and assert the version changes
> 
> Curious to hear what you think — happy to collaborate on the approach!

Thanks @pankajastro, I will adjust the PR 👍

michal-mrazek:
@pankajastro I started using the `run_dag()` as you suggested. I also switched from `jaffle shop` to `simple` dbt project, since we are actually running it and I didn't want it to take ages :) 

If you have any other suggestions, happy to include them :)

tests/test_airflow_versioning.py
```diff
@@ -0,0 +1,163 @@
...
+
+    version_1 = DagVersion.get_latest_version(test_dag_id)
+
+    # Add new dbt model (cosmos dag should change)
+    (dbt_project_test_dir / "models" / "dummy_model.sql").write_text("SELECT 2 AS id, 'example' AS name")
+
+    # Re-parse DAG, serialize and run
+    dagbag = DagBag(dag_folder=dbt_project_test_dir)
+    dag_v2 = dagbag.dags[test_dag_id]
+    serialize_dag(dag_v2)
```
pankajastro, line 118: Why do we need the `serialize_dag` fixture? We could remove it entirely. Here’s an example test where we run the DAG: https://github.com/astronomer/astronomer-cosmos/blob/main/tests/operators/test_local.py#L675

pankajastro:
> @pankajastro I started using the run_dag() as you suggested. I also switched from jaffle shop to simple dbt project, since we are actually running it and I didn't want it to take ages :)

> If you have any other suggestions, happy to include them :)

Thanks for addressing it. We run jaffle_shop in many tests, so it may be ok to run it. I have one more [feedback](https://github.com/astronomer/astronomer-cosmos/pull/2177/files#r2610561613) otherwise looking good

tests/test_airflow_versioning.py
michal-mrazek: @pankajastro I tried running without the serializer, but I didnt manage to make it work. It is a bit different usecase since we need to run it twice. Maybe, since you have more exp with cosmos than me, can you also try to remove the serialization step? :)

pankajkoti (commented):
Thanks for contributing this valuable PR @michal-mrazek 

A few tests are failing in the CI. Would you be able to investigate those, please?

Also, issue #2131 mentions in point 2 to also add a test for "changes that should not lead to DAG change, no new version". Do you think we can add that test as well as part of this PR?

michal-mrazek:
Hello @pankajkoti , hopefully I fixed the tests now :)

I did not find a scenarion when "changes that should not lead to DAG change, no new version". To by knowledge, all changes lead to a new version, or am I wrong here? :)

tatiana:
@michal-mrazek, it feels we're really close to merging this PR. If you get a chance, please, fix the last failing checks. If we can do this before tomorrow, we can release it as part of 1.13.0, or we'll merge this after this release

michal-mrazek:
@tatiana I am on it right now, so we can merge it today :)

michal-mrazek:
I think the tests should pass now, if you can pls trigger it 🙏

PR #2183, opened 2025-12-04
Refactor Watcher `InvocationMode.SUBPROCESS` log parser

tatiana:
While the PR #2152 introduced the capability of users running the `ExecutionMode.WATCHER` with dbt installed in a separate virtualenv, and still get updates in "real time", it came with two downsides:

1. Spread the watcher logic into multiple areas of the Cosmos code base (inside the `utils` folder, and also inside the `hooks/subprocess` - which should be focused only on running suprocess);
2. It changed the subprocess hook to always invoke the method `_store_dbt_resource_status_from_log(line, **kwargs)` for each log line made available in stdout, even when users are not using `ExecutionMode.WATCHER` .

This PR refactors the implementation to keep watcher implementation details inside the `operators/` folder. It also allows operators that subclass the `AbstractDbtLocalBase` to define the attribute `_process_log_line_callable`. Only classes that define this function will parse the log lines with the desired logic. At the moment, only the `DbtProducerWatcherOperator` implements the logic of uploading the status to XCom.


cosmos/operators/local.py
```diff
@@ -181,6 +181,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
         "compiled_sql": "sql",
         "freshness": "json",
     }
+    _process_log_line_callable: Callable[[str, Any], None] | None = None
```
pankajastro, line 184: Why do we need this as a class level attribute?

pankajastro (approved):
Looks good. I have posted a minor [comment](https://github.com/astronomer/astronomer-cosmos/pull/2183/files#r2592190835).

cosmos/operators/local.py
tatiana: It was just a strategy to decide when to define so that we could override it in subclasses. In the local execution mode, it is the node, for instance. What alternatives do you think could be better?

tatiana:
It seems the only coverage concern is a line that was already not covered before this PR, and was highlighted due to the file rename:
https://app.codecov.io/gh/astronomer/astronomer-cosmos/pull/2183/blob/cosmos/operators/_watcher/state.py?dropdown=coverage

Therefore, I'm going ahead and merging the change.

PR #2185, opened 2025-12-05
Update pre-commit default language version

pankajastro:
I'm using Python 3.11 virtual environment  and getting below error
```
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/Users/pankaj/Documents/astro_code/astronomer-cosmos/devenv/bin/python3.11', '-mvirtualenv', '/Users/pankaj/.cache/pre-commit/repou5hulpgs/py_env-python3.10', '-p', 'python3.10')
return code: 1
stdout:
    RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.10'
stderr: (none)
Check the log at /Users/pankaj/.cache/pre-commit/pre-commit.log
```

Changed default_language_version.python from python3.10 to python3 to allow pre-commit to use any available Python 3 interpreter

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 default_language_version:
-  python: python3.10
+  python: python3
 
 repos:
   - repo: local


.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 default_language_version:
-  python: python3.10
+  python: python3
```
tatiana, line 4: This fails if the person has Python 3.9 installed their local environment. We can pin to another version

pankajastro: Python 3.9 has reached EOL, and we decided to drop support for Python 3.9

pankajastro: I have a Python 3.11 virtual environment, and I'm hitting an issue. I try to follow the Airflow setting https://github.com/apache/airflow/blob/main/.pre-commit-config.yaml, but please let me know if you want me to test for some other version

tatiana: @pankajastro I fully understand, but developers may still have Python 3.9 installed in their local environment, and - this happened to me - unless we tell pre-commit, it may get the incorrect version of Python, which will conflict with pre-commit hooks that do not support that version.
It would be great if we set the minimum acceptable version, or at least the same version that the CI uses. I'm happy for that to be Python 3.11.
To set just Python 3 is not sufficient, I'm afraid

pankajastro: > developers may still have Python 3.9 installed in their local environment

I think they should move to Python 3.10 or newer, given pre-commit needs Python 3.10 or newer https://github.com/pre-commit/pre-commit/blob/main/setup.cfg#L26

> It would be great if we set the minimum acceptable version

I checked the [docs](https://pre-commit.com/) but could not find a way to set the Python min version. Maybe if we want, we can have a local pre-commit

> at least the same version that the CI uses

Which CI job are you referring to? I think we run with multiple Python versions, except fews jobs 

>  I'm happy for that to be Python 3.11.

If we set the default to Python 3.11, then it may cause issues for the contributor having Python version 3.10, or 3.12

Please do let me know how we should handle this.

tatiana:
Closing this since the issue was solved in #2190

pankajastro:
> Closing this since the issue was solved in #2190

I'm not sure how #2190 resolves this—could you clarify?

pankajkoti:
@pankajastro @tatiana -- I am not sure if this PR could solve the issue, or if we need another fix. But my local commits with branches cut from the latest `main` are failing and I have Python 3.12 installed on my local. For now, I am pushing the commits with `--no-verify` and relying on the CI pre-commit bot to take care of pre-commit fixes for me in the PRs opened.

pankajastro:
> @pankajastro @tatiana -- I am not sure if this PR could solve the issue, or if we need another fix. But my local commits with branches cut from the latest `main` are failing and I have Python 3.12 installed on my local. For now, I am pushing the commits with `--no-verify` and relying on the CI pre-commit bot to take care of pre-commit fixes for me in the PRs opened.

@pankajkoti I tested this PR locally using a Python 3.11 virtual environment, and it worked. I'm also using `--no-verify` for now

tatiana:
@pankajastro @pankajkoti https://github.com/astronomer/astronomer-cosmos/pull/2196

PR #2188, opened 2025-12-08
Use lazy import for `generate_mapping_docs` by moving it in `setup` function 

pankajastro:
closes: https://github.com/astronomer/oss-integrations-private/issues/286


pankajastro:
The CI is green now, so we do not need this

PR #2192, opened 2025-12-10
Update telemetry to v3 format with query parameters

pankajkoti:
- Migrate from v2 path-based URL to v3 query parameter format
- Change endpoint to event-collection/{event_type} with query params
- Add comments documenting v1 and v2 URL formats for record purposes
- Update tests to match the new v3 URL format

closes: https://github.com/astronomer/oss-integrations-private/issues/277

### DIFF
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -180,8 +180,13 @@ DBT_TEARDOWN_ASYNC_TASK_ID = "dbt_teardown_async"
 
 PRODUCER_WATCHER_TASK_ID = "dbt_producer_watcher"
 
-TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/{cosmos_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}/{cosmos_task_count}/{execution_modes}"
-TELEMETRY_VERSION = "v2"
+# Telemetry URL for v1 (for record purposes)
+# URL format: astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/{cosmos_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}/{cosmos_task_count}
+# Telemetry URL for v2 (commented for record purposes)
+# TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/{cosmos_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}/{cosmos_task_count}/{execution_modes}"
+# Telemetry URL for v3 - uses query params instead of path params
+TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/event-collection/{event_type}?{query_string}"
+TELEMETRY_VERSION = "v3"
 TELEMETRY_TIMEOUT = 1.0
 
 _AIRFLOW3_MAJOR_VERSION = 3
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -180,12 +180,12 @@ DBT_TEARDOWN_ASYNC_TASK_ID = "dbt_teardown_async"
 
 PRODUCER_WATCHER_TASK_ID = "dbt_producer_watcher"
 
-# Telemetry URL for v1 (for record purposes)
+# Telemetry URL for v1 (for record purposes) — shipped with Cosmos 1.8.0 through 1.10.x.
 # URL format: astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/{cosmos_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}/{cosmos_task_count}
-# Telemetry URL for v2 (commented for record purposes)
+# Telemetry URL for v2 (commented for record purposes) — Cosmos 1.11.0 added execution mode tracking via the extra path segment.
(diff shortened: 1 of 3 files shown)


cosmos/telemetry.py
```diff
@@ -76,7 +77,6 @@ def emit_usage_metrics_if_enabled(event_type: str, additional_metrics: dict[str,
     if should_emit():
         metrics = collect_standard_usage_metrics()
         metrics["event_type"] = event_type
-        metrics["variables"].update(additional_metrics)  # type: ignore[attr-defined]
```
pankajkoti, line 79: we no longer need this. `additional_metrics` are added to the `metrics` dictionary below, which get `urlencode`d as `query_string` and appended to the URL in the `emit_usage_metrics` method

tests/test_telemetry.py
```diff
@@ -142,5 +142,5 @@ def test_emit_usage_metrics_if_enabled_succeeds(
         "k1": "v1",
         "k2": "v2",
         "event_type": "any",
-        "variables": {"k2": "v2"},
+        "variables": {},
```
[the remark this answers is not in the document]
pankajkoti: Yes, this is not a problem.

```
>>> from urllib.parse import urlencode
>>> x = {"a": 1, "b": 2, "c": {}}
>>> urlencode(x)
'a=1&b=2&c=%7B%7D'
```

cosmos/telemetry.py
```diff
@@ -42,9 +42,10 @@ def emit_usage_metrics(metrics: dict[str, object]) -> bool:
 
     The metrics must contain the necessary fields to build the TELEMETRY_URL.
     """
+    event_type = metrics.pop("event_type")
```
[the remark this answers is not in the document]
pankajkoti: Thanks, I have added event_type in the debug log.

tests/test_telemetry.py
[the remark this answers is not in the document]
pankajkoti: We're free to pass custom query params with the v3 telemetry URL format and variables is an example query param.

pankajkoti:
@tatiana @pankajastro The Integration tests are failing because the route `https://astronomer.gateway.scarf.sh/astronomer-cosmos/<telemetry_version>/event-collection/<event_type>` is not yet created in Scarf. Once we're happy with this URL route, I can go ahead and create one and re-run the failing tests in the CI.

cosmos/constants.py
```diff
@@ -180,8 +180,13 @@ def _missing_value_(cls, value):  # type: ignore
...
 PRODUCER_WATCHER_TASK_ID = "dbt_producer_watcher"
 
-TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/{cosmos_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}/{cosmos_task_count}/{execution_modes}"
-TELEMETRY_VERSION = "v2"
+# Telemetry URL for v1 (for record purposes)
+# URL format: astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/{cosmos_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}/{cosmos_task_count}
+# Telemetry URL for v2 (commented for record purposes)
+# TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/{cosmos_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}/{cosmos_task_count}/{execution_modes}"
+# Telemetry URL for v3 - uses query params instead of path params
+TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/event-collection/{event_type}?{query_string}"
```
pankajastro: just want to understand — what is the motivation behind adding "event-collection" in the path?

pankajastro (approved):
Looks good — I’m happy for it to be merged once CI is green.

cosmos/constants.py
pankajkoti: just so that we give some name to this new endpoint :) Happy to discuss on this further if you have more thoughts there -- like if we should not have that or have a different name that sounds more suitable

cosmos/constants.py
```diff
@@ -180,8 +180,13 @@ def _missing_value_(cls, value):  # type: ignore
 
 PRODUCER_WATCHER_TASK_ID = "dbt_producer_watcher"
 
-TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/{cosmos_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}/{cosmos_task_count}/{execution_modes}"
-TELEMETRY_VERSION = "v2"
+# Telemetry URL for v1 (for record purposes)
+# URL format: astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}/{cosmos_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}/{cosmos_task_count}
```
tatiana, line 184: @pankajkoti, Please, could we add a comment explaining which versions of Cosmos use this pattern - and the two other ones

cosmos/constants.py
tatiana: @pankajkoti, maybe we could have just in the path:
```
"https://astronomer.gateway.scarf.sh/astronomer-cosmos/{telemetry_version}
```
And the event type as a query_string, WDYT?

I feel that `event-collection` is unnecessary.

tatiana (approved):
Hi @pankajkoti thanks a lot for working on this. I added some minor comments inline, happy for the PR to be merged after they are addressed.

cosmos/constants.py
pankajkoti: We agreed upon removing event-collection and keeping event-type so that we denote that it is a required path param instead of an optional query param

cosmos/constants.py
pankajkoti: added the Cosmos versions against those telemetry versions

Issue #2110, opened 2025-11-11
Track operators usage via Scarf


tatiana:
Leverage Airflow listeners to capture the utilisation of operators even if they are outside of `DbtDag` and `DbtTaskGroup`. The main information we want to know are:
- which is the execution mode
- which is the command (run, build, test, etc)
- invocation mode
- if dbt deps is being run
- if callback is being used

pankajkoti:
Addressed as part of PR #2195

PR #2195, opened 2025-12-12
Add telemetry task listener metrics for Cosmos operators

pankajkoti:
The PR introduces the Airflow task-instance listener and emits task-level telemetry for Cosmos operators (operator name, invocation mode, execution mode, dbt command, install_deps flag, has_callback, subclass detection)

related: https://github.com/astronomer/astronomer-cosmos/issues/2110
related: #2192

### DIFF
--- a/cosmos/listeners/task_instance_listener.py
+++ b/cosmos/listeners/task_instance_listener.py
@@ -151,7 +151,7 @@ def _build_task_metrics(task_instance: TaskInstance, status: str) -> dict[str, o
 
 @hookimpl
 def on_task_instance_success(
-    previous_state: Any, task_instance: TaskInstance, session: "Session"
+    previous_state: Any, task_instance: TaskInstance, session: Session
 ) -> None:  # type: ignore[override]
     if not _is_cosmos_task(task_instance):
         return
@@ -163,7 +163,7 @@ def on_task_instance_success(
 
 @hookimpl
 def on_task_instance_failed(
-    previous_state: Any, task_instance: TaskInstance, session: "Session"
+    previous_state: Any, task_instance: TaskInstance, session: Session
 ) -> None:  # type: ignore[override]
     if not _is_cosmos_task(task_instance):
         return
--- a/cosmos/listeners/task_instance_listener.py
+++ b/cosmos/listeners/task_instance_listener.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from airflow.listeners import hookimpl
 
 if TYPE_CHECKING:
     from airflow.models.taskinstance import TaskInstance
+    from sqlalchemy.orm import Session
 
 from cosmos import telemetry
 from cosmos.constants import InvocationMode
@@ -149,7 +150,9 @@ def _build_task_metrics(task_instance: TaskInstance, status: str) -> dict[str, o
 
 
 @hookimpl
-def on_task_instance_success(previous_state, task_instance, session):  # type: ignore[override]
+def on_task_instance_success(
+    previous_state: Any, task_instance: TaskInstance, session: "Session"
+) -> None:  # type: ignore[override]
     if not _is_cosmos_task(task_instance):
         return
 
@@ -159,7 +162,9 @@ def on_task_instance_success(previous_state, task_instance, session):  # type: i
 
 
 @hookimpl
-def on_task_instance_failed(previous_state, task_instance, session):  # type: ignore[override]
+def on_task_instance_failed(
+    previous_state: Any, task_instance: TaskInstance, session: "Session"
+) -> None:  # type: ignore[override]
     if not _is_cosmos_task(task_instance):
         return
 
--- a/cosmos/listeners/task_instance_listener.py
+++ b/cosmos/listeners/task_instance_listener.py
@@ -60,6 +60,23 @@ def _invocation_mode(task_instance: TaskInstance) -> str | None:
     return str(mode)
 
 
+def _dbt_command(task_instance: TaskInstance) -> str | None:
+    """Return the dbt sub-command encoded on Cosmos operators."""
+
+    task = task_instance.task
+    if not isinstance(task, AbstractDbtBase):
+        return None
+
+    command = getattr(task, "base_cmd", None)
+    if command is None:
+        return None
+
+    if isinstance(command, (list, tuple)):
+        return " ".join(str(part) for part in command if part is not None)
+
+    return str(command)
+
+
 def _build_task_metrics(task_instance: TaskInstance, status: str) -> dict[str, object]:
     """Build telemetry payload for task completion events."""
 
@@ -75,6 +92,10 @@ def _build_task_metrics(task_instance: TaskInstance, status: str) -> dict[str, o
         "priority_weight": task_instance.priority_weight,
         "map_index": task_instance.map_index,
     }
+
+    dbt_command = _dbt_command(task_instance)
+    if dbt_command:
+        metrics["dbt_command"] = dbt_command
 
     dag_run = getattr(task_instance, "dag_run", None)
     if dag_run is not None:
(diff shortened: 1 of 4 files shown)


cosmos/listeners/task_instance_listener.py
```diff
@@ -0,0 +1,173 @@
...
+    duration = getattr(task_instance, "duration", None)
+    if duration is not None:
+        metrics["duration"] = duration
+
+    return metrics
+
+
+@hookimpl
+def on_task_instance_success(
+    previous_state: Any, task_instance: TaskInstance, session: Session
```
pankajastro: Have you tested this on Airflow 3? It looks like the function signature has changed in Airflow 3. https://github.com/apache/airflow/blob/7b30cc0fb667e0a78147c2d01bd6478591fddc3a/airflow-core/src/airflow/listeners/spec/taskinstance.py#L38

cosmos/listeners/task_instance_listener.py
```diff
@@ -0,0 +1,173 @@
...
+    install_deps = _install_deps(task_instance)
+    if install_deps is not None:
+        metrics["install_deps"] = install_deps
+
+    metrics["has_callback"] = _has_callback(task_instance)
+
+    dag_run = getattr(task_instance, "dag_run", None)
+    if dag_run is not None:
+        metrics["dag_run_id"] = dag_run.run_id
+        dag_hash = getattr(dag_run, "dag_hash", None)
```
pankajastro: In the dagrun listener we hash using md5 for airflow3: https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/listeners/dag_run_listener.py#L75

pankajastro (commented):
looks good

cosmos/listeners/task_instance_listener.py
pankajkoti: yes, I guess not needed for task instance . removed dag_hash emission

cosmos/listeners/task_instance_listener.py
pankajkoti: Yes, good catch, I had previously tested for Airflow 2 using Astro CLI and relied on tests in CI for Airflow 3, which did not complain. But I have aligned the signature for both Airflow 2 and Airflow 3 now, and also tested on both using Astro CLI; it is working fine.

tests/listeners/test_task_instance_listener.py
```diff
@@ -0,0 +1,250 @@
...
+        self._task_module = module
+        if install_deps is not None:
+            self.install_deps = install_deps
+        if callback is not None:
+            self.callback = callback
+        if runner_callbacks is not None:
+            self._dbt_runner_callbacks = runner_callbacks
+
+    def build_and_run_cmd(
+        self, context, cmd_flags, run_as_async=False, async_context=None, **kwargs
```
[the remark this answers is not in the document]
pankajkoti: probably for another PR as it's a broader change for existing arguments

cosmos/listeners/task_instance_listener.py
```diff
@@ -0,0 +1,165 @@
...
+        return parts[2]
+    # TODO: When users subclass Cosmos operators in external modules, encode execution mode directly on the task
+    # so telemetry does not rely on module inspection.
+    return None
+
+
+def _operator_module(task_instance: TaskInstance) -> str:
+    """Return the module path for the operator backing the given task instance."""
+
+    return getattr(task_instance.task, "_task_module", None) or task_instance.task.__class__.__module__
```
tatiana, line 43: Which version of Airflow is related to each of these ways if retrieving the task module? It may be worth to document it
```suggestion
    return getattr(task_instance.task, "_task_module",  task_instance.task.__class__.__module__)
```

cosmos/listeners/task_instance_listener.py
```diff
@@ -0,0 +1,165 @@
...
+def _execution_mode_from_task(task_instance: TaskInstance) -> str | None:
+    """Extract Cosmos execution mode from the task's module path."""
+
+    module = _operator_module(task_instance)
+    parts = module.split(".")
+    if len(parts) >= 3 and parts[0] == "cosmos" and parts[1] == "operators":
+        return parts[2]
+    # TODO: When users subclass Cosmos operators in external modules, encode execution mode directly on the task
+    # so telemetry does not rely on module inspection.
+    return None
```
tatiana, line 37: Do we have a follow up ticket to track subclasses execution mode?

cosmos/listeners/task_instance_listener.py
```diff
@@ -0,0 +1,165 @@
...
+
+def _invocation_mode(task_instance: TaskInstance) -> str | None:
+    """Return the invocation mode recorded in Cosmos operators."""
+
+    mode = getattr(task_instance.task, "invocation_mode", None)
+    if mode is None:
+        return None
+    if isinstance(mode, InvocationMode):
+        return mode.value
+    return str(mode)
```
tatiana, line 60: In which scenarios would we land in this line? Are there any circumstances when the Invocation mode is neither None or InvocationMode?

cosmos/listeners/task_instance_listener.py
```diff
@@ -0,0 +1,165 @@
...
+
+    metrics: dict[str, object] = {
+        "dag_id": task_instance.dag_id,
+        "task_id": task_instance.task_id,
+        "status": status,
+        "operator_name": task_instance.task.__class__.__name__,
+        "is_cosmos_operator_subclass": _is_cosmos_subclass(task_instance),
+        "invocation_mode": _invocation_mode(task_instance),
+        "execution_mode": _execution_mode_from_task(task_instance),
+        "map_index": task_instance.map_index,
```
tatiana, line 122: What is the `map_index`and why do we have it here?

tatiana (approved):
Hi @pankajkoti , thanks for implementing this! I left minor comments in line.

The significant risk with this change is that our telemetry logic fails, causing the task to fail. We may need to add some exception handling there if issues surface, but it is a good starting point.

It would be great to have follow up tickets for the work we did not include here yet, and assign to the next milestone

cosmos/listeners/task_instance_listener.py
pankajkoti: I guess we need an `or` there. I copied it from the dag_run_listener here: https://github.com/astronomer/astronomer-cosmos/blob/9ba73954c0ab8c81c3bbdbef6d3e3ecfe43b9d30/cosmos/listeners/dag_run_listener.py#L42

cosmos/listeners/task_instance_listener.py
pankajkoti: map_index is the one that gets attached to task instances when using dynaming task mapping. I thought it would be nice to capture that since it's already available on the task instance.

cosmos/listeners/task_instance_listener.py
pankajkoti: only keeping it as a defensive safeguard when custom subclasses may set invocation_mode to a string instead of the enum

cosmos/listeners/task_instance_listener.py
pankajkoti: Logged this ticket https://github.com/astronomer/astronomer-cosmos/issues/2209

pankajkoti:
I am proceeding with the merge, having addressed the questions inline. Please let me know if you'd like me to address something further @tatiana 

cc: @pankajastro

cosmos/listeners/task_instance_listener.py
tatiana: @pankajkoti I do not believe we should add what the `map_index is`, but I think we should add if it were a mapped task (boolean). And we need to make this change consistent in https://github.com/astronomer/ap-vendor/pull/1102/changes#r2631088199

pankajkoti: Created a follow-up PR to address this suggestion: https://github.com/astronomer/astronomer-cosmos/pull/2210

PR #2196, opened 2025-12-12
Adjust `pre-commit` so Python 3.10 or higher can be used

tatiana:
I incorrectly closed #2185, thinking that the `- --py310-plus` syntax used in #2190 was `pre-commit`, not pyupgrade specific.

This PR aims to allow anyone using `Python versions => 3.10` to be able to run `pre-commit`, while failing if a previous version is used.


.pre-commit-config.yaml
```diff
@@ -4,6 +4,29 @@ default_language_version:
   python: python3.10
```
tatiana:
```suggestion
  python: python3
```

PR #2198, opened 2025-12-12
Collect cosmos profile metrics

### LINKED ISSUE #2108: Track profile config with Scarf
We want to measure
- Adoption of profile YAML
- Adoption  per profile mapping class
- Track which database is being used

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/2108

This PR adds telemetry collection for Cosmos profile configuration metrics to track how users configure dbt profiles (via mapping or yaml file), which profile mapping classes are used, and which database types are deployed.

**Key changes:**
- Adds a new `get_profile_metrics` function to extract profile strategy, mapping class, and database type from task instances
- Integrates profile metrics into the task telemetry payload
- Adds early return optimization in `_get_profile_path` when profiles_yml_filepath is provided

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -353,6 +353,9 @@ class ProfileConfig:
         Check if profile object version is exist then reuse it
         Otherwise, create profile yml for requested object and return the profile path
         """
+        if self.profiles_yml_filepath:
+            return Path(self.profiles_yml_filepath)
+
         assert self.profile_mapping  # To satisfy MyPy
         current_profile_version = self.profile_mapping.version(self.profile_name, self.target_name, use_mock_values)
         cached_profile_path = get_cached_profile(current_profile_version)
--- a/cosmos/listeners/task_instance_listener.py
+++ b/cosmos/listeners/task_instance_listener.py
@@ -7,7 +7,6 @@ from airflow.listeners import hookimpl
 if TYPE_CHECKING:
     from airflow.models.taskinstance import TaskInstance
 
-    from cosmos.config import ProfileConfig
 
 from cosmos import telemetry
 from cosmos.constants import InvocationMode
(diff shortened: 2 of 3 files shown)


cosmos/config.py
```diff
@@ -353,6 +353,9 @@ def _get_profile_path(self, use_mock_values: bool = False) -> Path:
         Check if profile object version is exist then reuse it
         Otherwise, create profile yml for requested object and return the profile path
         """
+        if self.profiles_yml_filepath:
+            return Path(self.profiles_yml_filepath)
```
pankajastro, line 357: Look like this was bug when `_get_profile_path` is called profile object contain yaml path instead of profile mapping

cosmos/listeners/task_instance_listener.py
```diff
@@ -119,6 +153,10 @@ def _build_task_metrics(task_instance: TaskInstance, status: str) -> dict[str, o
         "is_cosmos_operator_subclass": _is_cosmos_subclass(task_instance),
         "invocation_mode": _invocation_mode(task_instance),
         "execution_mode": _execution_mode_from_task(task_instance),
+        "map_index": task_instance.map_index,
```
pankajastro:
```suggestion
```

tatiana (approved):
Looks good, @pankajastro !

Since we'll be extending this to also emit profile when parsing the DAG, we can cover the missing line by the time we do this follow up PR:

Image

Do we have a ticket to do this follow-up work?

pankajkoti (approved):

pankajastro:
> Looks good, @pankajastro !
> 
> Since we'll be extending this to also emit profile when parsing the DAG, we can cover the missing line by the time we do this follow up PR:
> 
> Image
> Do we have a ticket to do this follow-up work?

Created PR: https://github.com/astronomer/astronomer-cosmos/pull/2215 @tatiana

PR #2207, opened 2025-12-16
Introduce `ExecutionMode.WATCHER_KUBERNETES` to use watcher with `KubernetesPodOperator`

### LINKED ISSUE #2042: Support `ExecutionMode.WATCHER` via virtualenv and/or kubernetespod
In Cosmos 1.11.0, the `DbtProducerWatcherOperator` tasks were implemented based on `ExecutionMode.LOCAL` operators:
https://github.com/astronomer/astronomer-cosmos/blob/be6de00870d05499eeb647a21e2a50a5eaf89626/cosmos/operators/watcher.py#L52

This can bring challenges in organisations where the same Airflow deployment uses different versions of Airflow, and users would like to allow each DAG to decide the desired set of dependencies - isolated from the Airflow deployment. Something similar to what can be done in `ExecutionMode.VIRTUALENV` or `ExecutionMode.KUBERNETES`.

From an implementation perspective, even though we may not be able to use [dbt Core callbacks](https://docs.getdbt.com/reference/programmatic-invocations#registering-callbacks) to accomplish this level of isolation, we could potentially leverage [dbt Core structured logging](https://docs.getdbt.com/reference/events-logging#structured-logging).

This has been a requested feature, and the need for something similar was raised in three ocasions so far:
- Airflow Summit: two engineers asked about this possibility
- During an Astronomer customer call
- Airflow [#airflow-dbt slack](https://apache-airflow.slack.com/archives/C059CC42E9W/p1760964809323639)

Image

tatiana:
Introduce `ExecutionMode.WATCHER_KUBERNETES`, allowing users to use the Producer/Consumer approach without requiring dbt to be installed in the same environment as Airflow.

This PR includes an example DAG illustrating the use-case.

Given an existing bug in the latest releases of `apache-airflow-providers-cncf-kubernetes` (https://github.com/apache/airflow/issues/59366), we had to subclass the `KubernetesPodManager` class and override a method. This may not be needed once the fix is merged and the new provider version is released. Given this issue, we have tested and successfully validated the usage of this execution mode with the following versions of `apache-airflow-providers-cncf-kubernetes` 
- 10.8.0
- 10.9.0
- 10.10.0
- 10.11.0
- 10.12.2 (tested on 30 January 2026, after this PR was merged)

**Why**

There is significant overhead each time a K8S Pod is spawned and when the dbt project is parsed. 

Cosmos `ExecutionMode.WATCHER` ([docs](https://astronomer.github.io/astronomer-cosmos/getting_started/watcher-execution-mode.html)) has delivered significant performance improvements for users who were previously using `ExecutionMode.LOCAL`; however, some users need to keep Airflow and dbt dependencies isolated.

This PR extends the `ExecutionMode.WATCHER` performance gains to Kubernetes.

Comparison of the same dbt project DAG run with Jaffle Shop (reproducible by modifying the DAG `jaffle_shop_watcher_kubernetes` included in this PR):
- `ExecutionMode.KUBERNETES`: 00:00:32.155
- `ExecutionMode.WATCHER_KUBERNETES`: 00:00:11.783

**Limitations**

- Does not work with older versions of the  `apache-airflow-providers-cncf-kubernetes` provider (`<=10.7.0`)
- Only works with`deferrable=False`.

** Pending work**

- The documentation will come in a follow-up PR.

**How to use**

- Users previously using `ExecutionMode.KUBERNETES` can just replace the `execution_mode` to use `ExecutionMode.WATCHER_KUBERNETES`.

**Example DAG**

This PR includes an example DAG that runs with our integration tests, `jaffle_shop_watcher_kubernetes`:
Screenshot 2026-01-06 at 11 12 59

The producer task runs dbt build using the `KubernetesPodOperator`:
Screenshot 2026-01-06 at 11 12 15
This task will populate XCom with status updates for each dbt node that is completed.

By default, the consumer tasks will be native Airflow deferrable sensors: 
Screenshot 2026-01-06 at 11 12 25

And, in case of retries, the consumer tasks use Cosmos  `ExecutionMode.KUBERNETES` KPO operators:
Screenshot 2026-01-06 at 11 18 42

Closes: https://github.com/astronomer/astronomer-cosmos/issues/2042


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, k8s-watcher-mess]
```
```suggestion
    branches: [main]
```

cosmos/airflow/graph.py
```diff
@@ -145,7 +145,7 @@ def _override_profile_if_needed(task_kwargs: dict[str, Any], profile_kwargs_over
         task_kwargs["profile_config"] = modified_profile_config
 
 
-def create_test_task_metadata(
+def create_test_task_metadata(  # noqa:C901
```
tatiana: Improve complexity

cosmos/airflow/graph.py
```diff
@@ -662,14 +666,30 @@ def _add_producer_watcher_and_dependencies(
...
+        operator_class=class_name,
         arguments=producer_task_args,
     )
     producer_airflow_task = create_airflow_task(producer_task_metadata, dag, task_group=task_group)
+    tasks_map[PRODUCER_WATCHER_TASK_ID] = producer_airflow_task
+
+    return producer_airflow_task
 
+
+def _add_watcher_dependencies(
```
tatiana, line 685: Create separate PR breaking down:
- `_add_producer_watcher_and_dependencies`

Into
- `_add_watcher_dependencies`
- `_add_watcher_producer_task`

cosmos/operators/_watcher/__init__.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
-__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher"]
+__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger", "_parse_compressed_xcom"]
```
tatiana, line 3:
```suggestion
__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger"]
```

cosmos/operators/_watcher/triggerer.py
```diff
@@ -37,7 +37,7 @@ def __init__(
 
     def serialize(self) -> tuple[str, dict[str, Any]]:
         return (
-            "cosmos._triggers.watcher.WatcherTrigger",
+            "cosmos.operators._watcher.triggerers.WatcherTrigger",
```
tatiana: Create a separate PR with this refactor

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,365 @@
...
+            # These lines raise an exception when using KubernetesPodOperatorCallback
+            context = extra_kwargs.get("context")
+            # assert context is not None  # Make MyPy happy
+            safe_xcom_push(task_instance=context["ti"], key=f"{unique_id.replace('.', '__')}_status", value=node_status)
+
+
+producer_task_context = None
+
+
+class WatcherKubernetesCallback(KubernetesPodOperatorCallback):  # type: ignore[misc]
```
tatiana, line 46: Create a `_watcher/base.py` with the logic that is common between the `watcher_kubernetes.py` and `watcher.py` modules

cosmos/converter.py
```diff
@@ -255,6 +255,8 @@ def __init__(
         **kwargs: Any,
     ) -> None:
 
+        logger.info("::group::Cosmos DAG/TaskGroup Parsing")
```
tatiana: Add a separate PR for this improvement, that aggregates DAG parsing logs

dev/dags/jaffle_shop_watcher_kubernetes.py
```diff
@@ -0,0 +1,98 @@
...
+    operator_args["trigger_rule"] = "all_success"
+
+dag = DbtDag(
+    dag_id="jaffle_shop_watcher_kubernetes",
+    start_date=datetime(2022, 11, 27),
+    doc_md=__doc__,
+    catchup=False,
+    # Cosmos-specific parameters:
+    project_config=ProjectConfig(),
+    profile_config=ProfileConfig(
```
tatiana: Change this example to use a `manifest.json` and avoid using `PostgresUserPasswordProfileMapping`

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,365 @@
...
+
+    @staticmethod
+    def progress_callback(*, line: str, client: client_type, mode: str, **kwargs: Any) -> None:
```
johnhoran: I've been playing about with how this could work if it were called from the triggerer instead of from the operator.  Currently the triggerer doesn't call progress_callback, so I'm going to try to get something upstreamed, but what I'll suggest would be to modify the signature of this method so the pod is passed to it.  
```
    def progress_callback(
        *,
        line: str,
        client: client_type,
        mode: str,
        container_name: str,
        timestamp: DateTime | None,
        pod: k8s.V1Pod,
        **kwargs,
    ) -> None:
```
and then the run_id, task_id, dag_id for setting xcoms can be extracted from the labels set on the pod. Obviously the pod manager needs to be modified to pass the pod too.

cosmos/airflow/graph.py
```diff
@@ -662,14 +666,30 @@ def _add_producer_watcher_and_dependencies(
                 "resource_type:unit_test",
             ]
 
+    if execution_mode == ExecutionMode.WATCHER:
+        class_name = "cosmos.operators.watcher.DbtProducerWatcherOperator"
+    else:
+        class_name = "cosmos.operators.watcher_kubernetes.DbtProducerKubernetesWatcherOperator"
```
johnhoran: I'd suggest renaming the class DBTProducerWatcherKubernetesOperator and then using 
```suggestion
    class_name = calculate_operator_class(execution_mode, "DbtProducerWatcher")
```

Primary reason is that overriding `calculate_operator_class` has been a very useful way of injecting some custom code we use with astronomer.

tatiana: Ah, thanks for letting us know, @johnhoran !

I just addressed your feedback in 2a68e694a66e6ec9b4fa9e0df7a28ccda9423876

cosmos/operators/watcher_kubernetes.py
tatiana: @johnhoran, please, do let us know how it goes!

My ultimate desire is not have this code at all over here.

I'll see if things continue working as expected if we override using the same changes you suggested to the upstream provider: https://github.com/apache/airflow/pull/59372

One limitation of the current implementation is that it only supports non-deferrable KPOs, and it would be great to support deferrable KPOs as well.

cosmos/converter.py
tatiana: Done: https://github.com/astronomer/astronomer-cosmos/pull/2235

cosmos/operators/_watcher/triggerer.py
tatiana: Here: https://github.com/astronomer/astronomer-cosmos/pull/2236

cosmos/operators/_watcher/__init__.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
-__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher"]
+__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger"]
```
tatiana:
```suggestion
__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger", "_parse_compressed_xcom"]
```

tatiana: Rollback this change

cosmos/airflow/graph.py
tatiana: Done in: https://github.com/astronomer/astronomer-cosmos/pull/2237

tests/operators/test_watcher_kubernetes.py
```diff
@@ -0,0 +1,142 @@
...
+if os.getenv("CI"):
+    operator_args["trigger_rule"] = "all_success"
+
+
+@pytest.mark.skipif(
+    AIRFLOW_VERSION < Version("3.1"),
+    reason="We are only testing watcher Kubernetes with Airflow 3.1 and more recent versions of the K8s provider",
+)
+@pytest.mark.integration
+def test_dbt_task_group_with_watcher_kubernetes():
```
tatiana: Update name of the test

cosmos/airflow/graph.py
```diff
@@ -203,11 +203,15 @@ def create_test_task_metadata(
...
 
     if (
-        execution_mode == ExecutionMode.WATCHER
-        and render_config is not None
+        render_config is not None
         and render_config.test_behavior == TestBehavior.AFTER_ALL
+        and execution_mode in (ExecutionMode.WATCHER, ExecutionMode.WATCHER_KUBERNETES)
     ):
-        operator_class = "cosmos.operators.local.DbtTestLocalOperator"
+        if execution_mode == ExecutionMode.WATCHER:
```
tatiana: Double check this logic

cosmos/airflow/graph.py
```diff
@@ -682,43 +689,42 @@ def _add_watcher_dependencies(
...
-                        task for task in node_tasks if not set(task.upstream_task_ids).intersection(taskgroup.children)
-                    ]
-                else:
-                    always_run_tasks = [task_or_taskgroup]
-
-                for task in always_run_tasks:
-                    task.trigger_rule = task_args.get("trigger_rule", "always")  # type: ignore[attr-defined]
-
-    return producer_airflow_task.task_id
+        if node_id != PRODUCER_WATCHER_TASK_ID:
```
```suggestion
        if node_id == PRODUCER_WATCHER_TASK_ID:
            continue
```

And de-dent the rest, to avoid huge diff

cosmos/airflow/graph.py
```diff
@@ -940,12 +947,12 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
...
     create_airflow_task_dependencies(nodes, tasks_map)
 
-    if execution_mode == ExecutionMode.WATCHER:
+    if execution_mode in (ExecutionMode.WATCHER, ExecutionMode.WATCHER_KUBERNETES):
         setup_operator_args = getattr(execution_config, "setup_operator_args", None) or {}
         _add_watcher_dependencies(
             dag=dag,
             producer_airflow_task=producer_task,
-            task_args=task_args,
+            task_args={**task_args},
```
```suggestion
            task_args=task_args,
```

cosmos/operators/watcher.py
```diff
@@ -67,19 +68,20 @@ def _store_dbt_resource_status_from_log(line: str, extra_kwargs: Any) -> None:
     try:
         log_line = json.loads(line)
     except json.JSONDecodeError:
-        logger.debug("Failed to parse log: %s", line)
+        logger.info("Failed to parse log: %s", line)
         log_line = {}
     node_info = log_line.get("data", {}).get("node_info", {})
     node_status = node_info.get("node_status")
     unique_id = node_info.get("unique_id")
+    if unique_id is not None:
```
tatiana: Revert to what was before

cosmos/operators/watcher.py
```diff
@@ -403,12 +405,13 @@ def execute(self, context: Context, **kwargs: Any) -> None:
                 method_name=self.execute_complete.__name__,
             )
 
-    def execute_complete(self, context: Context, event: dict[str, str]) -> None:
+    def execute_complete(self, context: Context, event: dict[str, Any], **kwargs: Any) -> None:
         status = event.get("status")
+        reason = event.get("reason")
```
tatiana: Check if this change is really needed

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,365 @@
...
+        line_content = line.strip()
+        logger.info(f"[xubi] {line_content}")
+        try:
+            k8s_timestamp, dbt_log = line_content.split(" ", 1)
+        except ValueError:
+            logger.debug(f"[XUBIRU] Failed to parse log: {line_content}")
```
tatiana: Change this log line

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,365 @@
...
+
+        # TODO: Handle and store all possible node statuses, not just the current success and failed
+        if node_status in ["success", "failed"]:
+            # These lines raise an exception when using KubernetesPodOperatorCallback
+            context = extra_kwargs.get("context")
+            # assert context is not None  # Make MyPy happy
+            safe_xcom_push(task_instance=context["ti"], key=f"{unique_id.replace('.', '__')}_status", value=node_status)
+
+
+producer_task_context = None
```
tatiana, line 43: Add comment explaining why this is necessary

cosmos/constants.py
```diff
@@ -192,3 +193,5 @@ def _missing_value_(cls, value):  # type: ignore
 TELEMETRY_TIMEOUT = 1.0
 
 _AIRFLOW3_MAJOR_VERSION = 3
+
+_K8s_WATCHER_MIN_K8S_PROVIDER_VERSION = Version("10.8.0")
```
tatiana, line 200: Consider adding a max version that was actually tested - and add a warning if it is newer

cosmos/converter.py
```diff
@@ -256,6 +256,8 @@ def __init__(
     ) -> None:
         logger.info("::group::Cosmos DAG parsing logs")
 
+        logger.info("::group::Cosmos DAG/TaskGroup Parsing")
```
tatiana:
```suggestion
```

dev/dags/jaffle_shop_kubernetes.py
```diff
@@ -97,17 +96,23 @@
             ),
         ),
         render_config=RenderConfig(dbt_project_path=AIRFLOW_PROJECT_DIR),
-        execution_config=ExecutionConfig(execution_mode=ExecutionMode.KUBERNETES, dbt_project_path=K8S_PROJECT_DIR),
+        execution_config=ExecutionConfig(
```
tatiana: Review, these changes probably don't need to happen

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,365 @@
...
+            "Try number #%s, poke attempt #%s: Pulling status from task_id '%s' for model '%s' and xcom '%s'",
+            try_number,
+            self.poke_retry_number,
+            self.producer_task_id,
+            self.model_unique_id,
+            f"{self.model_unique_id.replace('.', '__')}_status",
+        )
+
+        if try_number > 1:
+            return bool(self._fallback_to_local_run(try_number, context))
```
pankajkoti:
```suggestion
            return bool(self._fallback_to_k8s_run(try_number, context))
```
I guess it should be this here?

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,365 @@
...
+
+        if status is None:
+
+            if producer_task_state == "failed":
+                if self.poke_retry_number > 0:
+                    raise AirflowException(
+                        f"The dbt build command failed in producer task. Please check the log of task {self.producer_task_id} for details."
+                    )
+                else:
+                    return True  # TODO: remove me
```
pankajkoti: need to remove this return I believe

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,365 @@
...
+
+    # minor change from original DbtConsumerWatcherSensor to use K8s run
+    def _fallback_to_k8s_run(self, try_number: int, context: Context) -> bool:
+        """
+        Handles logic for retrying a failed dbt model execution.
+        Reconstructs the dbt command by cloning the project and re-running the model
+        with appropriate flags, while ensuring flags like `--select` or `--exclude` are excluded.
+        """
+        logger.info(
+            "Retry attempt #%s – Running model '%s' from project '%s' using ExecutionMode.LOCAL",
```
pankajkoti:
```suggestion
            "Retry attempt #%s – Running model '%s' from project '%s' using ExecutionMode.WATCHER_KUBERNETES",
```

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,365 @@
...
+
+producer_task_context = None
+
+
+class WatcherKubernetesCallback(KubernetesPodOperatorCallback):  # type: ignore[misc]
+
+    @staticmethod
+    def progress_callback(*, line: str, client: client_type, mode: str, **kwargs: Any) -> None:
+        line_content = line.strip()
+        logger.info(f"[xubi] {line_content}")
```
pankajkoti: what does `[xubi]` do here?

cosmos/operators/watcher_kubernetes.py
pankajkoti: same question what does `[XUBIRU]` do here?

cosmos/operators/watcher_kubernetes.py
pankajkoti: Yes, a comment to explain is great to have. Also, is it possible to have this as an operator attribute instead of a global variable?

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,365 @@
...
+
+        if reason == "producer_failed":
+            raise AirflowException(
+                f"Watcher producer task '{self.producer_task_id}' failed before reporting model results. Check its logs for the underlying error."
+            )
+
+    # def _use_event(self) -> bool:
+    #     if not self.invocation_mode:
+    #         self._discover_invocation_mode()
+    #     return self.invocation_mode == InvocationMode.DBT_RUNNER and EventMsg is not None
```
pankajkoti: Do we not need/use this here? Can we remove it if not needed?

pankajkoti (commented):
Looks good to me from the current progress. Have some comments in-line.

Great work on the persistence here on the complex task! 👏🏽

cosmos/operators/watcher_kubernetes.py
tatiana: Hi @johnhoran , I just changed the implementation to be equivalent to what we did in https://github.com/apache/airflow/pull/59372

The only two differences are:

1. Cosmos supports multiple versions of the K8s provider, including 10.8, 10.9  and 10.10 via these lines (this change does not make sense in your PR, since it will work only for the latest K8s provider)

```
            # Cosmos implementation difference when compared to proposal to fix the issue in the upstream provider:
            # https://github.com/apache/airflow/pull/59372/
            if Version(airflow_k8s_provider_version) >= Version("10.10.0"):
                from airflow.providers.cncf.kubernetes.utils.pod_manager import parse_log_line
            elif (
                Version(airflow_k8s_provider_version) >= _K8s_WATCHER_MIN_K8S_PROVIDER_VERSION
            ):  # Successfully tested with Airflow 3.1.0 and K8s provider 10.8.0 and 10.9.0
                parse_log_line = self.parse_log_line
            else:
                raise ValueError(f"Unsupported K8s provider version: {airflow_k8s_provider_version}")
            # Cosmos custom implementation finishes here.
```

2. I added additional type annotations required by our checks.

cosmos/operators/watcher_kubernetes.py
tatiana: It was just a debugging statement during development, it is removed!

cosmos/operators/watcher_kubernetes.py
tatiana: It was just a debugging statement during development, it is removed!

cosmos/operators/watcher_kubernetes.py
tatiana: This was refactored and no longer exists

cosmos/airflow/graph.py
tatiana: Done in 544643ac35da00a0f7747eb24487b4f5f5700e2d

cosmos/operators/watcher_kubernetes.py
tatiana: Done in https://github.com/astronomer/astronomer-cosmos/pull/2245

dev/dags/jaffle_shop_watcher_kubernetes.py
tatiana: Done

tests/operators/test_watcher_kubernetes.py
tatiana: Done in 13fd0d26d16e3e1b07b88ee38d8aefb92f2455a4

.github/workflows/test.yml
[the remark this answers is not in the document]
tatiana: This is a duplicate of https://github.com/astronomer/astronomer-cosmos/pull/2207/changes#r2631214012

cosmos/airflow/graph.py
tatiana: Addressed in 989e9094440ec06ad835ebeefddbcb15608b39ed

cosmos/airflow/graph.py
tatiana: Addressed in 62d87bcf1e5209b1401bcb48ab36ee85b21711d6

cosmos/operators/watcher_kubernetes.py
```diff
@@ -0,0 +1,184 @@
...
+        Invoke this callback to process pod container logs.
+
+        :param line: the read line of log.
+        :param client: the Kubernetes client that can be used in the callback.
+        :param mode: the current execution mode, it's one of (`sync`, `async`).
+        :param container_name: the name of the container from which the log line was read.
+        :param timestamp: the timestamp of the log line.
+        :param pod: the pod from which the log line was read.
+        """
+        if not "context" in kwargs:
```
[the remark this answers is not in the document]
tatiana: Addressed in ba7f0dc66f969b13e33b6a3b39a7aa92e971f96c

cosmos/operators/watcher_kubernetes.py
tatiana: Done

.github/workflows/test.yml
```diff
@@ -639,7 +639,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [ "3.12" ]
-        airflow-version: [ "2.10", "3.0" ]
+        airflow-version: [ "2.10", "3.1" ]
```
pankajastro: This could be done in a separate PR, but since the Kubernetes tests don’t take much time, I think we should run them for Airflow 3.0, 3.1, and possibly any other missing versions as well.

cosmos/airflow/_override.py
```diff
@@ -0,0 +1,190 @@
+import math
+import time
+from collections.abc import Callable
+from datetime import timedelta
+
+import pendulum
+from airflow.providers.cncf.kubernetes import __version__ as airflow_k8s_provider_version
```
pankajastro, line 7: since `_override.py` is located at the `cosmos.airflow` level, should we verify that users running other execution modes won’t encounter issues if the CNCF provider is not installed?

cosmos/operators/watcher_kubernetes.py
tatiana: This is no longer needed after the refactor https://github.com/astronomer/astronomer-cosmos/pull/2245

cosmos/airflow/_override.py
tatiana: Just confirmed it works fine, since we don't expose this in `cosmos/airflow/__init__.py`:
```
$ pip freeze | grep kubernetes
kubernetes==32.0.1
kubernetes_asyncio==32.3.2

$ python -c "import cosmos.airflow"
[2026-01-06T12:50:49.831879Z] {configuration.py:1101} WARNING - section/key [openlineage/namespace] not found in config
[2026-01-06T12:50:49.959748Z] {local.py:145} WARNING - To enable emitting Openlineage events, upgrade to Airflow 2.7 or install astronomer-cosmos[openlineage].

$
```

dev/dags/jaffle_shop_kubernetes.py
tatiana: Done in a6a8c1ba287d42460baec7d24705970d605d5c52

.github/workflows/test.yml
tatiana: @pankajastro I agree with your thought, we can improve this in a separate PR.

I'll give it a try to revert to 3.0 and see if it works.

The implementation for `ExecutionMode.WATCHER_KUBERNETES` is dependent on one of the following versions of `apache-airflow-providers-cncf-kubernetes`:

- 10.8.0
- 10.9.0
- 10.10.0
- 10.11.0

It does not work with older versions due to significant changes in the provider interface.

Given how we install dependencies and use `constraints.txt`, it seemed the simplest way to run tests with `ExecutionMode.WATCHER_KUBERNETES` in the CI.

If we want to support other Airflow versions, we'll need to review how we install `apache-airflow-providers-cncf-kubernetes` using `constraints` for each version we'd like to run tests against.

tatiana:
Thanks a lot for the reviews, @johnhoran @pankajkoti @pankajastro - I think I addressed most, if not all, of them

cosmos/constants.py
pankajastro: I think we may need to document this limitation

cosmos/operators/watcher.py
```diff
@@ -216,8 +216,8 @@ class DbtConsumerWatcherSensor(BaseConsumerSensor, DbtRunLocalOperator):  # type
     def __init__(
         self,
         *,
+        project_dir: str,
```
pankajastro, line 219: You are changing this from kwargs to args. Do we consider it a breaking change?

pankajastro (approved):
LGTM

tests/operators/test_watcher_kubernetes.py
```diff
@@ -0,0 +1,305 @@
...
+        assert result is True
+        mock_build_and_run_cmd.assert_called_once()
+
+    def test_use_event_returns_false(self):
+        """
+        DbtConsumerWatcherKubernetesSensor should return False for _use_event(),
+        meaning it uses XCom-based status retrieval instead of events.
+        """
+        sensor = self.make_sensor()
+        assert sensor._use_event() is False
```
pankajkoti:
```suggestion
        assert sensor.use_event() is False
```
The method was renamed, right?

tests/operators/test_watcher_kubernetes.py
```diff
@@ -0,0 +1,305 @@
...
+        context = self.make_context(ti)
+
+        result = sensor.poke(context)
+
+        assert result is True
+        mock_build_and_run_cmd.assert_called_once()
+
+    def test_use_event_returns_false(self):
+        """
+        DbtConsumerWatcherKubernetesSensor should return False for _use_event(),
```
pankajkoti:
```suggestion
        DbtConsumerWatcherKubernetesSensor should return False for use_event(),
```

tests/operators/test_watcher_kubernetes.py
pankajkoti: Wondering why the test didn't complain 🤔

cosmos/operators/watcher_kubernetes.py
pankajkoti: just want to think this out loud. So when a DAG run is created, it creates new pods right for the producer task? I mean, concurrent DAG runs won't cause a race condition here, right? With overriding or using stale contexts for the global variable?

pankajkoti (approved):
LGTM. Couple of questions inline

cosmos/operators/watcher_kubernetes.py
tatiana: @pankajkoti This global variable only matters during the producer task execution. Regardless of whether users are using Airflow Kubernetes or Celery executor, there is an independent process executing each producer task. Different task instances also represent different processes. This global environment variable is bound to the process, so one task execution will not affect another task execution.

cosmos/operators/watcher.py
tatiana: This change was introduced due to MyPy checks. The parent class already had this as an arg, so I believe it is a bit of grey territory if we consider it a breaking change or not. Given that `ExecutionMode.WATCHER` is in experimental mode, I believe we can go with this change without further concerns.

cosmos/constants.py
tatiana: Thanks, @pankajastro , it is documented in the PR description & there will be a follow up PR with the docs, it will include this!

.github/workflows/test.yml
tatiana: Attempted reverting to AF 3.0 in 5ec7e1715f1e597f917a13fa157099219a138653 - waiting to confirm if checks pass.

tests/operators/test_watcher_kubernetes.py
tatiana: Addressed in 495bafc30f53f7d85f410d91eb031c946b6637da

tests/operators/test_watcher_kubernetes.py
tatiana: Addressed in 495bafc30f53f7d85f410d91eb031c946b6637da

cosmos/operators/watcher_kubernetes.py
pankajkoti: okay, thanks for clarifying that

tests/operators/test_watcher_kubernetes.py
tatiana: I was surprised as well. I found out the issue and fixed in b06bacc74fdc0a265f958e07abaecebea7b1260d, the unittests were not running for the watcher K8s, because the file was being skipped because it led to import errors. So I split integration & unit tests in separate files

tests/operators/test_watcher_kubernetes.py
tatiana: Fixed in 495bafc30f53f7d85f410d91eb031c946b6637da

.github/workflows/test.yml
tatiana: It worked, and I reverted

PR #2210, opened 2025-12-18
Replace `map_index` with `is_mapped_task` boolean in task telemetry

pankajkoti:
Change task telemetry to emit `is_mapped_task` boolean instead of raw `map_index` integer value for clearer intent

related: #2195 
related: #2110


tatiana (approved):
Thanks for addressing the feedback so quickly, @pankajkoti , looks great!

PR #2223, opened 2025-12-23
Add DAG run telemetry metrics for load mode, invocation, and other render_config parameters

### LINKED ISSUE #2109: Track render config via Scarf
Measure adoption of:
- [x] each type of load_mode (differentiate DBT_LS and DBT_LS_CACHE)
- [ ] if using dbt ls, we also want to track 
    - [ ] dbt and Airflow are in the same virtualenv
- [x] track if node_converters are being used
- [x] track source node behaviour
- [x] track test behaviour
- [x] is dbt deps being installed
- [ ] time it took to parse the project
- [ ] if not using dbt ls, time it took to filter nodes
- [x] tasks count
- [x] cosmos task count
- [x] used_automatic_load_mode
- [x] actual_load_mode
- [x] invocation_mode
- [x] total_dbt_models
- [x] selected_dbt_models

pankajkoti:
Adds 9 new telemetry metrics to DAG run events to better understand how Cosmos is configured and used:

  - `used_automatic_load_mode` - whether the user specified LoadMode.AUTOMATIC
  - `actual_load_mode` - the resolved load method (dbt_ls, dbt_manifest, custom, etc.)
  - `invocation_mode` - subprocess or dbt_runner
  - `install_deps` - whether dbt deps installation is enabled
  - `uses_node_converter` - whether custom node converters are used
  - `test_behavior` - how dbt tests are handled (after_each, after_all, none, etc.)
  - `source_behavior` - how dbt sources are rendered
  - `total_dbt_models` - total number of dbt models in project
  - `selected_dbt_models` - number of models selected for rendering
  
 These metrics are collected during DAG conversion and stored in `dag.params` to survive Airflow's serialisation process, then emitted by the DAG run listener on success/failure events.

closes: #2109

### DIFF
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -413,16 +413,16 @@ class DbtToAirflowConverter:
             logger.warning(f"Failed to compute used_automatic_load_mode: {e}")
 
         try:
-            metadata["actual_load_mode"] = self.dbt_graph.load_method.value
+            metadata["actual_load_mode"] = str(self.dbt_graph.load_method.value)
         except Exception as e:
             logger.warning(f"Failed to compute actual_load_mode: {e}")
 
         try:
             invocation_mode = None
             if execution_config.invocation_mode:
-                invocation_mode = execution_config.invocation_mode.value
+                invocation_mode = str(execution_config.invocation_mode.value)
             elif render_config.invocation_mode:
-                invocation_mode = render_config.invocation_mode.value
+                invocation_mode = str(render_config.invocation_mode.value)
             metadata["invocation_mode"] = invocation_mode
         except Exception as e:
             logger.warning(f"Failed to compute invocation_mode: {e}")
@@ -441,12 +441,12 @@ class DbtToAirflowConverter:
             logger.warning(f"Failed to compute uses_node_converter: {e}")
 
         try:
-            metadata["test_behavior"] = render_config.test_behavior.value
+            metadata["test_behavior"] = str(render_config.test_behavior.value)
         except Exception as e:
             logger.warning(f"Failed to compute test_behavior: {e}")
 
         try:
-            metadata["source_behavior"] = render_config.source_rendering_behavior.value
+            metadata["source_behavior"] = str(render_config.source_rendering_behavior.value)
         except Exception as e:
             logger.warning(f"Failed to compute source_behavior: {e}")
 
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -33,6 +33,8 @@ from cosmos.versioning import _create_folder_version_hash
 
 logger = get_logger(__name__)
 
+from cosmos.constants import DbtResourceType
+
 
 def migrate_to_new_interface(
     execution_config: ExecutionConfig, project_config: ProjectConfig, render_config: RenderConfig
@@ -279,6 +281,9 @@ class DbtToAirflowConverter:
             cache_identifier = cache._create_cache_identifier(dag, task_group)
             cache_dir = cache._obtain_cache_dir_path(cache_identifier=cache_identifier)
 
+        # Store the initial load method before it gets resolved by dbt_graph.load()
+        initial_load_method = render_config.load_method
+
         previous_time = time.perf_counter()
         self.dbt_graph = DbtGraph(
             project=project_config,
@@ -293,6 +298,9 @@ class DbtToAirflowConverter:
         self.dbt_graph.load(method=render_config.load_method, execution_mode=execution_config.execution_mode)
 
         self._add_dbt_project_hash_to_dag_docs(dag)
+        self._store_cosmos_telemetry_metadata_on_dag(
+            dag, render_config, execution_config, project_config, operator_args, initial_load_method
+        )
 
         current_time = time.perf_counter()
         elapsed_time = current_time - previous_time
@@ -370,3 +378,91 @@ class DbtToAirflowConverter:
             logger.debug(f"Appended dbt project hash {dbt_project_hash} to DAG {dag.dag_id} documentation")
         except Exception as e:
             logger.warning(f"Failed to append dbt project hash to DAG documentation: {e}")
+
+    def _store_cosmos_telemetry_metadata_on_dag(  # noqa: C901
+        self,
+        dag: DAG | None,
+        render_config: RenderConfig,
+        execution_config: ExecutionConfig,
+        project_config: ProjectConfig,
+        operator_args: dict[str, Any],
+        initial_load_method: LoadMode,
+    ) -> None:
+        """
+        Store Cosmos configuration metadata on the DAG for telemetry purposes.
+
+        This metadata is used by the DAG run listener to emit telemetry metrics
+        about how Cosmos is configured and used.
+
+        :param dag: The Airflow DAG to store metadata on. If None, no action is taken.
+        :param render_config: The render configuration
+        :param execution_config: The execution configuration
+        :param project_config: The project configuration
+        :param operator_args: The operator arguments
+        :param initial_load_method: The load method specified by the user (before automatic resolution)
+        """
+        if dag is None:
+            return
+
+        metadata = {}
+
+        # Compute each metric individually with error handling
+        try:
+            metadata["used_automatic_load_mode"] = initial_load_method == LoadMode.AUTOMATIC
+        except Exception as e:
+            logger.warning(f"Failed to compute used_automatic_load_mode: {e}")
+
+        try:
+            metadata["actual_load_mode"] = self.dbt_graph.load_method.value
+        except Exception as e:
+            logger.warning(f"Failed to compute actual_load_mode: {e}")
+
+        try:
+            invocation_mode = None
+            if execution_config.invocation_mode:
+                invocation_mode = execution_config.invocation_mode.value
+            elif render_config.invocation_mode:
+                invocation_mode = render_config.invocation_mode.value
+            metadata["invocation_mode"] = invocation_mode
+        except Exception as e:
+            logger.warning(f"Failed to compute invocation_mode: {e}")
+
+        try:
+            install_deps = operator_args.get("install_deps")
+            if install_deps is None:
+                install_deps = project_config.install_dbt_deps
+            metadata["install_deps"] = bool(install_deps) if install_deps is not None else True
+        except Exception as e:
+            logger.warning(f"Failed to compute install_deps: {e}")
+
+        try:
+            metadata["uses_node_converter"] = render_config.node_converters is not None
+        except Exception as e:
+            logger.warning(f"Failed to compute uses_node_converter: {e}")
+
+        try:
+            metadata["test_behavior"] = render_config.test_behavior.value
+        except Exception as e:
+            logger.warning(f"Failed to compute test_behavior: {e}")
+
+        try:
+            metadata["source_behavior"] = render_config.source_rendering_behavior.value
+        except Exception as e:
+            logger.warning(f"Failed to compute source_behavior: {e}")
+
+        try:
+            metadata["total_dbt_models"] = sum(
+                1 for node in self.dbt_graph.nodes.values() if node.resource_type == DbtResourceType.MODEL
+            )
+        except Exception as e:
+            logger.warning(f"Failed to compute total_dbt_models: {e}")
+
+        try:
+            metadata["selected_dbt_models"] = sum(
+                1 for node in self.dbt_graph.filtered_nodes.values() if node.resource_type == DbtResourceType.MODEL
+            )
+        except Exception as e:
+            logger.warning(f"Failed to compute selected_dbt_models: {e}")
+
+        dag._cosmos_telemetry_metadata = metadata
+        logger.debug(f"Stored Cosmos telemetry metadata on DAG {dag.dag_id}: {metadata}")
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -464,5 +464,7 @@ class DbtToAirflowConverter:
         except Exception as e:
             logger.warning(f"Failed to compute selected_dbt_models: {e}")
 
-        dag._cosmos_telemetry_metadata = metadata
-        logger.debug(f"Stored Cosmos telemetry metadata on DAG {dag.dag_id}: {metadata}")
+        # Store metadata in dag.params which is preserved during serialization
+        # Using a key that's unlikely to conflict with user params
+        dag.params["__cosmos_telemetry_metadata__"] = metadata
+        logger.debug(f"Stored Cosmos telemetry metadata in DAG {dag.dag_id} params: {metadata}")
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -297,9 +297,7 @@ class DbtToAirflowConverter:
         self.dbt_graph.load(method=render_config.load_method, execution_mode=execution_config.execution_mode)
 
         self._add_dbt_project_hash_to_dag_docs(dag)
-        self._store_cosmos_telemetry_metadata_on_dag(
-            dag, render_config, project_config, operator_args, initial_load_method
-        )
+        self._store_cosmos_telemetry_metadata_on_dag(dag, render_config, project_config, initial_load_method)
 
         current_time = time.perf_counter()
         elapsed_time = current_time - previous_time
@@ -384,7 +382,6 @@ class DbtToAirflowConverter:
         dag: DAG | None,
         render_config: RenderConfig,
         project_config: ProjectConfig,
-        operator_args: dict[str, Any],
         initial_load_method: LoadMode,
     ) -> None:
         """
@@ -396,8 +393,6 @@ class DbtToAirflowConverter:
         :param dag: The Airflow DAG to store metadata on. If None, no action is taken.
         :param render_config: The render configuration
         :param project_config: The project configuration
-        :param operator_args: The operator arguments
-        :param initial_load_method: The load method specified by the user (before automatic resolution)
         """
         if dag is None:
             return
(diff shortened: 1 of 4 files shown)


cosmos/converter.py
```diff
@@ -370,3 +378,93 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+        """
+        if dag is None:
+            return
+
+        metadata = {}
+
+        # Compute each metric individually with error handling
+        try:
+            metadata["used_automatic_load_mode"] = initial_load_method == LoadMode.AUTOMATIC
+        except Exception as e:
```
tatiana: Could we capture a more specific exception here? Which error do we expect to face?

cosmos/converter.py
```diff
@@ -370,3 +378,93 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+            logger.warning(f"Failed to compute used_automatic_load_mode: {e}")
+
+        try:
+            metadata["actual_load_mode"] = str(self.dbt_graph.load_method.value)
+        except Exception as e:
```
tatiana: Could we capture a more specific exception here? Which error do we expect to face?

cosmos/converter.py
```diff
@@ -370,3 +378,93 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+            logger.warning(f"Failed to compute actual_load_mode: {e}")
+
+        try:
+            invocation_mode = None
+            if execution_config.invocation_mode:
```
tatiana: Is the `execution_config.invocation_mode` affecting DAG parsing? If yes, this is likely a bug, the DAG parsing should rely on the `render_config` only

cosmos/converter.py
```diff
@@ -370,3 +378,93 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+                invocation_mode = str(execution_config.invocation_mode.value)
+            elif render_config.invocation_mode:
+                invocation_mode = str(render_config.invocation_mode.value)
+            metadata["invocation_mode"] = invocation_mode
+        except Exception as e:
```
tatiana: Same feedback as https://github.com/astronomer/astronomer-cosmos/pull/2223/changes#r2650801255

cosmos/converter.py
```diff
@@ -370,3 +378,93 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+            logger.warning(f"Failed to compute invocation_mode: {e}")
+
+        try:
+            install_deps = operator_args.get("install_deps")
```
tatiana: Cosmos DAG parsing should not rely on `install_deps` from `operator_args`. It should rely exclusively in the `render_config` or `project_config` one

cosmos/converter.py
```diff
@@ -370,3 +378,93 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+            if install_deps is None:
+                install_deps = project_config.install_dbt_deps
+            metadata["install_deps"] = bool(install_deps) if install_deps is not None else True
+        except Exception as e:
```
tatiana: Same feedback as https://github.com/astronomer/astronomer-cosmos/pull/2223/changes#r2650801255

We should avoid capturing `Exception` unless strictly necessary. The same applies to all `except Exception` below

cosmos/converter.py
```diff
@@ -370,3 +378,93 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
             logger.debug(f"Appended dbt project hash {dbt_project_hash} to DAG {dag.dag_id} documentation")
         except Exception as e:
             logger.warning(f"Failed to append dbt project hash to DAG documentation: {e}")
+
+    def _store_cosmos_telemetry_metadata_on_dag(  # noqa: C901
+        self,
+        dag: DAG | None,
+        render_config: RenderConfig,
+        execution_config: ExecutionConfig,
```
tatiana: In theory, we should not need to pass `execution_config` to add `render_config` information.

tatiana (commented):
Thanks for working on this, @pankajkoti. I'm glad the DAG params are working!

I left some feedback inline.

Please confirm which versions of Airflow you tested it against, and whether you were able to confirm the events reaching Scarf?

tests/listeners/test_dag_run_listener.py
```diff
@@ -183,3 +184,116 @@ def test_on_dag_run_failed(mock_emit_usage_metrics_if_enabled, caplog):
     assert "Running on_dag_run_failed" in caplog.text
     assert "Completed on_dag_run_failed" in caplog.text
     assert mock_emit_usage_metrics_if_enabled.call_count == 1
+
+
+@pytest.mark.skipif(
```
tatiana, line 189: Do we have any ideas on how to fix this? Would it be worth having a follow-up ticket? Does this mean we're not running integration tests for emitting telemetry with AF 3.1 and above?

tests/listeners/test_dag_run_listener.py
```diff
@@ -183,3 +184,116 @@ def test_on_dag_run_failed(mock_emit_usage_metrics_if_enabled, caplog):
...
+    assert metrics["install_deps"] is True
+    assert metrics["uses_node_converter"] is False
+    assert metrics["test_behavior"] == "after_each"
+    assert metrics["source_behavior"] == "none"
+
+
+@pytest.mark.skipif(
+    AIRFLOW_VERSION >= Version("3.1.0"),
+    reason="TODO: Fix create_dag_run to work with AF 3.1 and remove this skip.",
+)
```
pankajkoti, line 248:
```suggestion
```

tests/listeners/test_dag_run_listener.py
```diff
@@ -183,3 +184,116 @@ def test_on_dag_run_failed(mock_emit_usage_metrics_if_enabled, caplog):
     assert "Running on_dag_run_failed" in caplog.text
     assert "Completed on_dag_run_failed" in caplog.text
     assert mock_emit_usage_metrics_if_enabled.call_count == 1
+
+
+@pytest.mark.skipif(
+    AIRFLOW_VERSION >= Version("3.1.0"),
+    reason="TODO: Fix create_dag_run to work with AF 3.1 and remove this skip.",
+)
```
pankajkoti, line 192:
```suggestion
```

cosmos/converter.py
pankajkoti: yes, capturing specific exceptions now

cosmos/converter.py
pankajkoti: yes, capturing specific exceptions now

cosmos/converter.py
pankajkoti: Yes, good catch, we don't use execution_config.invocation_mode during DAG parsing. Removed this.

https://github.com/astronomer/astronomer-cosmos/blob/0fa0163c02d709e0849e007e40f165ef0743d29c/cosmos/dbt/graph.py#L623

cosmos/converter.py
pankajkoti: Yes, using render_config now which gets used during DAG parsing https://github.com/astronomer/astronomer-cosmos/blob/0fa0163c02d709e0849e007e40f165ef0743d29c/cosmos/dbt/graph.py#L388

cosmos/converter.py
pankajkoti: yes, correct. removed this argument execution_config.

tests/listeners/test_dag_run_listener.py
pankajkoti: We are getting this error `AttributeError: 'DbtDag' object has no attribute 'create_dagrun'` https://github.com/astronomer/astronomer-cosmos/actions/runs/20595308844/job/59148647042?pr=2223#step:7:2707

I saw we have this reported earlier with the code comment: https://github.com/astronomer/astronomer-cosmos/blob/0fa0163c02d709e0849e007e40f165ef0743d29c/tests/listeners/test_dag_run_listener.py#L103 and also have disabled other existing tests on Airflow 3.1 for the same reason earlier e.g. https://github.com/astronomer/astronomer-cosmos/blob/0fa0163c02d709e0849e007e40f165ef0743d29c/tests/listeners/test_dag_run_listener.py#L136

pankajkoti:
> Thanks for working on this, @pankajkoti. I'm glad the DAG params are working!
> 
> I left some feedback inline.
> 
> Please confirm which versions of Airflow you tested it against, and whether you were able to confirm the events reaching Scarf?

Yes, I am glad too that the DAG params worked here. Thanks for the review, it helped.

Addressed the review feedback. And I tested this with both Airflow 2 and 3, events are reaching Scarf on both, and we get 200 OK responses.

tests/listeners/test_dag_run_listener.py
```diff
@@ -183,3 +184,116 @@ def test_on_dag_run_failed(mock_emit_usage_metrics_if_enabled, caplog):
...
+    assert "install_deps" in metrics
+    assert "uses_node_converter" in metrics
+    assert "test_behavior" in metrics
+    assert "source_behavior" in metrics
+    assert "total_dbt_models" in metrics
+    assert "selected_dbt_models" in metrics
+
+    # Verify some expected values
+    assert metrics["used_automatic_load_mode"] is True
+    assert metrics["invocation_mode"] == "dbt_runner"
```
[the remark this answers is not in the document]
pankajkoti: that is intentional to check execution_mode.invocation_mode is not used during DAG parsing.

tests/listeners/test_dag_run_listener.py
```diff
@@ -183,3 +184,116 @@ def test_on_dag_run_failed(mock_emit_usage_metrics_if_enabled, caplog):
...
+    assert "uses_node_converter" in metrics
+    assert "test_behavior" in metrics
+    assert "source_behavior" in metrics
+    assert "total_dbt_models" in metrics
+    assert "selected_dbt_models" in metrics
+
+    # Verify some expected values for failed case
+    assert metrics["used_automatic_load_mode"] is False
+    assert metrics["invocation_mode"] == "dbt_runner"
+    assert metrics["install_deps"] is True
```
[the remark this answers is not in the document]
pankajkoti: That is intentional to check that operator_args don't play a role during DAG parsing.

cosmos/converter.py
```diff
@@ -33,6 +33,8 @@
 
 logger = get_logger(__name__)
 
+from cosmos.constants import DbtResourceType
```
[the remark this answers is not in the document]
pankajkoti: yes, good catch, addressed.

cosmos/converter.py
```diff
@@ -294,6 +297,9 @@ def __init__(
         self.dbt_graph.load(method=render_config.load_method, execution_mode=execution_config.execution_mode)
 
         self._add_dbt_project_hash_to_dag_docs(dag)
+        self._store_cosmos_telemetry_metadata_on_dag(
+            dag, render_config, project_config, operator_args, initial_load_method
```
tatiana: We should not need `operator_args` for the DAG render config 
```suggestion
            dag, render_config, project_config, initial_load_method
```

tatiana:
```suggestion
            dag, render_config, project_config, initial_load_method
```

cosmos/converter.py
```diff
@@ -372,3 +378,87 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
             logger.debug(f"Appended dbt project hash {dbt_project_hash} to DAG {dag.dag_id} documentation")
         except Exception as e:
             logger.warning(f"Failed to append dbt project hash to DAG documentation: {e}")
+
+    def _store_cosmos_telemetry_metadata_on_dag(  # noqa: C901
+        self,
+        dag: DAG | None,
+        render_config: RenderConfig,
+        project_config: ProjectConfig,
+        operator_args: dict[str, Any],
```
tatiana:
```suggestion
```

cosmos/converter.py
```diff
@@ -372,3 +378,87 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+        """
+        Store Cosmos configuration metadata on the DAG for telemetry purposes.
+
+        This metadata is used by the DAG run listener to emit telemetry metrics
+        about how Cosmos is configured and used.
+
+        :param dag: The Airflow DAG to store metadata on. If None, no action is taken.
+        :param render_config: The render configuration
+        :param project_config: The project configuration
+        :param operator_args: The operator arguments
```
tatiana:
```suggestion
```

cosmos/converter.py
```diff
@@ -372,3 +378,87 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+        try:
+            invocation_mode = None
+            if render_config.invocation_mode:
+                invocation_mode = str(render_config.invocation_mode.value)
+            metadata["invocation_mode"] = invocation_mode
+        except (AttributeError, TypeError) as e:
+            logger.warning(f"Failed to compute invocation_mode: {e}")
+
+        try:
+            install_deps = render_config.dbt_deps
```
tatiana: It may come from `project_config. install_dbt_deps `, otherwise

cosmos/converter.py
```diff
@@ -372,3 +378,87 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+        """
+        if dag is None:
+            return
+
+        metadata = {}
+
+        # Compute each metric individually with error handling
+        try:
+            metadata["used_automatic_load_mode"] = initial_load_method == LoadMode.AUTOMATIC
+        except (AttributeError, TypeError) as e:
```
tatiana: In which scenario would we get `AttributeError` or `TypeError` here? We may not need any exception handling in this line

cosmos/converter.py
```diff
@@ -372,3 +378,87 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+            logger.warning(f"Failed to compute used_automatic_load_mode: {e}")
+
+        try:
+            metadata["actual_load_mode"] = str(self.dbt_graph.load_method.value)
```
tatiana: Are there any scenarios where we expect this code to be invoked when `self.dbt_graph` would not have been set?

cosmos/converter.py
```diff
@@ -372,3 +378,87 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+        except (AttributeError, TypeError) as e:
+            logger.warning(f"Failed to compute actual_load_mode: {e}")
+
+        try:
+            invocation_mode = None
+            if render_config.invocation_mode:
```
tatiana: If `render_config` is not `None`, then it will have an `invocation_mode` attribute:
https://github.com/astronomer/astronomer-cosmos/blob/0fa0163c02d709e0849e007e40f165ef0743d29c/cosmos/config.py#L86

The same applies to `dbt_deps` below:
https://github.com/astronomer/astronomer-cosmos/blob/0fa0163c02d709e0849e007e40f165ef0743d29c/cosmos/config.py#L90

And also:
- `uses_node_converter`
- `test_behavior`
- `source_behavior`

Would it be worth simplifying the code by checking whether `render_config` is `None` and setting those values accordingly, without the need of so many `try`/`except` statements?

cosmos/converter.py
tatiana: We may want to handle this close to the other dbt_graph properties:
- `total_dbt_models`
- `selected_dbt_models`

Applying the same feedback as described in https://github.com/astronomer/astronomer-cosmos/pull/2223/changes#r2655266426

tests/listeners/test_dag_run_listener.py
tatiana: HI @pankajkoti , that's a shame :(
Please, could you confirm if we have any tickets in the backlog to tackle this, and if not, create one and add to the next iteration, please?

tests/listeners/test_dag_run_listener.py
```diff
@@ -183,3 +184,116 @@ def test_on_dag_run_failed(mock_emit_usage_metrics_if_enabled, caplog):
...
+            DBT_ROOT_PATH / "jaffle_shop",
+        ),
+        profile_config=profile_config,
+        execution_config=ExecutionConfig(invocation_mode=InvocationMode.SUBPROCESS),
+        render_config=RenderConfig(
+            load_method=LoadMode.AUTOMATIC,
+            test_behavior=TestBehavior.AFTER_EACH,
+            source_rendering_behavior=SourceRenderingBehavior.NONE,
+        ),
+        operator_args={"install_deps": True},
```
tatiana, line 210: Do we want to pass the `install_deps` somewhere else setting to `False` and confirm that we're capturing the `install_deps` from the correct place?

tatiana (approved):
Thanks a lot, @pankajkoti , I left some minor feedback in line, but since you already tested and confirmed this worked with Airflow 2 and 3, I'm approving it - feel free to merge once the feedback is addressed.

cosmos/converter.py
pankajkoti: yes. Based on the current implementation there is no invocation when self.dbt_graph is not set. Yet I have added a safety check that the value is not None and grouped those logically together.

cosmos/converter.py
pankajkoti: yes done so, good suggestion

tests/listeners/test_dag_run_listener.py
pankajkoti: ~yes, I set it here to False. And set the render_config.dbt_deps=True and asserting that.~

tests/listeners/test_dag_run_listener.py
pankajkoti: created an issue for this: https://github.com/astronomer/astronomer-cosmos/issues/2244

tests/listeners/test_dag_run_listener.py
pankajkoti: yes, I addressed this comment in the following existing test that we have test_on_dag_run_failed_with_telemetry_metadata. Over there I set render_config.dbt_deps as False, operator_args:install_deps as True and asserting that DAG parsing uses the False value and is emitted so.

PR #2230, opened 2025-12-29
Add dbt Core 1.11 to test matrix

tatiana:
There have been [a few releases](https://pypi.org/project/dbt-core/#history) of dbt Core 1.11 since December 19th:
Screenshot 2025-12-29 at 10 17 54

This PR aims to validate the latest dbt version against Cosmos tests to identify potential issues.

We temporarily enabled running tests against this branch to confirm that GH Actions Jobs were working. Example of CI job run:
https://github.com/astronomer/astronomer-cosmos/actions/runs/20570466441

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -128,7 +128,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         airflow-version: ["2.6", "2.7", "2.8", "2.9", "2.10", "2.11", "3.0", "3.1"]
-        dbt-version: ["1.10", "1.11"]
+        dbt-version: [ "1.11" ]
         exclude:
           # Apache Airflow versions prior to 2.9.0 have not been tested with Python 3.12.
           # Official support for Python 3.12 and the corresponding constraints.txt are available only for Apache Airflow >= 2.9.0.


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, support-dbt-1.11]
```
```suggestion
    branches: [main]
```

tatiana: Before we merge, we should remove the reference to the current branch

.github/workflows/test.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         airflow-version: ["2.6", "2.7", "2.8", "2.9", "2.10", "2.11", "3.0", "3.1"]
-        dbt-version: ["1.10"]
+        dbt-version: ["1.11"]
```
```suggestion
        dbt-version: ["1.10", "1.11"]
```

.github/workflows/test.yml
```diff
@@ -128,7 +128,7 @@ jobs:
...
```
```suggestion
        dbt-version: ["1.10", "1.11"]
```

.github/workflows/test.yml
```diff
@@ -640,7 +640,7 @@ jobs:
       matrix:
         python-version: [ "3.12" ]
         airflow-version: [ "2.10", "3.0" ]
-        dbt-version: [ "1.10" ]
+        dbt-version: [ "1.11" ]
```
pankajkoti, line 643:
```suggestion
        dbt-version: [ "1.10", "1.11" ]
```

pankajkoti (approved):
Looks good. Inline suggestions to also preserve 1.10, along with 1.11, to run the Unit Integration and Kubernetes tests since it's still a supported version. WDYT?

.github/workflows/test.yml
tatiana: Happy for this, @pankajkoti !

.github/workflows/test.yml
```diff
@@ -128,7 +128,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         airflow-version: ["2.6", "2.7", "2.8", "2.9", "2.10", "2.11", "3.0", "3.1"]
-        dbt-version: ["1.10"]
+        dbt-version: ["1.10", "1.11"]
```
```suggestion
        dbt-version: [ "1.11" ]
```
@pankajkoti I don't think we should run Cosmos integration tests for both versions of dbt Core. Previously, we were only running against the latest minor. It is expensive to run them. We currently run 4x8 integration test jobs (32 in total). By adding this, we'd be running 64 jobs, without any clear benefit.

The main part of Cosmos that is sensitive to the dbt version is the async implementation - and for those we should certainly test against as many dbt versions as possible.

If we notice that things stop working with a specific dbt 1.10 version, we can enable testing against those versions, but that doesn't justify the cost for all tests.

.github/workflows/test.yml
```diff
@@ -640,7 +640,7 @@ jobs:
       matrix:
         python-version: [ "3.12" ]
         airflow-version: [ "2.10", "3.0" ]
-        dbt-version: [ "1.10" ]
+        dbt-version: [ "1.10", "1.11" ]
```
```suggestion
        dbt-version: [ "1.11" ]
```

.github/workflows/test.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         airflow-version: ["2.6", "2.7", "2.8", "2.9", "2.10", "2.11", "3.0", "3.1"]
-        dbt-version: ["1.10"]
+        dbt-version: ["1.10", "1.11"]
```
```suggestion
        dbt-version: ["1.10"]
```
@pankajkoti Cosmos unit tests do not rely on the `dbt-core` version. They should all be agnostic of dbt being installed.

pankajkoti: yes, makes sense for unit tests.

tatiana:
Thanks for the quick review, @pankajkoti !

> Looks good. Inline suggestions to also preserve 1.10, along with 1.11, to run the Unit Integration and Kubernetes tests, since it's still a supported version. WDYT?

For unit tests, it does not make sense because they should not rely on `dbt` being installed.

For integration tests, it would be too expensive to run them against all supported versions of Python and Airflow (currently, the matrix spans 32 jobs). By having two dbt versions, we'd have 64 - if we were to add all dbt versions that didn't reach EOL. It would be 128).

In most cases, Cosmos does not rely on specific features of dbt versions - except for two scenarios:
- the async implementation
- dbt 1.5
Both of which have dedicated integration test jobs.

I'm happy for us to have a more extensive dbt version testing in the CI. That said, I don't believe the solution is to multiply the existing Py & Airflow matrix by also having multiple dbt versions.

WDYT of logging a dedicated ticket for us to review and develop a better strategy? We could run a subset of integration tests with a varying matrix of AF & dbt, but pin the Python version. It feels this is outside of the scope of the current ticket.

pankajkoti:
Sounds all good to me @tatiana 👍🏽

tatiana:
Logged: https://github.com/astronomer/astronomer-cosmos/issues/2234

PR #2236, opened 2025-12-29
Refactor `ExecutionMode.WATCHER` so the logic is centralised

tatiana:
Since https://github.com/astronomer/astronomer-cosmos/pull/2084, the `ExecutionMode.WATCHER` logic started spreading throughout the Cosmos code base.

When implementing Airflow providers, it is common practice to split Airflow code into operators and triggerers folders. However, in the specific case of Cosmos, I believe this approach makes it harder to read and maintain the project and that it is better to centralise the execution mode code.

Historically, Cosmos execution modes were defined in the operators folder because the code for each execution mode was operator-only. With the introduction of the async and watcher execution modes, the execution mode implementation became more complex, requiring a custom triggerer for the watcher. As a result, the watcher execution mode code began spreading throughout the Cosmos codebase, complicating code distribution more than needed.

This PR brings the watcher triggerer logic closer to the operator, so that most of the watcher-specific logic lives near the execution mode implementation.

In Cosmos 2.0, we can review how we name the Cosmos execution mode folder, potentially renaming this folder from operators to execution_modes.

### DIFF
--- a/cosmos/operators/_watcher/__init__.py
+++ b/cosmos/operators/_watcher/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger", "_parse_compressed_xcom"]
+__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger"]
 
 from cosmos.operators._watcher.state import build_producer_state_fetcher, get_xcom_val, safe_xcom_push
 from cosmos.operators._watcher.triggerer import WatcherTrigger, _parse_compressed_xcom
--- a/cosmos/operators/_watcher/__init__.py
+++ b/cosmos/operators/_watcher/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
-__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher"]
+__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger", "_parse_compressed_xcom"]
 
 from cosmos.operators._watcher.state import build_producer_state_fetcher, get_xcom_val, safe_xcom_push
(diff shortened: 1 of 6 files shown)


cosmos/operators/_watcher/__init__.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
-__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher"]
+__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger", "_parse_compressed_xcom"]
```
tatiana:
```suggestion
__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger"]
```

cosmos/operators/_watcher/__init__.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
-__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher"]
+__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger"]
```
tatiana:
```suggestion
__all__ = ["get_xcom_val", "safe_xcom_push", "build_producer_state_fetcher", "WatcherTrigger", "_parse_compressed_xcom"]
```

pankajkoti (approved):
If not already done, would be nice to sanity test this once with running an Airflow Dag and confirm by seeing in the UI that the sensors are getting deferred and sensing XCOMs correctly and that the trigger complets it's lifecycle

tatiana:
Thanks a lot, @pankajkoti !

> If not already done, would be nice to sanity test this once by running an Airflow Dag and confirm by seeing in the UI that the sensors are getting deferred and sensing XCOMs correctly and that the trigger complets it's lifecycle

Things are working as expected.

During the execution of the `example_watcher`, I was able to observe tasks being deferred:
Screenshot 2025-12-29 at 16 29 07

By the end, all tasks were successful:
Screenshot 2025-12-29 at 16 33 08

PR #2237, opened 2025-12-29
Improve watcher producer task priority both in scheduling and in the UI

tatiana:
Some users noted that the producer task would appear after the consumer tasks in the Airflow UI. We also received a report that the producer task was sometimes scheduled after the consumer tasks, even though it has a higher priority weight.

I discussed these issues with @ashb, and he advised that we instantiate the producer task before the consumer tasks. This PR aims to accomplish this by breaking down the method `_add_producer_watcher_and_dependencies` into two methods:
- `_add_watcher_producer_task`
- `_add_watcher_dependencies`

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -649,7 +649,7 @@ def _add_watcher_producer_task(
     render_config: RenderConfig | None = None,
 ) -> BaseOperator:
     """
-    Create the producer task for the watcher execution modeand add it to the tasks_map.
+    Create the producer task for the watcher execution mode and add it to the tasks_map.
     The producer task is the task that will be used to produce the events for the watcher execution mode.
     """
     producer_task_args = task_args.copy()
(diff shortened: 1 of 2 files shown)


cosmos/airflow/graph.py
```diff
@@ -844,6 +859,15 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
     if execution_mode == ExecutionMode.AIRFLOW_ASYNC:
         # This property is only relevant for the setup task, not the other tasks:
         virtualenv_dir = task_args.pop("virtualenv_dir", None)
+    elif execution_mode == ExecutionMode.WATCHER:
+        setup_operator_args = getattr(execution_config, "setup_operator_args", None) or {}
+        producer_task = _add_watcher_producer_task(
```
tatiana, line 867:
```suggestion
        # We are intentionally creating the producer task ahead of the consumer tasks
        # Airflow priority weight is not being respected in multiple versions of the library, including 3.1
        # To instantiate the producer before helps having it before on the DAG topological order and scheduling this task before the consumer tasks
        producer_task = _add_watcher_producer_task(
```

tatiana: @pankajkoti just added the comment, as we agreed

pankajkoti (approved):
Looks solid to me. Should not harm us and only benefit, I believe, if our tests on DAG rendering/running conclude that

PR #2254, opened 2026-01-07
Remove dbt-core 2.0 from pyrpject.toml test matrix

pankajastro:
dbt Core 2.0 has not been released, so let's remove it from the test matrix

https://pypi.org/project/dbt-core/


pankajastro:
dbt Fusion is dbt 2.0

Issue #2255, opened 2026-01-08
[Feature] Support starrocks profile.


kurkim0661:
### Description

I’m using the StarRocks datasource, but there isn’t a BaseProfileMapping implementation that supports a StarRocks profile.

Since StarRocks supports the MySQL protocol, I tried using that, but it forces the profile type to mysql, which prevents me from using the dbt-starrocks adapter.

### Use case/motivation

As StarRocks usage keeps growing, adding a proper implementation for it would be really helpful.

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

pankajastro:
@kurkim0661 Please feel free to submit a PR.

PR #2256, opened 2026-01-08
feat: Add support for StarRocks profile mapping

kurkim0661:
## Description
Introduces the StarrocksUserPasswordProfileMapping class, enabling Cosmos to automatically map Airflow connections to dbt profiles for StarRocks.

### Test environment
- dbt-starrocks==1.11.0
- apache-airflow-providers-mysql==5.7.4
- airflow==2.10.4
- python==3.12

### Concerns
1. When using get_automatic_profile_mapping, it compares airflow_connection_type with conn_type. For StarRocks, we use the mysql type, so it overlaps with regular MySQL connections. Since MySQL is searched first, existing MySQL users are fine—but StarRocks won’t be found via the automatic search as things stand.
2. The same issue happens with get_env_var_name: it also relies on airflow_connection_type, so it ends up colliding with MySQL as well.

Both issues shouldn’t affect existing users, but they could cause problems for users who want to use StarRocks.

## Related Issue(s)
Closes: https://github.com/astronomer/astronomer-cosmos/issues/2255

## Checklist

### DIFF
--- a/cosmos/profiles/starrocks/user_pass.py
+++ b/cosmos/profiles/starrocks/user_pass.py
@@ -52,9 +52,6 @@ class StarrocksUserPasswordProfileMapping(BaseProfileMapping):
             "password": self.get_env_var_format("password"),
         }
 
-        if "schema" in self.profile_args:
-            profile["schema"] = self.profile_args["schema"]
-
         return self.filter_null(profile)
 
     @property


tatiana (approved):
@kurkim0661, this looks great, thank you very much for the contribution! We'll be releasing it as part of Cosmos 1.13. The implementation, tests, and PR description are excellent, including the limitations bit. Excellent work!

PR #2271, opened 2026-01-15
Support cross-referencing models across dbt projects using dbt-loom

### LINKED ISSUE #2107: [Feature] PoC with dbt Loom to accomplish dbt Mesh cross-project resolution
### Description

First, follow the dbt mesh tutorial in dbt Platform with a trial account:
https://docs.getdbt.com/guides/mesh-qs

Second, try to reproduce running this tutorial but using Airflow + Cosmos + dbt Loom:
https://github.com/nicholasyager/dbt-loom

We want to understand the feasibility of resolving cross-project references.

Example from the [dbt Docs](https://docs.getdbt.com/docs/mesh/govern/project-dependencies):
```
with monthly_revenue as (
  
    select * from {{ ref('jaffle_finance', 'monthly_revenue') }}

),
```

This type of cross-project resolution would need to work for:
- Model reference
- Source reference
- Macro reference

### Use case/motivation

_No response_

### Related issues

_No response_

pankajkoti:
This PR adds support for dbt-loom, enabling Cosmos to work with multi-project dbt architectures where downstream projects reference models from upstream projects.

When using dbt-loom, downstream projects reference upstream models via `{{ ref('upstream_project', 'model_name') }}`. dbt-loom injects these external model references into the downstream project's namespace by reading the upstream project's manifest.json

Cosmos now automatically detects and skips external nodes (those without original_file_path) during DAG generation, while still creating tasks for the project's own models. This works for both:
`LoadMode.DBT_LS` - parsing via dbt ls
`LoadMode.DBT_MANIFEST` - parsing via manifest file

The PR adds the example Projects (in `dev/dags/dbt/`):
`dbt_loom_upstream_platform/` - staging & intermediate models with seeds
`dbt_loom_downstream_finance/` - finance fact tables referencing upstream models
`dbt_loom_dags.py` - combined DAG with chained task groups

The PR also adds a comprehensive guide for multi-project setups in `docs/configuration/multi-project.rst`

closes: https://github.com/astronomer/astronomer-cosmos/issues/2107

---------


docs/configuration/multi-project.rst
```diff
@@ -0,0 +1,442 @@
...
+- **Maintains proper dependency tracking** within each project
+
+This means you don't need any special configuration - Cosmos works with dbt-loom projects automatically.
+
+Requirements
+------------
+
+For dbt-loom to work with Cosmos:
+
+1. **For DAG parsing**: The upstream project's ``manifest.json`` must be accessible
```
tatiana, line 48: It may be worth highlighting that dbt Loom needs to be installed in the same Python virtualenv as the dbt being used by Cosmos

tatiana:
Hi @pankajkoti ! This work is very exciting - it's really cool to be able to have a dbt Mesh feature in Cosmos without the need to lock into a proprietary platform.

I'd love it if you could address the following points, as we discussed yesterday:
1. Inside the `dev/dags/dbt` folder, put both new dbt projects into the folder `cross-project`. There, I suggest one of the projects is named upstream (or parent) and the other is named downstream (or child), so it is very clear why we have those projects
2. Please, could you also confirm that Airflow Assets/Datasets emitted via Cosmos by the upstream DAG could trigger the downstream DAG?
3. Please, could you confirm if the syntax is exactly the same as dbt Mesh, and if there are any other lacking feaeture parities
4. Were you able to test the cross-project reference, considering users were using `LoadMode.DBT_MANIFEST` both on the upstream and downstream projects?
5. Would it be worth highlighting in our example or in the docs the fact that both projects can be run with different profiles?
6. Is there a way of running nodes of different projects in the same Cosmos `DbtTaskGroup`?

dev/dags/dbt/dbt_loom_downstream_finance/dbt_loom.config.yml
```diff
@@ -0,0 +1,12 @@
+manifests:
+  - name: dbt_loom_upstream_platform
+    type: file
+    config:
+      # Use environment variable for flexibility, with fallback to relative path
+      # path: '{{ env_var("PLATFORM_MANIFEST_PATH", "../dbt_loom_upstream_platform/target/manifest.json") }}'
+      # In production, set PLATFORM_MANIFEST_PATH to absolute path
+
+      # For GitHub Actions Integration Tests CI run, set the path to the manifest.json file
+      path: /home/runner/work/astronomer-cosmos/astronomer-cosmos/dev/dags/dbt/dbt_loom_upstream_platform/target/manifest.json
```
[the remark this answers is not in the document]
pankajkoti: No, I want it to keep this as is -- for CI runs to succeed

cosmos/dbt/graph.py
```diff
@@ -315,14 +315,24 @@ def parse_dbt_ls_output(project_path: Path | None, ls_stdout: str) -> dict[str,
                 project_path.parent / node_dict["package_name"] if node_dict.get("package_name") else project_path  # type: ignore
             )
 
+            # dbt-core defined the node path via "original_file_path", dbt fusion identifies it via "path"
+            # External nodes (e.g., from dbt-loom) may not have a file path - skip them
+            # Only skip if it looks like a valid node (has unique_id) but no file path
```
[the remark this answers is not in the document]
pankajkoti: I think it's okay to have those as separate lines instead of making one big line.

pankajkoti:
Thanks for the detailed review @tatiana . I have addressed the feedback.

> 1. Inside the `dev/dags/dbt` folder, put both new dbt projects into the folder `cross-project`. There, I suggest one of the projects is named upstream (or parent) and the other is named downstream (or child), so it is very clear why we have those projects

Done so, thanks! 

> 2. Please, could you also confirm that Airflow Assets/Datasets emitted via Cosmos by the upstream DAG could trigger the downstream DAG?

Yes, verified this, works smoothly. Added snapshots in the docs as we discussed earlier today.

> 3. Please, could you confirm if the syntax is exactly the same as dbt Mesh, and if there are any other lacking feaeture parities

Yes, confiremd that the syntax is exactly the same as dbt Mesh.

> 4. Were you able to test the cross-project reference, considering users were using `LoadMode.DBT_MANIFEST` both on the upstream and downstream projects?

Yes, I tested this with both projects using `LoadMode.DBT_MANIFEST` and it works well. Also have added an example DAG for the same to be run in our CI.

> 5. Would it be worth highlighting in our example or in the docs the fact that both projects can be run with different profiles?

Yes. highlighted this in our docs now

> 6. Is there a way of running nodes of different projects in the same Cosmos `DbtTaskGroup`?

No, this is not currently supported. Each DbtTaskGroup (or DbtDag) is configured with a single ProjectConfig that points to one dbt project.
When Cosmos parses a dbt-loom project, it encounters external nodes injected from upstream projects. However, Cosmos intentionally skips these external nodes during DAG generation -- they don't have local file paths and are meant to be run by their own project. Cosmos only creates Airflow tasks for the models that belong to the current project. Supporting multiple projects in a single DbtTaskGroup would require significant changes, I believe.

Requesting re-review, please!

dev/dags/dbt/cross_project/downstream/models/fct_revenue.sql
```diff
@@ -0,0 +1,34 @@
...
+    p.category as product_category,
+    count(distinct o.order_id) as num_orders,
+    sum(oi.quantity) as units_sold,
+    sum(oi.line_total) as gross_revenue,
+    sum(oi.quantity * pr.cost_price) as cost_of_goods_sold,
+    sum(oi.line_total) - sum(oi.quantity * pr.cost_price) as gross_profit
+from orders o
+inner join order_items oi on o.order_id = oi.order_id
+inner join products pr on oi.product_id = pr.product_id
+cross join (select 'All' as category) p  -- Simplified for demo
```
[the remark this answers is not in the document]
pankajkoti: Thanks, addressed to remove the cross join with ALL.

cosmos/dbt/graph.py
```diff
@@ -315,14 +315,25 @@ def parse_dbt_ls_output(project_path: Path | None, ls_stdout: str) -> dict[str,
...
+            # Check for both None and empty string since dbt-loom may set either
+            node_file_path = node_dict.get("original_file_path") or node_dict.get("path")
+            if not node_file_path and node_dict.get("unique_id"):
+                logger.debug(
+                    "Skipping node `%s` because it has no file path (likely an external reference from dbt-loom or similar)",
```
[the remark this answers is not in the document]
pankajkoti: The current messaging is good.

dev/dags/dbt/cross_project/downstream/dbt_loom.config.yml
```diff
@@ -0,0 +1,14 @@
+manifests:
+  - name: upstream
+    type: file
+    config:
+      # Use environment variable for flexibility, with fallback to relative path
+      # path: '{{ env_var("PLATFORM_MANIFEST_PATH", "../upstream/target/manifest.json") }}'
+      # In production, set PLATFORM_MANIFEST_PATH to absolute path
+
+      # For GitHub Actions Integration Tests CI run, set the path to the manifest.json file
+      path: /home/runner/work/astronomer-cosmos/astronomer-cosmos/dev/dags/dbt/cross_project/upstream/target/manifest.json
```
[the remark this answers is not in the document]
pankajkoti: No, I want to keep as is for CI runs to work.

cosmos/dbt/graph.py
```diff
@@ -315,14 +315,25 @@ def parse_dbt_ls_output(project_path: Path | None, ls_stdout: str) -> dict[str,
                 project_path.parent / node_dict["package_name"] if node_dict.get("package_name") else project_path  # type: ignore
             )
 
+            # dbt-core defined the node path via "original_file_path", dbt fusion identifies it via "path"
+            # External nodes (e.g., from dbt-loom) may not have a file path - skip them
+            # Only skip if it looks like a valid node (has unique_id) but no file path
+            # Check for both None and empty string since dbt-loom may set either
+            node_file_path = node_dict.get("original_file_path") or node_dict.get("path")
```
tatiana, line 322: @pankajkoti, I have an impression that a few other dbt nodes may actually have this behaviour, including:
- Source nodes
- Metrics
- Macros
- Ephemeral models

Would there be another, safer way to determine whether the node came from dbt Loom? Do they by any chance inject another specific metadata?

tatiana: Maybe `meta. external_location`? or some other meta property?

pyproject.toml
```diff
@@ -36,7 +36,8 @@ dependencies = [
     "packaging>=22.0",
     "pydantic>=1.10.0",
     "virtualenv",
-    "openlineage-integration-common"  # Without this dependency, Airflow datasets/assets don't work
+    "openlineage-integration-common",  # Without this dependency, Airflow datasets/assets don't work
+    # "dbt-loom",
```
tatiana: Should we remove this line?

cosmos/dbt/graph.py
```diff
@@ -315,14 +315,25 @@ def parse_dbt_ls_output(project_path: Path | None, ls_stdout: str) -> dict[str,
...
+            if not node_file_path and node_dict.get("unique_id"):
```
tatiana: This needs to be tested - ideally, we'd rely on some other metadata - but - for now - this may be the safest way moving forward. It could be a good contribution to dbt-loom to add this as a metadata.
```suggestion
            if "Uses cross-project refs to upstream via dbt-loom" in node_dict.get("description", "") and (not node_file_path and node_dict.get("unique_id")):
```

tatiana (approved):
Hi @pankajkoti, this looks great, thank you very much!

I have one last concern, discussed in the thread https://github.com/astronomer/astronomer-cosmos/pull/2271/changes#r2736991273.

Once you address this and the checks are passing, please feel free to merge the PR.

cosmos/dbt/graph.py
pankajkoti: While this is currently not possible as dbt-loom does not provide an identifier, I have quickly logged an issue in our repo: https://github.com/astronomer/astronomer-cosmos/issues/2299 for now.

pyproject.toml
pankajkoti:
```suggestion
```

pankajkoti: removed, yes, thanks

cosmos/dbt/graph.py
pankajkoti: Thanks for raising this @tatiana.

I tested other dbt nodes and see that all of them have the original_file_path in the dbt ls output

Ephemeral models:
Image

exposures:
Image

seeds:
Image

snapshots:
Image

sources:
Image

Macros are not dbt resouces themselves and they are not listed as a line item in the dbt ls output. 

Essentially, all of the dbt resource types in https://docs.getdbt.com/reference/global-configs/resource-type need to be written in a file and based on above example, I believe, its relative path gets populated in the `original_file_path`

Based on our discussion yesterday, since dbt-loom only injects models, I have added a safety check for the resource_type=model before skipping such inferred external nodes.

cosmos/dbt/graph.py
tatiana: Thanks a lot, @pankajkoti , for making this safer by also adding the resource type check. I believe we're good to go

cosmos/dbt/graph.py
tatiana: Thanks a lot, @pankajkoti , for the deep analysis - I'm happy for us to go with the current implementation.
If this breaks some use-case, I believe the affected people will let us know - and we can find some solution 🤞

PR #2276, opened 2026-01-21
Drop support for Airflow < 2.10

jedcunningham:
## Description

Neither Runtime or Apache Airflow support 2.9 or earlier Airflow versions any longer, so we can drop support per our policy.

This lets us drop 2.6-2.9, and a number of special cases to support those versions.

## Related Issue(s)

None.

## Breaking Change?

🤷 

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,12 +2,12 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, drop_old_airflow_versions]
   # Also run on pull requests originating from forks. Although this is insecure by default, we need it to run
   # integration tests on forked PRs. As a guardrail, we’ve added an Authorize step to each job, which requires manually
   # approving the workflow run for each pushed commit. Approval only happens after a careful code review of the changes.
   pull_request_target:
-    branches: [main] # zizmor: ignore[dangerous-triggers]
+    branches: [main, drop_old_airflow_versions] # zizmor: ignore[dangerous-triggers]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}


tatiana:
Hi @jedcunningham, thanks a lot for working on this.

We will review and update Cosmos' policy towards Airflow compatibility in this document:
https://github.com/astronomer/astronomer-cosmos/blob/main/docs/compatibility-policy.rst#currently-supported-versions

I just checked the Astro Runtime lifecycle schedule. It seems we offer basic support to AF 2.7 until the end of this month and basic support to AF 2.9 until October 2026:
https://www.astronomer.io/docs/runtime/runtime-version-lifecycle-policy#astro-runtime-lifecycle-schedule
Screenshot 2026-01-22 at 14 06 59


From a maintenance cost perspective, I'd say it is low to continue support at 2.9 for a few months, but let's confirm with @yetudada, and we'll let you know if we need any adjustments to the PR's scope.

In the meantime, so we can see a relatively green CI, I was wondering if you could temporarily update the GH using something similar to: https://github.com/astronomer/astronomer-cosmos/pull/2161/changes/0fd4b54254edf87105b8d39856ae7970e156699b
It would be great to see a dedicated GH Job run all tests against this branch and confirm that the relevant tests pass.

jedcunningham:
@tatiana I think there is a bit of ambiguity now. Historically we had a single "support" date for Runtime, which was when we stopped releasing new images with bugfixes and CVE fixes, etc. Now that is "end of maintenance". There was no equivalent of the current "basic support".

Basic suppport: "Astronomer Support will continue to accept support tickets for an additional period after a Runtime version becomes unmaintained."

I'd argue cosmos should stop supporting a version when it hits "end of maintenance" instead of "end of basic support".

My 2c, let me know :)

tatiana:
@jedcunningham We are in a sort of delicate moment migrating some customers from older versions of Airflow, so we may decide to be more conservative on Cosmos than in the Airflow community / Astro Runtime - for now. But I agree with you, we should review and focus on what is actively developed and maintained. We'll make sure we involve the right stakeholders before deciding on this one

tatiana:
Hi @jedcunningham — thanks so much for putting this PR together, really appreciate the work you’ve put into it! 🙌

I spoke with @yetudada today, and we aligned on the following approach:
- Cosmos support for Airflow versions will be compatible at least through Astronomer’s “End of basic support.”
I'll update the project compatibility policy accordingly. 

Based on that, our current plan is:
- Cosmos 1.14.0 (currently targeted for March/April 2026) will drop support for Airflow 2.7
- The first minor/major Cosmos release after October 2026 will drop support for Airflow 2.9
And so on.

If you’re open to it, it would be great to adjust the scope of this PR to focus specifically on removing support for Airflow 2.7. If not, no worries at all — we can close this PR and open more targeted ones as we get closer to those milestones.

Thanks again!

jedcunningham:
No problem, opened #2286 to fix up the wording.

I'll also open another PR for the ones we can drop now.

PR #2278, opened 2026-01-22
Skip storing telemetry metadata when emission is disabled

pankajkoti:
The _store_dag_telemetry_metadata method in DbtToAirflowConverter was performing unnecessary processing even when telemetry was disabled. This included compressing metadata and storing it in DAG params, which adds overhead without benefit when telemetry collection is turned off.

The PR adds an early return check in `_store_dag_telemetry_metadata` to skip all processing when `should_emit()` returns False. This ensures telemetry-related operations only occur when telemetry is actually enabled.

related: https://github.com/astronomer/astronomer-cosmos/pull/2223
related: https://github.com/astronomer/astronomer-cosmos/issues/2109


tatiana (approved):

PR #1625, opened 2025-03-18
Render ephemeral type DBT model & no freshness DBT source with test as EmptyOperator

okayhooni:
## Description
- dbt models with `ephemeral` materialization type that serve only as CTEs should not be rendered as `DbtRunOperator` tasks, as they unnecessarily occupy Airflow worker slots, even for a short period.
- Updated the logic to render these ephemeral models as EmptyOperator tasks instead, ensuring they are processed quickly by the Airflow scheduler without being assigned to an Airflow worker.
- Similarly, DBT source nodes without a freshness check but with downstream tests are rendered as EmptyOperator to avoid unnecessary worker slot resource consumption.

## Breaking Change?
No

## Checklist
- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2279, opened 2026-01-22
Fix running empty models or ephemeral nodes in `ExecutionMode.WATCHER`

### LINKED ISSUE #2266: [Bug] Watcher mode sensor tasks timeout for ephemeral models without --debug
### Astronomer Cosmos Version

1.12.1a1

### dbt-core version

1.0.0.40.14

### Versions of dbt adapters

dbt-adapters==1.19.0
dbt-common==1.36.0
dbt-core==1.10.13
dbt-extractor==0.5.1
dbt-postgres==1.9.0
dbt-protos==1.0.382
dbt-redshift==1.9.0
dbt-semantic-interfaces==0.9.0

### LoadMode

AUTOMATIC

### ExecutionMode

WATCHER

### InvocationMode

None

### airflow version

3.0.6

### Operating System

Linux

_No response_

### Deployment

Amazon (AWS) MWAA

### Deployment details

_No response_

### Description

When using `ExecutionMode.WATCHER`, sensor tasks created for dbt ephemeral models will always timeout because the producer never pushes XCom for them. This is because dbt only emits `NodeFinished` events for ephemeral models at **debug log level**, which are not received by the producer's callback unless `--debug` is passed to dbt.

### Expected Behavior

Either:
- Ephemeral models should not generate watcher sensor tasks (since they're compiled inline as CTEs, not built as separate artifacts)
- OR the sensor should immediately succeed without waiting for XCom

### Actual Behavior

The watcher sensor task for the ephemeral model times out after 1 hour (default `timeout`) with:
```
AirflowSensorTimeout: Trigger timeout
```

### Root Cause Analysis

The issue is in how dbt emits events for ephemeral models vs. how the producer handles them:

1. **dbt's event emission:** For ephemeral models, dbt emits `NodeFinished` events (event code `Q025`) only at **debug log level**. From dbt's perspective, ephemeral models are "compiled" into referencing models, not "built" as separate nodes.

2. **Producer callback filtering:** In `watcher.py`, the producer only handles events with `name == "NodeFinished"`:
   ```python
   def _callback(event_message: EventMsg) -> None:
       name = event_message.info.name
       if name in {"MainReportVersion", "AdapterRegistered"}:
           self._handle_startup_event(event_message, startup_events)
       elif name == "NodeFinished":
           self._handle_node_finished(event_message, context)
   ```

3. **The gap:** Without `--debug`, dbt doesn't emit the `NodeFinished` event for ephemeral models at info level, so the producer's callback never fires → no XCom pushed → sensor waits forever.

### Relevant log output

```shell
No logs without --debug, the producer never mentions my ephemeral model. With --debug when it works we see the following

Finished running node model.analytics.my_ephemeral_model
...
Timing info for model.analytics.my_ephemeral_model (compile)
```

### How to reproduce

1. Create a dbt project with an ephemeral model:
   ```sql
   -- models/facts/my_ephemeral_model.sql
   {{ config(materialized='ephemeral') }}
   SELECT 1 as id
   ```

2. Create a DbtDag with `ExecutionMode.WATCHER`:
   ```python
   cosmos.DbtDag(
       dag_id="my_dag",
       execution_config=cosmos.ExecutionConfig(
           execution_mode=cosmos.ExecutionMode.WATCHER
       ),
       render_config=cosmos.RenderConfig(
           select=["models/facts"],
       ),
       # ...
   )
   ```

3. Run the DAG **without** `--debug` flag

### Workaround

Enable dbt's `--debug` flag:
```python
operator_args={
    "dbt_cmd_flags": ["--debug"],
}
```

**However**, this is not ideal because:
1. It significantly increases log verbosity
2. It's not documented as a requirement
3. Users shouldn't need to know dbt's internal event emission levels

### Related Issues and PRs

- **#1133** - [Feature Request] EmptyTask for ephemeral materializations in local mode
- **#1625** - PR that attempted to render ephemeral nodes as `EmptyOperator` (addresses the underlying issue)

PR #1625's approach seems correct - ephemeral models should be rendered as `EmptyOperator` since they don't produce build artifacts and there's nothing to "watch" for.

### Additional Context

The dbt event system uses different log levels for different materializations:
- **table/view/incremental:** `NodeFinished` emitted at info level
- **ephemeral:** `NodeFinished` emitted at debug level only

This is by design in dbt - ephemeral models are treated as "compilation steps" rather than "execution steps." Cosmos should align with this by not creating watcher tasks for ephemeral models.

### Contact Details

Not putting an email here without knowing it'll be shared publicly or not!

tatiana:
Avoid consumer tasks that hang indefinitely when using `ExecutionMode.WATCHER` when the associated dbt models are either ephemeral or consist of empty SQL models that are not run by dbt.

## Context

There are circumstances when there is a discrepant number of nodes in the output when we run `dbt ls` and `dbt build`, using the same selectors.

In the following example (`tests/sample/dbt_project_with_empty_model`), we can observe that `dbt ls` returned two models, while the `dbt build` returned a single one:
```
$ dbt ls
10:48:32  Running with dbt=1.11.2
10:48:32  Registered adapter: postgres=1.10.0
10:48:32  Unable to do partial parsing because saved manifest not found. Starting full parse.
10:48:32  Found 2 models, 464 macros
micro_dbt_project.add_row
micro_dbt_project.empty_model

$ dbt build
10:50:21  Running with dbt=1.11.2
10:50:21  Registered adapter: postgres=1.10.0
10:50:21  Found 2 models, 464 macros
10:50:21  
10:50:21  Concurrency: 4 threads (target='dev')
10:50:21  
10:50:21  1 of 1 START sql view model public.add_row ..................................... [RUN]
10:50:21  1 of 1 OK created sql view model public.add_row ................................ [CREATE VIEW in 0.06s]
10:50:21  
10:50:21  Finished running 1 view model in 0 hours 0 minutes and 0.20 seconds (0.20s).
10:50:21  
10:50:21  Completed successfully
10:50:21  
10:50:21  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 NO-OP=0 TOTAL=1
```

So far, we observed this happening in two scenarios:
1. Ephemeral nodes (https://github.com/astronomer/astronomer-cosmos/issues/2266)
2. If the dbt model is not executable (e.g. it is an empty SQL file), both the `dbt build` and the `dbt run` will not display it in their info logs.

Until Cosmos 1.12.1, Cosmos assumed these two commands would return the same number of nodes, and we implemented the `LoadMode.MANIFEST` assuming the same. In the case of `ExecutionMode.LOCAL`, this was not a big issue, because dbt does not run when we select the particular model it's excluding:

```
$  dbt build --select empty_model 
10:53:03  Running with dbt=1.11.2
10:53:03  Registered adapter: postgres=1.10.0
10:53:03  Found 2 models, 464 macros
10:53:03  Nothing to do. Try checking your model configs and model specification args
```

The downside in the case of `ExecutionMode.LOCAL`  is that we waste Airflow resources by potentially parsing a dbt project that wouldn't need to be parsed in those particular tasks. The PR #1625 aims to address this.

However, in the case of `ExecutionMode.WATCHER`, this became a big problem, as the behaviour caused consumer nodes representing ephemeral nodes or empty models to hang indefinitely after the producer task completed successfully. The producer task was not aware of them and would not populate XCom, whereas the consumer tasks would keep checking for updates.

Closes: https://github.com/astronomer/astronomer-cosmos/issues/2266
Closes: https://github.com/astronomer/oss-integrations-private/issues/315
Closes: https://astronomer.zendesk.com/agent/tickets/87180

## About the solution

Ideally, probably, we would know upfront which nodes `dbt build` decides to execute, and we would not render them as Airflow tasks. However, I do not believe this is a simple problem, since there may be other circumstances when `dbt build` skips nodes from being executed - and any custom logic we implement in Cosmos will be affected by changes dbt Core/Fusion implements upstream.

Therefore, it feels - for now - the safest solution is:
- Continue adding those nodes to Cosmos
- Mark them as successful, logging a specific message, if they were not actually run by the dbt command. This is identified by checking the `run_results.json` file.

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -203,18 +203,11 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
         status = event.get("status")
         reason = event.get("reason")
 
-        if status == "success":
-            try:
-                run_results_status = self._get_status_from_run_results(context["ti"], context)
-            except (KeyError, TypeError):
-                logger.debug("Could not fetch run results.")
-                run_results_status = None
-
-            if reason == "model_not_run" and run_results_status is None:
-                logger.info(
-                    "Model '%s' was skipped by the dbt command. This may happen if it is an ephemeral model or if the model sql file is empty.",
-                    self.model_unique_id,
-                )
+        if status == "success" and reason == "model_not_run":
+            logger.info(
+                "Model '%s' was skipped by the dbt command. This may happen if it is an ephemeral model or if the model sql file is empty.",
+                self.model_unique_id,
+            )
 
         if status != "failed":
             return
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -146,7 +146,10 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
         node_result = next((r for r in results if r.get("unique_id") == self.model_unique_id), None)
 
         if not node_result:  # pragma: no cover
-            logger.warning("No matching result found for unique_id '%s'", self.model_unique_id)
+            logger.warning(
+                "The dbt node with unique_id '%s' was not executed by the dbt command run in the producer task. This may happen if it is an ephemeral model or if the model sql file is empty.",
+                self.model_unique_id,
+            )
             return None
 
         logger.info("Node Info: %s", run_results_json)
@@ -198,10 +201,19 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
 
     def execute_complete(self, context: Context, event: dict[str, str]) -> None:
         status = event.get("status")
+        reason = event.get("reason")
+
+        if status == "success":
+            if reason == "model_not_run":
+                status = self._get_status_from_run_results(context["ti"], context)
+                logger.info(
+                    "Model '%s' was skipped by the dbt command. This may happen if it is an ephemeral model or if the model sql file is empty.",
+                    self.model_unique_id,
+                )
+
         if status != "failed":
             return
 
-        reason = event.get("reason")
         if reason == "model_failed":
             raise AirflowException(
                 f"dbt model '{self.model_unique_id}' failed. Review the producer task '{self.producer_task_id}' logs for details."
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -204,7 +204,12 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
         reason = event.get("reason")
 
         if status == "success":
-            run_results_status = self._get_status_from_run_results(context["ti"], context)
+            try:
+                run_results_status = self._get_status_from_run_results(context["ti"], context)
+            except KeyError:
+                logger.debug("Could not fetch run results.")
+                run_results_status = None
+
             if reason == "model_not_run" and run_results_status is None:
                 logger.info(
                     "Model '%s' was skipped by the dbt command. This may happen if it is an ephemeral model or if the model sql file is empty.",
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -205,11 +205,12 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
 
         if status == "success":
             if reason == "model_not_run":
-                status = self._get_status_from_run_results(context["ti"], context)
-                logger.info(
-                    "Model '%s' was skipped by the dbt command. This may happen if it is an ephemeral model or if the model sql file is empty.",
-                    self.model_unique_id,
-                )
+                run_results_status = self._get_status_from_run_results(context["ti"], context)
+                if run_results_status is None:
+                    logger.info(
+                        "Model '%s' was skipped by the dbt command. This may happen if it is an ephemeral model or if the model sql file is empty.",
+                        self.model_unique_id,
+                    )
 
         if status != "failed":
             return
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -204,13 +204,12 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
         reason = event.get("reason")
 
         if status == "success":
-            if reason == "model_not_run":
-                run_results_status = self._get_status_from_run_results(context["ti"], context)
-                if run_results_status is None:
-                    logger.info(
-                        "Model '%s' was skipped by the dbt command. This may happen if it is an ephemeral model or if the model sql file is empty.",
-                        self.model_unique_id,
-                    )
+            run_results_status = self._get_status_from_run_results(context["ti"], context)
+            if reason == "model_not_run" and run_results_status is None:
+                logger.info(
+                    "Model '%s' was skipped by the dbt command. This may happen if it is an ephemeral model or if the model sql file is empty.",
+                    self.model_unique_id,
+                )
 
         if status != "failed":
             return
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -206,7 +206,7 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
         if status == "success":
             try:
                 run_results_status = self._get_status_from_run_results(context["ti"], context)
-            except KeyError:
+            except (KeyError, TypeError):
                 logger.debug("Could not fetch run results.")
                 run_results_status = None
 
--- a/tests/operators/test_watcher.py
+++ b/tests/operators/test_watcher.py
@@ -1088,7 +1088,7 @@ def test_dbt_dag_with_watcher_and_empty_model(caplog):
         render_config=RenderConfig(emit_datasets=False, test_behavior=TestBehavior.NONE),
         operator_args={
             "trigger_rule": "all_success",
-            "execution_timeout": datetime.delta(seconds=120) if AIRFLOW_VERSION != Version("2.8") else None,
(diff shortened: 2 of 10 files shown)


cosmos/operators/_watcher/base.py
```diff
@@ -198,10 +201,19 @@ def execute(self, context: Context, **kwargs: Any) -> None:
 
     def execute_complete(self, context: Context, event: dict[str, str]) -> None:
         status = event.get("status")
+        reason = event.get("reason")
+
+        if status == "success":
+            if reason == "model_not_run":
+                status = self._get_status_from_run_results(context["ti"], context)
```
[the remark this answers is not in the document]
tatiana: Addressed in 86b0880c3925f33ad7cc9541c8d06b8fb1bd37b2

cosmos/operators/_watcher/base.py
```diff
@@ -198,10 +201,24 @@ def execute(self, context: Context, **kwargs: Any) -> None:
...
     def execute_complete(self, context: Context, event: dict[str, str]) -> None:
         status = event.get("status")
+        reason = event.get("reason")
+
+        if status == "success":
+            try:
+                run_results_status = self._get_status_from_run_results(context["ti"], context)
+            except (KeyError, TypeError):
+                logger.debug("Could not fetch run results.")
+                run_results_status = None
```
tatiana: We may decide that this part is not necessary - and that we are happy with the triggerer decision, without checking the `run_results.json`. The advantage of this simplification is that, without this part of the code, the solution should work for `ExecutionMode.WATCHER_KUBERNETES` as well.

pankajkoti: Yeah, I am not following well with this part of the code in execute_complete. The control would be returned from the triggerer to execute_complete, so additional checks if needed can happen in the triggerer I feel.

Also did not follow the sentence
"The advantage of this simplification is that, without this part of the code, the solution should work for ExecutionMode.WATCHER_KUBERNETES as well."

Could we please clarify which simplication we mean and how it helps the ExecutionMode.WATCHER_KUBERNETES with or without this code?

tatiana: @pankajkoti The code that triggers this analysis isn't aware of the Airflow context, so it wasn't a trivial change to invoke this method from there. It would require a bigger refactor.

I think my main concern in introducing this double verification that checks `run_results.json` was to validate, via a different path, that the task was not actually executed by dbt. I observed that when the problem happens, `run_results.json` does not reference the model that was "skipped".

The problem with using `run_results.json` is that this implementation is specific to `ExecutionMode.WATCHER`. The reason why this does not work with `ExecutionMode.WATCHER_KUBERNETES` is that the K8s pod currently does not inject the `run_results.json` into the XCom - it's an improvement for the future.

I'll remove this check for now, and we can review if it will be necessary in the future.

I tested triggering the DAG with and without this code section, so we should be good.

pankajkoti: yes, happy then if we could remove this for the time being and add later if need be.

pankajkoti (approved):
LGTM if https://github.com/astronomer/astronomer-cosmos/pull/2279/changes#r2717497880 can be considered and CI is happy. Thanks for fixing this!

cosmos/operators/_watcher/base.py
tatiana: Done in 0578b0f72b2d6f29ec1ca4c24697a6c16550f07f

tests/operators/test_watcher.py
```diff
@@ -1041,6 +1041,85 @@ def test_dbt_dag_with_watcher():
...
+        start_date=datetime(2023, 1, 1),
+        dag_id="watcher_dag",
+        execution_config=ExecutionConfig(
+            execution_mode=ExecutionMode.WATCHER,
+            invocation_mode=InvocationMode.DBT_RUNNER,
+        ),
+        render_config=RenderConfig(emit_datasets=False, test_behavior=TestBehavior.NONE),
+        operator_args={
+            "trigger_rule": "all_success",
+            "execution_timeout": datetime.delta(seconds=120) if AIRFLOW_VERSION != Version("2.8") else None,
```
tatiana:
```suggestion
            "execution_timeout": timedelta(seconds=120) if AIRFLOW_VERSION != Version("2.8") else None,
```

tatiana:
I just cut an alpha release from this branch:
https://github.com/astronomer/astronomer-cosmos/releases/tag/astronomer-cosmos-v1.13.0a2
https://pypi.org/project/astronomer-cosmos/1.13.0a2/

Miskho:
Hi, thank you for your great work @tatiana. 

However I am still currently encountering problems with this execution mode. I am experimenting on GCP Composer (`composer-3-airflow-3.1.0-build.7`) where I installed the latest astronomer-cosmos (1.13.0a2) version. My project is very small: 5 models, few MBs of data.

I am still continuously spammed by `warning` logs of format:
```
Triggerer's async thread was blocked for a while, likely due to the highly utilized environment....
```

and the number of blocking tasks keep increasing until the triggerer restarts. Everything is fine, until the DbtDag is scheduled again, until this behaviour keeps repeating.

Before these warnings and during the initial build of the project executed in the dbt_produced_watcher, I get few minutes of these `error` level logs
```
textPayload: "XCom not found dag_id=dbt run_id=scheduled__2026-01-27T10:00:00+00:00 task_id=dbt_producer_watcher key=nodefinished_model__redacted map_index=-1 detail={'detail': {'reason': 'not_found', 'message': "XCom with key='nodefinished_model__redacted' map_index=-1 not found for task 'dbt_producer_watcher' in DAG run 'scheduled__2026-01-27T10:00:00+00:00' of 'dbt'"}} status_code=404"
```

These logs occur only for the parent-less nodes (tasks). All other (dependent) models do not throw this error.
 It seems that the XCom is not populated ahead of time for those first tasks, but is correctly afterwards? Is this possible?
 
 Thank you very much.
 
 EDIT: Honestly its seems that ephemeral problems might still be the problem, when i swithced the downstream empemeral materialisations to views. The warning messages are no longer present, although the error ones are still present even in that scenario.

Issue #1978, opened 2025-09-16
Improve retries in ExecutionMode.WATCHER


pankajastro:
**Current Behavior**
When a user retries a Build task, the associated Watcher task does not automatically start watching the retried build. As a result, the Watcher either:

**Expected Behavior**
- Prevent users from retrying the Build task manually unless they restart the entire workflow, ensuring Watcher behavior remains predictable or
- Upon retrying a Build task, the Watcher task should either watch the new retry

tatiana:
@pankajastro @pankajastro We may be able to leverage:
- https://docs.getdbt.com/reference/commands/retry

If there is a way of clearing the state only of the relevant sensors and bringing them back to "sense"

tatiana:
Let's update the task description. Part of what is discussed here was addressed in https://github.com/astronomer/astronomer-cosmos/pull/2114.

In the ideal world, during retries, users would be able to:
- Retry the watcher producer task (using [dbt retry](https://docs.getdbt.com/reference/commands/retry), continuing from where the pipeline stopped
- The consumer tasks would be reset back to sensing changes from the producer, instead of using `ExecutionMode.LOCAL`

tatiana:
Related to discussion with customer: https://astronomer.slack.com/archives/C0607HRA2SC/p1765221813498099?thread_ts=1764677139.702229&cid=C0607HRA2SC

tatiana:
When we start working on this task, we may want to check https://github.com/astronomer/oss-integrations-private/issues/359 as well.

pankajkoti:
I have created a document to describe the approaches considered: https://docs.google.com/document/d/1r07ZkjKg84xsBhkJovApnDl0yMqky7w5_SHWXRUmeHM/edit?tab=t.0

pankajkoti:
Holding off the work here based on comment: https://github.com/astronomer/astronomer-cosmos/pull/2584#issuecomment-4407329907

pankajkoti:
The primary objective of this story is to do a PoC on implementing `dbt retry`, which has been validated with PR #2584. Since we decided not to take the PR to life until we find a reliable approach to generating the artefacts before and after retries that includes all the status and run details of all nodes in those, we're deferring the implementation of `dbt retry` in Cosmos.

With that, I'm closing the ticket, and we can re-open/revisit with a new ticket if we have a concrete need for it when the current retry story implementation in `main`/proposed `1.14.2` release is felt insufficient.

PR #2283, opened 2026-01-23
Allow watcher producer retries without erroring

### LINKED ISSUE #2282: [Improvement] Support retrying DbtTaskGroup at the task group level with Watcher
The retry behaviour in the Watcher mode is [detailed in our documentation](https://astronomer.github.io/astronomer-cosmos/getting_started/watcher-execution-mode.html#how-retries-work). When this was implemented, we expected customers to retry only the failed tasks by clearing those tasks (potentially automatically selecting downstream tasks), rather than clearing the entire TaskGroup. To be honest, I wasn’t aware that retrying at the TaskGroup level was an option in Airflow.

We should change the Cosmos implementation so that when the producer node is retried, it succeeds instead of erroring — even though it won’t perform any work during the retry. The consumer tasks would then be re-run as expected, executing their corresponding dbt commands. 

In the longer term, we’ll have a more robust retry solution once we implement [ticket #1978](https://github.com/astronomer/astronomer-cosmos/issues/1978). 

This would allow TaskGroup-level retries.

tatiana:
Aims to overcome retries at a TaskGroup level, is reported in: https://github.com/astronomer/astronomer-cosmos/issues/2282

During retries, the watcher will not execute any operation, but will not fail. We are intentionally changing the behaviour previously implemented in  #2114.

The behaviour will be improved once the following ticket is implemented: https://github.com/astronomer/astronomer-cosmos/issues/1978

Closes: #2282

### DIFF
--- a/docs/getting_started/watcher-execution-mode.rst
+++ b/docs/getting_started/watcher-execution-mode.rst
@@ -241,7 +241,9 @@ This behavior is designed to support TaskGroup-level retries, as reported in `#2
 
 - The producer task should still be configured with ``retries=0`` (which Cosmos enforces by default) to avoid unintended duplicate ``dbt build`` runs.
 
-This behavior will be further improved once `#1978 <https://github.com/astronomer/astronomer-cosmos/issues/1978>`_ is implemented.
+- By default, Cosmos sets ``retries`` to ``0`` in``DbtProducerWatcherOperator``. Users can override this by declaring the ``retries`` parameter in ``setup_operator_args``.
+
+The overall retry behavior will be further improved once `#1978 <https://github.com/astronomer/astronomer-cosmos/issues/1978>`_ is implemented.
 
 -------------------------------------------------------------------------------
 
--- a/docs/getting_started/watcher-execution-mode.rst
+++ b/docs/getting_started/watcher-execution-mode.rst
@@ -219,15 +219,29 @@ How retries work
 ~~~~~~~~~~~~~~~~
 
 When the ``dbt build`` command run by ``DbtProducerWatcherOperator`` fails, it will notify all the ``DbtConsumerWatcherSensor``.
-Cosmos always sets the producer's Airflow task retries to ``0``; this ensures the failure surfaces immediately and avoids kicking off a second full ``dbt build`` run.
 
 The individual watcher tasks, that subclass ``DbtConsumerWatcherSensor``, can retry the dbt command by themselves using the same behaviour as ``ExecutionMode.LOCAL``.
-This is also the reason why we set ``retries`` to ``0`` in the ``DbtProducerWatcherOperator`` task because rerunning the producer would repeat the full dbt build and duplicate
-watcher callbacks which may not be processed by the consumers if they have already processed output XCOMs from the first run of the producer.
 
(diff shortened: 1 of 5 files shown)


pankajkoti (approved):
LGTM. It could happen that users just clear the producer task and expect that it will be re-run. Considering that it would be important to add in our docs, that if they clear the producer task expecting everything to be re-run, they should also select all the **downstream tasks + recursive** as well while clearing the producer task.

tatiana:
Thanks a lot, @pankajkoti , please, could you review the docs - I changed in 5db165eb3ddab64c35caf9f4dad52557f5918d40

docs/getting_started/watcher-execution-mode.rst
```diff
@@ -219,15 +219,29 @@ How retries work
...
+.. versionadded:: 1.12.2
+
+When the ``DbtProducerWatcherOperator`` is triggered for a retry (try_number > 1), it will not re-run the dbt build command and will succeed. In previous versions of Cosmos, the producer task would fail during retries.
+This behavior is designed to support TaskGroup-level retries, as reported in `#2282 <https://github.com/astronomer/astronomer-cosmos/issues/2282>`_.
+
+**Why this matters:**
+
+- In earlier versions, attempting to retry the producer task would raise an ``AirflowException``, causing the retry to fail immediately.
+- Now, the producer gracefully skips execution on retries, logging an informational message explaining that the retry was skipped to avoid running a second ``dbt build``.
+- This allows users to retry entire TaskGroups without the producer task blocking the retry flow.
```
pankajkoti:
```suggestion
- This allows users to retry entire TaskGroups and/or DAGs without the producer task blocking the retry flow.
```
something like this to also cover when DAGs are cleared.

docs/getting_started/watcher-execution-mode.rst
```diff
@@ -219,15 +219,29 @@ How retries work
 ~~~~~~~~~~~~~~~~
 
 When the ``dbt build`` command run by ``DbtProducerWatcherOperator`` fails, it will notify all the ``DbtConsumerWatcherSensor``.
-Cosmos always sets the producer's Airflow task retries to ``0``; this ensures the failure surfaces immediately and avoids kicking off a second full ``dbt build`` run.
 
 The individual watcher tasks, that subclass ``DbtConsumerWatcherSensor``, can retry the dbt command by themselves using the same behaviour as ``ExecutionMode.LOCAL``.
-This is also the reason why we set ``retries`` to ``0`` in the ``DbtProducerWatcherOperator`` task because rerunning the producer would repeat the full dbt build and duplicate
-watcher callbacks which may not be processed by the consumers if they have already processed output XCOMs from the first run of the producer.
```
pankajkoti, line 226: I am thinking more here.We still set https://github.com/astronomer/astronomer-cosmos/blob/1d608c7ea93c139518e8f89a68ca040e32fd48af/cosmos/operators/watcher.py#L98 retries to 0 for the producer tasks. And probably we want to continue doing that. Because we don't want to retry ourselves when the producers fail, but are okay to retry when the user has explicitly triggered a retry or cleared the DAG/TaskGroup.

I think it maybe nice to still reflect this setting of default retries to 0  in our docs.

pankajkoti (approved):
LGTM. Thanks a lot for considering the comments!

PR #2287, opened 2026-01-26
Add integration tests using `InvocationMode.SUBPROCESS` and validate output

tatiana:
Between Cosmos 1.11 and Cosmos 1.12, the Cosmos `ExecutionMode.WATCHER changed its logging behaviour to always output JSON, which is a much worse experience for end-users.

We had a contribution from @tiovader to fix this issue (#2241), and we will release it in Cosmos 1.13, planned for this week.

To avoid regressions in how Cosmos outputs user-friendly logs, this PR extends our integration tests to cover this behaviour and further improves the implementation.

PR #2241, opened 2025-12-30
Restore plain text output when using `ExecutionMode.WATCHER`

tiovader:
## Description

This PR refactors dbt log processing in **WATCHER** mode to improve readability at task logs. It reverts the JSON log output introduced in #2152 back to plain text, while still preserving real-time consumer updates when using `SUBPROCESS` invocation mode.

Key changes include:

* Refactoring `FullOutputSubprocessHook` to ensure log outputs are processed by the `process_log_line` callable when provided.
* Introducing a new `_process_json_log_line` function, which consolidates the previous `_store_dbt_resource_status_from_log` logic and adds logging behavior:
  * Properly handles both JSON and non-JSON log lines.
  * Extracts dbt logging info from structured JSON output, handling dynamic log levels if needed.
  * Extracts dbt node execution status to XCom, preserving the original `_store_dbt_resource_status_from_log` behavior.
* Updating `DbtProducerWatcherOperator` to conditionally set `log_format="json"` and the log line processing callable only when running in `SUBPROCESS` invocation mode, avoiding unintended side effects in other execution paths.

## SUBPROCESS
As I've observed it works fine and even outputs CLI colors at airflow logs

image

```python
# [START example_watcher_subprocess]
example_watcher_subprocess = DbtDag(
    # dbt/cosmos-specific parameters
    execution_config=ExecutionConfig(execution_mode=ExecutionMode.WATCHER, invocation_mode=InvocationMode.SUBPROCESS),
    project_config=ProjectConfig(DBT_PROJECT_PATH),
    profile_config=profile_config,
    render_config=RenderConfig(exclude=["raw_payments"]),
    operator_args=operator_args,
    # normal dag parameters
    schedule="@daily",
    start_date=datetime(2023, 1, 1),
    catchup=False,
    dag_id="example_watcher_subprocess",
    default_args={"retries": 0},
)
# [END example_watcher_subprocess]
```

## DBT_RUNNER
works fine as previous version.

image

```python
# [START example_watcher]
example_watcher = DbtDag(
    # dbt/cosmos-specific parameters
    execution_config=ExecutionConfig(execution_mode=ExecutionMode.WATCHER, invocation_mode=InvocationMode.DBT_RUNNER),
    project_config=ProjectConfig(DBT_PROJECT_PATH),
    profile_config=profile_config,
    render_config=RenderConfig(exclude=["raw_payments"]),
    operator_args=operator_args,
    # normal dag parameters
    schedule="@daily",
[...]

PR #2301, opened 2026-01-29
Fix duplicate log lines in watcher subprocess execution and format timestamps

pankajkoti:
Changes:
- Remove duplicate logging of dbt messages in store_dbt_resource_status_from_log(). The message was being logged twice: once in the else block and again at the end of the function. Now it's only logged once with the appropriate log level.
- Format timestamps to match dbt runner format (HH:MM:SS) instead of full ISO format. This provides consistent log output between subprocess and dbt_runner modes. Example: "13:16:05  Running with dbt=1.10.11"

- Add unit tests to verify:
  - Log messages are logged exactly once (no duplicates)
  - Timestamps are formatted correctly as HH:MM:SS

- Extend integration tests (test_dbt_dag_with_watcher, test_dbt_dag_with_watcher_and_subprocess) to verify no duplicate log messages in both dbt_runner and subprocess modes

related: #2287
related: #2241

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -1,6 +1,6 @@
 import json
 import logging
-from datetime import timedelta
+from datetime import datetime, timedelta
 from typing import Any
 
 from airflow.exceptions import AirflowException
@@ -42,8 +42,6 @@ def store_dbt_resource_status_from_log(line: str, extra_kwargs: Any) -> None:
         log_line = {}
     else:
         logger.debug("Log line: %s", log_line)
-        if "info" in log_line and "msg" in log_line["info"]:
-            logger.info(log_line["info"]["msg"])
         node_info = log_line.get("data", {}).get("node_info", {})
         node_status = node_info.get("node_status")
         unique_id = node_info.get("unique_id")
@@ -60,8 +58,19 @@ def store_dbt_resource_status_from_log(line: str, extra_kwargs: Any) -> None:
     log_info = log_line.get("info", {})
     msg = log_info.get("msg")
     level = log_info.get("level", "INFO").upper()
+    ts = log_info.get("ts")
     if msg is not None:
-        logger.log(getattr(logging, level, logging.INFO), msg)
+        if ts:
+            # Format timestamp to match dbt runner format (HH:MM:SS)
+            try:
+                # Parse ISO format timestamp (e.g., "2025-01-29T13:16:05.123456Z")
+                dt = datetime.fromisoformat(ts.replace("Z", "+00:00"))
+                formatted_ts = dt.strftime("%H:%M:%S")
+            except (ValueError, AttributeError):
+                formatted_ts = ts
+            logger.log(getattr(logging, level, logging.INFO), "%s  %s", formatted_ts, msg)
+        else:
+            logger.log(getattr(logging, level, logging.INFO), msg)
 
 
 class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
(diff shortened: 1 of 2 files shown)


cosmos/operators/_watcher/base.py
```diff
@@ -42,8 +42,6 @@ def store_dbt_resource_status_from_log(line: str, extra_kwargs: Any) -> None:
         log_line = {}
     else:
         logger.debug("Log line: %s", log_line)
-        if "info" in log_line and "msg" in log_line["info"]:
-            logger.info(log_line["info"]["msg"])
```
pankajkoti, line 46: This was causing duplicate logs and is a good choice to remove than the one below since the one that logs below logs it with the appropriate dbt log level.

tatiana (approved):
Thanks a lot, @pankajkoti. It was probably my mistake during rebasing.

Looks great! I do not believe we need timestamp parity. From my perspective, this makes our codebase more complex without any clear benefit, because we can already see the Airflow timestamp when we're invoking from Airflow. If there was an easy way, I'd remove that timestamp from the dbtRunner implementation 😂 

That said, I believe the duplicate issue is the biggest problem and I'm happy for things to be the same, if you prefer - we'll just need to add tests to cover the additional logic.

pankajkoti:
Previously, we were seeing duplicate logs in the Watcher Subprocess execution
Screenshot 2026-01-29 at 7 27 19 PM


Now, after the change in this PR, we have only one log, and it also includes the timestamp to be consistent with the dbt Runner logs output

Subprocess output:
Screenshot 2026-01-29 at 7 27 07 PM


dbt Runner output:
Screenshot 2026-01-29 at 7 27 00 PM

cosmos/operators/_watcher/base.py
```diff
@@ -60,8 +58,19 @@ def store_dbt_resource_status_from_log(line: str, extra_kwargs: Any) -> None:
...
+            # Format timestamp to match dbt runner format (HH:MM:SS)
+            try:
+                # Parse ISO format timestamp (e.g., "2025-01-29T13:16:05.123456Z")
+                dt = datetime.fromisoformat(ts.replace("Z", "+00:00"))
+                formatted_ts = dt.strftime("%H:%M:%S")
+            except (ValueError, AttributeError):
+                formatted_ts = ts
+            logger.log(getattr(logging, level, logging.INFO), "%s  %s", formatted_ts, msg)
+        else:
+            logger.log(getattr(logging, level, logging.INFO), msg)
```
tatiana, line 73: @pankajkoti  I suggest we do not put effort into adding the timestamp, it does not give us much value - it's just more code & tests for us to maintain. If we want to keep it, though, let's add unit tests

pankajkoti:
> Looks great! I do not believe we need timestamp parity. From my perspective, this makes our codebase more complex without any clear benefit, because we can already see the Airflow timestamp when we're invoking from Airflow. If there was an easy way, I'd remove that timestamp from the dbtRunner implementation 😂
> 
> That said, I believe the duplicate issue is the biggest problem and I'm happy for things to be the same, if you prefer - we'll just need to add tests to cover the additional logic.

Yes, I am a bit torn on this one. I am slightly inclined to preserve the timestamp for the following reasons:
1. Only watcher subprocess logs do not have a timestamp value. Other execution modes' subprocess runs still include the timestamp
2. This timestamp is the real timestamp given by dbt event. Airflow logs might have slight delays (at times, maybe?) and reorderings of logs are possible at a time, based on how the flushing works(?).
3. If users have a regex configured for filtering/processing logs forwarded to their log store, the consistency in logs could still benefit, I believe.

Hence, I am thinking of preserving this for now. But if we feel it is a maintenance burden happy for it to be removed in a subsequent PR.

cosmos/operators/_watcher/base.py
pankajkoti: responded in https://github.com/astronomer/astronomer-cosmos/pull/2301#issuecomment-3818825442

PR #2307, opened 2026-01-30
Support defining custom callbacks alongside the `WATCHER_KUBERNETES` callback

johnhoran:
## Description

I am also setting callbacks on the watcher producer, and the code as it is tries to set them exclusively.  So just checking to see if the task already has some and adding the watcher pattern to the callback if it does.

## Checklist


tatiana:
Thanks a lot for this improvement, @johnhoran !

PR #2309, opened 2026-01-30
Refactor to avoid potential future `UnboundLocalError` for `producer_task` in `calculate_tasks_map`

RinZ27:
Fixed a potential scope issue where `producer_task` could be accessed before initialization in the `calculate_tasks_map` function. While I was digging through the watcher implementation, it became clear that the variable was only defined within a specific branch but referenced later in a separate block. Initializing it at the top and simplifying the downstream check makes the code much more robust against accidental crashes.

Updating the logic to verify the presence of the task object itself instead of re-checking the execution mode keeps the flow consistent. This ensures the dependency setup only triggers when the producer is actually available.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,19 +1,63 @@
 Changelog
 =========
 
-1.13.0a2 (2026-01-22)
+1.13.0 (2026-01-30)
 ---------------------
-
-(Includes more changes)
 
 Features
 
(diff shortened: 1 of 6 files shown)


tatiana (approved):
Thanks for improving the code, @RinZ27 ! Refactors like this improve code maintainability and help avoid issues in the future. We'll release this as part of the next Cosmos release, very likely 1.13.1.

RinZ27:
Happy to contribute to the project and glad to hear it'll be in the 1.13.1 release. @tatiana

PR #2293, opened 2026-01-27
Add cluster policy for cosmos

astro-anand:
## Description

Adding a cluster policy that support retries on a different worker queue for watcher sensor tasks. This does not currently work due to a circular import issue when applying the cluster policy. 

## Breaking Change?

No

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

## current issues

Circular import - I think the cluster policy may need to be packaged separately

```
Traceback (most recent call last):
  File "/usr/local/bin/airflow", line 5, in <module>
    from airflow.__main__ import main
  File "/usr/local/lib/python3.12/site-packages/airflow/__init__.py", line 74, in <module>
    settings.initialize()
  File "/usr/local/lib/python3.12/site-packages/airflow/settings.py", line 781, in initialize
    load_policy_plugins(POLICY_PLUGIN_MANAGER)
  File "/usr/local/lib/python3.12/site-packages/airflow/settings.py", line 200, in load_policy_plugins
    pm.load_setuptools_entrypoints("airflow.policy")
  File "/usr/local/lib/python3.12/site-packages/pluggy/_manager.py", line 416, in load_setuptools_entrypoints
    plugin = ep.load()
             ^^^^^^^^^
  File "/usr/local/lib/python3.12/importlib/metadata/__init__.py", line 205, in load
    module = import_module(match.group('module'))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/airflow/astronomer_cosmos/cosmos/__init__.py", line 17, in <module>
    from cosmos.airflow.dag import DbtDag
  File "/usr/local/airflow/astronomer_cosmos/cosmos/airflow/dag.py", line 11, in <module>
    from cosmos.converter import DbtToAirflowConverter, airflow_kwargs, specific_kwargs
  File "/usr/local/airflow/astronomer_cosmos/cosmos/converter.py", line 35, in <module>
    from cosmos.listeners.task_instance_listener import _get_profile_config_attribute
  File "/usr/local/airflow/astronomer_cosmos/cosmos/listeners/task_instance_listener.py", line 15, in <module>
[...]

PR #2315, opened 2026-02-02
Document the cluster policy config for ExecutionMode.WATCHER sensor task

pankajastro:
Adds documentation for the `watcher_retry_queue` Airflow config to route WATCHER consumer sensor retries onto a different worker queue with more resources.

**Changes:**
- Document `watcher_retry_queue` behavior and configuration in the WATCHER execution mode guide.
- Add `watcher_retry_queue` to the Cosmos configuration reference, including env var and default value.

Related PR: https://github.com/astronomer/astronomer-cosmos/pull/2293


docs/configuration/cosmos-conf.rst
```diff
@@ -261,6 +261,20 @@ This page lists all available Airflow configurations that affect ``astronomer-co
         :start-after: [START cosmos_init_imports]
         :end-before: [END cosmos_init_imports]
 
+.. _watcher_retry_queue:
+
+`watcher_retry_queue`_:
```
tatiana:
```suggestion
`watcher_dbt_execution_queue`_:
```
I've been thinking about this, and I believe we should rename this configuration to `watcher_dbt_execution_queue`. The advantage is that, if this is defined, we could set it directly for both the producer task and the consumers during retry. WDYT? I understand this change is outside of the scope of the docs, but it may be worth to do it in a separate PR and make this PR consistent with this change.

docs/configuration/cosmos-conf.rst
```diff
@@ -261,6 +261,20 @@ This page lists all available Airflow configurations that affect ``astronomer-co
...
+    (Introduced in Cosmos 1.14.0): In watcher execution mode, when the producer watcher task fails, the consumer tasks run the individual models on retry.
+    Since these tasks are sensors that require low memory/CPU on their first try, this setting allows retries to run on a queue with larger resources,
+    which is often necessary for larger dbt projects. When a watcher sensor task is retried (try_number >= 2 in Airflow 2.x, or try_number >= 1 in Airflow 3.x),
```
tatiana: Considering we change the behaviour, as suggested in https://github.com/astronomer/astronomer-cosmos/pull/2315/changes#r2764587891, this would also need to be rephrased:
```suggestion
    (Introduced in Cosmos 1.14.0) When using watcher execution mode, tasks may need to run dbt or not, depending on their type (producer vs. consumer) and the retry number. When running the dbt command, tasks use more resources (CPU and memory) than when behaving as sensors. The computational cost of running these tasks can vary widely. For example, a Cosmos watcher sensor consumes approximately 200MB, compared to 700MB consumed by a dbt build task running a project with almost 200 dbt models. This configuration allows users to define which queue to use when dbt commands are run, optimising their Airflow deployment. Internally, Cosmos leverages the [Airflow cluster policy feature](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/cluster-policies.html). As of now, this configuration will be used:
    - for watcher producer tasks, during their first execution
    - for watcher consumer tasks, from their first retry onwards
```

docs/configuration/cosmos-conf.rst
```diff
@@ -261,6 +261,20 @@ This page lists all available Airflow configurations that affect ``astronomer-co
...
+    it will automatically be assigned to the specified queue.
+
+    This behavior is enforced by Cosmos via an Airflow policy (``task_instance_mutation_hook``) that mutates ``task_instance.queue`` at runtime for retry attempts.
+    As a result, the configured ``watcher_retry_queue`` can overwrite any queue set directly on the operator, but only for retries; the initial run continues to use the operator's original queue.
+
```
tatiana, line 293: I believe we sohuld detail how this configuration is implemented  in the watcher-execution-mode dedicated page, instead of here, where people have an overview of the existing configuration.

docs/getting_started/watcher-execution-mode.rst
```diff
@@ -246,6 +246,40 @@ This behavior is designed to support TaskGroup-level retries, as reported in `#2
 
 The overall retry behavior will be further improved once `#1978 <https://github.com/astronomer/astronomer-cosmos/issues/1978>`_ is implemented.
 
+**Watcher retry queue**
+
+.. versionadded:: 1.14.0
+
+In watcher execution mode, consumer sensor tasks are lightweight sensors that wait for the producer task to complete. On their first attempt, they require minimal CPU and memory resources. However, when these tasks retry, they execute the dbt command for the node, which may require significantly more resources.
```
tatiana: Maybe we should include here the information I suggested in https://github.com/astronomer/astronomer-cosmos/pull/2315/changes#r2764702735, with concrete examples of memory utilisation:
```suggestion
In watcher execution mode, by default, consumer sensor tasks are lightweight sensors that wait for the producer task to complete. On their first attempt, they require minimal CPU and memory resources. However, when these tasks retry, they execute the dbt command for the node, which may require significantly more resources.
```

tatiana (commented):
@pankajastro, while reviewing the docs, I noticed we can significantly improve the feature. I suggest we improve the feature before publishing any documentation, and only add the documentation after we review the feature's behaviour.

The main feature changes I'm thinking about are:
- Rename `watcher_retry_queue` to `watcher_dbt_execution_queue`
- Use this setting not only for the consumer, during retries, but also to set the default node pool for the producer, if defined

This woudl alllow users to be able to use their "default" queue for running watcher and only about setting the "dbt execution" node queue in a single place - and Cosmos would handle which tasks need to use it, regardless if they are consumer or producer tasks.

pankajastro:
> @pankajastro, while reviewing the docs, I noticed we can significantly improve the feature. I suggest we improve the feature before publishing any documentation, and only add the documentation after we review the feature's behaviour.
> 
> The main feature changes I'm thinking about are:
> 
> * Rename `watcher_retry_queue` to `watcher_dbt_execution_queue`
> * Use this setting not only for the consumer, during retries, but also to set the default node pool for the producer, if defined
> 
> This woudl alllow users to be able to use their "default" queue for running watcher and only about setting the "dbt execution" node queue in a single place - and Cosmos would handle which tasks need to use it, regardless if they are consumer or producer tasks.

Created PR: https://github.com/astronomer/astronomer-cosmos/pull/2331 to implement the suggestions

docs/configuration/cosmos-conf.rst
pankajastro: Fixed in https://github.com/astronomer/astronomer-cosmos/pull/2331

docs/configuration/cosmos-conf.rst
pankajastro: I have updated docs watcher-execution-mode dedicated page PTAL

tatiana (approved):
Thanks a lot for improving the behaviour and updating the documentation in this area, @pankajastro . This is a very important feature that will further support the adoption of `ExecutionMode.WATCHER`

Issue #2257, opened 2026-01-08
[Feature] Support use of YAML selectors when using LoadMode.DBT_MANIFEST


YourRoyalLinus:
### Description

Support the use of selectors from a `selectors.yaml` file when using `LoadMode.DBT_MANIFEST`. 

### Use case/motivation

At my company, we recently migrated to using Cosmos and Airflow (Astronomer hosted) for orchestration of our dbt jobs. We were previously using FiveTran transformations (lol). 

To avoid dependency conflicts, our Airflow and dbt project repos were kept separate, and we deployed the latest dbt project to our Astronomer prod environment after each successful merge to main in our dbt repo. We are using `selectors.yaml` to serve as the bridge between our dbt jobs in Airflow and the underlying dbt workflows. Changes to selectors in our dbt project affect which models get rendered as tasks in our dbt dags. This required the use of `LoadMode.DBT_LS` and reliance on the `dbt_ls_cache` to keep render performance stable.

Our current development workflow has us pushing new changes to our dbt repo many times a day, which resulted in frequent invalidations of the `dbt_ls_cache`. With only 50 total dags, around half of which were dbt dags, we required a Large scheduler to handle the high dag parse cpu requirements when the `dbt_ls_cache` was invalidated and all of our dbt dags needed to be fully reparsed at once. We tweaked our Airflow configuration to alleviate the issue, but it reached a point where we were uncomfortable tweaking further due to the potential impact on our end users.

As a workaround, I implemented behavior to allow us to parse yaml selectors into their equivalent `select` and `exclude` components as outlined here https://docs.getdbt.com/reference/node-selection/yaml-selectors. This behavior was easy to implement into the existing Cosmos logic in a fairly non-invasive manner. I have a PR with a working implementation that I can submit and link to this issue. 

If the process to make contributions to this project goes well, I have been given approval to spend additional time to work on contributing additional improvements to Cosmos that were are either currently maintaining or have planned that I think would benefit others.

### Related issues

This appears to be a separate a follow up to this issue: https://github.com/astronomer/astronomer-cosmos/issues/718, which was closed by this PR: https://github.com/astronomer/astronomer-cosmos/pull/755.

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

YourRoyalLinus:
[...]

PR #2316, opened 2026-02-02
Error handle invalid YAML with `LoadMode.DBT_MANIFEST` and `RenderConfig.selector`

### LINKED ISSUE #2312: [Feature] Isolate YamlSelector parse errors
### Description

Currently a single yaml selector parsing error will cause all DAGs using `LoadMode.DBT_MANIFEST` with `RenderConfig.selector` to fail to parse. It is possible to prevent one misconfigured selector from failing all related DAG parses.

### Use case/motivation

Instead of raising an exception during parsing, parser errors can be captured and a Cosmos exception can be raised when they're accessed. This isolates DAG parsing failures only to DAGs attempting to use the invalid selector. This greatly improves UX, especially when non-engineers are changing selector definitions and what is a valid `selector.yml` file does not always translate to a valid Cosmos DbtGraph.

### Related issues

https://github.com/astronomer/astronomer-cosmos/issues/2257

YourRoyalLinus:
## Description

 Ensure that DAG parsing only fails for DAGs that are attempting to render using a misconfigured yaml selector.

This was done by collecting any errors during yaml selector parsing and raising an exception when the selector is accessed rather than when the issue is encountered. 

The exception to this is for invalid yaml structures. In those cases, we can't be certain whether we have a valid selector name or definition. When the graph attempts to grab the parsed selector, it would always get a selector not found exception, hiding the error messages related to the invalid structure. 

As a result, if there is an invalid yaml structure, all DAGs using `LoadMode.DBT_MANIFEST` and `RenderConfig.selector` will fail during dag parse. As long as a user never directly modifies the `manifest.json`, this should not occur.

When doing local QA, I created separate example DAGs, however keeping them did not seem valuable as the unit tests cover the expected behavior.

Follow up from #2257

## Related Issue(s)
closes #2312 

## Breaking Change?
N/A

## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,0 @@
-@astronomer/astro-cosmos-admins @astronomer/astro-cosmos-codeowners
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Validate workflow files
-        uses: reviewdog/action-actionlint@f4bb4d50c4c9e45cc64f413d2bc438ea1cf2d9d5 # v1.70.0
+        uses: reviewdog/action-actionlint@0d952c597ef8459f634d7145b0b044a9699e5e43 # v1.71.0
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           reporter: github-check
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -503,6 +503,7 @@ jobs:
           SNOWFLAKE_SCHEMA: ${{ secrets.SNOWFLAKE_SCHEMA }}
           SNOWFLAKE_WAREHOUSE: ${{ secrets.SNOWFLAKE_WAREHOUSE }}
           SNOWFLAKE_DATABASE: ${{ secrets.SNOWFLAKE_DATABASE }}
+          AIRFLOW_CONN_GCP_GS_CONN: ${{ secrets.AIRFLOW_CONN_GCP_GS_CONN }}
           RESOURCE_PREFIX: ${{ steps.set-resource-prefix.outputs.prefix }}
 
       - name: Upload coverage to Github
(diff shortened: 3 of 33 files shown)


YourRoyalLinus:
I also noticed a bug that prevents the cache from invalidating when a DAG initially swaps `LoadModes` from `DBT_LS` to `DBT_MANIFEST or vice versa. The fix is extremely small (8 simple loc + 2 new unit tests). 

Would you prefer a separate issue and PR for tracking or is there a threshold that's small enough to lump those changes into this PR? My preference is to separate them, for many reasons, but I don't want to pile on new issues and PRs unnecessarily.

cosmos/dbt/selector.py
```diff
@@ -774,11 +774,25 @@ def get_parsed(self, selector_name: str, default: Any = None) -> dict[str, Any]
...
+        if parsed_selector and isinstance(parsed_selector, dict):
+            errors = parsed_selector.get("errors", [])
+
+            if errors:
+                if len(errors) == 1:
+                    error_message = f"Error parsing selector '{selector_name}': {errors[0]}"
+                else:
+                    error_list = "\n  - ".join(errors)
+                    error_message = f"Errors parsing selector '{selector_name}':\n  - {error_list}"
+                raise CosmosValueError(error_message)
```
tatiana, line 787: We can simplify this:
```suggestion
            if errors:
                error_list = "\n  - ".join(errors)
                error_message = f"Error(s) parsing selector '{selector_name}':\n  - {error_list}"
                raise CosmosValueError(error_message)
```

tatiana:
Hi @YourRoyalLinus — thanks so much for following up and further improving the solution! The implementation itself, along with the documentation and tests, is very thorough.

> I also noticed a bug that prevents the cache from invalidating when a DAG initially swaps `LoadModes` from `DBT_LS` to `DBT_MANIFEST or vice versa. The fix is extremely small (8 simple loc + 2 new unit tests).
> 
> Would you prefer a separate issue and PR for tracking, or is there a threshold that's small enough to lump those changes into this PR? My preference is to separate them, for many reasons, but I don't want to pile on new issues and PRs unnecessarily.

A separate PR for this fix would be great.

If you could address the minor comment I left and rebase this PR, I'm happy to approve it once the checks pass. We can release this as part of the next Cosmos release, likely 1.13.1 or 1.14.0.

tatiana:
I'm merging since the check that failed is unrelated to the changes in this PR, thanks a lot, @YourRoyalLinus !

Issue #2311, opened 2026-01-30
[Feature] Watcher to support AFTER_EACH test


michal-mrazek:
### Description

Hi,
would it be possible to support AFTER_EACH test behaviour for WATCHER mode? same way as running models. Test could push a record to XCOM. 

In current scenario, we dont specify test_behavior. When test fails, dbt model run is green as well as test task is green. This is cconfusing for end users.

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

michal-mrazek:
I believe that DbtTestWatcherOperator should inherit from DbtConsumerWatcherSensor. And then if we push to xcom all states as mentioned https://github.com/astronomer/astronomer-cosmos/issues/2291, we would also get a xcom for test nodes. 

Let me know if I can help with this :)

tatiana:
This was completed in https://github.com/astronomer/astronomer-cosmos/pull/2447

Issue #2291, opened 2026-01-27
[Feature] Consider all node status.


maoxingda:
### Description

https://github.com/astronomer/astronomer-cosmos/blob/3355910a6c1faa10625a7db3721a6cf6960facb3/cosmos/operators/_watcher/base.py#L49

<img width="1462" height="944" alt="Image" src="https://github.com/user-attachments/assets/78872e22-8452-486e-b10c-68b54ee9b3bb" />

### Use case/motivation

**sql query timeout.**

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

tatiana:
Thanks a lot for pointing this, @maoxingda ! Which status did you face issues with Cosmos Watcher mode? Would you by any chance like to contribute this fix?

tatiana:
I believe we'll need to change more parts of the code to fix this, such as:
https://github.com/astronomer/astronomer-cosmos/blob/3355910a6c1faa10625a7db3721a6cf6960facb3/cosmos/operators/_watcher/triggerer.py#L133-L166

We'll aim to fix this in Cosmos 1.13.1.

maoxingda:
**As shown in the screenshot, the `node_status` changes over time in the query timeout scenario.**

<img width="1460" height="948" alt="Image" src="https://github.com/user-attachments/assets/b7ec27c0-d8a2-4a59-92ed-8f2b57687a36" />

maoxingda:
> Thanks a lot for pointing this, [@maoxingda](https://github.com/maoxingda) ! Which status did you face issues with Cosmos Watcher mode? Would you by any chance like to contribute this fix?

I also don’t really know how to comprehensively handle the various `node_status` values provided by dbt.

tatiana:
This is a duplicate: https://github.com/astronomer/astronomer-cosmos/issues/2181 - we'll be updating the progress in the other ticket

Issue #2181, opened 2025-12-03
Track all state in real-time consumer for ExecutionMode.WATCHER & InvocationMode.SUBPROCESS


pankajastro:
The recent PR [#2152](https://github.com/astronomer/astronomer-cosmos/pull/2152)
 introduced real-time consumer updates when using ExecutionMode.WATCHER and InvocationMode.SUBPROCESS. Currently, the consumer tracks only FAILED and SUCCESS step states.

Goal:
Enhance the real-time tracking to include the STARTED state as well. This improvement will provide:
- Better visibility into ongoing steps
- A more user-friendly experience for monitoring execution progress

https://github.com/dbt-labs/dbt-core/blob/019ebdcc98a7a39551bdea5d8f0a4e633529291d/core/dbt/artifacts/schemas/results.py#L58-L67

tatiana:
@pankajastro are there any other states we are not currently tracking?
Are we currently printing the logs in the dbt logs related to that dbt model in the consumer tasks? If not, it may be an opportunity to improve this.

pankajastro:
> [@pankajastro](https://github.com/pankajastro) are there any other states we are not currently tracking? Are we currently printing the logs in the dbt logs related to that dbt model in the consumer tasks? If not, it may be an opportunity to improve this.

I think possible values are coming from:

https://github.com/dbt-labs/dbt-core/blob/eba90863ed4043957330ea44ca267db1a2d81fcd/core/dbt/contracts/results.py#L75-L88

Yes, the idea is to log them in the watcher trigger once we have those state stores in XCom

PR #2318, opened 2026-02-03
feat: Add watcher mode support for dbt test node states

michal-mrazek:
## Description

When using ExecutionMode.WATCHER, dbt test nodes now have their results aggregated per parent model and pushed as a single XCom value ("pass" or "fail"), enabling downstream consumer sensors to observe the test outcome for each model. This lays the groundwork for a future DbtTestConsumerWatcherSensor.

I also introduced some helpful function for handling node status. This can later be easily enhanced. 

I removed using hardcoded string in TriggerEvent - this was prone for typos.  

I think that documentation should be changed in following PR - where we change the behaviour of test tasks. For now, there is no change for end users.

Future work:
- Implement the TestOperator to read the xcom
- also implement for k8 execution mode

## Related Issue(s)

 Related to: #2311 

## Breaking Change?

No

## Checklist

### DIFF
--- a/tests/operators/test_watcher.py
+++ b/tests/operators/test_watcher.py
@@ -604,7 +604,89 @@ class TestStoreDbtStatusFromLog:
 
         assert ti.store.get("test__pkg__my_test_status") == "fail"
 
-    def test_store_dbt_resource_status_from_log_handles_invalid_json(self, caplog):
+    def test_store_dbt_resource_status_from_log_aggregates_test_results_when_tests_per_model_provided(self):
+        """When tests_per_model is non-empty and a test node finishes, the function should
+        accumulate results and push a single aggregated *_tests_status XCom once all tests
+        for the model have reported — instead of pushing individual *_status keys per test.
+        """
+        ti = _MockTI()
+        ctx = {"ti": ti}
+
+        tests_per_model = {
+            "model.pkg.orders": ["test.pkg.not_null_orders_id", "test.pkg.unique_orders_id"],
+        }
+        test_results_per_model: dict[str, list[str]] = {}
+
+        # First test passes — not all tests reported yet, no XCom push
+        log_line_1 = json.dumps(
+            {
+                "data": {
+                    "node_info": {
+                        "node_status": "pass",
+                        "unique_id": "test.pkg.not_null_orders_id",
+                        "resource_type": "test",
+                    }
+                }
+            }
+        )
+        store_dbt_resource_status_from_log(
+            log_line_1,
+            {"context": ctx},
+            tests_per_model=tests_per_model,
+            test_results_per_model=test_results_per_model,
+        )
+        assert "test__pkg__not_null_orders_id_status" not in ti.store  # no per-test key
+        assert "model__pkg__orders_tests_status" not in ti.store  # not yet aggregated
+
+        # Second test passes — all tests reported, aggregated XCom should be pushed
+        log_line_2 = json.dumps(
+            {
+                "data": {
+                    "node_info": {
+                        "node_status": "pass",
+                        "unique_id": "test.pkg.unique_orders_id",
+                        "resource_type": "test",
+                    }
+                }
+            }
+        )
+        store_dbt_resource_status_from_log(
+            log_line_2,
+            {"context": ctx},
+            tests_per_model=tests_per_model,
+            test_results_per_model=test_results_per_model,
+        )
+        assert "test__pkg__unique_orders_id_status" not in ti.store  # no per-test key
+        assert ti.store.get("model__pkg__orders_tests_status") == "pass"
+
+    def test_store_dbt_resource_status_from_log_aggregates_fail_when_any_test_fails(self):
+        """When at least one test fails, the aggregated status should be 'fail'."""
+        ti = _MockTI()
+        ctx = {"ti": ti}
+
+        tests_per_model = {
+            "model.pkg.orders": ["test.pkg.not_null_orders_id", "test.pkg.unique_orders_id"],
+        }
+        test_results_per_model: dict[str, list[str]] = {}
+
+        for uid, status in [
+            ("test.pkg.not_null_orders_id", "pass"),
+            ("test.pkg.unique_orders_id", "fail"),
+        ]:
+            log_line = json.dumps(
+                {"data": {"node_info": {"node_status": status, "unique_id": uid, "resource_type": "test"}}}
+            )
+            store_dbt_resource_status_from_log(
+                log_line,
+                {"context": ctx},
+                tests_per_model=tests_per_model,
+                test_results_per_model=test_results_per_model,
+            )
+
+        assert ti.store.get("model__pkg__orders_tests_status") == "fail"
+        # No per-test status keys should exist
+        assert "test__pkg__not_null_orders_id_status" not in ti.store
+        assert "test__pkg__unique_orders_id_status" not in ti.store
         """Test that invalid JSON doesn't raise an exception."""
         ti = _MockTI()
         ctx = {"ti": ti}


cosmos/operators/_watcher/state.py
```diff
@@ -14,6 +14,25 @@
...
+
+def is_node_status_failed(status: str | None) -> bool:
+    """Check if the dbt node status indicates failure (works for both models and tests)."""
+    return status in FAILED_STATUSES
+
+
+def is_node_status_terminal(status: str | None) -> bool:
+    """Check if the dbt node status is terminal (success or failed)."""
+    return is_node_status_success(status) or is_node_status_failed(status)
+
```
tatiana: Excellent improvement, @michal-mrazek! It may be good to specify if these functions relate to Airflow or dbt states.

cosmos/operators/_watcher/base.py
```diff
@@ -48,8 +54,9 @@ def store_dbt_resource_status_from_log(line: str, extra_kwargs: Any) -> None:
 
         logger.debug("Model: %s is in %s state", unique_id, node_status)
 
-        # TODO: Handle and store all possible node statuses, not just the current success and failed
-        if node_status in ["success", "failed"]:
+        # Handle terminal statuses for both models (success/failed) and tests (pass/fail)
+        # TODO: handle all possible statuses including skipped, warn, etc.
+        if is_node_status_terminal(node_status):
```
tatiana: This is a step forward with https://github.com/astronomer/astronomer-cosmos/issues/2291 🎉

cosmos/operators/_watcher/state.py
```diff
@@ -14,6 +14,25 @@
 
 ProducerStateFetcher = Callable[[], str | None]
 
+# dbt uses different status values for different node types (models/tests):"
+SUCCESS_STATUSES = frozenset({"success", "pass"})
+FAILED_STATUSES = frozenset({"failed", "fail", "error"})
```
tatiana: It may be good for us to name these constants in a way we know if if these are dbt or Airflow status states.

cosmos/operators/_watcher/triggerer.py
```diff
@@ -136,34 +141,34 @@ async def run(self) -> AsyncIterator[TriggerEvent]:
         while True:
             producer_task_state = await self._get_producer_task_status()
```
tatiana, line 152: IIR, this returns the Airflow task state, and not necessarily the dbt task state. Could you confirm, please?

cosmos/operators/_watcher/state.py
michal-mrazek: yes, agree, will rename :)

cosmos/operators/_watcher/state.py
michal-mrazek: Agree, lets change it :)

cosmos/operators/_watcher/triggerer.py
michal-mrazek: Yes, this is airflow state. I renamed all variables that operate with dbt state - added the "dbt" prefix for better clarity.

michal-mrazek:
@tatiana, during testing subprocess/dbt_runner, I realized that there is one caveat. 

There can be multiple tests (so dbt nodes) for one model. Moreover, the test node name is more complex. So to make the AFTER_EACH work lately we have to map airflow task to multiple dbt test nodes:
1. Aggregate during the xcom push
2. Aggregate during the xcom pull

I tend to incline to option 2. But since this is architecture decision, I would be curious about your opinion :) 

This is not strictly related to this PR, but we have to keep this in mind. But yeah, works with both subprocess/dbt_runner :) 

Screenshot 2026-02-05 at 10 49 34 PM

michal-mrazek:
@tatiana, once we merge this, I will start working on the test operator for after_each. I think it makes sense, we can reuse a lot of code, we will just have to pull for multiple xcoms :)

michal-mrazek:
@tatiana @pankajkoti Can you pls check if you have some time? :) I will then follow up with the next PR :)

tatiana:
Hi @michal-mrazek,

Thanks again for all the work here - I agree the refactor to clean up the states, and the follow-up changes to ensure each test actually runs (rather than being an empty task) are both necessary and valuable.

There is one aspect of the current implementation that concerns me, though. As it stands, the producer task pushes `XCom`s not only for every `model`, `seed`, and `snapshot`, but also one `XCom` per test - as you pointed out earlier in the PR discussion.

From the dbt projects I’ve seen, it’s common to have tens or hundreds of models, and in some cases even thousands. Each model often has 3–10 tests. With this change, instead of roughly n*`XCom`s, we could end up with up to 10×n `XCom`s per DAG run - a significant increase in cardinality, which will increase the database load.

Given that:
- The producer runs in a single process, and
- Typicaly there is a single test task consuming all tests for a given model,
I strongly suggest grouping each model’s tests into a single XCom, as you proposed in the PR discussion.

For example, in the `InvocationMode.SUBPROCESS` case, `store_dbt_resource_status_from_log` could be made aware of the node’s `resource_type` and - somewhere in the code - we'd aggregate test results per model into a single XCom entry. While this would make the producer logic slightly more complex, it would:
- Significantly reduce database load and
- Simplify the consumer logic.

We already store the dbt graph topology in the `dbt_graph` attribute of `DbtDag` or `DbtTaskGroup`. During producer task initialisation, we could process this structure into something like:
```
{model_name: [<list of test names>]}
```

This would make it straightforward to group and push test results as a single XCom per model.

Would you be able to adjust the implementation in this direction? I believe it would give us a cleaner and more scalable solution.

One way for us to get this PR merged sooner than later would be, for now, to just do the relevant checks and restrict emitting XCom to `model`, `seed`, and `snapshot`. And, in a follow-up PR, tackle the test nodes' complexity.

michal-mrazek:
@tatiana thanks, yes, I agree with this :) If we aggregate the test results before pushing to xcom, the following implementation of test operators should be simple. Good catch with the xcom overload :) I will try to push the changes in the following days, I think we can handle the tests xcom here

cosmos/operators/_watcher/base.py
```diff
@@ -43,16 +49,19 @@ def store_dbt_resource_status_from_log(line: str, extra_kwargs: Any) -> None:
...
         unique_id = node_info.get("unique_id")
 
-        logger.debug("Model: %s is in %s state", unique_id, node_status)
+        logger.debug("Model: %s is in %s state", unique_id, dbt_node_status)
 
-        # TODO: Handle and store all possible node statuses, not just the current success and failed
-        if node_status in ["success", "failed"]:
+        # Handle terminal statuses for both models (success/failed) and tests (pass/fail)
+        # TODO: handle all possible statuses including skipped, warn, etc.
+        if is_dbt_node_status_terminal(dbt_node_status):
```
pankajastro, line 133: Is it possible to track all node states here? I'm working on https://github.com/astronomer/astronomer-cosmos/issues/2072, and we want to display more real-time information in the Consumer task log. I previously submitted a PR https://github.com/astronomer/astronomer-cosmos/pull/2392, but it was closed in favour of this PR.

michal-mrazek: What dbt statuses should also be included pls? :)

pankajastro: probably https://github.com/dbt-labs/dbt-core/blob/019ebdcc98a7a39551bdea5d8f0a4e633529291d/core/dbt/artifacts/schemas/results.py#L52-L67

michal-mrazek:
Added the dbt test aggregation before pushing to xcom. This was a bit more complex than I expected :) 
Screenshot 2026-02-23 at 2 08 30 PM

We need to pass the tests_per_model but also keep the state of existing tests (but be aware of multithreading, so lock was introduced). I will change the PR description.

Both subprocess and dbt_runner were tested - k8 should follow in another PR, I feel like this PR is already growing a lot.

cosmos/airflow/graph.py
```diff
@@ -678,12 +678,15 @@ def _add_watcher_producer_task(
...
     render_config: RenderConfig | None = None,
     execution_mode: ExecutionMode = ExecutionMode.WATCHER,
+    tests_per_model: dict[str, list[str]] | None = None,
 ) -> BaseOperator:
     """
     Create the producer task for the watcher execution mode and add it to the tasks_map.
     The producer task is the task that will be used to produce the events for the watcher execution mode.
     """
     producer_task_args = task_args.copy()
+    if tests_per_model is not None:
```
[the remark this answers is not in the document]
tatiana: @michal-mrazek, we'll need to make sure this works with `watcher_kubernetes` as well before releasing 1.14 - but I'd be happy for us to address this in a separate PR, if you could commit to helping on this

tatiana:
Thanks a lot for advancing the implementation further, @michal-mrazek!

I just tried out commit 0826b5bf90aaa6b5b18ee00cc00e9247ef1f9dc7, and I noticed that we're still generating one XCom per test - could you adjust this, please?
Screenshot 2026-02-24 at 13 45 48

I used the  DAG that's in the repo `example_watcher.py` and `airflow standalone` to try it out.

michal-mrazek:
@tatiana That is strange 🤔 
```
git log --oneline -5
0826b5bf (HEAD -> feat/watcher-test-operator-support, origin/feat/watcher-test-operator-support) tests
5f016214 comments
9f5332f9 handle the dbt test aggregation
bda49418 rename all dbt node variables for cleaner code
da0ac024 feat: Add watcher mode support for dbt test node states
```

Screenshot 2026-02-24 at 4 58 56 PM
I just pulled the branch and run via docker compose you guys provide 🤔

michal-mrazek:
@tatiana I did not manage to match your results with this branch. I specifically desinged tests for this case https://github.com/astronomer/astronomer-cosmos/pull/2318/changes#diff-b6f62de0effcda8f032c7203047c74394bfd2ee93ab65392462714a9910dbdfeR687 Can you please pull the branch and try it from scratch? 🙏

cosmos/operators/_watcher/base.py
tatiana: @pankajastro, as we agreed, you'll be working on this once this PR is merged as part of https://github.com/astronomer/astronomer-cosmos/issues/2181

tatiana:
Thanks a lot, @michal-mrazek , I'm not sure about what happened. Thanks for the amazing work here. I created a new Airflow deployment locally, and I was able to confirm the expected behaviour 🎉 

Screenshot 2026-03-03 at 13 11 06

cosmos/airflow/graph.py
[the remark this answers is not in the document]
michal-mrazek: Created an issue so we dont forget it :) https://github.com/astronomer/astronomer-cosmos/issues/2427

PR #2320, opened 2026-02-04
[Experimental feature] enable orjson parser for whole project

corsettigyg:
## Description

At GetYourGuide, as we started having more and more cosmos dags, we noticed that - even though we use the dbt manifest parsing method - the total dag parsing time started increasing remarkably, which started affecting our metrics.

On a further investigation, the main offender is that the size of our manifest (which is used for many different dags) increased over 30 mb which makes the parsing a somehow heavy process, especially when we have over 50+ cosmos dags running, some of them with hundreds of models.

To work around that, I started playing with the [orjson](https://github.com/ijl/orjson) library, which is a rust-based json library for python. They advertise being much faster and "more correct" than the standard json library. Using our manifest, I executed a benchmark on it and got the following results

Screenshot 2026-02-04 at 16 49 43

which basically showed an improvement of 40% on the parsing time.

## Breaking Change?

Should not be right now as it is disabled by default.

## Checklist

### DIFF
--- a/docs/guides/run_dbt/airflow-worker/watcher-execution-mode.rst
+++ b/docs/guides/run_dbt/airflow-worker/watcher-execution-mode.rst
@@ -341,7 +341,16 @@ Test behavior
 
 By default, the watcher mode runs tests alongside models via the ``dbt build`` command being executed by the producer ``DbtProducerWatcherOperator`` operator.
 
-As a starting point, this execution mode does not support the ``TestBehavior.AFTER_EACH`` behavior, since the tests are not run as individual tasks. Since this is the default ``TestBehavior`` in Cosmos, we are injecting ``EmptyOperator`` as a starting point to ensure a seamless transition to the new mode.
+.. versionchanged:: 1.14.0
+
+Starting with Cosmos 1.14.0, ``TestBehavior.AFTER_EACH`` is fully supported in ``ExecutionMode.WATCHER``.
+Each test task is rendered as a ``DbtTestWatcherOperator`` (a ``DbtConsumerWatcherSensor`` subclass) that watches
+the aggregated test results published by the producer via XCom. This means test tasks now behave as real sensors
+rather than no-op placeholders.
+
+In Cosmos versions prior to 1.14.0, ``TestBehavior.AFTER_EACH`` was not supported by the watcher mode because tests
+were not run as individual tasks. Since ``TestBehavior.AFTER_EACH`` is the default ``TestBehavior`` in Cosmos,
+``EmptyOperator`` tasks were injected as placeholders to ensure a seamless transition to the new mode.
 
 The ``TestBehavior.BUILD`` behavior is embedded in the producer ``DbtProducerWatcherOperator`` operator.
 
--- a/docs/guides/run_dbt/container/watcher-kubernetes-execution-mode.rst
+++ b/docs/guides/run_dbt/container/watcher-kubernetes-execution-mode.rst
@@ -165,7 +165,7 @@ The following limitations from ``ExecutionMode.WATCHER`` also apply to ``Executi
 
 * **Individual dbt Operators**: Only ``DbtSeedWatcherKubernetesOperator``, ``DbtSnapshotWatcherKubernetesOperator``, and ``DbtRunWatcherKubernetesOperator`` are implemented. The ``DbtTestWatcherKubernetesOperator`` is currently a placeholder.
 
-* **Test behavior**: The ``TestBehavior.AFTER_EACH`` is not supported. Tests are run as part of the ``dbt build`` command by the producer task.
+* **Test behavior**: Unlike ``ExecutionMode.WATCHER`` (which fully supports ``TestBehavior.AFTER_EACH`` since Cosmos 1.14.0), ``ExecutionMode.WATCHER_KUBERNETES`` does not yet support ``TestBehavior.AFTER_EACH``. Tests are run as part of the ``dbt build`` command by the producer task, and test tasks are rendered as ``EmptyOperator`` placeholders. This is tracked in `#1974 <https://github.com/astronomer/astronomer-cosmos/issues/1974>`_.
 
 * **Source freshness nodes**: The ``dbt build`` command does not run source freshness checks.
 


tatiana:
This is super exciting, @corsettigyg ! If this were the default, do you have any concerns regarding backwards compatibility - other than the dependency on this Rust JSON parser?

corsettigyg:
hey @tatiana 😎  honestly, I would argue it is a pretty safe change in general. I have tested it on airflow 2.9 & airflow 2.11 and did not see any issues. The orjson does behave almost exactly like the classic json package. Of course, I came across orjson by chance so cannot say much of the project, but in the local sandbox I have been playing around with it, it has been working exactly the same with very good performances. I'd say to keep it as an experimental feature and if feedback is positive, use it as the only json parser from cosmos 2.0

On my local airflow, I have been testing this branch on our cosmos dags and could not see any issues. Only observation is that the scheduler is suffering less to parse them hahaha

obs: to keep complexity <10, had to create the `_load_manifest_from_file` function

tatiana:
@corsettigyg Sorry for commenting while this task is still in draft, but I had an idea for how we could expose this feature - and I thought it was good to share it early in the implementation.

At the moment, it seems we're importing the `json` library in a few places in the Cosmos codebase:
```
$ grep -rn "import json" cosmos
cosmos/operators/local.py:5:import json
cosmos/operators/watcher.py:4:import json
cosmos/operators/_watcher/triggerer.py:5:import json
cosmos/operators/_watcher/base.py:1:import json
cosmos/dbt/graph.py:7:import json
cosmos/plugin/airflow3.py:4:import json
cosmos/cache.py:5:import json
cosmos/telemetry.py:3:import json
cosmos/profiles/bigquery/service_account_keyfile_dict.py:5:import json
cosmos/profiles/snowflake/user_pass.py:5:import json
cosmos/profiles/snowflake/user_encrypted_privatekey_env_variable.py:5:import json
cosmos/profiles/snowflake/user_privatekey.py:5:import json
cosmos/profiles/snowflake/user_encrypted_privatekey_file.py:5:import json
cosmos/profiles/base.py:9:import json
cosmos/listeners/dag_run_listener.py:5:import json
```

What if we had a `cosmos/dependencies.py` or something similar, where we did:
```
try:
    import orjson as json
except ImportError:  # pragma: no cover
    import json
```

And we change the 15 import paths to import json to be `from cosmos. dependencies import json`?

And we did not have yet-another-configuration in Cosmos to set this up? This would allow us to remove:
- the new config
- the new installation plugin

This way, people could benefit from this optimisation - in all parts of Cosmos - by just installing [orjson](https://github.com/ijl/orjson).

corsettigyg:
hey @tatiana , dont worry, your observations are always more than welcome even if it is a draft 🚀  

This is a very good take actually. I will try to do something like that, so the overall performance impact can reverberate across the whole codebase. orjson has some small peculiarities though that do not make it into an one to one equivalent of the standard json package though, but it is worth the try. will try to do it next week

corsettigyg:
@tatiana I tried creating a global enabler for the orjson across the project. Some parts might not be that elegant and I had to use AI to come up with a solution in some components I am not familar with, but based on some basic tests I did, it is functional. 
I was wondering if it would be a good idea to enable the flag to `true` by default , run the integration tests & unit tests in the cicd with it, then put it back to `false` again.  This way, it is an extra layer of validation.

I created a simple benchmark with claude to test some areas touched by the change to have a global idea of performance enhancement. 
```
  1. Manifest parsing        – cosmos/dbt/graph.py  load_from_dbt_manifest()
  2. dbt ls line parsing     – cosmos/dbt/graph.py  run_command_with_dbt_runner()
  3. Compress + serialize    – cosmos/operators/local.py  store_compiled_sql() / watcher events
  4. Decompress + deserialize– cosmos/dbt/graph.py  _load_dbt_ls_cache_from_disk()
```
The results are pretty good

```
========================================================================
  Generating ~2 MB manifest ...
  Generated 4.1 MB  (886 nodes)  in 0.1s
  Running 5 iterations per scenario
========================================================================
  [Manifest parse        ] json= 101.1ms  orjson=  40.9ms  (2.5x)
  [dbt ls line parse     ] json=  28.9ms  orjson=  12.5ms  (2.3x)
  [Compress+serialize    ] json= 199.0ms  orjson=  36.8ms  (5.4x)
  [Decompress+deser.     ] json= 104.2ms  orjson=  49.1ms  (2.1x)

========================================================================
  Generating ~10 MB manifest ...
  Generated 23.4 MB  (5594 nodes)  in 0.4s
  Running 5 iterations per scenario
========================================================================
  [Manifest parse        ] json= 600.7ms  orjson= 287.3ms  (2.1x)
  [dbt ls line parse     ] json=  29.1ms  orjson=  29.5ms  (1.0x)
  [Compress+serialize    ] json=   1.17s  orjson= 215.7ms  (5.4x)
  [Decompress+deser.     ] json= 624.5ms  orjson= 299.2ms  (2.1x)

========================================================================
  Generating ~50 MB manifest ...
  Generated 120.4 MB  (29133 nodes)  in 2.2s
  Running 5 iterations per scenario
========================================================================
  [Manifest parse        ] json= 829.9ms  orjson= 340.5ms  (2.4x)
  [dbt ls line parse     ] json=   5.0ms  orjson=   2.2ms  (2.2x)
  [Compress+serialize    ] json=   1.44s  orjson=   1.05s  (1.4x)
  [Decompress+deser.     ] json=   1.01s  orjson= 393.9ms  (2.6x)

========================================================================
  Generating ~100 MB manifest ...
  Generated 241.7 MB  (58557 nodes)  in 4.7s
  Running 5 iterations per scenario
========================================================================
  [Manifest parse        ] json=   1.71s  orjson= 650.8ms  (2.6x)
  [dbt ls line parse     ] json=   5.4ms  orjson=  42.1ms  (0.1x)
  [Compress+serialize    ] json=   2.91s  orjson=   2.13s  (1.4x)
  [Decompress+deser.     ] json=   2.02s  orjson=   1.02s  (2.0x)
```

For now, I will enable it for review, but I am more than welcome to receive feedback on the implementation or changes I'd have to do 😄

corsettigyg:
@tatiana so I have rebased and fixed the issues in the code. To keep both json and orjson, I had to do some hacks here and there but overall the tests are passing.  Not a beautiful solution code-wise and it would be much easier to just fully replace to orjson instead of keeping this switch OR just use orjson where it would in fact provide a performance boost, so I am open for suggestions, if any 😄

evanvolgas:
Wow! This is a good idea @corsettigyg and I can tell how much work you put into this. Really appreciate that :) 

I can also confirm that currently JSON parsing is slower than ideal (we are seeing ~2 seconds for our files) and in ways that we've had to hack around. orjson would materially make a difference. 

How can I help on this PR?

evanvolgas:
A few additional thoughts on this PR, if I may:

  1. `dump()` silently drops `kwargs` when orjson is active (Copilot noticed this too). If someone calls `json.dump(..., default=my_serializer)`, orjson will ignore default without warning. The loads/dumps* functions fall back to stdlib when kwargs is non-empty; dump() should do the same.
  2. `dbt_ls_cache_key_args` changes separators — the PR quietly changes `json.dumps_str(env_vars, sort_keys=True)` to add `separators=(",", ":")`. This alters the cache key format for existing users, invalidating their `dbt ls` cache on upgrade. Minor, but worth a note in the changelog.
  3. The `enable_orjson_parser` setting path — it fails loudly with ImportError at call time if orjson isn't installed, which is good. But users who set the flag and then pip install astronomer-cosmos (without the extra) will only get the error during DAG parsing, not at startup.

corsettigyg:
hey @evanvolgas , I'd say that for now, the problem I see in this PR is that there is too much complexity on trying to use orjson everywhere while keeping the old json as an alternative/backup. I was actually considering to close this PR and re-open as the initial simple switch I idealised. If you have suggestions on how to improve the process, feel free to open PRs on my fork and they will end up reflected here.

after playing with the implementation for some time however, my 2 cents here is: **implement orjson only where it will give results and keep json on the other places**. In many places, orjson has some limitations and has almost no value. the unit tests & integration tests are strong enough to deal with the internal switch without introducing it as experimental IMO. 

Once the maintainers have time to check it,I'd be willing to hear the desired direction here, since creating a complex switch with so many edge cases feels overly complex, which is why I stopped working on it for the time being

tatiana:
Hi @corsettigyg,

Apologies for the delay in getting back to you on this PR.

We've had a huge amount of demand recently, including work not directly visible here, and we're a bit short on hands. I completely agree that extending the usage of orjson to all areas made the code more complex than intended - I regret suggesting it in the first place.

I support closing this PR and opening a new one scoped only to `manifest.json`.

Additionally, could you share some numbers on how this change affects parsing `manifest.json`? Specifically, it would be helpful to see:

* The size of the file (if you have multiple manifest.json files, 2–3 examples would be great)
* The time Cosmos takes to parse the project using `json` vs `orjson`, based on the DAG processor logs, similar to:
```
Cosmos performance (<cache_id>) - [<hostname>|<pid>]: It took 0.068s to parse the dbt
```

Once we have these numbers alongside the original change, I’m happy to move forward and commit to releasing it as part of Cosmos 1.5.0.

Thanks again for your patience and for helping improve this!

corsettigyg:
@tatiana sure, opened a new PR for it 😄

PR #2288, opened 2026-01-26
Drop support for Airflow under 2.9

jedcunningham:
## Description

Neither Runtime or Apache Airflow support 2.8 or earlier Airflow versions any longer, so we can drop support per our policy.

This lets us drop 2.6-2.8, and a number of special cases to support those versions.

## Related Issue(s)

Related: #2276 

## Breaking Change?

🤷 

## Checklist

- [x] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2321, opened 2026-02-05
Fix tests after removing support to Airflow versions under 2.9 support

tatiana:
Relates to #2288

Confirm that all tests that should be passing are passing, and make any further adjustments, as needed.

The proposal is that we first merge #2288, and then this PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, drop_airflow_under_29]
+    branches: [main]
   # Also run on pull requests originating from forks. Although this is insecure by default, we need it to run
   # integration tests on forked PRs. As a guardrail, we’ve added an Authorize step to each job, which requires manually
   # approving the workflow run for each pushed commit. Approval only happens after a careful code review of the changes.


tatiana:
All the tests in the relevant test branch worked 🎉 
https://github.com/astronomer/astronomer-cosmos/actions/runs/21709968850

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, drop_airflow_under_29]
```
```suggestion
    branches: [main]
```

cosmos/settings.py
```diff
@@ -64,8 +62,6 @@
 # this setting allows retries to run on a queue with larger resources, which is often necessary for larger dbt projects
 watcher_retry_queue = conf.get("cosmos", "watcher_retry_queue", fallback=None)
 
-AIRFLOW_IO_AVAILABLE = Version(airflow_version) >= Version("2.8.0")
```
pankajastro: Not sure, but do we need to highlight this in the changelog?

tatiana: Yes, to remove support for Airflow <= 2.8 is a breaking change - we should document it.

tatiana:
> LGTM
> 
> Q: Can we remove `openlineage-airflow` from `pryproject.toml`?

@pankajastro Very good observation - let's log a dedicated ticket for this, and prioritise it alongside our remaining work

Issue #2313, opened 2026-02-02
Test the cluster policy and dbt docs


pankajastro:
PR: https://github.com/astronomer/astronomer-cosmos/pull/2293 added the cluster policy plugin; this change also touches the plugin loading part. 

The aim here is to test both dbt docs and cluster policy from our side as well

pankajastro:
If env `AIRFLOW__COSMOS__WATCHER_DBT_EXECUTION_QUEUE` is set to a non-existent queue, it remains queued indefinitely

https://github.com/user-attachments/assets/db8ba513-4ca9-47d9-ac9e-eedb02f8a49d

If the env `AIRFLOW__COSMOS__WATCHER_DBT_EXECUTION_QUEUE` is set to an existing queue, both the producer task and the retry sensor will use it.

https://github.com/user-attachments/assets/c407ba22-b0a4-4485-b8b9-905653824eef

<img width="1714" height="989" alt="Image" src="https://github.com/user-attachments/assets/48cf3c1f-c888-48bb-bdbd-c1baa17698fa" />

Sensor in first run using default queue

<img width="1717" height="992" alt="Image" src="https://github.com/user-attachments/assets/2af5acf4-ad6f-49a0-adb3-5ee1af7deba7" />

pankajastro:
DBT Docs is working as expected

<img width="1695" height="996" alt="Image" src="https://github.com/user-attachments/assets/7e9a65b5-86c2-4c80-9a2d-2a4037f36b4c" />
<img width="1681" height="996" alt="Image" src="https://github.com/user-attachments/assets/50fd11dd-1a0c-40b7-b478-fb5862cca835" />

tatiana:
Looks great, thanks a lot, @pankajastro !

PR #2331, opened 2026-02-05
Rename watcher-mode sensor retry queue and reuse it for producer tasks

pankajastro:
Implement the [suggestion](https://github.com/astronomer/astronomer-cosmos/pull/2315#pullrequestreview-3751871740) from PR: #2315

This PR renames the `watcher_retry_queue` configuration setting to `watcher_dbt_execution_queue` to better reflect its expanded purpose: not only handling retries for consumer watcher sensors, but also for running the producer task itself.

**Changes:**
- Renamed `watcher_retry_queue` to `watcher_dbt_execution_queue` in settings and all references
- Added queue assignment logic to `DbtProducerWatcherOperator` to use the configured queue
- Updated all test patches and test names to reflect the new setting name

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -12,12 +12,12 @@ from airflow.exceptions import AirflowException
 
 from cosmos.config import ProfileConfig
 from cosmos.operators._watcher import _parse_compressed_xcom, safe_xcom_push
+from cosmos.settings import watcher_dbt_execution_queue
 
 try:
     from airflow.providers.standard.operators.empty import EmptyOperator
 except ImportError:  # pragma: no cover
     from airflow.operators.empty import EmptyOperator  # type: ignore[no-redef]
-
 
 from cosmos.constants import (
     PRODUCER_WATCHER_DEFAULT_PRIORITY_WEIGHT,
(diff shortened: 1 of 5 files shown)


cosmos/operators/watcher.py
```diff
@@ -96,6 +100,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         default_args["retries"] = 0
         kwargs["default_args"] = default_args
         kwargs["retries"] = 0
+        kwargs["queue"] = watcher_dbt_execution_queue or DEFAULT_QUEUE
```
pankajkoti, line 103: Can we please test this on Astro to ensure that Astro sets the DEFAULT_QUEUE and gets used when the watcher_dbt_execution_queue is not set?

pankajkoti (approved):
LGTM. One quick suggestion inline for sanity testing.

tatiana (approved):
Looks great, thank you very much, @pankajastro !
We'll be further testing this in Astro as part of https://github.com/astronomer/astronomer-cosmos/issues/2313, right?

cosmos/operators/watcher.py
pankajastro: Astro have default worker queue. I'll test it in issue https://github.com/astronomer/astronomer-cosmos/issues/2313

PR #2341, opened 2026-02-09
Fix failing integration tests

pankajkoti:


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,pankajkoti-patch-1]
```
```suggestion
    branches: [main]
```

pankajkoti:
Leaving it here if we want to get in a quick fix until the better alternative is worked upon in https://github.com/astronomer/astronomer-cosmos/pull/2343

Also happy for it to be closed if we no longer need it. 

cc: @tatiana @pankajastro

tatiana:
Closing this, since we followed a different approach to tackle the problem: #2343

PR #2172, opened 2025-11-27
Pin the pydantic<2.0 for Airflow 2.6

pankajastro:
I accidentally removed the pinned pydantic<2.0 in PR https://github.com/astronomer/astronomer-cosmos/pull/2161 while cleaning some redundant code, and CI also did not complain. This PR bring it back. 
This PR also remove manual installation of `pyopenssl`

PR #1987, opened 2025-09-19
Use click<8.3.0 for hatch as click 8.3 breaks hatch

pankajkoti:
`click 8.3.0` release broke hatch run.

We're seeing the following error in CI:
```
 + zstandard==0.25.0
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/hatch/c │
│ li/__init__.py:221 in main                                                   │
│                                                                              │
│   218                                                                        │
│   219 def main():  # no cov                                                  │
│   220 │   try:                                                               │
│ ❱ 221 │   │   hatch(prog_name='hatch', windows_expand_args=False)            │
│   222 │   except Exception:  # noqa: BLE001                                  │
│   223 │   │   import sys                                                     │
│   224                                                                        │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/click/c │
│ ore.py:1462 in __call__                                                      │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/click/c │
│ ore.py:1383 in main                                                          │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/click/c │
│ ore.py:1850 in invoke                                                        │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/click/c │
│ ore.py:1246 in invoke                                                        │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/click/c │
│ ore.py:814 in invoke                                                         │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/click/d │
│ ecorators.py:34 in new_func                                                  │
│                                                                              │
│ /opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/hatch/c │
│ li/run/__init__.py:147 in run                                                │
│                                                                              │
│   144 │   elif not env_name:                                                 │
│   145 │   │   env_name = 'system'                                            │
│   146 │                                                                      │
│ ❱ 147 │   ctx.invoke(                                                        │
│   148 │   │   run_command,                                                   │
│   149 │   │   args=[command, *final_args],                                   │
[...]

PR #2343, opened 2026-02-09
Define pinned `requirements.txt` to run Cosmos tests

tatiana:
**Context**

Over the past few months, we have experienced recurring CI failures caused by dependency conflicts. These issues are not triggered by changes in Cosmos itself, but rather by unpinned and dynamically resolved dependencies.

These conflicts have been disruptive, time-consuming to debug, and difficult to reproduce locally.

Some recent examples:

- https://github.com/astronomer/astronomer-cosmos/pull/2268
- https://github.com/astronomer/astronomer-cosmos/pull/2172
- https://github.com/astronomer/astronomer-cosmos/pull/2076
- https://github.com/astronomer/astronomer-cosmos/pull/1987
- https://github.com/astronomer/astronomer-cosmos/pull/1909
- https://github.com/astronomer/astronomer-cosmos/pull/1902

**Current Situation**

We have tried several approaches to mitigate dependency issues:
- Dynamically resolving dependencies while restricting requirements as little as possible
- Handling Airflow version–specific dependencies individually
- Sometimes relying on Airflow constraints (which do not include provider versions), sometimes not
- Attempting to isolate dependencies in `pre-install-airflow.sh`

More recently, however, Airflow began being reinstalled via `integration-setup.sh`, leading to overlapping installations and Python virtual environments with conflicting packages.

Additionally, if we try to recreate the environment locally using the exact dependency list printed by the CI at the end of the “Install packages and dependencies” step, the environment cannot be created due to dependencies conflicts.

This results in:
- Non-deterministic Python package installation
- CI failures that are difficult (or impossible) to reproduce locally
- Significant frustration and lost engineering time

**Proposed Approach**

This PR proposes pinning dependencies more strictly to make our CI environments deterministic and reproducible.

I do not believe this is a perfect solution. Some trade-offs include:
- We will no longer automatically test against newly released dependency versions
- Pinned versions may require manual updates, including potential security updates

This PR also does not introduce pip constraints files (since we currently rely on pip install -r). However, it can be viewed as a step toward the approach discussed in: https://github.com/astronomer/astronomer-cosmos/issues/967

I discussed this proposal with @pankajastro and @pankajkoti, and we agreed it is worth trying. If this strategy does not work well in practice, we can revert to the current approach.

**Known Side Effects**

The changes in this PR introduce a few limitations:

- Tests no longer run the cross-project example DAG (dbt Loom) using the LoadMode.DBT_LS approach, as dbt Loom is now installed in a separate virtual environment due to dependency conflicts.
- The S3 load manifest task is no longer executed in the corresponding DAG for Airflow 2.9, due to conflicts between aiobotocore and other dependencies.

### DIFF
--- a/tests/test_example_dags.py
+++ b/tests/test_example_dags.py
@@ -64,6 +64,8 @@ def get_dag_bag() -> DagBag:  # noqa: C901
         if AIRFLOW_VERSION == Version("2.9.0"):
             # aiobotocore can't be installed with all the other Cosmos test dependencies in Airflow 2.9
             file.writelines("cosmos_example_manifest_dag.py\n")
+            # This DAG is taking too long to run int the CI (https://github.com/astronomer/astronomer-cosmos/actions/runs/21902660682/job/63234728594)
+            file.writelines("example_cosmos_python_models.py\n")
 
         # Disabling these DAGs temporarily due to an Airflow 3 bug on processing DatasetAlias that contain non-ASCII characters:
         # https://github.com/apache/airflow/issues/51566


scripts/test/pre-install-airflow.sh
```diff
@@ -13,56 +13,34 @@ if [ "$GITHUB_ACTIONS" = "true" ] && [ -z "${VIRTUAL_ENV}" ]; then
...
+  uv pip install -r requirements/requirements-airflow-3.1-dbt-1.11.txt
 else
   uv pip install "apache-airflow-providers-amazon[s3fs]" --constraint /tmp/constraint.txt
   uv pip install "apache-airflow-providers-cncf-kubernetes" --constraint /tmp/constraint.txt
-  uv pip install "apache-airflow-providers-google" --constraint /tmp/constraint.txt
+  uv pip install "apache-airflow-providers-google" "apache-airflow==$AIRFLOW_VERSION" --constraint /tmp/constraint.txt
   uv pip install "apache-airflow-providers-microsoft-azure" --constraint /tmp/constraint.txt
+  uv pip install -U "dbt-core~=$DBT_VERSION" dbt-postgres dbt-bigquery dbt-vertica dbt-databricks pyspark
+  uv pip install 'dbt-duckdb' "airflow-provider-duckdb>=0.2.0" apache-airflow==$AIRFLOW_VERSION
+  uv pip install dbt-loom
```
pankajkoti, line 61: When do we need the else block?

requirements/requirements-airflow-2.10-dbt-1.11.txt
pankajkoti, line 1: Should we call this directory constraints?

pankajkoti, line 1: How are these files generated?

scripts/test/pre-install-airflow.sh
pankajkoti, line 61: We need to ensure that dbt-loom is still installed. Should it be installed in the venv-subprocess/bin?

requirements/requirements-airflow-2.10-dbt-1.11.txt
tatiana: I'm not sure, because at the moment, these are all the dependencies except Cosmos, not only constraints. They are installed with `pip install -r`, which references requirements

tatiana: I checked successful CI runs and tried to use the pip freeze dependencies available there. Unfortunately, they can't be installed, because they conflict. So I started removing the dependencies that were conflicting, and we don't really care about the versions (such as `google*`, `grpc`, `paradime`, `opentelemetry`) and so on - until I was able to install everything via a single `uv pip install` command locally.

Once that succeeded, I ran pip freeze, removed the line that installs Cosmos, cleared the virtual environment, and attempted to reinstall everything from scratch.

I run repeatedly sequences of commands similar to these two:
```
hatch env remove tests.py3.13-3.1-1.11
AIRFLOW_HOME=`pwd` hatch run tests.py3.13-3.1-1.11:test-integration-setup
```

For each version of Airflow we support.

scripts/test/pre-install-airflow.sh
tatiana: I will be necessary when we support new versions and need to regenerate a `requirements.txt` file that is compliant with those versions.

pankajastro:
Auto-generation for requirements would be great if possible; otherwise, LGTM.

scripts/test/pre-install-airflow.sh
tatiana: ATM it is installed in all requirements.txt we define. If in future we find ourselves in positions where we can't resolve the conflicts in the same Python virtualenv, we could move them to the separate venv we create in the `pre-install-airflow.sh` (`venv-subprocess`)

tatiana:
> I am +1 on this approach of having pre-generated constraint files. We should probably also plan to include it as a constraint artefact in our upcoming releases.

Thanks a lot, @pankajkoti! I believe this approach has both pros and cons, but at the very least, it’s a concrete attempt on our side to simplify the CI issues and dependency conflicts that have been consuming a lot of time over the past few months—an area where AI hasn’t been particularly helpful so far.

For now, I propose sticking with `requirements.txt` files rather than full-fledged `constraints`. I don’t think our team should take on the responsibility of maintaining constraints for the wider community. Even Airflow’s attempt at this is limited (to Airflow-only and excludes providers). Trying to align versions across Airflow, dbt-core, Airflow providers, and dbt adapters across multiple versions of Python, and having to maintain this, seems significantly more challenging.

Instead, I suggest we limit ourselves to `requirements.txt` files used exclusively for our test environment. By installing all dependencies declared in the same file with a single uv pip install command, I expect the pinned versions to be respected and that we’d end up with an environment as deterministic as reasonably possible. These requirements will naturally evolve over time, and I also expect that, in the longer term, we’ll want to allow some dependencies to resolve without fully pinned requirements.

tests/test_example_dags_no_connections.py
```diff
@@ -46,6 +46,9 @@ def get_dag_bag() -> DagBag:
             print(f"Adding {file_name} to .airflowignore")
             file.write(f"{file_name}\n")
 
+        if AIRFLOW_VERSION == Version("2.9.0"):
+            file.writelines("example_cosmos_python_models\n")
```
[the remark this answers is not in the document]
pankajkoti: +1 I guess. Or we can remove it since it may not have been ignored and tests are passing?

pankajkoti (approved):
Thanks for picking this up and for your persistence in getting this approach to work!

requirements/requirements-airflow-2.9-dbt-1.11.txt
```diff
@@ -0,0 +1,323 @@
...
+openpyxl==3.1.5
+ordered-set==4.1.0
+orderly-set==5.5.0
+packaging==26.0
+pandas==2.1.4
+pandas-gbq==0.33.0
+parsedatetime==2.6
+pathspec==0.12.1
+pendulum==3.0.0
+pip==26.0.1
```
tatiana:
```suggestion
```

requirements/requirements-airflow-2.10-dbt-1.11.txt
```diff
@@ -0,0 +1,376 @@
...
+ordered-set==4.1.0
+orderly-set==5.5.0
+packaging==24.1
+pandas==2.1.4
+pandas-gbq==0.23.1
+paradime-io==4.17.0
+parsedatetime==2.6
+pathspec==0.12.1
+pendulum==3.0.0
+pip==26.0.1
```
tatiana:
```suggestion
```

scripts/test/pre-install-airflow.sh
```diff
@@ -13,64 +13,77 @@ if [ "$GITHUB_ACTIONS" = "true" ] && [ -z "${VIRTUAL_ENV}" ]; then
...
-  CONSTRAINT_URL="https://raw.githubusercontent.com/apache/airflow/constraints-$AIRFLOW_VERSION.0/constraints-$PYTHON_VERSION.txt"
-fi;
-
-curl -sSL $CONSTRAINT_URL -o /tmp/constraint.txt
-# Workaround to remove PyYAML constraint that will work on both Linux and MacOS
-sed '/PyYAML==/d' /tmp/constraint.txt > /tmp/constraint.txt.tmp
-mv /tmp/constraint.txt.tmp /tmp/constraint.txt
+# Remove artifacts used locally
+rm -f $AIRFLOW_HOME/airflow.cfg
+rm -f $AIRFLOW_HOME/airflow.db
```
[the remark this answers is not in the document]
tatiana: This didn't work in the CI

tests/test_example_dags_no_connections.py
```diff
@@ -46,6 +47,9 @@ def get_dag_bag() -> DagBag:
             print(f"Adding {file_name} to .airflowignore")
             file.write(f"{file_name}\n")
 
+        if AIRFLOW_VERSION == Version("2.9.0"):
+            file.writelines("example_cosmos_python_models.py\n")
+
```
tatiana:
```suggestion
```

PR #2352, opened 2026-02-12
Accept int or float for cosmos_debug_max_memory_mb in integration test

pankajkoti:
It appears XCom JSON serialization can deserialize whole numbers as int (e.g. 10.0 -> 10),
causing intermittent AssertionError on some CI envs (e.g. py3.10, Airflow 3.1, dbt 1.11 CI run [here](https://github.com/astronomer/astronomer-cosmos/actions/runs/21912893530/job/63271736980?pr=2335#step:7:1620)).
Relax assertion to isinstance(memory_value, (int, float)) and use float() for the > 0 check.


tatiana (approved):
Looks great, thanks for the improvement, @pankajkoti !

PR #2353, opened 2026-02-12
Remove outdated docs for dbt docs plugin airflow 3

pankajastro:
Removes an outdated documentation note claiming Cosmos’ Airflow plugin cannot host dbt docs on Airflow 3,


tatiana (approved):
Excellent, thanks, @pankajastro !

Issue #2231, opened 2025-12-29
[Bug] Cosmos introduces profile `config` session that is incompatible with dbt Fusion


tatiana:
### Astronomer Cosmos Version

1.12

### dbt-core version

 dbt-fusion 2.0.0-preview.76

### Versions of dbt adapters

When trying to run this DAG:
```
"""
An example DAG that uses Cosmos to render a dbt project into an Airflow DAG.
"""

import os
from datetime import datetime, timedelta
from pathlib import Path

from cosmos import DbtDag, ExecutionConfig, ProfileConfig, ProjectConfig, RenderConfig
from cosmos.constants import ExecutionMode
from cosmos.profiles import PostgresUserPasswordProfileMapping

DEFAULT_DBT_ROOT_PATH = Path(__file__).parent / "dbt"
DBT_ROOT_PATH = Path(os.getenv("DBT_ROOT_PATH", DEFAULT_DBT_ROOT_PATH))
DBT_PROJECT_NAME = os.getenv("DBT_PROJECT_NAME", "jaffle_shop")
DBT_PROJECT_PATH = DBT_ROOT_PATH / DBT_PROJECT_NAME


profile_mapping = PostgresUserPasswordProfileMapping(conn_id="example_conn", profile_args={"schema": "public"})


profile_config = ProfileConfig(
    profile_name="default",
    target_name="dev",
    profile_mapping=profile_mapping,
)


operator_args = {
    "install_deps": True,  # install any necessary dependencies before running any dbt command
    "execution_timeout": timedelta(seconds=120),
}

# Currently airflow dags test ignores priority_weight and  weight_rule, for this reason, we're setting the following in the CI only:
if os.getenv("CI"):
    operator_args["trigger_rule"] = "all_success"


from cosmos.constants import InvocationMode

# [START example_watcher]
example_watcher = DbtDag(
    # dbt/cosmos-specific parameters
    execution_config=ExecutionConfig(
        execution_mode=ExecutionMode.WATCHER,
        invocation_mode=InvocationMode.SUBPROCESS,
        dbt_executable_path="/tmp/venv-isolated/bin/dbt",
    ),
    project_config=ProjectConfig(DBT_PROJECT_PATH),
    profile_config=profile_config,
    render_config=RenderConfig(exclude=["raw_payments"]),
    operator_args=operator_args,
    # normal dag parameters
    schedule="@daily",
    start_date=datetime(2023, 1, 1),
    catchup=False,
[...]

Issue #2232, opened 2025-12-29
Enable running Cosmos integration tests against dbt Fusion


tatiana:
If we cannot enable all tests, we should at least allow running all example DAGs in dbt Fusion.

Follow-up #2231

tatiana:
Timebox to two days:
- Identify which are the problems
- If not all of them can be fixed in two days, log follow up tickets, which we can prioritise in future sprints

pankajkoti:
Can we please also test bigquery adapter with Fusion when enabling these tests and fixing bugs if any?

This was asked in Astronomer Slack: https://astronomer.slack.com/archives/C04EXKRQFD3/p1770338083665259 & Zendesk ticket: https://astronomer.zendesk.com/agent/tickets/88136

romanoww:
> Can we please also test bigquery adapter with Fusion when enabling these tests and fixing bugs if any?
> 
> This was asked in Astronomer Slack: https://astronomer.slack.com/archives/C04EXKRQFD3/p1770338083665259 & Zendesk ticket: https://astronomer.zendesk.com/agent/tickets/88136

Our team is also interested in having the bigquery adapter enabled for dbt fusion, following this thread.

tatiana:
Most Cosmos example DAGs run with Postgres. I just attempted to run a Postgres DAG with the latest fusion release (2.0.0-preview.110), and it seems the Postgres adaptor isn't available yet (I just subscribed to the open ticket https://github.com/dbt-labs/dbt-core/issues/13123):
```
[2026-02-12 15:07:37] ERROR - dbt-fusion 2.0.0-preview.110
   Loading ../cosmos/profile/ac4e9cde9bc05d574c157e795dcbcc6b60246a73ca1d92d4fc669e90a1e494e0/profiles.yml
   Loading packages.yml
panic: panicked at fs/sa/crates/dbt-adapter/src/metadata/***.rs:181:9:
not yet implemented: PostgresAdapter::create_schemas_if_not_exists
```

This is the Postgres DAG I attempted to run:
```
"""
An example DAG that uses Cosmos to render a dbt project into an Airflow DAG.
"""

import os
from datetime import datetime, timedelta
from pathlib import Path

from cosmos import DbtDag, ExecutionConfig, ProfileConfig, ProjectConfig, RenderConfig
from cosmos.constants import ExecutionMode
from cosmos.profiles import PostgresUserPasswordProfileMapping

[...]

PR #2356, opened 2026-02-12
Improve dbt fusion support and tests

tatiana:
Fix running dbt Fusion with Cosmos when using profile mapping. Add an integration test to run dbt Fusion against BigQuery, complementing the previously existing dbt Fusion against Snowflake.

Main changes to the implementation:

1. Use dbt recommended way to disable telemetry (`send_anonymous_usage_stats`), via environment variables (https://docs.getdbt.com/reference/global-configs/usage-stats), in a way it should be compatible with dbt Core and dbt Fusion:
   - `DO_NOT_TRACK=1`
   - `DBT_SEND_ANONYMOUS_USAGE_STATS=False`

2. Move the config section inside the profile, instead of being on the top-level of the `profiles.yml`

Before:
```
config:
  send_anonymous_usage_stats: false
bigquery_profile:
  target: dev
  outputs:
    dev:
      type: bigquery
      method: service-account-json
      project: astronomer-dag-authoring
      dataset: release_17
      threads: 1
```

After this change:
```
bigquery_profile:
  target: dev
  outputs:
    dev:
      type: bigquery
      method: service-account-json
      project: astronomer-dag-authoring
      dataset: release_17
      threads: 1
  config:
    send_anonymous_usage_stats: false
```

Main changes to tests:
- Change Snowflake dbt Fusion integration test to also use dbt Fusion to render the dbt DAG (before it was only being used during execution)
- Add `config` `send_anonymous_usage_stats` in the `snowflake_profile`
- Introduce DbtDag that also tests dbtFusion against BQ

Moving forward, dbt plans to deprecate `config` in `profiles.yml` and encourages users to use `flags` in `dbt_project.yml`: https://github.com/dbt-labs/dbt-core/issues/9183

Since this is a dbt configuration change in a file that Cosmos doesn't currently mess up (`dbt_profile.yml`), it feels outside our control to make any further changes.

Closes: https://github.com/astronomer/astronomer-cosmos/issues/2231
Related: https://github.com/astronomer/astronomer-cosmos/issues/2232

Why the checks are red: because there is a change to the GH Action `test.yml` that applies only to this branch, and not the tests running against the `main` branch configuration:
https://github.com/astronomer/astronomer-cosmos/actions/runs/21957764192/job/63426578539?pr=2356
This is expected and will work after merging.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, improve-dbt-fusion-tests]
+    branches: [main]
   # Also run on pull requests originating from forks. Although this is insecure by default, we need it to run
   # integration tests on forked PRs. As a guardrail, we’ve added an Authorize step to each job, which requires manually
   # approving the workflow run for each pushed commit. Approval only happens after a careful code review of the changes.


dev/dags/dbt/jaffle_shop/profiles.yml
```diff
@@ -22,6 +22,8 @@ snowflake_profile:
       schema: "{{ env_var('SNOWFLAKE_SCHEMA') }}"
       warehouse: "{{ env_var('SNOWFLAKE_WAREHOUSE') }}"
       database: "{{ env_var('SNOWFLAKE_DATABASE') }}"
+  config:
+    send_anonymous_usage_stats: false
```
pankajkoti, line 26: Looks like CI is not happy with this

tatiana: This is really annoying, because it seems to be working on my computer 🤓 💻 

There is a dedicated ticket for this, I'll figure out what's going on:
https://github.com/astronomer/astronomer-cosmos/issues/2231

tatiana: Good news, I was able to reproduce again!
https://github.com/astronomer/astronomer-cosmos/issues/2231#issuecomment-3891877472

tatiana: And I believe I fixed it in https://github.com/astronomer/astronomer-cosmos/pull/2356/changes/52221d84a46ba4ff5f24a90251c53fca6d95e946

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, improve-dbt-fusion-tests]
```
```suggestion
    branches: [main]
```

PR #2314, opened 2026-02-02
Release 1.14.0

pankajastro:
1.14.0 (2026-04-07)
---------------------

Breaking Changes

* Drop support for Airflow versions earlier than **2.9** by @jedcunningham in #2288
* Fix inclusion of package models and selection/exclusion behavior by @pankajkoti in #2357
* ``ExecutionMode.WATCHER``: The per-node ``*_status`` XCom value is now a dict (``{"status": "<status>", "outlet_uris": [...]}``) instead of a plain string. Any custom code that reads these internal XCom keys directly will need to be updated by @pankajkoti in #2507

Features

* Add cluster policy support for ``ExecutionMode.WATCHER`` sensor retries by @astro-anand in #2293
* Add debug mode to track memory utilization by @tatiana in #2327
* Add FQN selection support for ``LoadMode.DBT_MANIFEST`` by @pankajastro in #2375
* Introduce interceptors for Cosmos tasks by @tatiana in #2419
* Add config to allow disabling dag versioning by @pankajkoti in #2470
* Implement TaskGroups by models folder by @maximilianoarcieri and @tatiana in #1566, #2469, and #2420
* feat: implement DbtTestWatcherOperator by @michal-mrazek in #2447
* Add source freshness aware execution for ``ExecutionMode.WATCHER`` by @pankajastro and @tatiana in #2467

    * Note: Like ``ExecutionMode.WATCHER``, this feature is experimental and its interface and implementation can change in the future.
* Add Airflow 3.2 support by @pankajastro and @pankajkoti in #2472

Enhancements

* Add watcher mode support for dbt test node states by @michal-mrazek in #2318
* Rename watcher-mode sensor retry queue and reuse it for producer tasks by @pankajastro in #2331
* Fix leaked semaphore warnings in Airflow 3 by resetting dbt adapters by @pankajkoti in #2335
* Improve dbt Fusion support and related tests by @tatiana in #2356
* Default Snowflake profile mappings to four threads by @tatiana in #2374
* Attempt to remove Pydantic as a dependency by @tatiana in #2377
* Log dbt-core and adapter versions in watcher consumer tasks by @pankajastro in #2412
* Log model errors in watcher consumer on dbt node failure by @pankajastro in #2431
* Reduce XCom read/write for tracking node state and errors in ConsumerWatcher task by @pankajastro in #2471
* Remove duplicate debug log in watcher subprocess path by @tatiana in #2494
[...]

PR #2357, opened 2026-02-13
Fix inclusion of package models, and selecting/excluding

### LINKED ISSUE #1528: [Bug] Issue with exlcuding packages
### Astronomer Cosmos Version

1.9.0a5

### dbt-core version

1.8

### Versions of dbt adapters

dbt snowflake 1.8

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

LOCAL

### InvocationMode

DBT_RUNNER

### airflow version

2.10

### Operating System

Unix

_No response_

### Deployment

Amazon (AWS) MWAA

### Deployment details

_No response_

### What happened?

Issue with select and exclude for packages. So for example excluding the dbt artifacts package from running. This still works when using dbt ls but not when switching to dbt manifest:

    my_cosmos_dag = DbtTaskGroup(
        project_config=ProjectConfig(
            dbt_project_path=dbt_project_path,
            env_vars={'ENV_NAME': env, 'dag_id': dag.dag_id, 'run_id': '{{run_id}}'},
            manifest_path=f'{dbt_project_path}/target/manifest.json',
        ),
        execution_config=ExecutionConfig(),
        operator_args={'dbt_cmd_flags': ['--exclude', 'test_type:unit', '--no-version-check'], 'full_refresh': True},
        render_config=RenderConfig(
            emit_datasets=False,
            dbt_deps=False,
            exclude=['dbt_artifacts'],
        ),

### Relevant log output

```shell

```

### How to reproduce

Try to install an additional package such as dbt artifacts and try to exclude them from the run.

### Anything else :)?

_No response_

### Contact Details

_No response_

pankajkoti:
Fixes DAG rendering when using manifest load mode with installed dbt packages (e.g. Elementary, dbt_artifacts): package models are now included and have correct paths (this was not working previously).
Additionally, you can now select/exclude by package which was not working previously.
Also, fixes selecting/excluding models by folder_name which was not working previously.

**Changes**

**Manifest mode – package nodes**
- Correct `file_path` for package nodes: In the manifest, package nodes use `original_file_path` relative to the package root (e.g. `models/edr/...`). Cosmos now resolves them as `project_path/dbt_packages/<package_name>/<original_file_path>` so package models appear in the DAG and path-based logic works.
- Root vs package: Uses manifest `metadata.project_name` to treat root project nodes as `project_path/original_file_path`. When `metadata.project_name` is missing (older/test manifests), all nodes are treated as root for backward compatibility.
- Docs: Parsing-methods doc updated to note that package models are included when present in the manifest and that the manifest should be generated after installing and enabling packages (dbt deps then dbt compile).

**Selectors**
- `package:package_name`: You can include or exclude all nodes from a package (e.g. `exclude=["package:dbt_artifacts"]` or `select=["package:elementary"]`), including when using manifest load mode.
- Bare identifiers: A bare name with no prefix (e.g. `exclude=["dbt_artifacts"]`) is resolved like in dbt: by package name or node name or a folder name containing models.
- Docs and tests: Selecting-excluding doc updated with the new options and an example; tests added for select/exclude by package and bare identifier.

**Robustness and errors**
Empty `dbt_project.yml`: `get_dbt_packages_subpath()` no longer assumes `yaml.safe_load()` returns a dict; handles `None` (e.g. empty file) so manifest loading does not raise when the project dir has an empty dbt_project.yml.
Manifest loading: Handles json.load() returning None and uses metadata only when it is a dict when reading project_name.

closes: #1528 
closes: https://github.com/astronomer/oss-integrations-private/issues/318

### DIFF
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -54,6 +54,7 @@ from cosmos.dbt.project import (
     copy_dbt_packages,
     create_symlinks,
     environ,
+    get_dbt_packages_subpath,
     get_partial_parse_path,
     has_non_empty_dependencies_file,
 )
@@ -1137,6 +1138,11 @@ class DbtGraph:
             manifest = json.load(fp)
 
             resources = {**manifest.get("nodes", {}), **manifest.get("sources", {}), **manifest.get("exposures", {})}
+            project_path = self.execution_config.project_path
+            packages_subpath = get_dbt_packages_subpath(project_path)
+            # Root project name from manifest metadata (dbt 1.x); when absent, treat all nodes as root (e.g. old/test manifests)
+            manifest_project_name = (manifest.get("metadata") or {}).get("project_name")
+
             for unique_id, node_dict in resources.items():
                 # External nodes (e.g., from dbt-loom) may not have a file path - skip them
                 # Check for both None and empty string since dbt-loom may set either
@@ -1148,12 +1154,23 @@ class DbtGraph:
                     )
                     continue
 
+                package_name = node_dict.get("package_name")
+                # For installed-package nodes, original_file_path is relative to the package root (e.g. models/edr/...).
+                # Resolve to project_path/dbt_packages/<package_name>/<original_file_path>.
+                # Root project nodes (manifest_project_name == package_name) use project_path/original_file_path.
+                # When manifest has no project_name (old/test manifests), treat every node as root for backward compatibility.
+                is_root_project_node = manifest_project_name is None or (package_name == manifest_project_name)
+                if package_name and not is_root_project_node:
+                    resolved_path = project_path / packages_subpath / package_name / _normalize_path(original_file_path)
+                else:
+                    resolved_path = project_path / _normalize_path(original_file_path)
+
                 node = DbtNode(
                     unique_id=unique_id,
-                    package_name=node_dict.get("package_name"),
+                    package_name=package_name,
                     resource_type=DbtResourceType(node_dict["resource_type"]),
                     depends_on=node_dict.get("depends_on", {}).get("nodes", []),
-                    file_path=self.execution_config.project_path / _normalize_path(original_file_path),
+                    file_path=resolved_path,
                     tags=node_dict.get("tags") or [],
                     config=node_dict.get("config") or {},
                     has_freshness=(
--- a/cosmos/dbt/graph.py
+++ b/cosmos/dbt/graph.py
@@ -352,6 +352,44 @@ def parse_dbt_ls_output(project_path: Path | None, ls_stdout: str) -> dict[str, 
     return nodes
 
 
+def _parse_manifest_resources_to_nodes(manifest: dict[str, Any], project_path: Path) -> dict[str, DbtNode]:
+    """Build a nodes dict from manifest resources (nodes, sources, exposures) with correct file_path for packages."""
+    resources = {**manifest.get("nodes", {}), **manifest.get("sources", {}), **manifest.get("exposures", {})}
+    packages_subpath = get_dbt_packages_subpath(project_path)
+    manifest_metadata = manifest.get("metadata")
+    manifest_project_name = manifest_metadata.get("project_name") if isinstance(manifest_metadata, dict) else None
+    nodes: dict[str, DbtNode] = {}
+    for unique_id, node_dict in resources.items():
+        original_file_path = node_dict.get("original_file_path")
+        if not original_file_path:
+            logger.debug(
+                "Skipping node `%s` because it has no file path (likely an external reference from dbt-loom or similar)",
+                unique_id,
+            )
+            continue
+        package_name = node_dict.get("package_name")
+        is_root_project_node = manifest_project_name is None or (package_name == manifest_project_name)
+        if package_name and not is_root_project_node:
+            resolved_path = project_path / packages_subpath / package_name / _normalize_path(original_file_path)
+        else:
+            resolved_path = project_path / _normalize_path(original_file_path)
+        resource_type = DbtResourceType(node_dict["resource_type"])
+        node = DbtNode(
+            unique_id=unique_id,
+            package_name=package_name,
+            resource_type=resource_type,
+            depends_on=node_dict.get("depends_on", {}).get("nodes", []),
+            file_path=resolved_path,
+            tags=node_dict.get("tags") or [],
+            config=node_dict.get("config") or {},
+            has_freshness=(
+                is_freshness_effective(node_dict.get("freshness")) if resource_type == DbtResourceType.SOURCE else False
+            ),
+        )
+        nodes[node.unique_id] = node
+    return nodes
+
+
 class DbtGraph:
     """
     A dbt project graph (represented by `nodes` and `filtered_nodes`).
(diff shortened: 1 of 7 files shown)


cosmos/dbt/graph.py
```diff
@@ -351,6 +352,44 @@ def parse_dbt_ls_output(project_path: Path | None, ls_stdout: str) -> dict[str,
     return nodes
 
 
+def _parse_manifest_resources_to_nodes(manifest: dict[str, Any], project_path: Path) -> dict[str, DbtNode]:
```
pankajkoti: this is a just a refactor to address the failing pre-commit hook complaining about the function being too complex.

cosmos/dbt/graph.py
```diff
@@ -1128,72 +1167,41 @@ def load_from_dbt_manifest(self) -> None:
         if not self.execution_config.project_path:
             raise CosmosLoadDbtException("Unable to load manifest without ExecutionConfig.dbt_project_path")
 
-        nodes = {}
```
pankajkoti, line 1134: this is a just a refactor to address the failing pre-commit hook complaining about the function being too complex.

cosmos/dbt/graph.py
```diff
@@ -1128,72 +1167,41 @@ def load_from_dbt_manifest(self) -> None:
...
-                # External nodes (e.g., from dbt-loom) may not have a file path - skip them
-                # Check for both None and empty string since dbt-loom may set either
-                original_file_path = node_dict.get("original_file_path")
-                if not original_file_path:
-                    logger.debug(
-                        "Skipping node `%s` because it has no file path (likely an external reference from dbt-loom or similar)",
-                        unique_id,
-                    )
-                    continue
+        nodes = _parse_manifest_resources_to_nodes(manifest, self.execution_config.project_path)
```
pankajkoti: this is a just a refactor to address the failing pre-commit hook complaining about the function being too complex.

cosmos/dbt/graph.py
```diff
@@ -1128,72 +1167,41 @@ def load_from_dbt_manifest(self) -> None:
...
+                exclude=selections["exclude"],
+            )
+        else:
+            self.nodes = nodes
+            self.filtered_nodes = select_nodes(
+                project_dir=self.execution_config.project_path,
+                nodes=nodes,
+                select=self.render_config.select,
+                exclude=self.render_config.exclude,
+            )
```
pankajkoti: this is a just a refactor to address the failing pre-commit hook complaining about the function being too complex.

cosmos/dbt/project.py
```diff
@@ -85,7 +85,8 @@ def get_dbt_packages_subpath(source_folder: Path) -> str:
             except yaml.YAMLError:
                 logger.info(f"Unable to read the {DBT_PROJECT_FILENAME} file")
             else:
-                subpath = dbt_project_file_content.get("packages-install-path", DBT_DEFAULT_PACKAGES_FOLDER)
+                if isinstance(dbt_project_file_content, dict):
+                    subpath = dbt_project_file_content.get("packages-install-path", DBT_DEFAULT_PACKAGES_FOLDER)
```
pankajkoti, line 89: this is a fix for a failing test that run on a sample dbt project that has an empty dbt_project.yml file. It didn't fail previously because models were previously ignored from packages folder when using manifest mode.

cosmos/dbt/selector.py
```diff
@@ -395,24 +401,38 @@ def load_from_statement(self, statement: str) -> None:
...
+                node_name.startswith(EXCLUDE_RESOURCE_TYPE_SELECTOR),
+                lambda: self._parse_exclude_resource_type_selector(item),
+            ),
+            (self._is_bare_identifier(node_name), lambda: self._parse_bare_identifier(node_name)),
+        ]
+        for condition, handler in dispatch:
+            if condition:
+                handler()
+                return
+        self._parse_unknown_selector(item)
```
pankajkoti: this is a just a refactor to address the failing pre-commit hook complaining about the function being too complex.

cosmos/dbt/project.py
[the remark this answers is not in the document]
pankajkoti: ignoring.

cosmos/dbt/graph.py
pankajkoti: reverted the refactor to simplify the reviewing of the PR. Added a noqa which is to be removed in a separate PR

cosmos/dbt/graph.py
pankajkoti: reverted the refactor to simplify the reviewing of the PR. Added a noqa which is to be removed in a separate PR

cosmos/dbt/graph.py
pankajkoti: reverted the refactor to simplify the reviewing of the PR. Added a noqa which is to be removed in a separate PR

cosmos/dbt/graph.py
pankajkoti: reverted the refactor to simplify the reviewing of the PR. Added a noqa which is to be removed in a separate PR

cosmos/dbt/selector.py
pankajkoti: reverted the refactor to simplify the reviewing of the PR. Added a noqa which is to be removed in a separate PR

cosmos/dbt/selector.py
```diff
@@ -461,6 +481,11 @@ def _parse_exposure_selector(self, item: str) -> None:
         exposure_name = item[index:].strip()
         self.exposures.append(exposure_name)
 
+    def _parse_package_selector(self, item: str) -> None:
+        index = len(PACKAGE_SELECTOR)
+        package_name = item[index:].strip()
```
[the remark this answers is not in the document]
pankajkoti: addressed

cosmos/dbt/selector.py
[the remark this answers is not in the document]
pankajkoti: requested re-review since the refactor has been reverted.

docs/configuration/selecting-excluding.rst
```diff
@@ -33,6 +33,7 @@ The ``select`` and ``exclude`` parameters are lists, with values like the follow
 - ``source:my_source.my_table``: include/exclude nodes that have the source ``my_source`` and the table ``my_table``
 - ``exposure:my_exposure``: include/exclude nodes that have the exposure ``my_exposure`` and are of resource_type ``exposure``
 - ``exposure:+my_exposure``: include/exclude nodes that have the exposure ``my_exposure`` and their parents
+- ``package:package_name``: include/exclude all nodes that belong to the given package (e.g. ``package:dbt_artifacts``). Especially useful with ``LoadMode.DBT_MANIFEST`` when excluding installed packages from the DAG. A bare name without a method prefix (e.g. ``dbt_artifacts`` or ``child``) is resolved like dbt: it matches nodes by package name, node name, or path segment (folder name). So ``select=['folder_a']`` or ``exclude=['folder_a']`` includes or excludes all models under a folder named ``folder_a``, including when using ``LoadMode.DBT_MANIFEST``.
```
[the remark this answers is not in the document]
pankajkoti: addressed

cosmos/dbt/selector.py
```diff
@@ -352,6 +360,8 @@ def __init__(self, project_dir: Path | None, statement: str):
         self.graph_selectors: list[GraphSelector] = []
         self.sources: list[str] = []
         self.exposures: list[str] = []
+        self.packages: list[str] = []
+        self.bare_identifiers: list[str] = []  # bare strings: match by package_name or node name (dbt-like)
```
pankajkoti: also folder_name

tatiana (approved):
This looks great, @pankajkoti , thanks for fixing these two issues.

As I mentioned on our call, a side-effect of this change is that the Cosmos DAG topology may change, since we were not rendering nodes that we now are. This is sometimes a side-effect of fixing issues, but could be perceived as a breaking change by some.

For this reason, I suggest we release this in 1.14.0 instead of 1.13.1 - and that we make this very explicit in the release notes in https://github.com/astronomer/astronomer-cosmos/pull/2314 - please, raise this in that PR.

cosmos/dbt/selector.py
pankajkoti:
```suggestion
        self.bare_identifiers: list[str] = []  # bare strings: match by package_name, node name or folder name (dbt ls-like)
```

PR #2364, opened 2026-02-13
Fix select & exclude type mismatch

tatiana:
`RenderConfig` defines the `select` and `exclude` attributes as `list[str]`, but `AbstractDbtBase` (the base class for all Cosmos operators) expects them as `str | None`. When `airflow/graph.py` passes `render_config.select` / `render_config.exclude` directly into `task_args`, the operators received lists instead of strings, causing a type mismatch.

This PR aims to solve this issue.

Closes: #838

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -145,7 +145,7 @@ def exclude_detached_tests_if_needed(
     for test_node in tests_detached_from_this_node:
         exclude_items.append(test_node.resource_name.split(".")[0])
     if exclude_items:
-        task_args["exclude"] = _convert_list_to_str(exclude_items)
+        task_args["exclude"] = _convert_list_to_str(exclude_items) or ""
 
 
 def _override_profile_if_needed(task_kwargs: dict[str, Any], profile_kwargs_override: dict[str, Any]) -> None:


cosmos/airflow/graph.py
```diff
@@ -121,12 +133,19 @@ def exclude_detached_tests_if_needed(
     """
     if detached_from_parent is None:
         detached_from_parent = {}
-    exclude: list[str] = task_args.get("exclude", [])  # type: ignore
+    current_exclude = task_args.get("exclude")
+    # Handle both list[str] (legacy) and str formats for backward compatibility
+    if isinstance(current_exclude, list):
+        exclude_items = list(current_exclude)
```
```suggestion
        exclude_items = current_exclude
```
I guess we don't need this cast here.

cosmos/airflow/graph.py
```diff
@@ -121,12 +133,19 @@ def exclude_detached_tests_if_needed(
...
+    else:
+        exclude_items = []
     tests_detached_from_this_node: list[DbtNode] = detached_from_parent.get(node.unique_id, [])  # type: ignore
     for test_node in tests_detached_from_this_node:
-        exclude.append(test_node.resource_name.split(".")[0])
-    if exclude:
-        task_args["exclude"] = exclude  # type: ignore
+        exclude_items.append(test_node.resource_name.split(".")[0])
+    if exclude_items:
+        task_args["exclude"] = " ".join(exclude_items)
```
```suggestion
        task_args["exclude"] = _convert_list_to_str(exclude_items)
```

pankajkoti (approved):
LGTM. Minor suggestions inline if you'd like to consider

PR #2328, opened 2026-02-05
Fix selecting model when it has the same name as folder

pankajastro:
closes: https://github.com/astronomer/astronomer-cosmos/issues/1093

Before Changes
Screenshot 2026-02-05 at 9 34 35 PM

After Changes
Screenshot 2026-02-05 at 9 29 41 PM

PR #2371, opened 2026-02-17
Docs: include/exclude node based on fqn

pankajastro:
Document the `fqn` include/exclude option introduced in PR: https://github.com/astronomer/astronomer-cosmos/pull/2328 and https://github.com/astronomer/astronomer-cosmos/pull/2375

### DIFF
--- a/docs/configuration/selecting-excluding.rst
+++ b/docs/configuration/selecting-excluding.rst
@@ -34,6 +34,10 @@ The ``select`` and ``exclude`` parameters are lists, with values like the follow
 - ``exposure:my_exposure``: include/exclude nodes that have the exposure ``my_exposure`` and are of resource_type ``exposure``
 - ``exposure:+my_exposure``: include/exclude nodes that have the exposure ``my_exposure`` and their parents
 - ``fqn:some_model``: include/exclude nodes based on their fully qualified names (FQN), which consist of the project name, folder path, and model name. For example, ``fqn:my_dbt_project.analytics.tables.my_model`` selects the model ``my_model`` in the ``analytics/tables`` folder of the ``my_dbt_project`` project.
+- ``package:package_name``: include/exclude all nodes that belong to the given package (e.g. ``package:dbt_artifacts``). The package name must be non-empty (use ``package:dbt_artifacts``, not ``package:``).
+- ``package:package_name+``: include/exclude all nodes in the package and their descendants (children).
+- ``+package:package_name``: include/exclude all nodes in the package and their ancestors (parents).
+- A bare name without a method prefix (e.g. ``dbt_artifacts`` or ``child``) is resolved like dbt: it matches nodes by package name, node name, or path segment (folder name). So ``select=['folder_a']`` or ``exclude=['folder_a']`` includes or excludes all models under a folder named ``folder_a``, including when using ``LoadMode.DBT_MANIFEST``.
 
 .. note::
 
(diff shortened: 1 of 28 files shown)


docs/configuration/selecting-excluding.rst
```diff
@@ -116,6 +116,19 @@ Examples:
...
 
+.. code-block:: python
+
+    from cosmos.airflow.dag import DbtDag
+    from cosmos.config import RenderConfig
+
+    jaffle_shop = DbtDag(
+        render_config=RenderConfig(
+            select=[
+                "fqn:fqn:jaffle_shop.foo.foo"
```
pankajkoti: Is the two times `fqn:fqn` correct and expected here?
```suggestion
                "fqn:jaffle_shop.foo.foo"
```

pankajastro: oops, thanks for pointing it

docs/configuration/selecting-excluding.rst
```diff
@@ -116,6 +116,19 @@ Examples:
         )
     )
 
+.. code-block:: python
+
+    from cosmos.airflow.dag import DbtDag
+    from cosmos.config import RenderConfig
```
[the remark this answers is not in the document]
pankajastro: The full path is optimised version for memory uses so make sense to use full path in the docs

pankajastro:
> @pankajastro the changes we introduced in #2328 only affected Cosmos task execution, and not Cosmos custom selectors that are used with `LoadMode.MANIFEST` and `LoadMode.CUSTOM` to decide which nodes will be rendered in the DAG, to define its topology.
> 
> Therefore, this change in the documentation is not accurate. If users attempted to do the following on a project loaded with `LoadMode.MANIFEST`, Cosmos would error:
> 
> ```
> DbtDag(
>     ...
>     select=["fqn:my_dbt_project.analytics.tables.my_model"]
> )
> ```
> 
> For the feature you're documenting to work, we'd need to modify: https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/dbt/selector.py
> 
> While this change would not be very complex, I'd say it is a new feature that would need to be implemented. One aspect for this feature would be to make sure we analyse a few manifest files to confirm the representation of the fqn field.
> 
> As an example, I just checked a local manifest, and I observed it looked like this:
> 
> ```
>   "nodes": {
>     "model.BIC_C5.AcctC5LuTrnf_C5HHAct_stg": {
>       "resource_type": "model",
>       "unique_id": "model.BIC_C5.AcctC5LuTrnf_C5HHAct_stg",
>       "database": "SANDBOX",
>       "schema": "TATIANAALCHUEYR",
>       "name": "AcctC5LuTrnf_C5HHAct_stg",
>       "package_name": "BIC_C5",
>       "fqn": [
>         "BIC_C5",
>         "AcctC5LuTrnf",
>         "AcctC5LuTrnf_C5HHAct_stg"
>       ]
> ```
> 
> Which means we'd need to join the list in each model's "fqn". Something to test would also be whether dbt supports relative `fqn` paths.
> 
> Therefore, I suggest we close this PR for now and log a ticket to add this feature.

Thanks for the review and pointing out this issue, @tatiana. I have drafted PR: https://github.com/astronomer/astronomer-cosmos/pull/2375 to address it

tatiana (approved):
Thanks, @pankajastro , for implementing the feature, happy to merge the docs now!

PR #2372, opened 2026-02-17
Set correct priority for queue in WatcherMode producer task

pankajastro:
This PR fixes the queue configuration priority for the WatcherMode producer task. Previously, the producer task would always use `watcher_dbt_execution_queue` or `DEFAULT_QUEUE`, ignoring any queue specified in `setup_operator_args`. The change ensures that if a queue is explicitly provided via `setup_operator_args`, it takes precedence over the global `watcher_dbt_execution_queue` configuration.

**Changes:**
- Modified queue assignment logic in `DbtProducerWatcherOperator.__init__` to respect `setup_operator_args` queue parameter
- Added test coverage verifying queue precedence when both `setup_operator_args` queue and `watcher_dbt_execution_queue` are set
- Added test coverage verifying fallback to `watcher_dbt_execution_queue` when only that is configured

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -103,7 +103,7 @@ class DbtProducerWatcherOperator(DbtBuildMixin, DbtLocalBaseOperator):
         default_args["retries"] = 0
         kwargs["default_args"] = default_args
         kwargs["retries"] = 0
-        kwargs["queue"] = watcher_dbt_execution_queue or DEFAULT_QUEUE
+        kwargs["queue"] = kwargs.get("queue") or watcher_dbt_execution_queue or DEFAULT_QUEUE
         super().__init__(task_id=task_id, *args, **kwargs)
 
         if self.invocation_mode == InvocationMode.SUBPROCESS:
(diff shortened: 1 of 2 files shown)


cosmos/operators/watcher.py
```diff
@@ -103,7 +103,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         default_args["retries"] = 0
         kwargs["default_args"] = default_args
         kwargs["retries"] = 0
-        kwargs["queue"] = watcher_dbt_execution_queue or DEFAULT_QUEUE
+        kwargs["queue"] = kwargs.get("queue") or watcher_dbt_execution_queue or DEFAULT_QUEUE
```
pankajkoti, line 106: It would be good to mention in the docs the order of precedence from the various ways of setting this value.

pankajkoti (approved):
LGTM

tatiana (approved):
Thanks for improving this, @pankajastro ! I agree with @pankajkoti on the docs update, if possible, on top of these improvements: https://github.com/astronomer/astronomer-cosmos/pull/2380

pankajastro:
> Thanks for improving this, @pankajastro ! I agree with @pankajkoti on the docs update, if possible, on top of these improvements: #2380

Yes, I'll update this on top of #2380

cosmos/operators/watcher.py
pankajastro: Yes, I'll update this on top of https://github.com/astronomer/astronomer-cosmos/pull/2380

Issue #1572, opened 2025-02-27
Revert installing `methodtools` as a dependency of Cosmos


tatiana:
Follow up to: https://github.com/astronomer/astronomer-cosmos/pull/1567#discussion_r1973340697
https://github.com/apache/airflow/issues/47147
https://github.com/apache/airflow/pull/47148

PR #2378, opened 2026-02-17
Remove methodtools as a dependency

tatiana:
Closes: https://github.com/astronomer/astronomer-cosmos/issues/1572


pankajkoti (approved):
Nice!

PR #2381, opened 2026-02-19
Preserve extra_context for watcher consumer task instances

### LINKED ISSUE #2250: [Feature] `DbtRunWatcherOperator` is missing the `extra_context` context.
### Description

https://github.com/astronomer/astronomer-cosmos/blob/acf6790ca581c9d26ec044fde490625fa7b70aad/cosmos/operators/_watcher/base.py#L73

After we changed the execution mode from `local` to `watcher`, the extra_context context is missing

### Use case/motivation

_No response_

### Related issues

_No response_

pankajkoti:
Store back `extra_context` on `BaseConsumerSensor` after popping from `kwargs`, so subclasses and user customisations can use it at runtime (`BaseSensorOperator` does not accept `extra_context` and hence, it is popped from `kwargs` before calling `super.init`).

closes: https://github.com/astronomer/astronomer-cosmos/issues/2250

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -157,9 +157,13 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
         **kwargs: Any,
     ) -> None:
         self.compiled_sql = ""
-        extra_context = kwargs.pop("extra_context") if "extra_context" in kwargs else {}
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Thanks, @pankajkoti , looks good! It may be worth it for us to cut 1.13.1a1 and share it with the user who reported the bug to confirm that it now works for his use-case now. I'll create the release branch now.

PR #2397, opened 2026-02-24
Update docs contributing code block

lzdanski:
## Description

Added more detail for virtual environment set up to run docs locally. This can help technical writers who might have more experience working with Sphinx in a different project with a different implementation.

## Related Issue(s)

N/A

## Breaking Change?

None expected.

## Checklist

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -155,7 +155,7 @@ To run the checks manually, run:
 Writing Docs
 ____________
 
-[Hatch](https://hatch.pypa.io/latest/) is a unified command-line tool for managing dependencies and environment isolation for Python developers. In Cosmos, we use a Hatchto declare the dependencies required for the project itself, as well as for tests and documentation builds.
+`Hatch <https://hatch.pypa.io/latest/>`_ is a unified command-line tool for managing dependencies and environment isolation for Python developers. In Cosmos, we use a Hatchto declare the dependencies required for the project itself, as well as for tests and documentation builds.
 
 If you don’t already have Hatch installed, please `install it <https://hatch.pypa.io/latest/install/>`_ before proceeding. As an example, on macOS, you can do so with:
 


docs/contributing.rst
```diff
@@ -155,15 +155,16 @@ To run the checks manually, run:
...
 ____________
 
-You can run the docs locally by running the following:
+You can run the docs locally in a virtual environment with the following:
 
 .. code-block:: bash
 
+    python3 -m venv docs_env
+    source docs_env/bin/activate
+    pip install hatch
```
pankajkoti: These commands are not required for subsequent runs, so they could be broken into a pre-requisite section, WDYT? Subsequent runs just need to activate the virtualenv using `source docs_env/bin/activate` and run `hatch run docs:serve`

tatiana (commented):
@lzdanski Since we’re using [Hatch](https://hatch.pypa.io/latest/), which already manages its own virtual environments, requiring contributors to create another virtualenv step may be unnecessary. It could introduce redundancy or confusion, as hatch handles environment isolation for us. Perhaps we can simplify the instructions by relying on Hatch’s built-in environment management rather than a new `docs_env`, and let users install Hatch however they prefer. I'd vote for us not to make the changes proposed in this PR.

docs/contributing.rst
```diff
@@ -155,15 +155,16 @@ To run the checks manually, run:
...
     hatch run docs:serve
 
-
-This will run the docs server in a virtual environment with the right dependencies. Note that it may take longer on the first run as it sets up the virtual environment, but will be quick on subsequent runs.
-
+This will run the docs server in a virtual environment called ``docs_env`` with the right dependencies. Note that it may take longer on the first run as it sets up the virtual environment, but will be quick on subsequent runs.
```
yetudada: I think it might be handy to give more information, step-by-step about  what these commands do for users i.e. that this creates a Python virtual environment called ``docs_env``, activates it, installs hatch which is used for x and then does y.

lzdanski:
Based on the reviews, I've made the following changes:

1. @tatiana - Thinking ahead to encouraging docs contributions, a deterrent to submitting docs updates can be the complexity of setting up the full local development environment, compared to how straightforward the local docs builds can be. I've split the docs contributing into two scenarios: 1. If someone already has their local Cosmos development environment set up, they can just run hatch. 2. If someone *only* wants to work on docs, and doesn't have anything set up, which includes creating the docs_venv.

2. Added the recommended changes from @pankajkoti + @yetudada.

docs/contributing.rst
```diff
@@ -155,15 +155,36 @@ To run the checks manually, run:
 Writing Docs
 ____________
 
-You can run the docs locally by running the following:
+If you already have your environment set up for local Cosmos development and have Hatch installed, you can run the docs locally in a virtual environment with the following:
```
tatiana:
```suggestion
[Hatch](https://hatch.pypa.io/latest/) is a unified command-line tool for managing dependencies and environment isolation for Python developers. In Cosmos, we use a Hatchto declare the dependencies required for the project itself, as well as for tests and documentation builds.

If you don’t already have Hatch installed, please [install it](https://hatch.pypa.io/latest/install/)before proceeding. As an example, on macOS, you can do so with:

`brew install hatch`
```

docs/contributing.rst
```diff
@@ -155,15 +155,36 @@ To run the checks manually, run:
...
+    # Creates a local instance of your docs
+    hatch run docs:serve
+
+#. For future docs local builds, you can reactivate the virtual environment and launch hatch:
 
+.. code-block:: bash
+    # Activate the virtual environment
+    source docs_env/bin/activate
+    # Create a local instance of your docs
+    hatch run docs:serve
```
tatiana:
```suggestion
```

docs/contributing.rst
```diff
@@ -155,14 +155,15 @@ To run the checks manually, run:
...
 ____________
 
-You can run the docs locally by running the following:
+[Hatch](https://hatch.pypa.io/latest/) is a unified command-line tool for managing dependencies and environment isolation for Python developers. In Cosmos, we use a Hatchto declare the dependencies required for the project itself, as well as for tests and documentation builds.
 
-.. code-block:: bash
+If you don’t already have Hatch installed, please [install it](https://hatch.pypa.io/latest/install/)before proceeding. As an example, on macOS, you can do so with:
 
-    hatch run docs:serve
+`brew install hatch`
```
lzdanski:
```suggestion
``brew install hatch``
```

docs/contributing.rst
```diff
@@ -155,14 +155,15 @@ To run the checks manually, run:
...
 ____________
 
-You can run the docs locally by running the following:
+[Hatch](https://hatch.pypa.io/latest/) is a unified command-line tool for managing dependencies and environment isolation for Python developers. In Cosmos, we use a Hatchto declare the dependencies required for the project itself, as well as for tests and documentation builds.
 
-.. code-block:: bash
+If you don’t already have Hatch installed, please [install it](https://hatch.pypa.io/latest/install/)before proceeding. As an example, on macOS, you can do so with:
 
-    hatch run docs:serve
+``brew install hatch``
```
tatiana:
```suggestion
.. code-block:: bash
    brew install hatch
```

docs/contributing.rst
```diff
@@ -155,14 +155,15 @@ To run the checks manually, run:
...
 
+.. code-block:: bash
+    hatch run docs:serve
```
tatiana, line 169:
```suggestion

You can run the docs locally by running the following:

.. code-block:: bash
    hatch run docs:serve
```

This will run the docs server in a virtual environment with the right dependencies. Note that it may take longer on the first run while it sets up the virtual environment, but subsequent runs will be quick.

tatiana: Sorry, @lzdanski , just noticed we removed some important bits that give additional context

docs/contributing.rst
tatiana: This was incorrectly formatted as Markdown; ^ fix for reStructuredText

docs/contributing.rst
```diff
@@ -155,14 +155,19 @@ To run the checks manually, run:
 Writing Docs
 ____________
 
-You can run the docs locally by running the following:
+[Hatch](https://hatch.pypa.io/latest/) is a unified command-line tool for managing dependencies and environment isolation for Python developers. In Cosmos, we use a Hatchto declare the dependencies required for the project itself, as well as for tests and documentation builds.
```
tatiana:
```suggestion
`Hatch <https://hatch.pypa.io/latest/>`_ is a unified command-line tool for managing dependencies and environment isolation for Python developers. In Cosmos, we use a Hatchto declare the dependencies required for the project itself, as well as for tests and documentation builds.
```

docs/contributing.rst
```diff
@@ -155,14 +155,19 @@ To run the checks manually, run:
...
+
+If you don’t already have Hatch installed, please [install it](https://hatch.pypa.io/latest/install/)before proceeding. As an example, on macOS, you can do so with:
```
tatiana:
```suggestion
If you don’t already have Hatch installed, please `install it <https://hatch.pypa.io/latest/install/>`_ before proceeding. As an example, on macOS, you can do so with:
```

docs/contributing.rst
tatiana: @yetudada would you like any additional changes before we merge it?

PR #2409, opened 2026-02-26
Fix: Attempt deterministic dbt project copy in test fixtures

pankajkoti:
Ignore inconsistent local files and directories like `.user.yml`, `.DS_Store`, `logs`, and `target` when copying the dbt project, so folder hashes (e.g. in `test_save_dbt_ls_cache`) are stable across clones and machines.

related: https://github.com/astronomer/astronomer-cosmos/pull/2406/


pankajkoti:
> Thanks a lot, @pankajkoti , this will certainly help us make the hash test more stable. Would it make sense to isolate the list with the filenames in a constant, so its easier to extend it over time?

yes. moved to a constant, thanks!

PR #2412, opened 2026-02-26
logs dbt-core and adapter in watcher consumer task

pankajastro:
Related to: https://github.com/astronomer/astronomer-cosmos/issues/2072

Screenshot 2026-02-27 at 1 04 53 AM


cosmos/operators/_watcher/triggerer.py
```diff
@@ -140,8 +140,26 @@ async def _get_producer_task_status(self) -> str | None:
 
         return await sync_to_async(fetch_state)()
 
+    async def _wait_and_log_startup_versions(self) -> None:
```
tatiana: @pankajastro What happens if the producer fails before invoking the dbt command, and this XCom is never populated? Should we check for the producer status as well?

cosmos/operators/_watcher/triggerer.py
```diff
@@ -140,8 +140,26 @@ async def _get_producer_task_status(self) -> str | None:
...
+        while True:
+            events = await self.get_xcom_val(DBT_STARTUP_EVENTS_XCOM_KEY)
+            if isinstance(events, list) and events:
+                for ev in events:
+                    name, msg = ev.get("name"), ev.get("msg") or ""
+                    if name == "MainReportVersion" and "dbt=" in msg:
+                        logger.info("%s", msg)
+                    elif name == "AdapterRegistered" and "Registered adapter: " in msg:
+                        logger.info("%s", msg)
+                return
```
tatiana: At the moment, we are injecting two events into this XCom - one at a time:
- "MainReportVersion"
- "AdapterRegistered"

Isn't there a risk that the consumer would consume only one of those events (e.g. MainReportVersion) because the other log hadn't been parsed (AdapterRegistered) - and the other event would be added just after this look was processed, which would mean the consumer would never print the second line?

It may make sense to accumulate the status in the producer node before injecting them to XCom, similar to what was done with the tests of a model in https://github.com/astronomer/astronomer-cosmos/pull/2318

tatiana (commented):
@pankajastro Thanks a lot for this, it will be great to add more visibility to the consumer logs! I left some feedback inline. My last question is:
- Does the current implementation work with both `InvocationMode.SUBPROCESS` and `InvocationMode.DBT_RUNNER`?

cosmos/operators/_watcher/triggerer.py
```diff
@@ -140,8 +140,26 @@ async def _get_producer_task_status(self) -> str | None:
...
+        """Wait for dbt_startup_events from producer (pushed early in runner callback; from log in subprocess) and log versions."""
+        while True:
+            events = await self.get_xcom_val(DBT_STARTUP_EVENTS_XCOM_KEY)
+            if isinstance(events, list) and events:
+                for ev in events:
+                    name, msg = ev.get("name"), ev.get("msg") or ""
+                    if name == "MainReportVersion" and "dbt=" in msg:
+                        logger.info("%s", msg)
+                    elif name == "AdapterRegistered" and "Registered adapter: " in msg:
+                        logger.info("%s", msg)
```
pankajkoti: Do we already log these when using the non-deferred sensors in subprocess mode? Or do we not support that yet and are planning to handle it separately?

cosmos/operators/_watcher/triggerer.py
pankajkoti: +1

cosmos/operators/watcher.py
```diff
@@ -117,6 +118,7 @@ def _serialize_event(event_message: EventMsg) -> dict[str, Any]:
         return MessageToDict(event_message, preserving_proto_field_name=True)  # type: ignore[no-any-return]
 
     def _handle_startup_event(self, event_message: EventMsg, startup_events: list[dict[str, Any]]) -> None:
+        # TODO: get dbt version
```
pankajkoti: Are we missing the dbt version here?

cosmos/operators/_watcher/triggerer.py
```diff
@@ -140,8 +140,26 @@ async def _get_producer_task_status(self) -> str | None:
...
+                return
+            producer_task_state = await self._get_producer_task_status()
+            if producer_task_state == "failed":
+                return
```
pankajkoti, line 183: It might be good to recheck here and pull the XCOMs again to avoid the race condition that XCOM was not pushed before we processed the above lines, and that it was pushed just before the failed event was available.

I think similarly, it might be good to add a success (& maybe other terminal states) state here as well and return after checking for the XCOMs, just in case there's a scenario where those events may not be available and that we do not wait here forever for those startup events to be available.

cosmos/operators/_watcher/triggerer.py
pankajastro: I intentionally did not add the dbt version logging mechanism in the Poke method (i.e., non-deferrable) because every time the sensor pokes, the logs would appear, which could result in many log outputs for the version.

cosmos/operators/watcher.py
pankajastro: No, I just missed removing it

cosmos/operators/_watcher/triggerer.py
pankajastro: Added them as debug log

cosmos/operators/_watcher/triggerer.py
pankajastro: aha, adjusted the logic to log both

cosmos/operators/_watcher/triggerer.py
pankajastro: I'm tracking status of producer and if it failed or success and startup event not available function return

cosmos/operators/_watcher/triggerer.py
pankajastro: I have adjusted it a bit post your feedback. PTAL

pankajastro:
> * Does the current implementation work with both `InvocationMode.SUBPROCESS` and `InvocationMode.DBT_RUNNER`?

Yes, this work for both `InvocationMode`

cosmos/operators/_watcher/triggerer.py
```diff
@@ -145,8 +145,40 @@ async def _get_producer_task_status(self) -> str | None:
...
+                        adapter_logged = True
+
+                    if producer_task_state in ["failed", "success"]:
+                        return
+
+                # exit only when both events were found
+                if main_logged and adapter_logged:
+                    return
+
+            await asyncio.sleep(self.poke_interval)
```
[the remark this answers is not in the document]
pankajastro: Fixed

cosmos/operators/_watcher/triggerer.py
```diff
@@ -145,8 +145,43 @@ async def _get_producer_task_status(self) -> str | None:
...
+                    return
+
+            # Check producer status after processing events so we never return before
+            # logging the full list. Also ensures we exit if producer finishes before
+            # ever pushing _DBT_STARTUP_EVENTS_XCOM_KEY.
+            producer_task_state = await self._get_producer_task_status()
+            if producer_task_state in ("failed", "success"):
+                return
+
+            await asyncio.sleep(self.poke_interval)
```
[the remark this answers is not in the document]
pankajastro: Fixed

cosmos/operators/_watcher/base.py
```diff
@@ -359,6 +379,15 @@ def use_event(self) -> bool:
...
         raise NotImplementedError("Subclasses should implement this method if `use_event` may return True")
 
+    def _log_startup_events(self, ti: Any) -> None:
+        dbt_startup_events: list[dict[str, Any]] = ti.xcom_pull(
+            task_ids=self.producer_task_id, key=_DBT_STARTUP_EVENTS_XCOM_KEY
+        )
+        if dbt_startup_events:  # pragma: no cover
+            for event in dbt_startup_events:
+                # Adding debug level to avoid redundant logs for non-deferrable mode
+                logger.debug("%s", event.get("msg"))
```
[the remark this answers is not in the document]
pankajastro: I think good suggestion, but we can't track this with `__dbt_startup_events_logged__` because other node events may override it. For the scope of this PR, I'll keep it like this

cosmos/operators/_watcher/triggerer.py
```diff
@@ -145,8 +150,48 @@ async def _get_producer_task_status(self) -> str | None:
...
+                        adapter_logged = True
+
+                if main_logged and adapter_logged:
+                    return
+
+            # Check producer status after processing events so we never return before
+            # logging the full list. Also ensures we exit if producer finishes before
+            # ever pushing _DBT_STARTUP_EVENTS_XCOM_KEY.
+            producer_task_state = await self._get_producer_task_status()
+            if producer_task_state in ("failed", "success"):
```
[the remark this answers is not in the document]
pankajastro: No, I just want to check the "failed", "success" state

cosmos/operators/_watcher/triggerer.py
```diff
@@ -145,8 +150,48 @@ async def _get_producer_task_status(self) -> str | None:
...
+    async def _log_startup_events(self) -> None:
+        """Wait for dbt_startup_events from producer (pushed early in runner callback; from log in subprocess) and log versions."""
+        main_logged = False
+        adapter_logged = False
+
+        while True:
+
+            events = await self.get_xcom_val(_DBT_STARTUP_EVENTS_XCOM_KEY)
+
+            if isinstance(events, list) and events:
```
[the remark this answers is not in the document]
pankajastro: yes, but in most of the cases it should not take more than iteration

cosmos/operators/_watcher/base.py
```diff
@@ -361,6 +381,15 @@ def use_event(self) -> bool:
...
+
```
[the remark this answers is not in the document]
pankajastro: This is to avoid noises for the non-deferrable mode

tatiana (approved):
Thanks a lot for the implementation and for addressing the feedback, @pankajastro ! I believe this is already a significant improvement - and if we notice it is not behaving as expected, we can always improve the implementation over time.

PR #2414, opened 2026-02-27
log dbt nodes error in consumer sensor for runner invocation

pankajastro:
Related to: https://github.com/astronomer/astronomer-cosmos/issues/2072

Screenshot 2026-03-03 at 12 37 55 PM

### DIFF
--- a/tests/operators/test_watcher.py
+++ b/tests/operators/test_watcher.py
@@ -1150,7 +1150,7 @@ class TestDbtConsumerWatcherSensor:
         context = self.make_context(ti)
 
         result = sensor._get_status_from_events(ti, context)
-        assert result is None, None
+        assert result == (None, None)
 
     def test_get_status_from_events_does_not_set_compiled_sql_from_event(self):
         """compiled_sql is no longer in the event payload; it is read from the canonical XCom key in poke()."""


cosmos/operators/_watcher/triggerer.py
```diff
@@ -98,32 +98,35 @@ async def get_xcom_val(self, key: str) -> Any | None:
...
-            status = data_json.get("data", {}).get("run_result", {}).get("status")
+            run_result = data_json.get("data", {}).get("run_result", {})
+            status = run_result.get("status")
+            model_error = run_result.get("message")
         else:
             status = await self.get_xcom_val(status_key)
 
         compiled_sql = await self.get_xcom_val(compiled_sql_key) if status is not None else None
-        return status, compiled_sql
+        return status, compiled_sql, model_error
```
tatiana, line 129: How does this change affect `ExecutionMode.WATCHER_KUBERNETES`?

cosmos/operators/_watcher/triggerer.py
```diff
@@ -145,7 +148,9 @@ async def run(self) -> AsyncIterator[TriggerEvent]:
 
         while True:
             producer_task_state = await self._get_producer_task_status()
-            node_status, compiled_sql = await self._parse_node_status_and_compiled_sql()
+            node_status, compiled_sql, model_error_message = await self._parse_node_status_and_compiled_sql()
```
tatiana, line 151: I was just thinking about something: given that we have this while loop that depends on data generated only by the end of the producer dbt command execution, does that mean consumer tasks never finish until the producer task finishes?

I understand that this behaviour was not introduced in this PR, but we should probably review it

michal-mrazek:
If we can get this work with subprocess, that would make our team very happy 👍

cosmos/operators/_watcher/base.py
```diff
@@ -328,7 +328,7 @@ def use_event(self) -> bool:
     def _get_status_from_events(self, ti: Any, context: Context) -> Any:
         raise NotImplementedError("Subclasses should implement this method if `use_event` may return True")
 
-    def poke(self, context: Context) -> bool:
+    def poke(self, context: Context) -> bool:  # noqa: C901
```
pankajkoti, line 331: Are we planning to address the noqa: C901 in a follow-up PR? Should we log a quick ticket for that so we don't miss it?

cosmos/operators/_watcher/base.py
```diff
@@ -381,4 +382,10 @@ def poke(self, context: Context) -> bool:
         elif status == "success":
             return True
         else:
+            # status == "failed" (or other failure status)
+            if self.use_event():
+                logger.error("%s", log_msg)
```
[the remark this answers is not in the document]
pankajkoti: +1

cosmos/operators/_watcher/base.py
```diff
@@ -381,4 +382,10 @@ def poke(self, context: Context) -> bool:
...
+            else:
+                # TODO: Address this post PR: https://github.com/astronomer/astronomer-cosmos/pull/2318
```
pankajkoti, line 389: Looks like PR 2318 is merged now. Happy to take another pass at the PR once this TODO is addressed.

cosmos/operators/_watcher/triggerer.py
pankajkoti: Good question, but I don't think that is the behaviour. We call `_get_producer_task_status` (which could return `None` if not available yet) for the case to handle that the producer failed early, and the corresponding XCOM for the model would never be available. 

But we have the `_parse_node_status_and_compiled_sql` call that gets the individual model's status and should help wrap up early before the producer finishes up entirely, as producer would have pushed the corresponding model's status as and when available before waiting for itself to fully finish up and the if conditionals below handle these cases.

cosmos/operators/_watcher/triggerer.py
```diff
@@ -145,7 +148,9 @@ async def run(self) -> AsyncIterator[TriggerEvent]:
...
+            if model_error_message:
+                logger.error("%s", model_error_message)
```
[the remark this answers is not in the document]
pankajkoti: +1. We should move this to where we're handling failure statuses and not always log it.

pankajastro:
> It's really exciting to see this Watcher improvement, @pankajastro. It will really help end-users with troubleshooting specific model issues.
> 
> Have you tested this with:
> 
> * `ExecutionMode.WATCHER` with `InvocationMethod.DBT_RUNNER`
> * `ExecutionMode.WATCHER` with `InvocationMethod.SUBPROCESS`
> * `ExecutionMode.KUBERNETES_WATCHER`
>   ?

This PR was aiming for dbt-runner only, as PR title suggest but I have created https://github.com/astronomer/astronomer-cosmos/pull/2431, which address for both subprocess and runner

pankajastro:
Moving this in draft in favour of https://github.com/astronomer/astronomer-cosmos/pull/2431

pankajastro:
closing in favour of https://github.com/astronomer/astronomer-cosmos/pull/2431

PR #2419, opened 2026-03-02
Introduce the concept of an `interceptor` in Cosmos tasks

tatiana:
Sometimes users may need to run commands before the dbt command is executed in Cosmos tasks.

This behaviour is similar to the `callback` feature, but instead of happening after dbt command execution, it happens before.

A use case is to calculate dbt or environment variables at runtime and use them during dbt command execution.

Example:
```
def my_interceptor(context, operator):
    # Modify vars (e.g. add execution date)
    if operator.vars is None:
        operator.vars = {}
    operator.vars["run_date"] = str(context["data_interval_start"].date())
    # Modify env
    if operator.env is None:
        operator.env = {}
    operator.env["MY_VAR"] = "value"

# Example of usage in an operator
dbt_build = DbtBuildLocalOperator(
    ...,
    interceptors=[my_interceptor],
)

# Example of usage in a DbtTaskGroup
dbt_task_group = DbtTaskGroup(
    ...,
    operator_args={"interceptors": [my_interceptor]},
)
  ```

The current implementation works with all existing execution modes:
- Local
- Virtualenv
- Docker
- Kubernetes
- AWS EKS
- AWS ECS
- GCP Cloud Run Job
- Azure Container Instance
- Async (e.g. BigQuery)
- Watcher
- Watcher Kubernetes

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -41,7 +41,7 @@ jobs:
       contents: read
     steps:
       # retrieve your distributions here
-      - uses: actions/download-artifact@v7
+      - uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3
         with:
           name: artifacts
           path: dist
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -779,7 +779,7 @@ jobs:
         run: |
           pip3 install coverage
       - name: Download all coverage artifacts
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3
         with:
           path: ./coverage
       - name: Combine coverage
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -85,7 +85,7 @@ repos:
           - --py310-plus
           - --keep-runtime-typing
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.2
+    rev: v0.15.4
     hooks:
       - id: ruff-check
         args: [ --fix ]

### CODE REVIEW

File: tests/operators/test_local.py [hunk 1]
```diff
@@ -1734,6 +1734,44 @@ def test_build_and_run_cmd_with_full_refresh_in_async_mode():
...
+    assert operator.vars == {"new_var": "new_var_value"}
+    assert operator.env == {"NEW_ENV_VAR": "new_env_var_value"}
+
+    call_kwargs = mock_run_command.call_args[1]
+    cmd = call_kwargs["cmd"]
+    env = call_kwargs["env"]
+
+    assert "--vars" in cmd
+    assert "--vars" == cmd[2]
+    assert "new_var: new_var_value" in cmd[3]
```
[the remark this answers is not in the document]
tatiana: Tests should be simple and have the least amount of logic possible. I prefer my version.

docs/configuration/operator-args.rst
```diff
@@ -89,6 +89,7 @@ dbt-related
 - ``dbt_executable_path``: Path to dbt executable.
 - ``env``: (Deprecated since Cosmos 1.3 use ``ProjectConfig.env_vars`` instead) Declare, using a Python dictionary, values to be set as environment variables when running ``dbt`` commands.
 - ``fail_fast``: ``dbt`` exits immediately if ``dbt`` fails to process a resource.
+- ``interceptors``: (new in v1.14) Optional list of callables run before building the dbt command. Each callable receives ``(context, operator)`` and may modify ``operator.vars`` and ``operator.env``; the modified values are then used when building and running the dbt command. Supported for all execution modes. Useful for injecting runtime vars or env (e.g. from Airflow connections or XCom) per task run.
```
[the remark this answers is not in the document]
tatiana: Since Cosmos documentation is not versioned, we need to be explicit about in which version a feature was introduced.

cosmos/operators/base.py
pankajkoti, line 288: Would it be possible to call the self.invoke_interceptors here? Instead of having it in each of the operators?

cosmos/operators/aws_ecs.py
```diff
@@ -123,6 +123,7 @@ def build_and_run_cmd(
         async_context: dict[str, Any] | None = None,
         **kwargs: Any,
     ) -> Any:
+        self.invoke_interceptors(context)
```
pankajkoti, line 126: wondering if it's possible to call this in the `operators/base.py` 's `build_cmd` method so that we have it only once instead of having to copy it in each of the operators?

pankajastro: +1

tatiana:
Thanks a lot for the review, @pankajkoti !

> And maybe you have thought of this already and have your rationale, but wondering if the name `injectors` could suit better instead of `interceptors` for the use case we're tackling here?

Interceptor typically refers to a mechanism that observes, modifies, or blocks an operation at a well-defined point in its lifecycle. It often implies hooking into a control flow to inspect or change something - exactly what this PR does (it lets users run code before dbt commands). Interceptors are common in middleware, request pipelines, event systems, and AOP (aspect-oriented programming).

Injectors, on the other hand, usually suggest adding or providing something (usually a dependency). In a dependency injection system, injectors provide objects with what they need; they typically don't observe or intervene in behaviour. The term focuses more on the insertion of dependencies than on the control of execution.

I think in this particular feature, interceptor is a better fit because:
1. It hooks into the execution flow, running beofre dbt commadns are built and executed, what interceptors are meant to do
2. It implies extension/modification of behaviour, not just insertion of data. Users could modify existing environment variables, dbt variables, and even the Airflow context, potentially changing the operator's behaviour.
3. Conceptually, interceptors sit along callbacks as lifecycle hooks

In contrast, injector would imply that the primary purpose is to insert something (like environment variables or config), which is part of what the new hook does, but not its defining behaviour.

We could evolve the current version of the interceptor to, depending on its output, not even run the dbt command. At the moment, users could accomplish this by raising an exception, for instance. I'm curious about the community's feedback and how they'll use it.

cosmos/operators/aws_ecs.py
tatiana: From a code-repetition perspective, I agree it would be better placed, but from a conceptual perspective, I believe the duplication makes sense.

The main reason why I believe we should not place this inside `build_cmd` is that this feature can be used for more than just building the command line being run. It can change the behaviour of what happens before - and even block dbt from running.

Ideally, we'd have interceptors just before executing the dbt command, as in the callback implementation. When I tried to implement this approach, however, I faced a few challenges while trying to implement this approach
- Extending variables was a more complex operation, since they were pre-build as part of `build_cmd` (`add_global_flags`)
- It would be even more specific for each execution mode, since it would be deep down on the actual execute method
- It was much more complex than necessary, for now

I strongly believe we have significant potential to refactor the current flow for building commands and running them in Cosmos, and maybe find an interface that would reduce duplication. The current code flow is very complex, and we are actually changing command flags not only in the `build_cmd` part but also in the `run_cmd` method, since many flags are only calculated by then. Unfortunately, this will represent a larger refactor that may break how users subclass and use Cosmos, so I believe we'll need Cosmos 2.0.

cosmos/operators/base.py
tatiana: Discussion in another thread: https://github.com/astronomer/astronomer-cosmos/pull/2419#discussion_r2873063922

cosmos/operators/aws_ecs.py
pankajkoti: Yes, happy with the reasoning, thanks for the details!

pankajkoti:
> Injectors, on the other hand, usually suggest adding or providing something (usually a dependency). In a dependency injection system, injectors provide objects with what they need; they typically don't observe or intervene in behaviour. The term focuses more on the insertion of dependencies than on the control of execution.

> In contrast, injector would imply that the primary purpose is to insert something (like environment variables or config), which is part of what the new hook does, but not its defining behaviour.

I see the association with dependency injection, which does come up in framework discussions. I am reading that injection as a term can be used more broadly across domains, for example, in Code injection, Command injection,  Script injection, SQL injection, Chaos injection, fault injection (in testing), environment variable injection, feature flag injection, sidecar injection in Kubernetes, or trace context (trace ID) injection in observability, all of which influence runtime behaviour. Even dependency injection itself ultimately changes behaviour by supplying different implementations.

I support `interceptor` :), I think it fits well, just wanted to share why `injector` also felt reasonable to me and why it came to mind.

PR #2424, opened 2026-03-03
[DRAFT] Add Reference docs section

lzdanski:
## Description

This PR moves the `Profiles` section into a sub-directory of the new `Reference` section in the docs.

This also includes the Configuration reference materials.

The "Profiles" sidebar navigation has this structure, and can't be collapsed in the sidebar navigation, because it uses refs generated by `profiles_mapping.py` in the toctree instead of the file names/directories.

Screenshot 2026-03-02 at 9 49 16 PM

**Remaining work**

- Refactor Profiles text docs (separate PR)
- Redirects
- Address test build errors - resulting from a comment in profiles page, now that they are in a different directory

## Checklist

### DIFF
--- a/docs/reference/configs/index.rst
+++ b/docs/reference/configs/index.rst
@@ -12,4 +12,3 @@ Configuration References
    Profile Config <profile-config>
    Execution Config <execution-config>
    Cosmos Config <cosmos-conf>
-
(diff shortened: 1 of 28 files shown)


lzdanski:
> @lzdanski Do we need to check in those profile mapping docs files, given that they’re auto-generated?

I don't think we need them. I've also updated the template to use the two code backticks anyways, which was triggering the failure.

lzdanski:
Closing PR due to a git rebase issue.

Issue #2411, opened 2026-02-26
Remove the pin on virtualenv<21 in CI


pankajkoti:
Once https://github.com/pypa/hatch/issues/2193 is resolved, remove the pin for virtualenv<21 in CI setups added in PR https://github.com/astronomer/astronomer-cosmos/pull/2410

related: https://github.com/pypa/hatch/issues/2193
related: https://github.com/astronomer/astronomer-cosmos/pull/2410

tatiana:
A new version of Hatch is compatible with the latest virtualenv package was released:
https://github.com/pypa/hatch/releases/tag/hatch-v1.16.5

So we should be able to remove that pin 🎉

pankajkoti:
Yes, had subscribed to the hatch issue and got notified earlier. Hence, got the issue in in this sprint!

PR #2426, opened 2026-03-03
Revert virtualenv pin for hatch installation in CI 

pankajkoti:
closes: https://github.com/astronomer/astronomer-cosmos/issues/2411

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main,revert-virtualenv-pin]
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,revert-virtualenv-pin]
```
```suggestion
    branches: [main]
```

pankajkoti:
successful run on the push target: https://github.com/astronomer/astronomer-cosmos/actions/runs/22620012701/job/65542433934

Issue #2415, opened 2026-02-27
[Bug] `TestBehavior.NONE` does not work with selectors in Watcher Exec Mode


anissarashid:
### Astronomer Cosmos Version

1.14.0a2

### dbt Core or Fusion version

1.10.11

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

WATCHER

### InvocationMode

None

### airflow version

2.10.5

### Operating System

n/a

### Deployment

Astronomer

### What happened?

The producer task in watcher execution mode uses `dbt build`, and when `TestBehavior` is set to `NONE`, the task runs the dbt command with `'--exclude', 'resource_type:test resource_type:unit_test',`. However, if we use a dbt selector, the `exclude` statement is ignored, so the dbt build will continue to run tests. 

For example, this is a command that the producer task will run: 
```
Running command: ['/usr/local/airflow/dbt_venv/bin/dbt', 'build', '--exclude', 'resource_type:test resource_type:unit_test', '--selector', 'cx', '--log-format', 'json', '--log-level', 'info', '--project-dir', '/tmp/tmp8aw2skza', '--profiles-dir', '/tmp/cosmos/profile/fb57e81874cbc32d98158fc13ea02670490dfffd51e63b4f61397bc7b9e6c6ca', '--profile', 'cosmos-dbt', '--target', 'prod']
[2026-02-27, 15:00:58 UTC] {subprocess.py:98} INFO - Command output:
[2026-02-27, 15:01:00 UTC] {base.py:71} INFO - 15:01:00  Running with dbt=1.10.11
[2026-02-27, 15:01:01 UTC] {base.py:71} INFO - 15:01:01  Registered adapter: ***flake=1.10.0
```

and from the dbt output, I can see that tests are still being run from the command, for example: 
```
[2026-02-27, 15:02:19 UTC] {base.py:71} INFO - 15:02:19  19 of 127 START test not_null_stg_playvox__business_role_business_role_id ...... [RUN]
```

I think the solution here would be to use a `dbt run` command instead of `build` when test behavior is set to none for the producer task.

### Relevant log output

```shell

```

### How to reproduce

create a DbtTaskGroup using:
* Watcher execution mode
* RenderConfig with a dbt `selector` and `TestBehavior.NONE`

observe that tests still run in the `dbt_producer_watcher` task

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Contact Details

anissa.rashid@findheadway.com

[...]

PR #2428, opened 2026-03-03
override the build command and use run when watcher execution test behavior is set to NONE or AFTER_ALL

anissarashid:
## Description
Watcher execution mode uses `dbt build` as its base operator, and uses `--exclude` flags when the test behavior is set to `NONE` or `AFTER_ALL` to exclude tests from the producer task. When we use `selector`, the `exclude` flags are ignored by dbt, so tests will be run in unintentionally in the producer task. 

This updates the watcher operator to conditionally use the `run` command, and override the `build_cmd` if set to true.

## Related Issue(s)

Closes: #2415

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -694,12 +694,17 @@ def _add_watcher_producer_task(
         producer_task_args["exclude"] = _convert_list_to_str(render_config.exclude)
 
         if render_config.test_behavior in [TestBehavior.NONE, TestBehavior.AFTER_ALL]:
-            additional_excludes = "resource_type:test resource_type:unit_test"
-            current_exclude = producer_task_args.get("exclude")
-            if current_exclude:
-                producer_task_args["exclude"] = f"{current_exclude} {additional_excludes}"
+            # Use `dbt run` instead of `dbt build` so tests are not executed by the
+            # producer (dbt ignores --exclude with --selector).
+            if render_config.selector:
+                producer_task_args["use_run_command"] = True
             else:
-                producer_task_args["exclude"] = additional_excludes
+                additional_excludes = "resource_type:test resource_type:unit_test"
+                current_exclude = producer_task_args.get("exclude")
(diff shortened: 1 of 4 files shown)


johnhoran:
Could I suggest looking at the https://docs.getdbt.com/reference/global-configs/resource-type?version=1.11 flag in the build command instead of switching to dbt run.

>The build task supports standard selection syntax (--select, --exclude, --selector), as well as a --resource-type flag that offers a final filter 

I tried running a dbt build with select and selectors and it ran all expected nodes excluding tests.
```
dbt build --resource-type "analysis exposure metric model saved_query seed semantic_model snapshot source" --select ....
```
DBT 1.8 also included a exclude-resource-type flag.  @tatiana I'm not sure what the lower bound of DBT support is in cosmos.  exclude-resource-type would be easier to maintain if its greater than 1.8.

pankajkoti:
Thank you, @anissarashid, for reporting the issue and offering to contribute this fix.

Thank you @johnhoran for bringing to attention the `--resource-type` flag being respected with dbt build. Since Cosmos supports dbt 1.5+ as documented in the policy https://github.com/astronomer/astronomer-cosmos/blob/75365d1ef4f36380277e4434fbd79147578e2980/docs/policy/compatibility-policy.rst?plain=1#L50, we unfortunately can't use `exclude-resource-type` yet.

As we would like to fix the issue https://github.com/astronomer/astronomer-cosmos/issues/2415 on priority and ship the fix in our upcoming release 1.14.0, this issue is assigned to me on priority. I have followed the approach to use `--exclude-type` and created a new PR to address this issue: https://github.com/astronomer/astronomer-cosmos/pull/2511

For your contributions here, with your permissions, I would also like to add you both as co-authors in the new PR. I have tried to get your emails from the public GitHub API, but please let me know if you'd like me to correct those that are currently reflected in the new PR's description. Thanks a lot again for your contributions! 🚀

PR #2430, opened 2026-03-04
Watcher skip on producer failures

### LINKED ISSUE #2429: [Bug] Watcher pattern producer max retries forced to zero
### Astronomer Cosmos Version

1.13.1

-

### Versions of dbt adapters

I'm just wondering what the motivation is to force the max-retries to zero on the producer in the watcher pattern.

https://github.com/astronomer/astronomer-cosmos/blob/5a9b0881751c84c3118b1fdb0424ab7ff1267bdf/cosmos/operators/watcher_kubernetes.py#L86-L88
https://github.com/astronomer/astronomer-cosmos/blob/5a9b0881751c84c3118b1fdb0424ab7ff1267bdf/cosmos/operators/watcher.py#L107-L109

I get that in the watcher mode if the producer fails then subsequent runs of the producer do noting and its the individual tasks that take over, and that a retry of the producer does nothing.  But it should still retry so that the whole task group doesn't fail if the individual tasks succeed.  Otherwise we will always need manual intervention to clear the producer task even if all the watcher tasks have rerun and succeeded.

### LoadMode

AUTOMATIC

### ExecutionMode

WATCHER_KUBERNETES

### InvocationMode

None

### airflow version

2.13

### Operating System

-

_No response_

### Deployment

Astronomer

### Deployment details

_No response_

### What happened?

A bug happened!

### Relevant log output

```shell

```

### How to reproduce

-

### Anything else :)?

_No response_

### Contact Details

_No response_

johnhoran:
## Description

Ran into an issue where the watcher_producer task failed (due to pod preemption), the sensor tasks retried and ran successfully but then the dagrun failed because the watcher_producer had failed.

The solution I'm proposing is that the producer should handle failures by converting its exceptions into `AirflowSkipException`, the watcher sensors should treat these as failures and act accordingly.  I've also added an empty operator, its purpose is to stop the skip state from propagating to any downstream tasks.

I also think the empty watcher gate should be downstream of all the DBT tasks, at least when depends_on_past is set to true, and that if `depends_on_past` is set to true for the tasks, then `wait_for_downstream` should also be true for the tasks that are upstream of the gate task.  The logic here is so we can consider the task group as one atomic function.  If wait_for_downstream isn't set, then its possible for one producer task to fail, and cause the sensors to retry, but allow another producer task to start running.  In the current state I guess we are basically relying on the person on call to be aware of this possibility and not clear tasks to avoid it.  

## Related Issue(s)
closes #2429
## Checklist


johnhoran:
I'm moved the gate task so it is downstream from the producer and all the watcher tasks.  This was because if somebody was using wait_for_downstream then the gate task would allow the producer to run on a subsequent run.  But I'm wondering if wait_for_downstream should be always enabled for the producer and watcher tasks if depends_on_past is true?  Otherwise we could have a situation where the producer failed for one run and the watchers have taken over, then the producer for the next run starts...

tatiana:
@johnhoran how do you feel this approach relates to the one in #2283? 
I still have the hope we'll be able to find a good solution for running retries more gracefully in the producer itself (https://github.com/astronomer/astronomer-cosmos/issues/1978#issuecomment-3627431680)

tatiana:
Hi @johnhoran - sorry for not following up after our chat on 19 March.

I’ve discussed this with the team, and there’s some hesitation about further modifying the `DbtDag` / `DbtTaskGroup` topology by introducing an additional “gateway” task.

In the meantime, we addressed the problematic default reported in #2429 with the fix in PR #2479.

I agree that having the watcher producer marked as skipped during retries - while still allowing its downstream tasks to run normally - would be a cleaner approach than marking it as success when it hasn’t actually done any work.

What do you think about marking the producer as skipped and configuring the downstream consumer tasks with `trigger_rule=TriggerRule.ALL_DONE`, so they continue to run even if the parent task is skipped?

johnhoran:
The consumer tasks aren't downstream of the producer one, otherwise they won't start until the producer task completes.  And making them downstream would be unfortunate because then you'd be waiting until the producer task completes before you start working though the consumer tasks.  

You could maybe have the very last consumers in the task group be downstream of the producer task and use NONE_FAILED trigger rule.  I'm still hesitant about this for selfish reasons just because some of our task groups are very flat so essentially it's the problem above.  

Other than that, I think it could work.  My other concern was around the `wait_for_downstream`, but that should be okay.

johnhoran:
Also, and I'm probably wrong about this.  But I thought the reason why retries was set to zero was so you could have https://github.com/astronomer/astronomer-cosmos/blob/513f0ed12835bf41c9a4aef5fcfa17e6f68aa84f/cosmos/operators/_watcher/triggerer.py#L245-L260

If the producer task fails, and then retries and gets marked as successful because its a retry and not doing anything, won't the consumer tasks also succeed?  Or at least there might be a race condition where if a consumer is running while the producer fails then its okay, but if it is waiting on an upstream task, then by the time it gets its turn the producer has retried and succeeded.

PR #2446, opened 2026-03-07
Minor code improvements

dnskr:
## Description

This PR consolidates several commits aimed at resolving minor defects and refactoring for better code quality.

## Related Issue(s)

N/A

## Breaking Change?

No

## Checklist


cosmos/profiles/databricks/token.py
```diff
@@ -45,7 +45,3 @@ def profile(self) -> dict[str, Any | None]:
             # token should always get set as env var
             "token": self.get_env_var_format("token"),
         }
-
-    def transform_host(self, host: str) -> str:
-        """Removes the https:// prefix."""
-        return host.replace("https://", "")
```
pankajkoti, line 51: Why are we removing this?

cosmos/profiles/trino/base.py
```diff
@@ -46,7 +46,3 @@ def mock_profile(self) -> dict[str, Any]:
         mock_profile = super().mock_profile
         mock_profile["port"] = 99999
         return mock_profile
-
-    def transform_host(self, host: str) -> str:
-        """Replaces http:// or https:// with nothing."""
-        return host.replace("http://", "").replace("https://", "")
```
pankajkoti, line 52: same question. Why are we removing this?

cosmos/config.py
```diff
@@ -331,8 +331,6 @@ def get_profile_type(self) -> str:
             target_type = profile["outputs"][self.target_name]["type"]
             return str(target_type)
 
-        return "undefined"
-
```
pankajkoti, line 335: Are we removing this because the code may never reach here?

cosmos/profiles/databricks/token.py
tatiana: @pankajkoti it seems it's because it was declared but never used:
```
$ grep -r transform_host cosmos                                  
cosmos/profiles/trino/base.py:    def transform_host(self, host: str) -> str:
cosmos/profiles/databricks/token.py:    def transform_host(self, host: str) -> str:
```

cosmos/profiles/trino/base.py
tatiana: Same here: https://github.com/astronomer/astronomer-cosmos/pull/2446/changes#r2904419251

cosmos/profiles/databricks/token.py
pankajkoti: @tatiana I believe it gets used here: https://github.com/astronomer/astronomer-cosmos/blob/313328b7aa761f6d1599fa3ab80df54303b185fb/cosmos/profiles/base.py#L311

cosmos/config.py
dnskr: Yes, this line of code is unreachable.

cosmos/profiles/databricks/token.py
dnskr: Good catch, thanks! I'll revert the change and see if I can refactor the code using a different pattern. The current implementation is not so obvious and cannot be detected by static analysis tools.

dnskr:
> Thanks a lot for the code cleanup, @dnskr , these look great! Please, could you fix the checks that are failing?

Thank you for the review!
The tests should pass now (tested locally) that I've reverted the commit [Remove unused transform_host methods](https://github.com/astronomer/astronomer-cosmos/pull/2446/changes/1687dcfc274c7d842480bb80019fb370c8a21eeb).

tatiana:
Very well spot, @pankajkoti - and thanks a lot @dnskr for addressing the feedback so quickly!

Issue #2441, opened 2026-03-06
[Feature] Implement DbtTestWatcherOperator


michal-mrazek:
### Description

Currently, DbtTestWatcherOperator is just empty operator. But since we are already aggregating test results in the producer task, we can just pull XCOM for test operator also with appropriate xcom naming.

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

tatiana:
Thanks for continuing the work started in https://github.com/astronomer/astronomer-cosmos/pull/2318, @michal-mrazek !

michal-mrazek:
@tatiana I think we also need the K8 execution mode but I have never got k8 locally working, can somebody from your team handle that? 🙏

Issue #2427, opened 2026-03-03
[Feature] dbt test node states for WATCHER kubernetes mode


michal-mrazek:
### Description

As part of https://github.com/astronomer/astronomer-cosmos/pull/2318, we are now aggregating tests status in watcher in dbt runner and subprocess. We should do the same for k8 execution mode.

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

vricciardulli:
Hi @michal-mrazek @tatiana I've attempted to support tests for `WATCHER_KUBERNETES` mode in https://github.com/astronomer/astronomer-cosmos/pull/2529

PR #2447, opened 2026-03-08
feat: implement DbtTestWatcherOperator 

michal-mrazek:
## Description

Implementation of the DbtTestWatcherOperator (formerly only EmptyOperator) with a DbtConsumerWatcherSensor subclass that watches aggregated dbt test results. All logic is handled in the BaseConsumerSensor that just checks for is_test_sensor property.

I also added WatcherEventReason enum for type-safe trigger event payloads and changed model->node.

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/2441

## Checklist

### DIFF
--- a/tests/operators/test_watcher.py
+++ b/tests/operators/test_watcher.py
@@ -314,7 +314,9 @@ def test_dbt_consumer_watcher_sensor_execute_complete(event, expected_message):
     )
     sensor.model_unique_id = "model.pkg.m"
 
-    context = {"dag_run": MagicMock()}
+    ti = MagicMock()
+    ti.xcom_pull.return_value = None
+    context = {"dag_run": MagicMock(), "ti": ti}
 
     if expected_message is None:
         sensor.execute_complete(context, event)
@@ -1110,7 +1112,8 @@ class TestDbtConsumerWatcherSensor:
 
         ti = MagicMock()
         ti.try_number = 1
-        ti.xcom_pull.return_value = "success"
+        # xcom_pull calls: _log_startup_events=None, _get_node_status="success", compiled_sql=None, _dbt_event=None
+        ti.xcom_pull.side_effect = [None, "success", None, None]
         context = self.make_context(ti)
 
         result = sensor.poke(context)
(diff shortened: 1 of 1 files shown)


michal-mrazek:
@tatiana I will rebase this ticket since we now have logging in sensors (which we are super grateful for!) and elt you know it is ready fore review :)

michal-mrazek:
@tatiana Ready for review :)

cosmos/operators/_watcher/base.py
```diff
@@ -334,7 +344,16 @@ def _fallback_to_non_watcher_run(self, try_number: int, context: Context) -> boo
...
+
+        For test sensors, re-execution is not supported in watcher mode; retries are skipped.
         """
+        if self.is_test_sensor:
+            logger.info(
+                "Test sensor retry detected (try #%s). Skipping — test re-execution is not supported in watcher mode.",
+                try_number,
+            )
+            return True
+
```
tatiana, line 356: @michal-mrazek Do you think it would make sense to route test ask retries through the local/K8s test operator, to keep things consistent with the model run execution path?

michal-mrazek: I think it would make sense. But I would probably keep it for another PR (and we can also include it in the following release, not this one :) ).

michal-mrazek: I added a note that retry is not yet supported and we will implement it later :)

michal-mrazek:
> @michal-mrazek Thank you very much for continuing this work stream. I would like us to merge these changes before we release Cosmos 1.14.0. I just gave some feedback in-line - please, let me know your thoughts.
> 
> Also, please, could you mention how you tested these changes?
> 
> * InvocationMode.SUBPROCESS x  InvocationMode.DBT_RUNNER
> * Local x K8s (based on the files changed, I assume only local)
> 
> For us to have feature parity in K8s & Local, should we update the ticket #2427 or create a new one?

@tatiana I tested on both DBT_RUNNER and SUBPROCESS. Not K8 though, so I think the #2427 is still actual and we should work on it :)

Btw do you have any timeline for 1.14.0 release?

tatiana:
@michal-mrazek thanks a lot, I believe we're very close to merging this - please, could you fix the unit tests?

michal-mrazek:
Adjusted tests, hopefully it will pass now :) Lets trigger it, I can fix in a minute if needed :)

PR #2449, opened 2026-03-09
Getting Started: How cosmos works

lzdanski:
## Description

- Adds a landing page with a mental model/extremely high level overview of what Cosmos is.
- Adds a section for the QuickStarts and any relevant overview/introductory concept materials
Screenshot 2026-03-13 at 1 58 40 PM

Screenshot 2026-03-13 at 1 58 54 PM

Screenshot 2026-03-13 at 1 58 35 PM

## Breaking Change?

None expected

## Checklist


docs/getting_started/how-cosmos-works.rst
```diff
@@ -0,0 +1,41 @@
...
+workflows.
+
+When using Cosmos you can write your data transformations using dbt and then use Airflow’s
+advanced orchestration capabilities to integrate your dbt projects into your end to end workflows, scheduled to
+run automatically.
+
+Workflow overview
+~~~~~~~~~~~~~~~~~
+
+Cosmos creates an interface between a dbt project and Airflow, allowing you to translate your dbt project into a Dag. Then, your Cosmos configuration provides the necessary configurations to Airflow so that it can schedule and initiate running your dbt code when you want.
```
pankajastro, line 17:
```suggestion
Cosmos creates an interface between a dbt project and Apache Airflow®, allowing you to translate your dbt project into a Dag. Then, your Cosmos configuration provides the necessary configurations to Apache Airflow® so that it can schedule and initiate running your dbt code when you want.
```

Maybe we should follow the trademark guide https://lists.apache.org/thread/0fmp9olljfkm3vmg3ns5qd1o95sm7235

docs/getting_started/how-cosmos-works.rst
```diff
@@ -0,0 +1,41 @@
...
+
+Cosmos is an open-source Python package developed by Astronomer under the Apache License 2.0. The
+package includes classes that automatically turn dbt Core and dbt Fusion projects into Apache Airflow®
+workflows.
+
+When using Cosmos you can write your data transformations using dbt and then use Airflow’s
+advanced orchestration capabilities to integrate your dbt projects into your end to end workflows, scheduled to
+run automatically.
+
+Workflow overview
```
pankajastro: I think we discussed changing this heading, no?

docs/getting_started/how-cosmos-works.rst
lzdanski: We only need to use the full name for the *first and prominent* mention. This means:

- All titles/headers with the full name "Apache Airflow"
- Within the text, the first mention on the page. So I don't think that these will need to be updated.

But you're right that this convention is missing something! It needs to link to Airflow with the first mention.

docs/getting_started/how-cosmos-works.rst
```diff
@@ -0,0 +1,41 @@
+.. _how-cosmos-works:
+
+How Cosmos works
+=================
+
+Cosmos is an open-source Python package developed by Astronomer under the Apache License 2.0. The
+package includes classes that automatically turn dbt Core and dbt Fusion projects into Apache Airflow®
```
lzdanski:
```suggestion
package includes classes that automatically turn dbt Core and dbt Fusion projects into `Apache Airflow® <https://airflow.apache.org/>`_
```

docs/getting_started/how-cosmos-works.rst
lzdanski: I don't recall changing this heading, but if there's something better, definitely want to switch it! This is a pretty generic fill-in

docs/_static/how-cosmos-works.png
tatiana, line 1: Thanks, Laura! Would it make sense to have the dbt project image generated by dbt in the beginning?
Image)

To show dbt UI and then Airflow UI is probably more powerful than to show two screenshots of Cosmos - one with the DAG before execution and the other after execution.

Regarding the text, it looks great; there would be just one adjustment. Currently the bottom text says:
> dbt commands run using the relevant dbt adapter, running in the environment configured in the execution mode, for example, the Airflow worker, or K8s

This is not true for the `ExecutionMode.AIRFLOW_ASYNC` - when the dbt transformations are run by a native Airflow operator (e.g. `BigQueryInsertJobOperator`

lzdanski: This diagram is taken from the ebook, and I don't have permissions to edit it, however, I've passed along the feedback and have removed the diagram for now. I'll re-add it after its been updated!

lzdanski:
I have removed the diagram from this PR - depending on timing, I might be able to add it back. Otherwise, will leave instructions for diagram owner to create follow-up PR with updated diagram next week.

lzdanski:
@tatiana - diagram has been updated by diagram owner, and included in PR. Will provide alternatives to team separately.

docs/getting_started/how-cosmos-works.rst
lzdanski: Included in: https://github.com/astronomer/astronomer-cosmos/pull/2462

Issue #2421, opened 2026-03-02
ParamValidationError: Invalid input for param cosmos_telemetry_metadata


pankajkoti:
With our introduction of Scarf telemetry PR https://github.com/astronomer/astronomer-cosmos/pull/2223, we are intermittenly observing ParamValidationError while parsing Cosmos DAGs with the below logs
```
Log message source details sources=["/Users/pankaj/Documents/astro_code/cosmos-oss-demo/logs/dag_id=micro_project_dag/run_id=manual__2026-03-02T06:03:47.383931+00:00/task_id=base_model_run/attempt=1.log"]

[2026-03-02 11:33:50] WARNING - section/key [openlineage/namespace] not found in config
[2026-03-02 11:33:50] WARNING - The `airflow.models.param.Param` attribute is deprecated. Please use `'airflow.sdk.Param'`. category=DeprecatedImportWarning
[2026-03-02 11:33:50] WARNING - To enable emitting Openlineage events, upgrade to Airflow 2.7 or install astronomer-cosmos[openlineage].
[2026-03-02 11:33:50] INFO - DAG bundles loaded: dags-folder
[2026-03-02 11:33:50] INFO - Filling up the DagBag from /Users/pankaj/Documents/astro_code/cosmos-oss-demo/dags/micro_project_dag.py
Cosmos DAG parsing logs
[2026-03-02 11:33:50] INFO - Trying to parse the dbt project using dbt ls cache cosmos_cache__micro_project_dag...
[2026-03-02 11:33:50] INFO - Cosmos performance: time to calculate cache identifier cosmos_cache__micro_project_dag for current version: 0.0008460829994874075
[2026-03-02 11:33:50] INFO - Cosmos performance [Pankajs-MacBook-Pro.local|18716]: The cache size for cosmos_cache__micro_project_dag is 1793
[2026-03-02 11:33:50] INFO - Cosmos performance: Cache hit for cosmos_cache__micro_project_dag - 0f73115ad1938ab6fc50ff586f98842d,d41d8cd98f00b204e9800998ecf8427e
[2026-03-02 11:33:50] INFO - Total nodes: 2
[2026-03-02 11:33:50] INFO - Total filtered nodes: 2
[2026-03-02 11:33:50] INFO - Cosmos performance (micro_project_dag) -  [Pankajs-MacBook-Pro.local|18716]: It took 0.00805s to parse the dbt project for DAG using LoadMode.DBT_LS_CACHE
[...]

PR #2450, opened 2026-03-09
Store Cosmos telemetry metadata in Airflow Variable instead of DAG param

pankajkoti:
- Store __cosmos_telemetry_metadata__ in Variable (key: cosmos_telemetry__<dag_id>) to avoid Param validation issues when DAG is re-parsed.
- Listener reads from Variable via get_cosmos_telemetry_metadata_from_variable(); returns empty dict if Variable is missing or invalid.
- Remove Param/ParamValidationError usage from the converter for telemetry.

closes: #2421

### DIFF
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -12,14 +12,7 @@ from collections.abc import Callable
 from typing import Any
 from warnings import warn
 
-from airflow.exceptions import ParamValidationError
 from airflow.models.dag import DAG
-
-try:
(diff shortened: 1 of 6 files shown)


tatiana:
@pankajkoti, wondering if https://github.com/apache/airflow/issues/51816 could pose challenges for this implementation. If possible, double-check with the latest micro versions of Airflow if things work as expected.

pankajkoti:
Closing in favour of #2466

PR #2443, opened 2026-03-06
Reference documentation restructure + refactor 

lzdanski:
## Description

This PR restructures the **Reference** Section, creates the new **Guides>Connect database** section from the existing [Profiles Overview page](https://astronomer.github.io/astronomer-cosmos/profiles/index.html). It replaces https://github.com/astronomer/astronomer-cosmos/pull/2424.
Screenshot 2026-03-11 at 10 03 06 AM

- It removes the redundant [**Profile Config** page](https://astronomer.github.io/astronomer-cosmos/reference/configs/profile-config.html) -> this refactoring would have had two different "Profiles" pages in the reference section 
- Adds the **Profiles** mappings to the **Reference** sidebar and hides the list of profile mappings from the Profiles Index overview.
Screenshot 2026-03-06 at 2 48 09 PM

- Moves the contents of the [Profiles Index Overview page](https://astronomer.github.io/astronomer-cosmos/profiles/index.html) into three new pages in the **Connect database** section. NOTE - due to issue in local build, the sidebar appears inconsistent in screenshots.

Screenshot 2026-03-06 at 2 48 17 PM

Screenshot 2026-03-06 at 2 48 26 PM

Screenshot 2026-03-06 at 2 48 35 PM

Screenshot 2026-03-06 at 2 48 51 PM
Screenshot 2026-03-06 at 2 49 00 PM
Screenshot 2026-03-06 at 2 49 07 PM

- **Discussion question**: disabling dbt event tracking might not fit on the current page in the "Connect Database", and would need to be relocated.
Screenshot 2026-03-06 at 2 49 15 PM

## Breaking Change?

Possible docs broken links. Moved files have been compared to the list of redirects created to reduce this possibility.

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2453, opened 2026-03-10
[Guides] How cosmos runs dbt

lzdanski:
## Description

- Adds a new page, **How Cosmos runs dbt**, which acts as a landing page for this subsection in the TOC. 

Screenshot 2026-03-09 at 9 07 28 PM

Screenshot 2026-03-09 at 9 07 40 PM

- Renames the "Operators" page to **dbt command operators**. I'm not sure if this is the right label, but "Operators" didn't seem quite accurate as this appears to be about dbt commands specifically
Screenshot 2026-03-09 at 9 08 51 PM

- **Moves Operator arguments into the correct section in the sidebar** 
Screenshot 2026-03-09 at 9 08 15 PM

Screenshot 2026-03-09 at 9 08 23 PM

Screenshot 2026-03-09 at 9 08 33 PM

- Refactors the **Operator Args** page into **Operator Args** and **Overriding Operator Args**. This PR does not move "Template Fields" out of the **Operator Args** doc and into "References", because it affects an area of the docs that has an open PR. I can finish that in a follow-up after https://github.com/astronomer/astronomer-cosmos/pull/2443 merges.

Screenshot 2026-03-09 at 9 08 44 PM

## Checklist

### DIFF
--- a/docs/guides/run_dbt/operators/operator-args.rst
+++ b/docs/guides/run_dbt/operators/operator-args.rst
@@ -152,7 +152,7 @@ Example usage of templated ``dbt_cmd_flags`` for microbatch models with event-ti
         },
     )
 
-The following template fields are only selectable when using the operators in a standalone context via the `operator_args` parameter (starting in Cosmos 1.4):
+The following template fields are only selectable when using the operators in a standalone context via the ``operator_args`` parameter (starting in Cosmos 1.4):
 
 - ``select``
 - ``exclude``


docs/guides/run_dbt/operators/index.rst
```diff
@@ -0,0 +1,15 @@
...
+
+Learn how to use operators with Cosmos.
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Operators
+
+   operator-args
+   overriding-operator-args
+   operators
```
tatiana: Probably operators should be the first item in this list

docs/guides/run_dbt/index.rst
```diff
@@ -0,0 +1,33 @@
+.. _how-cosmos-runs-dbt:
+
+How Cosmos runs dbt
+===================
+
+After the dbt project has been parsed and turned into an Airflow dag or task group, you can execute it.
+
+Cosmos’ ``DbtDag`` and ``DbtTaskGroup`` follow the Airflow best practice of creating tasks that are as atomic as
```
tatiana: @lzdanski with `WATCHER` and `AIRFLOW_ASYNC`, we can no longer state that the tasks are atomic

docs/guides/run_dbt/index.rst
```diff
@@ -0,0 +1,33 @@
...
+possible, meaning that there is a separate task for each dbt command. This leads to improved visibility and the
```
tatiana: Most people are used to use a dbt command to run multiple dbt nodes - so I believe it is clearer to say "dbt node" instead of command.
```suggestion
possible, meaning that there is a separate task for each dbt node. This leads to improved visibility and the
```

docs/guides/run_dbt/index.rst
```diff
@@ -0,0 +1,33 @@
...
+possibility of fine-grained control over your dbt commands. For example, you can set task parameters like pool
+or retries, on individual Cosmos tasks. Or, you can make downstream tasks run as soon as a specific Cosmos task has
+finished successfully.
+
+Cosmos uses different kinds of configurations to control how the dbt commands are executed within the Airflow dag or task group.
```
tatiana:
```suggestion
Cosmos uses different kinds of configurations to control how the dbt nodes are executed within the Airflow Dag or task group.
```

docs/guides/run_dbt/index.rst
```diff
@@ -0,0 +1,33 @@
...
+
+Check out the available :ref:`execution-modes` and the detailed :ref:`execution-config` for more information about how to set up your Cosmos execution.
+
+
+Running dbt commands
+~~~~~~~~~~~~~~~~~~~~
+
+After specifying where you want Cosmos to run dbt commands, you can additionally specify the following:
+
+- :ref:`callbacks`: Tell Cosmos how to handle artifacts produced by dbt while executing dbt code.
```
tatiana, line 30: We recently introduced a concept very similar to `callback`, but it runs before the dbt command. We may want to add it to this list, even though it does not have a dedicated page:
https://github.com/astronomer/astronomer-cosmos/pull/2419/changes

docs/guides/run_dbt/index.rst
```diff
@@ -0,0 +1,33 @@
+.. _how-cosmos-runs-dbt:
+
+How Cosmos runs dbt
+===================
+
+After the dbt project has been parsed and turned into an Airflow dag or task group, you can execute it.
```
tatiana: I noticed we do not highlight that users can also run dbt Commands using operators directly, rather than via `DbtTaskGroup` or `DbtDag`. It may make sense to at least mention this here.

docs/guides/run_dbt/operators/operators.rst
```diff
@@ -1,7 +1,7 @@
 .. _operators:
 
-Operators
-=========
+dbt command operators
+=====================
 
 Cosmos exposes individual operators that correspond to specific dbt commands, which can be used just like traditional
 `Apache Airflow® <https://airflow.apache.org/>`_ operators. Cosmos names these operators using the format ``Dbt<dbt-command><execution-mode>Operator``. For example, ``DbtBuildLocalOperator``.
```
tatiana: It would be very valuable if we could add the code snipped referencing the relevant parts of the existing example DAG:
https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/example_operators.py

lzdanski: It looks like the code examples are already present for `Clone` and `Seed` in the following sections, so they aren't shown in this PR. However, I'm not seeing anything in the `example_operators.py` file for `build`. Is there an additional operator you want me to add from the example? https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/example_operators.py. 

Image


I've updated this sentence to say "Clone" and "Seed" and added a link to the example dag.

docs/guides/run_dbt/index.rst
lzdanski: Added a new entry and link to existing materials in `operator-args`, until there's a doc for this feature

docs/guides/run_dbt/operators/operator-args.rst
tatiana:
```suggestion
The following template fields are only selectable when using the operators in a standalone context via the `operator_args` parameter (starting in Cosmos 1.4):
```

docs/guides/run_dbt/index.rst
```diff
@@ -0,0 +1,35 @@
...
+===================
+
+Cosmos can run dbt commands directly using operators, or, after the dbt project has been parsed and turned into an Airflow Dag or task group, you can execute it.
+
+For most execution modes, Cosmos’ ``DbtDag`` and ``DbtTaskGroup`` follow the Airflow best practice of creating tasks that are as atomic as
+possible, meaning that there is a separate task for each dbt node. This leads to improved visibility and the
+possibility of fine-grained control over your dbt commands. For example, you can set task parameters like pool
+or retries, on individual Cosmos tasks. Or, you can make downstream tasks run as soon as a specific Cosmos task has
+finished successfully. The exceptions to using the Airflow best practice of creating atomic tasks are :ref:`watcher-execution-mode`,
+which uses dbt threading and deferrable sensors, and :ref:`async-execution-mode`, which uses the Airflow defferable operator.
```
tatiana:
```suggestion
In many execution modes, Cosmos’ ``DbtDag`` and ``DbtTaskGroup '' create a separate task for each dbt node (model, seed, snapshot).
This leads to improved visibility and the
possibility of fine-grained control over your dbt commands. For example, you can set task parameters like pool
or retries on individual Cosmos tasks. Or, you can make downstream tasks run as soon as a specific Cosmos task has finished successfully.
Running one dbt command per task can bring performance challenges, since each invocation of a dbt command incurs overhead. To improve performance, newer versions of Cosmos have introduced alternatives that offer the same level of granularity while centralising the execution of the dbt command in a single task. Check :ref:`watcher-execution-mode` and :ref:`async-execution-mode`, for more information.
```

tatiana: The previous version stated that the new approach is not compatible with Airflow tasks, which I believe is not the most important point in this section. I rewrote this part so we focus on the options and, if performance matters, how users can improve it.

docs/guides/run_dbt/operators/operator-args.rst
```diff
@@ -193,15 +152,15 @@ Example usage of templated ``dbt_cmd_flags`` for microbatch models with event-ti
         },
     )
 
-The following template fields are only selectable when using the operators in a standalone context (starting in Cosmos 1.4):
+The following template fields are only selectable when using the operators in a standalone context via the `operator_args` parameter (starting in Cosmos 1.4):
```
tatiana:
```suggestion
The following template fields are only selectable when using the operators in a standalone context via the ``operator_args`` parameter (starting in Cosmos 1.4):
```

PR #2467, opened 2026-03-16
[Experimental] Add source freshness aware execution for ExecutionMode.WATCHER

### LINKED ISSUE #2053: [Feature] Support custom source nodes in `ExecutionMode.WATCHER`
Since Cosmos 1.6, it [supports the rendering of source nodes](https://www.astronomer.io/blog/native-support-for-source-node-rendering-in-cosmos).

We noticed some Cosmos users use this feature alongside [overriding Cosmos source nodes](https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html#customizing-how-nodes-are-rendered-experimental) as sensors or another operator, allowing them to skip the following branch of the DAG if the source is not fresh.

This use case is not currently supported by ``ExecutionMode.WATCHER`` in Cosmos 1.11.0 because the ``dbt build`` command does not run [source freshness checks](https://docs.getdbt.com/reference/commands/build#source-freshness-checks).

The goal of this ticket is to investigate this use case further and find a solution.

Internal note: For more context on an example use case, check the Astro customer call: https://us-35700.app.gong.io/call?id=9016142599412191287. @anyapriya can give you more context if you need it.

pankajastro:
## Add source freshness awareness for `ExecutionMode.WATCHER`

  **Related issue:**  #2053                                                                   
   
  ### What this PR does                                                                             
                                                                                                  
  When `SourceRenderingBehavior` is not `NONE`, the `DbtProducerWatcherOperator` now runs           
  `dbt source freshness` before the main `dbt build`. For every source that is stale (`error`
  or `warn`), all downstream models that transitively depend on that source are:                    
                                                                                                    
  - pushed a synthetic `"skipped"` XCom so their consumer sensors raise `AirflowSkipException`
  immediately                                                                                       
  - added to the `--exclude` selector of the subsequent `dbt build` so dbt does not attempt to run
  them                                                                                              
   
  The dependency traversal (BFS from stale source → transitive model dependents) runs at task       
  execution time using nodes from `dag.dbt_graph.nodes`. Test nodes are intentionally excluded    
  from the skip/exclude logic — dbt skips them automatically when their parent model is excluded.   
                                                                                                    
  `dbt source freshness` is run via the existing `run_command` infrastructure (temp directory,      
  profile handling, subprocess vs dbt runner routing) rather than a raw subprocess call.            
                                                                                                    
  ### Known limitations
                                                                                                    
  > [!WARNING]                                                                                    
  > This PR does **not** fully close #2053. See limitations below.
                                                                                                    
  - **`"skip"` is the only supported action** — when a stale source is detected, downstream models  
    are always skipped. Configurable behavior (e.g. warn-only or custom handling) is not supported  
    in this release.                                                                                
  - **`_freshness_callback` is internal** — not part of the public API and not exposed for user     
    customisation. A future release may promote this to a stable, user-facing interface.            
  - **Custom source node overrides not supported** — the original issue also requested support for  
    overriding Cosmos source nodes with custom SQL-based table state checks (similar to dbt's       
    [state-aware runs](https://docs.getdbt.com/docs/deploy/state-aware-about)). This will be        
    tracked separately.                                                                             
  - **Incompatible with `--selector`** — the exclude logic appends to `self.exclude` using model    
    names and does not interact correctly with dbt's `--selector` argument (see #2415).             
  - **`LoadMode.DBT_MANIFEST` freshness data is ignored** — `dbt source freshness` is always        
    re-run at task execution time; any freshness metadata in a pre-compiled manifest is not used. 

Screenshot 2026-03-20 at 5 54 15 AM

https://github.com/user-attachments/assets/11c4341c-a9a9-4fdb-9fd9-1ae16143ffa9

Co-authored-by: Claude <noreply@anthropic.com>


cosmos/operators/watcher.py
```diff
@@ -159,6 +175,116 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+    def _get_stale_sources(self) -> list[str]:
+        """Run ``dbt source freshness`` and return the unique_ids of stale sources.
+
+        Runs the command as a subprocess against ``self.project_dir`` using the same profile
+        configuration as the build step. Returns an empty list (and logs a warning) if the
+        freshness check cannot be performed or the artifact is missing.
+        """
+        if not self.profile_config:
+            logger.warning("check_source_freshness=True but no profile_config is set; skipping freshness check.")
+            return []
```
```suggestion
```
This does not seem like a realistic use case, since Cosmos would have thrown an error in the converter if there were no profile_config.

cosmos/operators/watcher.py
```diff
@@ -159,6 +175,116 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+            if self.profile_config.profile_name:
+                cmd += ["--profile", self.profile_config.profile_name]
+            if self.profile_config.target_name:
+                cmd += ["--target", self.profile_config.target_name]
+
+            logger.info("Running dbt source freshness: %s", cmd)
+            proc = subprocess.run(cmd, env=env, capture_output=True, text=True, cwd=str(self.project_dir))
+            logger.info("dbt source freshness stdout:\n%s", proc.stdout)
+            if proc.returncode != 0:
+                logger.warning("dbt source freshness exited with code %d; stderr:\n%s", proc.returncode, proc.stderr)
```
tatiana: It feels like we duplicated a lot of the logic from `run_command`:  https://github.com/astronomer/astronomer-cosmos/blob/5e5a27458f3495a15054162870a926e65768ff7e/cosmos/operators/local.py#L628

We should really try to reuse the existing method. We may be able to use a custom callback. I'm happy for us to refactor the existing code if needed.

Additionally, we should not be executing dbt commands directly from the dbt project folder - we always have tasks running them from dedicated temporary directories. The existing `run_command` already handles all these lessons we learned over time.

cosmos/operators/watcher.py
```diff
@@ -159,6 +175,116 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+
+        sources_json_path = Path(self.project_dir) / "target" / "sources.json"
```
tatiana: We should not use the main dbt project directory, but rather a temporary directory from which the dbt command should be run. This change is not enough, but illustrates part of the potential fix:
```suggestion
        sources_json_path = Path(tmp_project_dir) / "target" / "sources.json"
```

cosmos/operators/watcher.py
```diff
@@ -159,6 +175,116 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+
+            dbt_executable = str(getattr(self, "dbt_executable_path", None) or "dbt")
+            cmd: list[str] = [
+                dbt_executable,
+                "source",
+                "freshness",
+                "--project-dir",
+                str(self.project_dir),
+                "--profiles-dir",
+                str(profile_path.parent),
```
tatiana: When using `LoadMode.DBT_MANIFEST`, should we retrieve the source freshness from the manifest? This may be necessary if we want to be consistent: https://www.astronomer.io/blog/native-support-for-source-node-rendering-in-cosmos/

cosmos/operators/watcher.py
```diff
@@ -159,6 +175,116 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+            ]
+            if self.profile_config.profile_name:
+                cmd += ["--profile", self.profile_config.profile_name]
+            if self.profile_config.target_name:
+                cmd += ["--target", self.profile_config.target_name]
+
+            logger.info("Running dbt source freshness: %s", cmd)
+            proc = subprocess.run(cmd, env=env, capture_output=True, text=True, cwd=str(self.project_dir))
```
tatiana: What about `InvocationMode.DBT_RUNNER`..?

cosmos/operators/watcher.py
```diff
@@ -159,6 +175,116 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+        No-op when ``check_source_freshness=False`` or no source dependency map is available.
+        On any error during the freshness check the method logs a warning and returns without
+        modifying the exclude list so that the ``dbt build`` proceeds normally (fail-open).
+        """
+        if not self.check_source_freshness or not self.source_to_dependent_models:
+            return
+
+        try:
+            stale_source_ids = self._get_stale_sources()
+        except Exception:
```
tatiana: Let's use specific exceptions

cosmos/operators/watcher.py
```diff
@@ -159,6 +175,116 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+            logger.info("Marking model '%s' as skipped (stale upstream source)", unique_id)
+            self._push_skipped_xcom_for_model(ti, unique_id)
+
+        # Append model names to the exclude selector for dbt build.
+        # Use only the last segment of the unique_id (the model name) as the dbt selector.
+        model_names = [uid.split(".")[-1] for uid in excluded_unique_ids]
+        additional_exclude = " ".join(model_names)
+        current_exclude: str = getattr(self, "exclude", None) or ""
+        self.exclude = f"{current_exclude} {additional_exclude}".strip() if current_exclude else additional_exclude
+        logger.info("dbt build --exclude updated to: %s", self.exclude)
```
tatiana: This method would not work if users are using `--selector`, as in https://github.com/astronomer/astronomer-cosmos/issues/2415, so we should probably update the documentation accordingly, once we add it.

tatiana (commented):
@pankajastro I think this is a solid step forward in supporting source nodes with the `ExecutionMode.WATCHER` - thanks a lot for the work on this!

I’ve left a few inline comments regarding the implementation.

While I do think we should continue iterating on this and move it toward a release as part of 1.14, it doesn’t fully address the original issue that prompted feature request #2053. In particular, the current solution doesn’t account for overriding Cosmos source nodes (as described here: https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html#customizing-how-nodes-are-rendered-experimental).

Supporting this was a key requirement when the issue was raised by the customer - they’re using custom SQL-based table state checks, effectively making Cosmos stateful, similar to dbt’s state-aware workflows (https://docs.getdbt.com/docs/deploy/state-aware-about).

So we change the PR description to clarify that it does not close the custom source node behaviour, as declared in https://github.com/astronomer/astronomer-cosmos/issues/2053.

One last comment: how do we want this feature to relate to the existing source behaviour (https://astronomer.github.io/astronomer-cosmos/guides/translate_dbt_to_airflow/managing-sources.html)?

cosmos/airflow/graph.py
```diff
@@ -679,6 +714,7 @@ def _add_watcher_producer_task(
     render_config: RenderConfig | None = None,
     execution_mode: ExecutionMode = ExecutionMode.WATCHER,
     tests_per_model: dict[str, list[str]] | None = None,
+    source_to_dependent_models: dict[str, list[str]] | None = None,
```
tatiana: Do we really need to inject this parameter into the tasks, changing the interface,  or could we calculate this auxiliary data structure from within the watcher tasks, leveraging the `self.graph` that is available either in the `DbtTaskGroup` or in the `DbtDag`? https://github.com/astronomer/astronomer-cosmos/blob/5e5a27458f3495a15054162870a926e65768ff7e/cosmos/converter.py#L317

pankajastro: Removed this change

cosmos/operators/watcher.py
pankajastro: Now using run_command so I believe that should take care of this

cosmos/operators/watcher.py
pankajastro: It may improve performance a bit, but I'm running this during task execution, so I feel running the command will yield a more accurate result as data in manifest can be outdated

cosmos/operators/watcher.py
pankajastro: Fixed

cosmos/operators/watcher.py
pankajastro: Thanks, updated this. PTAL

cosmos/operators/watcher.py
pankajastro: Removed

pankajastro:
> @pankajastro I think this is a solid step forward in supporting source nodes with the `ExecutionMode.WATCHER` - thanks a lot for the work on this!
> 
> I’ve left a few inline comments regarding the implementation.
> 
> While I do think we should continue iterating on this and move it toward a release as part of 1.14, it doesn’t fully address the original issue that prompted feature request #2053. In particular, the current solution doesn’t account for overriding Cosmos source nodes (as described here: https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html#customizing-how-nodes-are-rendered-experimental).
> 
> Supporting this was a key requirement when the issue was raised by the customer - they’re using custom SQL-based table state checks, effectively making Cosmos stateful, similar to dbt’s state-aware workflows (https://docs.getdbt.com/docs/deploy/state-aware-about).
> 
> So we change the PR description to clarify that it does not close the custom source node behaviour, as declared in #2053.
> 
> One last comment: how do we want this feature to relate to the existing source behaviour (https://astronomer.github.io/astronomer-cosmos/guides/translate_dbt_to_airflow/managing-sources.html)?

Thanks, @tatiana, for the review feedback.

I’ve updated the PR description to include the current limitations and have addressed your earlier comments.

I also revised the description to ensure that issue https://github.com/astronomer/astronomer-cosmos/issues/2053 does not get closed.

If the source behavior is not None, we will run freshness checks and take action based on the results.

This PR does not yet expose a callable to override the behavior. However, it lays the groundwork for introducing a callable in the future and allowing customization of the default behavior.

cosmos/operators/watcher.py
pankajastro: Yes, this would be a limitation. I'll create a separate PR to add doc

cosmos/operators/watcher.py
pankajastro: The intention here is that if source freshness fails, do not fail, and the workflow should proceed with the build. I have added a code comment

cosmos/operators/_watcher/base.py
```diff
@@ -377,7 +378,7 @@ def _get_status_from_run_results(self, ti: Any, context: Context) -> Any:
             )
             return None
 
-        logger.info("Node Info: %s", run_results_json)
+        logger.info("Node Info: %s", str(run_results_json))
```
tatiana: We probably want to change this to `debug` - happy to be done in a follow up PR, since the change as not introduced in this PR
```suggestion
        logger.debug("Node Info: %s", str(run_results_json))
```

cosmos/operators/watcher.py
```diff
@@ -151,6 +214,85 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+        - **SUBPROCESS mode**: key ``<uid>_status``, value is the plain string ``"skipped"``.
+        """
+        uid_key = unique_id.replace(".", "__")
+        if self.invocation_mode == InvocationMode.DBT_RUNNER:
+            payload = base64.b64encode(
+                zlib.compress(json.dumps({"data": {"run_result": {"status": "skipped"}}}).encode())
+            ).decode()
+            safe_xcom_push(task_instance=ti, key=f"nodefinished_{uid_key}", value=payload)
+        else:
+            safe_xcom_push(task_instance=ti, key=f"{uid_key}_status", value="skipped")
```
tatiana: I feel we should unify how information is added in XCom and stop using two methods, regardless of the invocat_mode. Again, we can do it in a separate PR.

cosmos/operators/watcher.py
```diff
@@ -151,6 +214,85 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+
+    def _run_source_freshness(self, context: Context) -> None:
+        """Run ``dbt source freshness`` via ``build_cmd`` and ``run_command`` (temp dir, profile, deps like ``dbt build``)."""
+        original_base_cmd = self.base_cmd
+        original_indirect_selection = getattr(self, "indirect_selection", None)
+        try:
+            self.base_cmd = ["source", "freshness"]
+            self.indirect_selection = None  # ``dbt source freshness`` does not support --indirect-selection
+            full_cmd, env = self.build_cmd(context=context, cmd_flags=[])
```
tatiana: Don't we need to respect any `--select`, `--exclude` or `--selector` passed by the user when running `dbt source freshness` as well?

cosmos/operators/watcher.py
```diff
@@ -151,6 +214,85 @@ def _finalize(self, context: Context, startup_events: list[dict[str, Any]]) -> N
...
+            self._run_source_freshness(context)
+            dbt_nodes = getattr(getattr(getattr(self, "dag", None), "dbt_graph", None), "nodes", None)
+            node_unique_ids, status = self._freshness_callback(
+                context, self.dag, self.task_group, dbt_nodes, self._sources_json
+            )
+            if node_unique_ids is None or node_unique_ids == []:
+                return
+            if status == "skip":
+                self._skipped_node_token(context, node_unique_ids)
+        except Exception:  # intentional: freshness check must never block dbt build
```
tatiana: Why don't we want to block `dbt build` if `dbt source --freshness`? Especially if the user opts in to render source freshness, I believe we should highlight the error and fail the producer.

tatiana (commented):
@pankajastro Thanks a lot for addressing all the feedback! I believe we're really close to getting this merged. I left some feedback inline. Please let me know if you'd like to have a last review over a call, so we can merge this sooner rather than later

tests/operators/test_watcher.py
```diff
@@ -2044,6 +2045,308 @@ def test_dbt_source_watcher_operator_template_fields():
     assert DbtSourceWatcherOperator.template_fields == DbtSourceLocalOperator.template_fields
 
 
+# ---------------------------------------------------------------------------
+# Tests for source freshness feature (_check_source_freshness=True)
+# ---------------------------------------------------------------------------
+
+from cosmos.constants import DbtResourceType
+from cosmos.dbt.graph import DbtNode
```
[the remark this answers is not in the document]
pankajastro: Fixed

cosmos/operators/watcher.py
pankajastro: Yeah, since we are enabling only when source rendering is enabled. I have removed this try/except

cosmos/operators/watcher.py
pankajastro: Fixed this, passing the flags

cosmos/operators/watcher.py
pankajastro: yes, I think make sense to clean in a separate PR

cosmos/operators/_watcher/base.py
pankajastro: yes, I think from history point of view make sense to do this in a separate PR

tatiana:
I managed to rebase this PR onto main. The original version is in the branch feature/watcher-source-freshness-backup.

When I attempted to run this using the example DAG `source_rendering_dag` and modifying it to use `ExecutionMode.WATCHER`, I noticed it is not working as expected. This happened before and after the rebase, so I don't believe it was introduced as part of the rebase.

I logged a ticket to address this separately in:
https://github.com/astronomer/astronomer-cosmos/issues/2520

cosmos/operators/_watcher/state.py
```diff
@@ -44,9 +45,14 @@ def is_dbt_node_status_failed(status: str | None) -> bool:
...
+def is_dbt_node_status_skipped(status: str | None) -> bool:
+    """Check if the dbt node status indicates it was skipped due to a stale upstream source."""
+    return status in DBT_SKIPPED_STATUSES
+
+
 def is_dbt_node_status_terminal(status: str | None) -> bool:
-    """Check if the dbt node status is terminal (success or failed)."""
-    return is_dbt_node_status_success(status) or is_dbt_node_status_failed(status)
+    """Check if the dbt node status is terminal (success, failed, or skipped)."""
+    return is_dbt_node_status_success(status) or is_dbt_node_status_failed(status) or is_dbt_node_status_skipped(status)
```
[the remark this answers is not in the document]
tatiana: Addressed in 32c637bf26ee208303363df0f8127a3db8b02973

cosmos/operators/watcher.py
```diff
@@ -46,6 +54,49 @@
...
+        for dep_id in node.depends_on:
+            dependents.setdefault(dep_id, set()).add(node_id)
+
+    # BFS from each stale source to collect all transitive dependents
+    _excludable_resource_types = {DbtResourceType.MODEL, DbtResourceType.SEED, DbtResourceType.SNAPSHOT}
+    visited: set[str] = set()
+    queue = list(stale_source_ids)
+    while queue:
+        current = queue.pop()
+        for dependent_id in dependents.get(current, set()):
```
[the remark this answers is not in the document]
tatiana: Addressed in 32c637bf26ee208303363df0f8127a3db8b02973

cosmos/operators/watcher.py
```diff
@@ -172,6 +228,61 @@ def _event_callback(event: Any) -> None:
...
+        else:
+            self.exclude = exclude_str
+
+    def _apply_source_freshness(self, context: Context) -> None:
+        """Run source freshness, invoke the callback, and mark affected nodes as skipped."""
+        self._run_source_freshness(context)
+        dag = context.get("dag")
+        task_group = getattr(context.get("task_instance"), "task", None)
+        task_group = getattr(task_group, "task_group", None)
+        nodes = getattr(dag, "_cosmos_nodes", None) if dag else None
```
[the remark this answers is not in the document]
tatiana: Addressed in 4fc12de9748dc8b8e7fc0c65f3262f9a4cd965d8

cosmos/operators/watcher.py
```diff
@@ -46,6 +54,49 @@
...
+        for dep_id in node.depends_on:
+            dependents.setdefault(dep_id, set()).add(node_id)
+
+    # DFS from each stale source to collect all transitive dependents
+    _excludable_resource_types = {DbtResourceType.MODEL, DbtResourceType.SEED, DbtResourceType.SNAPSHOT}
+    visited: set[str] = set()
+    queue = list(stale_source_ids)
+    while queue:
+        current = queue.pop()
+        for dependent_id in dependents.get(current, set()):
```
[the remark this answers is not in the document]
tatiana: Fixed in b32cb01fb6c17d6770a607aae26152b1fce26b97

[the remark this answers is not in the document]
tatiana: Fixed in b32cb01fb6c17d6770a607aae26152b1fce26b97

cosmos/airflow/graph.py
```diff
@@ -687,6 +687,8 @@ def _add_watcher_producer_task(
     producer_task_args = task_args.copy()
     if tests_per_model is not None:
         producer_task_args["tests_per_model"] = tests_per_model
+    if render_config is not None and render_config.source_rendering_behavior is not SourceRenderingBehavior.NONE:
```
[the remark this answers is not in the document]
tatiana: Fixed in 0895c55739c57a724e925c9f0ae9521df20f67c0

cosmos/operators/watcher.py
```diff
@@ -172,6 +234,70 @@ def _event_callback(event: Any) -> None:
...
+        ti = context["ti"]
+
+        for unique_id in node_unique_ids:
+            logger.info(
+                "Marking resource '%s' as skipped (stale upstream source)",
+                unique_id,
+            )
+            self._push_skipped_xcom_for_model(ti, unique_id)
+
+        model_names = sorted({uid.rsplit(".", 1)[-1] for uid in node_unique_ids})
```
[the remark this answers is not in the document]
tatiana: Fixed in 84fb1667cbfa9952995d34db63b02e691b1e7bc1

cosmos/operators/watcher.py
```diff
@@ -172,6 +234,72 @@ def _event_callback(event: Any) -> None:
...
+        # Resolve graph nodes from the DbtDag or DbtTaskGroup converter, falling back gracefully.
+        nodes = None
+        if dag is not None:
+            dbt_graph = getattr(dag, "dbt_graph", None)
+            nodes = getattr(dbt_graph, "filtered_nodes", None)
+        if nodes is None and task_group is not None:
+            tg_dbt_graph = getattr(task_group, "dbt_graph", None)
+            nodes = getattr(tg_dbt_graph, "filtered_nodes", None)
+
+        node_ids_to_skip, _ = self._freshness_callback(context, dag, task_group, nodes, self._sources_json)
```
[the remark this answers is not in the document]
tatiana: Fixed in bb288c091e807d83e78dd93a1e601df600ba7b6a

cosmos/operators/local.py
```diff
@@ -674,6 +688,10 @@ def run_command(  # noqa: C901
                     cwd=tmp_project_dir,
                     context=context,
                 )
+                if context.get("_check_source_freshness"):
+                    self._sources_json = _read_target_sources_json(tmp_dir_path)
+                    if self._sources_json is not None:
+                        return result
```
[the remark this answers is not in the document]
tatiana: Addressed in 8322a8303e92ed1446d3badd65fc8cdb279430e1

cosmos/operators/watcher.py
```diff
@@ -2,17 +2,25 @@
 
 import contextlib
 import functools
-from collections.abc import Callable, Sequence
+from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
```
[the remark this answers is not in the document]
tatiana: Addressed in 8e8c4e94b0084a6f6973a590eeae92d504db449f

cosmos/operators/watcher.py
```diff
@@ -250,12 +407,17 @@ class DbtSnapshotWatcherOperator(DbtSnapshotMixin, DbtConsumerWatcherSensor):  #
...
-    Executes a dbt source freshness command, synchronously, as ExecutionMode.LOCAL.
+class DbtSourceWatcherOperator(BaseConsumerSensor, DbtSourceLocalOperator):  # type: ignore[misc]
+    """Watches for source freshness results from the producer task.
+
+    When the producer has ``_check_source_freshness`` enabled it runs
+    ``dbt source freshness`` and pushes per-source status to XCom.
+    This sensor reads that status.  On retry (or when the producer did
+    not provide a result) it falls back to running ``dbt source freshness``
+    locally for its specific source.
     """
```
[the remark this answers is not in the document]
tatiana: Addressed in a8f828b53b033766588c91ee6e51615bc46847e5

cosmos/operators/watcher.py
```diff
@@ -250,12 +407,17 @@ class DbtSnapshotWatcherOperator(DbtSnapshotMixin, DbtConsumerWatcherSensor):  #
...
-    """
-    Executes a dbt source freshness command, synchronously, as ExecutionMode.LOCAL.
+class DbtSourceWatcherOperator(BaseConsumerSensor, DbtSourceLocalOperator):  # type: ignore[misc]
+    """Watches for source freshness results from the producer task.
+
+    When the producer has ``_check_source_freshness`` enabled it runs
+    ``dbt source freshness`` and pushes per-source status to XCom.
+    This sensor reads that status.  On retry (or when the producer did
+    not provide a result) it falls back to running ``dbt source freshness``
+    locally for its specific source.
```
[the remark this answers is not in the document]
tatiana: Addressed in 73538c762a147965e740323c797cc106d7cfe130

cosmos/operators/_watcher/triggerer.py
```diff
@@ -243,4 +248,9 @@ async def run(self) -> AsyncIterator[TriggerEvent]:
 
             # Sleep briefly before re-polling
             await asyncio.sleep(self.poke_interval)
-            logger.debug("Polling again for node '%s' status...", self.model_unique_id)
+            logger.info(
+                "Polling again for node '%s': status=%s, producer_state=%s",
+                self.model_unique_id,
+                dbt_node_status,
+                producer_task_state,
+            )
```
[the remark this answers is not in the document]
tatiana: Addressed in fd9e2a2f91f4fe6b39924a163021319d8a6fcb50

tests/operators/test_watcher.py
```diff
@@ -1922,3 +1926,163 @@ def test_fallback_raises_on_retry(self):
...
+            )
+        return producer
+
+    def test_init_stores_check_source_freshness_flag(self):
+        producer = self._make_producer(check_source_freshness=True)
+        assert producer._check_source_freshness is True
+
+    def test_init_default_check_source_freshness_is_false(self):
+        producer = self._make_producer(check_source_freshness=False)
+        assert producer._check_source_freshness is False
```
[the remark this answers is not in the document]
tatiana: Fixed in 270ad194511d69b6ac800ef56b37487b2558fbbd

tatiana (approved):
I've picked up this work from @pankajastro, rebased and made a few adjustments, which include:
- source nodes were running the source freshness, even though it had been run by the producer
- added a warning to the successful status
- fixed issues when sources had tests after rebase: https://github.com/astronomer/astronomer-cosmos/issues/2520

I logged a ticket with some of the improvements we agreed to do after:
https://github.com/astronomer/astronomer-cosmos/issues/2523

I also re-enabled the integration test (https://github.com/astronomer/astronomer-cosmos/issues/1802) and temporarily removed the non-ASCII model (https://github.com/astronomer/astronomer-cosmos/issues/2524). We'll address that in a follow up PR, since this one grew a lot.

PR #2470, opened 2026-03-17
Add config to allow disabling dag versioning

### LINKED ISSUE #2386: Add a config to allow disabling dag versioning
A user has [reported](https://github.com/astronomer/astronomer-cosmos/pull/1907#issuecomment-3917286831) that the dag versioning support we added in https://github.com/astronomer/astronomer-cosmos/pull/1907 is impacting their parsing performance. They are on Airflow 2.x. They would like a configurable option to allow disabling this computation for dag versioning.

pankajkoti:
Adds a Cosmos setting to disable DAG versioning (dbt project hash in `doc_md`) to improve DAG parsing performance, especially on Airflow 2.x.

New `airflow.cfg` setting: `enable_dag_versioning` in the `[cosmos]` section (default: `True`).
When False, Cosmos skips computing the dbt project hash and does not update dag.doc_md. Parsing is faster; on Airflow 3, DAG versioning will no longer reflect dbt project changes.

## Changes
 - `cosmos/settings.py`: Added `enable_dag_versioning = conf.getboolean("cosmos", "enable_dag_versioning", fallback=True)`.
- `cosmos/converter.py`: In `_add_dbt_project_hash_to_dag_docs`, return early when `settings.enable_dag_versioning` is `False`.
- `tests/test_converter.py`: Added `test_dag_versioning_skipped_when_setting_disabled` to assert the hash is not computed and `doc_md` is unchanged when the setting is disabled.
- `docs/reference/configs/cosmos-conf.rst`: Documented `enable_dag_versioning` (default, env var, behaviour when disabled).

## Usage
No DAG code changes. Configure once via `airflow.cfg` or with the corresponding environment variable:

**airflow.cfg:**
```
[cosmos]
enable_dag_versioning = False
```
or

**Environment variable:**
```
export AIRFLOW__COSMOS__ENABLE_DAG_VERSIONING=false
```

closes: #2386

### DIFF
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -386,6 +386,8 @@ class DbtToAirflowConverter:
         """
         if dag is None:
             return
+        if not settings.enable_dag_versioning:
+            return
 
         try:
             dbt_project_hash = _create_folder_version_hash(self.dbt_graph.project_path)
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -30,7 +30,14 @@ except ImportError:
 from cosmos import cache, settings
 from cosmos.airflow.graph import build_airflow_graph
 from cosmos.config import ExecutionConfig, ProfileConfig, ProjectConfig, RenderConfig
-from cosmos.constants import DbtResourceType, ExecutionMode, InvocationMode, LoadMode
+from cosmos.constants import (
+    _AIRFLOW3_MAJOR_VERSION,
+    AIRFLOW_VERSION,
+    DbtResourceType,
+    ExecutionMode,
+    InvocationMode,
+    LoadMode,
+)
 from cosmos.dbt.executable import get_system_dbt, is_dbt_installed_in_same_environment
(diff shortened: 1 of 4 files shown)


cosmos/converter.py
```diff
@@ -386,6 +386,8 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
         """
         if dag is None:
             return
+        if not settings.enable_dag_versioning:
```
tatiana, line 398: @pankajkoti maybe we should add another condition related to only add this if it's Airflow >= 3, WDYT?

tatiana (approved):
Great work, @pankajkoti , a minor feedback inline!

cosmos/converter.py
pankajkoti: Yes, great suggestion. Addressed in the newer commit.

pmuntianu:
thanks a lot team for bringing this flag. This was fast! ❤️

PR #2471, opened 2026-03-17
Reduce XCom read/write for tracking node state and errors in ConsumerWatcher task

pankajastro:
This PR aims to reduce XCom load in the watcher flow by limiting which dbt log/events are persisted for consumer-side logging and by removing the per-event XCom read used for deduplication.

**Changes:**
- Introduces an allowlist of dbt event names and only pushes per-node dbt events to XCom when the event is allowlisted.
- Removes the “skip duplicate event” XCom read-before-write behavior.

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -44,7 +44,10 @@ except ImportError:  # pragma: no cover
 
 logger = get_logger(__name__)
 
-# List of dbt error/failure events we are interested in, used to display logs in consumer tasks.
+# Subset of dbt event types that represent errors/failures.
+# Used (together with node status lifecycle events like NodeStart/NodeCompiling/
+# NodeExecuting/NodeFinished) to build _DBT_EVENT_ALLOWLIST, which controls which
+# events are surfaced in consumer tasks.
 # Source: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/events/types.py
 _DBT_ERROR_EVENTS_TYPES = frozenset(
     {
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -44,9 +44,9 @@ except ImportError:  # pragma: no cover
 
 logger = get_logger(__name__)
 
-# List of dbt event we are interested in to display logs in consumer tasks
-# https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/events/types.py
-_DBT_EVENTS_TYPE = frozenset(
+# List of dbt error/failure events we are interested in, used to display logs in consumer tasks.
+# Source: https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/events/types.py
+_DBT_ERROR_EVENTS_TYPES = frozenset(
     {
         "InvalidOptionYAML",
         "LogDbtProjectError",
@@ -72,13 +72,10 @@ _DBT_EVENTS_TYPE = frozenset(
         "FlushEventsFailure",
         "TrackingInitializeFailure",
         "ArtifactUploadError",
-        # Non-error state
-        "NodeStart",
-        "NodeCompiling",
-        "NodeExecuting",
-        "NodeFinished",
     }
 )
+
+_DBT_NODE_STATUS_EVENT_TYPES = frozenset({"NodeStart", "NodeCompiling", "NodeExecuting", "NodeFinished"})
 
 
 def _process_dbt_log_event(task_instance: Any, dbt_log: dict[str, Any] | EventMsg) -> None:
@@ -103,7 +100,7 @@ def _process_dbt_log_event(task_instance: Any, dbt_log: dict[str, Any] | EventMs
         finish_time = getattr(node_info, "node_finished_at", None) if node_info else None
         msg = getattr(dbt_log.info, "msg", None)
 
-    if event_name not in _DBT_EVENTS_TYPE:
+    if event_name not in (_DBT_ERROR_EVENTS_TYPES | _DBT_NODE_STATUS_EVENT_TYPES):
         return None
 
     if unique_id:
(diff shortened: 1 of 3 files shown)


cosmos/operators/_watcher/base.py
```diff
@@ -44,36 +44,67 @@
...
         info = dbt_log.get("info", {})
 
+        event_name = info.get("name")
         node_info = data.get("node_info")
         status = node_info.get("node_status") if node_info else None
         unique_id = node_info.get("unique_id") if node_info else None
         start_time = node_info.get("node_started_at") if node_info else None
         finish_time = node_info.get("node_finished_at") if node_info else None
         msg = data.get("msg") or info.get("msg") or None
     else:  # Runner
```
pankajkoti, line 101: What is the instance type of dbt_log here for dbtRunner? Like it is `dict` for the subprocess branch above.

cosmos/operators/_watcher/base.py
```diff
@@ -44,36 +44,67 @@
...
-    # Special case when node status is the string "None"; only process messages that contain an error or fail word
-    if status in ["None"] and msg is not None:
-        # Check if there is error log message
-        for sensitive_word in sensitive_words:
-            if sensitive_word in msg.lower():
-                break
-        else:
-            return None
+    if event_name not in _DBT_EVENTS_TYPE:
+        return None
```
pankajkoti: can we not move this block to the top of the method so we avoid further processing and return early?

cosmos/operators/_watcher/base.py
```diff
@@ -84,11 +115,6 @@ def _process_dbt_log_event(task_instance: Any, dbt_log: dict[str, Any] | EventMs
         }
 
         xcom_key = f"{unique_id.replace('.', '__')}_dbt_event"
-        # Avoid redundant XCom writes (and global lock contention) by only pushing
-        # when the event payload has changed.
-        existing_event = get_xcom_val(task_instance=task_instance, key=xcom_key, task_ids=PRODUCER_WATCHER_TASK_ID)
-        if existing_event == dbt_event:
-            return None
```
pankajkoti, line 91: Do we not need this anymore? Why was this needed before?

cosmos/operators/_watcher/base.py
```diff
@@ -44,36 +44,67 @@
 
 logger = get_logger(__name__)
 
+# List of dbt event we are interested in to display logs in consumer tasks
+# https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/events/types.py
+_DBT_EVENTS_TYPE = frozenset(
```
pankajkoti: Are there event types that we're not considering? Which are those, for example?

pankajkoti (approved):
It is looking good to me, but I have some questions in line for my understanding and curiosity. Are we confident that with a higher number of threads, it won't cause an issue?

I am happy to get this in once the above question and in-line questions are answered, and if we get another pair of eyes for review here.

cosmos/operators/_watcher/base.py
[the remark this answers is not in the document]
pankajkoti: +1 to the concern here that it will drift with dbt versions.

cosmos/operators/_watcher/base.py
```diff
@@ -44,36 +44,67 @@
...
+        "FlushEventsFailure",
+        "TrackingInitializeFailure",
+        "ArtifactUploadError",
+        # Non-error state
+        "NodeStart",
+        "NodeCompiling",
+        "NodeExecuting",
+        "NodeFinished",
+    }
+)
```
tatiana: @pankajastro, please, could you split this constant into two:
- `_DBT_ERROR_EVENTS_TYPES`
- `_DBT_NODE_STATUS_EVENT_TYPES`
 
And rename `_DBT_EVENTS_TYPE`, so it is clear that we are not mapping all dbt events to Cosmos (maybe `_LOGGED_DBT_EVENTS_TYPES`); in a way, it is a combination of the previous two.

In the comments, please specify which dbt version we used to define this list. I'm happy for you to consider Copilot's suggestion as well

cosmos/operators/_watcher/base.py
pankajastro: We are only considering these events means if the event belong to these types, we will display corresponding logs in the consumer

cosmos/operators/_watcher/base.py
pankajastro: The selected events are rarely duplicated. The most common duplicate event names are SkippingDetails, LogStartLine, Formatting, etc., anyway, I'm skipping now

cosmos/operators/_watcher/base.py
pankajastro: Fixed

cosmos/operators/_watcher/base.py
pankajastro: For the runner it is `EventMsg` and subprocess json

cosmos/operators/_watcher/base.py
pankajastro: We can't move it to the top because we parsed it differently for runner and subprocess. I have moved it to the respective if/else block. PTAL

cosmos/operators/_watcher/base.py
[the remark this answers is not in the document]
pankajastro: Yes, this approach can become outdated over time.

If new error event types or node state types are introduced, they won’t be captured in the consumer until this list is updated. That said, such changes are relatively infrequent, and we can extend the list as needed based on user reports.

An alternative would be to detect events dynamically (e.g., by matching substrings like "error" or "fail" in the event name), but this is more error-prone and could lead to inaccuracy.

cosmos/operators/_watcher/base.py
```diff
@@ -84,11 +115,6 @@ def _process_dbt_log_event(task_instance: Any, dbt_log: dict[str, Any] | EventMs
         }
 
         xcom_key = f"{unique_id.replace('.', '__')}_dbt_event"
-        # Avoid redundant XCom writes (and global lock contention) by only pushing
-        # when the event payload has changed.
-        existing_event = get_xcom_val(task_instance=task_instance, key=xcom_key, task_ids=PRODUCER_WATCHER_TASK_ID)
-        if existing_event == dbt_event:
-            return None
         safe_xcom_push(task_instance=task_instance, key=xcom_key, value=dbt_event)
```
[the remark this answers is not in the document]
pankajastro: resolving as there is open discussion hereL https://github.com/astronomer/astronomer-cosmos/pull/2471#discussion_r2954883869

tatiana (approved):

Issue #2317, opened 2026-02-03
[Bug] Missing debug level log


maoxingda:
### Astronomer Cosmos Version

1.13.0

### dbt-core version

1.11

### LoadMode

AUTOMATIC

### ExecutionMode

WATCHER

### InvocationMode

SUBPROCESS

### airflow version

3.1

### Operating System

mac

### Deployment

Official Apache Airflow Helm Chart

### What happened?

<img width="810" height="727" alt="Image" src="https://github.com/user-attachments/assets/adbedfc8-6af6-407f-84b3-16ea97f94d1b" />

### Relevant log output

```shell

```

### How to reproduce

see screenshot

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

tatiana:
@maoxingda Thanks for reporting this issue! Please could you share an example DAG and the debug message you expected to see?

maoxingda:
The output is expected to be consistent with that of `dbt --debug sub_command`.

In real-world usage scenarios, we need to extract the complete executable SQL from the debug logs.

maoxingda:
<img width="1460" height="948" alt="Image" src="https://github.com/user-attachments/assets/57ca430d-90d3-40a4-bc2c-79edf040f10f" />

tatiana:
Hi @maoxingda , dbt debug messages are logged at Python `DEBUG` level and require `AIRFLOW__LOGGING__LOGGING_LEVEL=DEBUG` to be visible, due to the way they were implemented:
https://github.com/astronomer/astronomer-cosmos/blob/31bc504397efa3e9d25ec73a94da918d90407061/cosmos/operators/_watcher/base.py#L86
Example:
<img width="1507" height="841" alt="Image" src="https://github.com/user-attachments/assets/a22d9c27-73f4-459a-94d7-4a1c9dd3ed26" />

I made a PR to update the documentation so this behaviour is clear:

PR #2491, opened 2026-03-23
Add troubleshooting note for dbt debug logs in `ExecutionMode.WATCHER`

tatiana:
Documents that dbt debug logs can be enabled by passing `--debug` via `dbt_cmd_global_flags` in `operator_args`, and that `AIRFLOW__LOGGING__LOGGING_LEVEL=DEBUG` must be set to make them visible when running dbt as a subprocess.

Relates to: https://github.com/astronomer/astronomer-cosmos/issues/2317


pankajkoti (approved):
Do those debug log lines come from https://github.com/astronomer/astronomer-cosmos/blob/31bc504397efa3e9d25ec73a94da918d90407061/cosmos/operators/_watcher/base.py#L231?

If yes, I guess we may have duplicate logs coming from that line and https://github.com/astronomer/astronomer-cosmos/blob/31bc504397efa3e9d25ec73a94da918d90407061/cosmos/operators/_watcher/base.py#L273. And, it should be okay? 

If possible, can we also please attach a screenshot to this PR showing that the debug logs now appear in the output?

tatiana:
@pankajkoti, thanks a lot for the review!

I had added the image to the ticket response. Please take a look:
https://github.com/astronomer/astronomer-cosmos/issues/2317#issuecomment-4110189525

I agree that there is an opportunity to remove some line duplication. I'll make a separate PR to address this.

PR #2498, opened 2026-03-23
Simplify and unify WATCHER implementation regardless of `InvocationMode`

tatiana:
This PR unifies `ExecutionMode.WATCHER` so it works almost identically with both `InvocationMode.SUBPROCESS` and `InvocationMode.DBT_RUNNER` - removing the previously separate code paths that depended on the invocation mode.                                                                                                                   
                                                                                                                                                                     
**Problem**                                                 
                                                                                                                                                                     
Before this PR, `ExecutionMode.WATCHER` had two completely different strategies depending on how dbt was invoked:
- `InvocationMode.DBT_RUNNER`: Subscribed to dbt's `EventMsg` callbacks, serialised each `NodeFinished` event as zlib-compressed base64 JSON under XCom key `nodefinished_<uid>`, and
pushed startup metadata under dbt_startup_events.                                                                                                             
- `InvocationMode.SUBPROCESS`: After dbt exited, read target/run_results.json and pushed the entire results blob once under XCom key run_results. Sensors could not react per-model 
until the full build finished.                                                                                                                                    

|                    | InvocationMode.DBT_RUNNER                                      | InvocationMode.SUBPROCESS                                   |
|--------------------|----------------------------------------------------------------|-------------------------------------------------------------|
| Producer mechanism | `_dbt_runner_callbacks` with structured `EventMsg` objects         | `_process_log_line_callable` + `--log-format json` parsing       |
| `XCom` key (status)  | `nodefinished_{uid}` (compressed+base64 JSON)                    | `{uid}_status` (plain string)                                 |
| Consumer path      | `_get_status_from_events()` (decompresses JSON)                  | `get_xcom_val(…_status)`                                      |
| Trigger path       | `_get_node_status()` → decompresses JSON                         | `_get_node_status()` → reads plain value                      |

This meant the consumer sensors needed to know the invocation mode at poke time (`use_event()` method), and the triggerer needed a `use_event` flag. The code was complex, duplicated, and the two modes delivered very different latency and observability.                                                                         
                                                                                              
  **Solution**                                                                                                                                                           
                                                                                              
Both invocation modes now feed through the same `store_dbt_resource_status_from_log` parser, which pushes per-model XCom keys `<unique_id>_status` as each `NodeFinished` event arrives:                                                                                                                                                    
                                                          
- `InvocatioMode.SUBPROCESS`: Each JSON log line from stdout is parsed directly by the callable wired to `run_subprocess`.                                                          ` 
- `InvocationMode.DBT_RUNNER`: Each `EventMsg` from the dbt callback is serialised to JSON via `google.protobuf.json_format.MessageToJson` (a transitive dbt-core dependency) and passed
 through the same parser. Stdout is redirected to suppress raw JSON duplication in Airflow logs.                                                                   
                                                                                            
Consumer sensors and the `WatcherTrigger` always poll the canonical `{uid}_status` key regardless of invocation mode.                                                  
                                                                    
**Key changes**                                                                                                                                                        
                                                                                                                                                                     
`cosmos/operators/watcher.py` — `DbtProducerWatcherOperator`
- Added `run_dbt_runner()` override: registers a `MessageToJson`-based callback that feeds `store_dbt_resource_status_from_log`. Callback is only registered when context
 is not None (guards against dbt deps calls during `_install_dependencies`, which would otherwise cause a phantom callback and `GenericExceptionOnRun` errors).        
- Added `contextlib.redirect_stdout(io.StringIO())` during the main build to suppress raw JSON from dbt's stdout (since events are handled via the callback).        
- Removed `EventMsg-based` `_handle_node_finished` / `_serialize_event` / `zlib-compressed` XCom strategy.                                                         
- Removed `InvocationMode-based` branching from `execute()`.                                                                                                           
                                                                                            
`cosmos/operators/_watcher/base.py`
- Removed `use_event()` abstract method and `_get_status_from_events()` / `_get_status_from_run_results()` — no longer needed.
- `_get_node_status()` now always reads `{uid}_status` XCom key (plus the `_tests_status` key for `DbtTestWatcherOperator`).                                               
- `poke()` calls `_log_startup_events()` for all non-test sensors regardless of invocation mode.                        
- `_process_dbt_log_event()` simplified: removed the `EventMsg` branch (dict-only now).                                                                                
- Added `_log_dbt_msg()` helper to improve log UX: shows human-readable dbt messages (e.g. `"Found 5 models, 3 seeds..."`) instead of raw JSON in Airflow task logs.   
Reduces cyclomatic complexity of `store_dbt_resource_status_from_log` from 11 to 9 (fixes Ruff C901).                                                                
- assert context is not None replaced with if context is None: return — avoids `AssertionError("")` → `GenericExceptionOnRun` when dbt's phantom callback fires without
 a context.                                                                                                                                                        
- `_log_startup_events()` now checks `isinstance(dbt_startup_events, list)` before iterating.                                                                          
                                                                                            
`cosmos/operators/_watcher/triggerer.py`
- Removed `use_event` parameter from `WatcherTrigger`.                                                                                                                 
- `_parse_dbt_node_status_and_compiled_sql()` simplified: always reads `{uid}_status` (no `nodefinished_*` key lookup).                                                  
                                                                                                                                                                   
`cosmos/operators/watcher_kubernetes.py `                                                                                                                            
- Removed `use_event()` stub.                                                                                                                                        
                                                                                            
Tests                                                                                                                                                          
- Removed tests for deleted code paths (`_handle_node_finished`, `_get_status_from_events`, `_get_status_from_run_results`, `use_event()`).                                
- Switched `test_dbt_dag_with_watcher` from `capsys` to `caplog` to avoid thread-safety issues with dbt's multi-threaded stdout writes.                                  
- Fixed `test_dbt_runner_caching_and_callbacks` to remove the now-unnecessary `EventMsg` patch.                                                                        
                                                                                                                                                                   
Test plan                                                         
                                                                                                                                                                   
- Unit tests: `hatch run tests.py3.11-2.10-1.9:test`
- Integration tests verify `ExecutionMode.WATCHER` runs end-to-end with both `SUBPROCESS` and `DBT_RUNNER` invocation modes                                              
- `airflow dags test example_watcher` completes without getting stuck or emitting GenericExceptionOnRun errors  
- Trigger watcher DAGs with `airflow standalone` and validate the output.


cosmos/operators/_watcher/base.py
```diff
@@ -82,31 +76,20 @@
...
-        msg = getattr(dbt_log.info, "msg", None)
+    event_name = info.get("name")
+    if event_name not in _DBT_EVENT_ALLOWLIST:
+        return None
+    node_info = data.get("node_info")
+    status = node_info.get("node_status") if node_info else None
+    unique_id = node_info.get("unique_id") if node_info else None
+    start_time = node_info.get("node_started_at") if node_info else None
+    finish_time = node_info.get("node_finished_at") if node_info else None
+    msg = data.get("msg") or info.get("msg") or None
```
tatiana:
```suggestion
    node_info = data.get("node_info") or {}
    status = node_info.get("node_status")
    unique_id = node_info.get("unique_id")
    start_time = node_info.get("node_started_at")
    finish_time = node_info.get("node_finished_at")
    msg = data.get("msg") or info.get("msg")
```

cosmos/operators/watcher.py
```diff
@@ -101,71 +84,65 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+
+            def _event_callback(event: Any) -> None:
+                from google.protobuf.json_format import MessageToJson
 
+                json_str = MessageToJson(event, preserving_proto_field_name=True)
+                parse(json_str, extra_kwargs)
+
+            self._dbt_runner_callbacks = [*(self._dbt_runner_callbacks or []), _event_callback]
+            with contextlib.redirect_stdout(io.StringIO()):
+                return super().run_dbt_runner(command, env, cwd, **kwargs)
```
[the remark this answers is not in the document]
tatiana: Addressed in ca56f7cb6790959024294634a98ac6f087d4f4a4

cosmos/operators/watcher.py
```diff
@@ -101,71 +84,65 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        context = kwargs.get("context")
+        if context is not None:
+            extra_kwargs: dict[str, Any] = {"project_dir": cwd, "context": context}
+            parse = self._make_parse_callable()
+
+            def _event_callback(event: Any) -> None:
+                from google.protobuf.json_format import MessageToJson
 
+                json_str = MessageToJson(event, preserving_proto_field_name=True)
+                parse(json_str, extra_kwargs)
```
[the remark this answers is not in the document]
tatiana: If we were to make this change, we'd be able to see the error line and traceback information, but the corresponding sensor tasks would probably hang indefinitely, waiting without ever being notified. If we just continued the flow, we might introduce other errors as well.

I believe I addressed the issue in: 319fd34d77a7fda5410e22ef86db957181dcf771

cosmos/operators/_watcher/triggerer.py
```diff
@@ -43,7 +40,6 @@ def __init__(
         dag_id: str,
         run_id: str,
         map_index: int | None,
-        use_event: bool,
         poke_interval: float = 5.0,
         is_test_sensor: bool = False,
```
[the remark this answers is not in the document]
tatiana: Addressed in c548c23aee6021e02b864d88f78702b10b909747

cosmos/operators/watcher.py
```diff
@@ -60,29 +46,41 @@
...
+    """Write-only sink that discards all data; used to suppress dbt stdout in DBT_RUNNER mode.
+
+    Preferred over ``io.StringIO()`` because StringIO buffers every byte written to it for the
+    lifetime of the context manager. On large projects dbt emits megabytes of JSON log lines,
+    so StringIO would grow unbounded and increase worker memory usage proportionally to project
+    size and verbosity. _NullWriter discards each write immediately with no allocation.
+    """
+
+    def write(self, *args: Any) -> int:
+        return 0
```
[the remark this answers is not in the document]
tatiana: Addressed in 47f7463b164268c117573baaa4d437a9f9183606

cosmos/operators/_watcher/base.py
```diff
@@ -193,6 +174,21 @@ def _store_startup_event_from_log(task_instance: Any, log_line: dict[str, Any])
     safe_xcom_push(task_instance=task_instance, key=_DBT_STARTUP_EVENTS_XCOM_KEY, value=current)
```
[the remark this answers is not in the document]
tatiana: Addressed in b1a85fa3267c2c50382a269775037047f41db4e7

tatiana:
@pankajastro @pankajkoti @michal-mrazek @johnhoran This PR is quite large, but I'd really love your review if you have a chance. You've all been working on the internals of the watcher, and I'd like your brains to check if I missed any important parts of the implementation and recent improvements we made.

michal-mrazek:
I am gonna test it. Just wanted to say now that unifying both invocation modes makes the code much better to develop, step in a good direction 🚀

cosmos/operators/_watcher/triggerer.py
```diff
@@ -43,16 +40,19 @@ def __init__(
...
+        # Accepted for upgrade-compatibility only: triggers serialized before the
```
pankajkoti, line 45: (optional, can be ignored) Probably good to mention the Cosmos version here. Like versions before 1.14.0 or something.

pankajkoti (approved):
Looks like a solid refactor to me! Thanks for adding all the code comments, making the review easier and the code maintainable. 

I did a sanity test on both SUBPROCESS and dbt Runner approaches, and it's looking fine to me with respect to pushing XCOMs, creating asset events, populating compiled SQLs, consistent logs and dag run durations. Also, the thread count was set to 12 in my BigQuery profile while testing, so it's looking concurrency safe. Good net 400+ lines being removed :)

cosmos/operators/_watcher/base.py
```diff
@@ -242,7 +249,8 @@ def store_dbt_resource_status_from_log(
         # TODO: handle all possible statuses including skipped, warn, etc.
         if is_dbt_node_status_terminal(dbt_node_status):
             context = extra_kwargs.get("context")
-            assert context is not None  # Make MyPy happy
+            if context is None:
```
michal-mrazek, line 252:
```suggestion
            if context is None:
                logger.warning(
                    "context is None for terminal node '%s' — XCom status will not be pushed. "
                    "This is unexpected and should never happen; check the caller is passing context correctly.",
                    unique_id,
                )
```

michal-mrazek (approved):
Looking good, I tested in local docker compose. Just added one small suggestion - it took me some time to figure out that the `if context is None` is just typing thing :)

cosmos/operators/_watcher/triggerer.py
```diff
@@ -43,16 +40,19 @@ def __init__(
...
+        # invocation-mode unification may still carry this kwarg. It is no longer
```
tatiana:
```suggestion
        # invocation-mode unification may still carry this kwarg (Cosmos < 1.14.0). It is no longer
```

cosmos/operators/_watcher/triggerer.py
tatiana: Great point, just addressed in unify-watcher-regardless-of-invocation-mode

cosmos/operators/_watcher/base.py
tatiana: Thanks a lot, @michal-mrazek, just applied your suggestion!

PR #2546, opened 2026-04-09
Fix mypy 1.20.0 type check failures

pankajkoti:
Update mirrors-mypy from v1.19.1 to v1.20.0 in pre-commit config. Change type: ignore error codes from [attr-defined] to [union-attr] in cosmos/airflow/graph.py. mypy 1.20.0 now correctly infers the union type DAGNode | Any and reports union-attr instead of attr-defined for attribute access on union members.

related: #2545

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -122,7 +122,7 @@ repos:
         pass_filenames: false
         files: ^docs/
   - repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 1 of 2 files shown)


pankajkoti:
Successful type-check job run in the CI for this PR: https://github.com/astronomer/astronomer-cosmos/actions/runs/24186882877/job/70593741210?pr=2546

pankajastro (approved):

PR #2553, opened 2026-04-11
Add aws and azure tarball upload options

TTMichaelA:
## Related Issue(s)

Related to PR #2497, this extends the ability to add all artifacts to a tarball to the upload functions for both aws and azure.

## Breaking Change?

No breaking changes anticipated. Should be silent for all present implementations, as the additional parameter is optional and the current behavior is preserved.

## Checklist

- [] I have made corresponding changes to the documentation (if required)

### DIFF
--- a/cosmos/io.py
+++ b/cosmos/io.py
@@ -25,6 +25,7 @@ def upload_to_aws_s3(
     bucket_name: str,
     aws_conn_id: str | None = None,
     source_subpath: str = DEFAULT_TARGET_PATH,
+    use_tarball: bool = False,
     **kwargs: Any,
 ) -> None:
     """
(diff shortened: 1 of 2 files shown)


pankajkoti (approved):
Thanks for the contribution @TTMichaelA . Just checking, were you able to verify the tarball option working end-to-end with GCP buckets and Azure containers?

TTMichaelA:
> Thanks for the contribution @TTMichaelA . Just checking, were you able to verify the tarball option working end-to-end with GCP buckets and Azure containers?

@pankajkoti to test, I wound up creating Docker containers for AWS and Azure, created a dummy manifest and run results, mocked an airflow context dict, called the functions with `use_tarball = True`, send the tarballs to the containers, and validated the files existed. Both seemed to work:

image

I have not run either in a proper production context as I did with the GCP version for the previous PR (as I don't have convenient environments available to me), but at least to the degree that I could/did, they seemed to work as expected.

pankajastro (commented):
LGTM

tatiana (approved):
Thanks a lot, @TTMichaelA ! We have those environments in Astronomer, and we can test this as part of its release. We'll ship this in 1.15.0

PR #2557, opened 2026-04-15
Avoid the Watcher consumers retrying to run a command that is being run by the producer

pankajkoti:
When a sensor times out while the producer is still running a long dbt build, the retry previously launched a standalone dbt run for the same model — creating a duplicate concurrent execution.

Now the sensor checks the producer task state before deciding:
- Producer terminated → fall back to local dbt run
- Producer still active → continue polling XCom instead of launching a duplicate run

closes: https://github.com/astronomer/oss-integrations-private/issues/359

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -28,6 +28,7 @@ from cosmos.operators._watcher.state import (
     is_dbt_node_status_skipped,
     is_dbt_node_status_success,
     is_dbt_node_status_terminal,
+    is_producer_task_terminated,
     safe_xcom_push,
     xcom_set_lock,
 )
@@ -583,6 +584,28 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
             if hasattr(self, "_override_rtif"):
                 self._override_rtif(context)
 
+    def _handle_retry(self, try_number: int, producer_task_state: str | None, context: Context) -> bool | None:
+        """Handle sensor retry by checking whether the producer is still active.
+
+        Returns True if the fallback ran successfully, or None if the sensor
+        should continue polling (producer still active).
+        """
+        if is_producer_task_terminated(producer_task_state):
+            # Producer finished — this is either an automatic retry after
+            # the producer completed or a manual task clear from the UI.
+            # Fall back to running the model locally.
+            return self._fallback_to_non_watcher_run(try_number, context)
+        # Producer is still active — the sensor likely timed out while the
+        # producer was still working.  Keep polling instead of launching a
+        # duplicate dbt run.
+        logger.info(
+            "Retry attempt #%s but producer '%s' is still %s — continuing to poll instead of fallback.",
+            try_number - 1,
+            self.producer_task_id,
+            producer_task_state or "unknown",
+        )
+        return None
+
     def poke(self, context: Context) -> bool:
         """
         Checks the status of a dbt node (model or aggregated tests) by pulling relevant XComs from the producer task.
@@ -605,10 +628,13 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
             self.model_unique_id,
         )
 
-        if try_number > 1:
-            return self._fallback_to_non_watcher_run(try_number, context)
-
         producer_task_state = self._get_producer_task_status(context)
+
+        if try_number > 1:
+            retry_result = self._handle_retry(try_number, producer_task_state, context)
+            if retry_result is not None:
+                return retry_result
+
         if not self.is_test_sensor:
             self._log_startup_events(ti)
         status = self._get_node_status(ti, context)
@@ -624,8 +650,13 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
         )
         _log_dbt_event(dbt_events)
 
-        if status is None:
+        return self._evaluate_node_status(status, producer_task_state, try_number, context)
 
+    def _evaluate_node_status(
+        self, status: Any, producer_task_state: str | None, try_number: int, context: Context
+    ) -> bool:
+        """Evaluate the dbt node status and return the poke result."""
+        if status is None:
             if producer_task_state == "failed":
                 if self.poke_retry_number > 0:
                     raise AirflowException(
@@ -636,13 +667,12 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
                     return self._fallback_to_non_watcher_run(try_number, context)
 
             self.poke_retry_number += 1
-
             return False
-        elif is_dbt_node_status_skipped(status):
+
+        if is_dbt_node_status_skipped(status):
             raise AirflowSkipException(
                 f"{self._resource_label} '{self.model_unique_id}' was skipped by the dbt command."
             )
-        elif is_dbt_node_status_success(status):
+        if is_dbt_node_status_success(status):
             return True
-        else:
-            raise AirflowException(f"{self._resource_label} '{self.model_unique_id}' finished with status '{status}'")
+        raise AirflowException(f"{self._resource_label} '{self.model_unique_id}' finished with status '{status}'")
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -587,20 +587,20 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
     def _handle_retry(self, try_number: int, producer_task_state: str | None, context: Context) -> bool | None:
         """Handle sensor retry by checking whether the producer is still active.
 
-        Returns True if the fallback ran successfully, or None if the sensor
-        should continue polling (producer still active).
+        Returns the fallback result if the producer has terminated, or None if
+        the sensor should continue polling (producer still active).
         """
         if is_producer_task_terminated(producer_task_state):
             # Producer finished — this is either an automatic retry after
             # the producer completed or a manual task clear from the UI.
-            # Fall back to running the model locally.
+            # Fall back to a non-watcher run.
             return self._fallback_to_non_watcher_run(try_number, context)
         # Producer is still active — the sensor likely timed out while the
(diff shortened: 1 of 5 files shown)


pankajkoti:
Try number = 2 for the sensor when producer is still running. 

Screenshot 2026-04-15 at 2 01 58 PM

cosmos/operators/_watcher/base.py
```diff
@@ -624,8 +650,13 @@ def poke(self, context: Context) -> bool:
         )
         _log_dbt_event(dbt_events)
 
-        if status is None:
+        return self._evaluate_node_status(status, producer_task_state, try_number, context)
 
+    def _evaluate_node_status(
```
pankajkoti, line 655: Moved this to a new method, as this was getting complained about code complexity

cosmos/operators/_watcher/base.py
```diff
@@ -583,6 +584,28 @@ def _cache_compiled_sql(self, ti: Any, context: Context) -> None:
             if hasattr(self, "_override_rtif"):
                 self._override_rtif(context)
 
+    def _handle_retry(self, try_number: int, producer_task_state: str | None, context: Context) -> bool | None:
```
tatiana, line 587: I believe it would be a cleaner interface if we renamed things to `is_producer_still_running`.
Retry is a consequence depending on other variables - including if the producer is actively running the task

cosmos/operators/_watcher/base.py
```diff
@@ -583,6 +584,28 @@ def _cache_compiled_sql(self, ti: Any, context: Context) -> None:
...
+            # Producer finished — this is either an automatic retry after
+            # the producer completed or a manual task clear from the UI.
+            # Fall back to running the model locally.
+            return self._fallback_to_non_watcher_run(try_number, context)
+        # Producer is still active — the sensor likely timed out while the
+        # producer was still working.  Keep polling instead of launching a
+        # duplicate dbt run.
+        logger.info(
+            "Retry attempt #%s but producer '%s' is still %s — continuing to poll instead of fallback.",
+            try_number - 1,
```
tatiana: Why are we setting `try_number - 1` here?

cosmos/operators/_watcher/state.py
```diff
@@ -55,6 +55,17 @@ def is_dbt_node_status_terminal(status: str | None) -> bool:
     return is_dbt_node_status_success(status) or is_dbt_node_status_failed(status) or is_dbt_node_status_skipped(status)
 
 
+# Airflow task states that indicate the producer has finished and will not
+# deliver any more XCom updates.  Used to decide whether a sensor retry
+# should fall back to running the model locally or keep polling.
+PRODUCER_TERMINAL_STATES = frozenset({"success", "failed", "skipped", "upstream_failed", "removed"})
```
tatiana: Let's keep constants at the top of the file - together with other ones - as per convention

cosmos/operators/_watcher/base.py
tatiana: @pankajkoti Would there be an alternative way to refactor?

The function name (_evaluate_node_status) does not seem accurate with what it is doing, which is actually to:
- check the status of the node
- check the status of the producer airflow task
- maybe execute dbt itself to run, as a fallback

These steps feel quite critical and relevant in the `poke` method, and it is probably hard to find a suitable name in a single method. There may be other parts of the function that could be refactored and isolated that are not so core to the logic (maybe the retrieval of status, or the logging handling?)

tatiana (commented):
I'm really happy we're improving this part, @pankajkoti ! I left some questions and advice on the implementation. While they will not change the overall logic - your fix seems accurate - I believe they will help with the code maintainability.

cosmos/operators/_watcher/base.py
pankajkoti: Airflow sets the try number to 2 for the first retry, and so on. To evaluate that way, we have it set that way. Would you like to suggest something different here instead?

cosmos/operators/_watcher/base.py
pankajkoti: I believe `_handle_retry` still makes sense as the name here. It's called from the `poke` method when `try_number > 1`, i.e., when we're in an Airflow retry scenario, and this method, as part of handling the Airflow retry, decides whether to actually fall back (producer terminated) or keep polling (producer still active).

cosmos/operators/_watcher/base.py
pankajkoti: I think the name broadly reflects the method’s primary responsibility: it evaluates the node status and determines the appropriate poke outcome. By "evaluate/evaluation", I’m including checking dependencies like the producer task state and deciding on actions such as fallback, retry, raise, or succeed. The producer state check and fallback are part of handling cases where _no direct status_ is available.

That said, I agree it’s doing a few important things, so I’m open to renaming if you have a more precise suggestion.

On extracting retrieval/logging: those are already separated into helper methods (`_log_startup_events`, `_get_node_status`, `_cache_compiled_sql`). Wrapping them further would likely add indirection without much benefit. The evaluation logic, on the other hand, forms a cohesive decision tree, which is why it felt reasonable to group it into a single method.

Happy to refactor further if you feel a different split would improve readability.

cosmos/operators/_watcher/base.py
pankajkoti: Changed the word retry ->  try and printing the try_number value here.

cosmos/operators/_watcher/state.py
pankajkoti: addressed

cosmos/operators/_watcher/state.py
```diff
@@ -55,6 +55,17 @@ def is_dbt_node_status_terminal(status: str | None) -> bool:
     return is_dbt_node_status_success(status) or is_dbt_node_status_failed(status) or is_dbt_node_status_skipped(status)
 
 
+# Airflow task states that indicate the producer has finished and will not
+# deliver any more XCom updates.  Used to decide whether a sensor retry
+# should fall back to running the model locally or keep polling.
```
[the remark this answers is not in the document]
pankajkoti: addressed

cosmos/operators/_watcher/base.py
```diff
@@ -583,6 +584,28 @@ def _cache_compiled_sql(self, ti: Any, context: Context) -> None:
...
+    def _handle_retry(self, try_number: int, producer_task_state: str | None, context: Context) -> bool | None:
+        """Handle sensor retry by checking whether the producer is still active.
+
+        Returns True if the fallback ran successfully, or None if the sensor
+        should continue polling (producer still active).
+        """
+        if is_producer_task_terminated(producer_task_state):
+            # Producer finished — this is either an automatic retry after
+            # the producer completed or a manual task clear from the UI.
+            # Fall back to running the model locally.
```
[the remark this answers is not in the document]
pankajkoti: addressed

pankajkoti:
> I'm really happy we're improving this part, @pankajkoti ! I left some questions and advice on the implementation. While they will not change the overall logic - your fix seems accurate - I believe they will help with the code maintainability.

Thanks for the review @tatiana. I left my responses on the feedback and addressed some other minor comments that I have marked as resolved.

cosmos/operators/_watcher/base.py
tatiana: Functions should be named based on what they do, not how they are used. The current name (`_handle_retry`) is related to where it is used, rather than to its name.
 
Naming functions after their action creates reusable, predictable code. If a function is named after its specific use case, it becomes difficult to reuse that same logic elsewhere.

cosmos/operators/_watcher/base.py
tatiana: This function combines decision-making, state retrieval, and side effects (running dbt). While it is small, it has a lot of responsibility. I'm worried about future readability and testability.

Maybe we could break it down into two smaller methods with clear responsibilities, something along the lines:
- `_evaluate_node_status`: implements the decision tree logic
- `_handle_node_decision(decision, ...)`: executes side effects - run dbt or others

cosmos/operators/_watcher/base.py
tatiana: Thanks, @pankajkoti !

cosmos/operators/_watcher/base.py
pankajkoti: I am a bit torn here. `_handle_retry` actually does describe what the method does; it handles the retry scenario. It checks the producer state, decides whether to fall back or keep polling, and acts accordingly. That is the method's responsibility. It's not named after its call site; it's named after the action it performs.

When the time comes for reusing (although this function is marked local to the module/class), I believe these are what the actions will be for handling retries for the case, no? Or do you see a different path there?

We already have `is_producer_task_terminated` for the state check. The suggested `is_producer_still_running` shouldn't call an action based on its name, so the fallback logic would have to move elsewhere, either into another method or back into `poke`. If it goes into poke, it's not reusable either.

cosmos/operators/_watcher/base.py
pankajkoti: Looking at the method again, it doesn’t actually do state retrieval. Both `status` and `producer_task_state` are passed in; the retrieval happens in `poke`. So this method is really decision-making and taking actions based on the decisions.

I feel splitting it would introduce indirection without reducing coupling; the action method would still need the same parameters (try_number, context) for the fallback path.

Right now, the full flow is visible in one place:
- status is None -> check producer ->  fallback or keep polling
- skipped ->  skip
- success -> done

If we split this into decision + handler, a reader would need to jump between methods and map decisions to actions, which I think hurts readability more than it helps at this size.

I’m happy to rename the method if the current name doesn’t reflect its behaviour well, open to suggestions there. But I’d prefer not to split it unless the logic grows or we see reuse emerging.

pankajkoti:
We now have conflicts here after the merge of PR #2559. It felt simpler to create a new PR than to resolve the conflicts here. I have created a new PR at https://github.com/astronomer/astronomer-cosmos/pull/2592, so closing this PR in favour of that one.

Issue #2052, opened 2025-10-27
Benchmark `ExecutionMode.WATCHER`


tatiana:
We have DAG run-time improvements for the watcher execution mode, but we do not have published data on memory and CPU utilisation. We should obtain and share these.

tatiana:
This is particularly important since some users reported issues with memory and vCPU

maoxingda:
May I ask what the current status is?

pankajkoti:
Action items:
* [ ] 1. Rerun the benchmark steps and confirm the times match the previous iteration's run - @pankajkoti & @tatiana 
* [ ] 2. If there is a difference between 2 runs (Pankaj vs Tatiana or Tatiana previous vs Tatiana now), understand the why? If no difference between the runs, get the local machines in sync to comparable times. - @pankajkoti & @tatiana 
* [ ] 3. Get [PR #6](https://github.com/astronomer/cosmos-benchmark/pull/7) to completion (without the metrics) - @tatiana 
* [ ] 4. Track the memory, CPU & run time in the distributed Airflow setup created with Helm chart (depends on https://github.com/astronomer/cosmos-benchmark/pull/7) -  @pankajkoti 
* [ ] 5. Decide which test cases to run to compare the metrics for benchmarking (e.g., cosmos watcher vs cosmos local, watcher multiple threads, cosmos vs dbt-core (run in Airflow bashoperator, dbt core cli command). Check whether to preinstall dbt deps or not, benchmark the performance when pre-installing vs not. - @pankajkoti
* [ ] 6. Where to publish the metrics (previously tracked in https://docs.google.com/spreadsheets/d/10c6hMjwBJZ1DybiJT8o0TApVPyk-oPuJCY01bj2zC0A/edit?gid=0#gid=0)? Starter idea is to create a new spreadsheet. - @pankajkoti 
* [ ] 7. Iteratively, create a PR in astronomer-cosmos documentation comparing the performance of WATCHER vs LOCAL with these results. - @pankajkoti

tatiana:
**1. Rerun the benchmark steps and confirm the times match the previous iteration's run**

I re-run and run into a few issues :( I created these PRs to be able to run again:
- https://github.com/astronomer/cosmos-benchmark/pull/20
- https://github.com/astronomer/cosmos-benchmark/pull/21

After this, I got into the same stage as Pankaj Koti, where the command timed out:
```
[...]

PR #2559, opened 2026-04-15
Fix watcher producer retries behaviour

### LINKED ISSUE #2554: [Bug] Watcher incorrectly reporting when producer fails
### Astronomer Cosmos Version

1.14.0

1.10.13

### Versions of dbt adapters

_No response_

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

WATCHER

### InvocationMode

SUBPROCESS

### airflow version

2.10.1

### Operating System

amazon linux

_No response_

### Deployment

Amazon (AWS) MWAA

### Deployment details

_No response_

### What happened?

I am testing the new Watcher with 1.14.0 on MWAA. Here is a scenario I noticed:

- Producer tasks fails on the exact same dbt tests (we have a model with 6 tests), probably some edge case with test aggregation that causes the producer to fails
- Producer retries and logs "Dbt WATCHER producer task does not support Airflow retries. Detected attempt #2; skipping execution to avoid running a second dbt build." and is green
- Following consumer tasks are skipped due to "was skipped by the dbt command. This may happen if it is an ephemeral model or if the model sql file is empty."

So something strange in test aggregation is causing that the part of our dbt pipeline does not work, but ends in green.

For us, this is breaking bug and we will not use watcher in 1.14.0 until we fix it. I will report more details once I more investigate

### Relevant log output

```shell

```

### How to reproduce

not sure yet

### Anything else :)?

_No response_

### Contact Details

_No response_

tatiana:
The producer retry behaviour change in Cosmos 1.14.0 had issues, as detailed in  #2430:                                                                          
- Consumer tasks were incorrectly marked as successful even when one of their models failed in the producer, because they only captured the producer status after it retried and "succeeded" (behaving as an empty operator on the second try)       
- The producer was marked as successful on retry, while the correct behaviour is to skip it                                                                     
- Producer XCom values were deleted during retries (standard Airflow behaviour), leading to incorrect status on consumer tasks 

Affted both:
- `ExecutionMode.WATCHER`
- `ExecutionMode.WATCHER_KUBERNETES`                      

**Behaviour before and after**
                                                 
Cosmos 1.14.0 - DAG `dev/failed_dags/example_watcher_failing_tests.py`

https://github.com/user-attachments/assets/9d02fbf4-d4d7-47c8-860e-6d172acd2ded

Cosmos 1.14.1a2 (cut from this branch) -  DAG `dev/failed_dags/example_watcher_failing_tests.py`

https://github.com/user-attachments/assets/0eb4e055-7242-459b-a3a2-3f84864d41c6
                             
**Main changes**

(1) Producer retry behaviour

Fix overall behaviour introduced in Cosmos 1.14.0:
- Producer raises `AirflowSkipException` on retry instead of re-running the dbt build 
- Consumer sensors detect the skipped producer and fall back to running dbt individually via `_fallback_to_non_watcher_run  `

(3) XCom backup/restore for retry resilience                                        
 
Overcome a limitation of Airflow in refreshing the `XCom` between DAG retrues. This is a workaround. On the mid-long term, this may be addressed as part of [AIP-103](https://cwiki.apache.org/confluence/display/AIRFLOW/AIP-103%3A+Task+State+Management) after Airflow 3.2 (the current latest version).

- Added incremental XCom backup: `each safe_xcom_push` accumulates key/value pairs in a buffer on the task instance, persisted to an Airflow Variable on each push
- On failure, the backup Variable is preserved so the retry can restore it      
- On success, the backup Variable is deleted to prevent stale data from accumulating 
- On retry (try_number > 1), XCom entries are restored from the Variable before the producer raises `AirflowSkipException`
- Uses the Airflow Variable API (not direct DB queries) for Airflow 2 and 3 compatibility                                                                   

An alternative to updating the backup variable per model update would be to do so by the end of the execution. The one problem with this approach is that it would not work for tasks killed by OOM or -9 from the OSS, but it would result in fewer Airflow metadata database calls.

**How to reproduce**

Run the following DAGs referencing the dbt project `watcher_failing_tests` that was added to Cosmos in this PR:

- `example_watcher_failing_tests.py` - watcher + dbt_runner (default)
- `example_watcher_failing_tests_subprocess.py` - watcher + subprocess
- `example_watcher_failing_multi_folder.py` - watcher + group_nodes_by_folder
- `example_watcher_failing_tests_kubernetes.py` - watcher_kubernetes
- `example_watcher_failing_tests_non_deferrable.py` - watcher + dbt_runner + non-deferrable

All reference dbt projects in dev/dags/dbt/.

I updated `scripts/test/kubernetes-setup.sh` to create the Docker image with the failing dbt project and to upload it to kind.

To use with `airflow standalone`, set `AIRFLOW__CORE__DAGS_FOLDER` to `dev/failed_dags/`.

**Pending work**

The following will be accomplished outside of the scope of this PR:
- Implement an approach to delete stale variables that backed up XCom values across DAG run retries
- Collect user feedback

Closes #2554   

Personal note: sometimes the DAG wouldn't work as expected (everything would be green), because something else had failed and the test fixture would be left hanging. I'd use this workaround:
```
docker exec dc59f59911eb psql -U postgres -d postgres -c "DROP SEQUENCE IF
       EXISTS public._cosmos_fail_once_seq;"
```


cosmos/operators/_watcher/triggerer.py
```diff
@@ -211,6 +212,16 @@ async def _log_startup_events(self) -> None:
 
             await asyncio.sleep(self.poke_interval)
 
+    def _build_success_event(self, compiled_sql: str | None) -> dict[str, Any]:
```
tatiana, line 215: Refactored pre-existing code into a new method due to `run` complexity

cosmos/operators/_watcher/__init__.py
```diff
@@ -1,12 +1,12 @@
...
-    "safe_xcom_push",
     "build_producer_state_fetcher",
-    "is_dbt_node_status_success",
+    "get_xcom_val",
     "is_dbt_node_status_failed",
+    "is_dbt_node_status_success",
     "is_dbt_node_status_terminal",
+    "safe_xcom_push",
     "WatcherTrigger",
 ]
```
tatiana, line 11: Just sorted alphabetically

cosmos/operators/_watcher/base.py
```diff
@@ -583,6 +593,23 @@ def _cache_compiled_sql(self, ti: Any, context: Context) -> None:
             if hasattr(self, "_override_rtif"):
                 self._override_rtif(context)
 
+    def _handle_no_dbt_node_status(self, producer_task_state: str | None, try_number: int, context: Context) -> bool:
+        """Handle the case where no dbt node status has been reported yet."""
+        if producer_task_state == "failed":
```
pankajkoti, line 605: Should we introduce constants for these states? failed/skipped?

cosmos/operators/_watcher/base.py
```diff
@@ -583,6 +593,23 @@ def _cache_compiled_sql(self, ti: Any, context: Context) -> None:
...
+            if self.poke_retry_number > 0:
+                raise AirflowException(
+                    f"The dbt build command failed in producer task. Please check the log of task {self.producer_task_id} for details."
+                )
+            else:
+                # This handles the scenario of tasks that failed with `State.UPSTREAM_FAILED`
+                return self._fallback_to_non_watcher_run(try_number, context)
+
+        if producer_task_state == "skipped":
```
pankajkoti, line 614: same question about constants for state here.

cosmos/operators/_watcher/xcom.py
```diff
@@ -0,0 +1,122 @@
...
+
+def _persist_backup(var_key: str, backup_buffer: dict[str, Any]) -> None:
+    """Write the current backup buffer to an Airflow Variable."""
+    if not backup_buffer:
+        return
+
+    from airflow.models import Variable
+
+    compressed = base64.b64encode(zlib.compress(json.dumps(backup_buffer, default=str).encode("utf-8"))).decode("utf-8")
+    Variable.set(var_key, compressed)
```
pankajastro, line 62: Wondering if we need lock here?

cosmos/operators/_watcher/xcom.py
```diff
@@ -0,0 +1,122 @@
...
+    var_key = getattr(ti, "_cosmos_xcom_backup_var_key", None)
+    if not isinstance(var_key, str):
+        return
+    try:
+        from airflow.models import Variable
+
+        Variable.delete(var_key)
+        logger.debug("Deleted XCom backup Variable '%s'", var_key)
+    except KeyError:
+        pass
```
pankajastro, line 95: Maybe worth adding a debug here too.

cosmos/operators/_watcher/xcom.py
```diff
@@ -0,0 +1,122 @@
...
+    if compressed is None:
+        logger.info("No XCom backup Variable found at '%s'", var_key)
+        return False
+
+    backup: dict[str, Any] = json.loads(zlib.decompress(base64.b64decode(compressed.encode("utf-8"))).decode("utf-8"))
+    for key, value in backup.items():
+        safe_xcom_push(task_instance=ti, key=key, value=value)
+    logger.info("Restored %d XCom entries from Variable '%s'", len(backup), var_key)
+
+    Variable.delete(var_key)
```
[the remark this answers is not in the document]
pankajastro: I think Copilot suggestion makes sense here

pankajastro (commented):
Looking great! I think we should also manually validate the DAG using both SUBPROCESS and RUNNER invocation modes.

cosmos/operators/_watcher/base.py
```diff
@@ -544,6 +544,16 @@ def execute_complete(self, context: Context, event: dict[str, Any]) -> None:
...
             )
 
+        if reason == WatcherEventReason.PRODUCER_SKIPPED:
+            logger.info(
+                "Watcher producer task '%s' was skipped. Falling back to running dbt for %s '%s'.",
+                self.producer_task_id,
+                self._resource_label.lower(),
+                self.model_unique_id,
+            )
+            self._fallback_to_non_watcher_run(try_number=2, context=context)
```
pankajkoti: Trying to understand here. So the retry triggered here as part of the fallback happens in the same current Airflow `try_number=1` context, right? It does not change the Airflow task state for the current `try_number` and start a new Airflow try, right? The `try_number=2` could confuse readability. Should we refactor this method `_fallback_to_non_watcher_run` to rename the `try_number` argument to `retry_attempt` if it does not alter the Airflow `try_number` or require the value of Airflow `try_number`?

cosmos/operators/_watcher/xcom.py
pankajkoti, line 62: Should we wrap this in a try-except block to handle exceptions like they're being handled in https://github.com/astronomer/astronomer-cosmos/pull/2573

cosmos/operators/_watcher/xcom.py
```diff
@@ -0,0 +1,122 @@
...
+    ti = context["ti"]
+    var_key = getattr(ti, "_cosmos_xcom_backup_var_key", None)
+    if not isinstance(var_key, str):
+        return
+    try:
+        from airflow.models import Variable
+
+        Variable.delete(var_key)
+        logger.debug("Deleted XCom backup Variable '%s'", var_key)
+    except KeyError:
```
pankajkoti, line 94: For this `Variable.delete` call here, should we also catch here `AirflowRuntimeError` like done for the `set` call in Should we wrap this in a try-except block to handle exceptions like they're being handled in https://github.com/astronomer/astronomer-cosmos/pull/2573

cosmos/operators/_watcher/xcom.py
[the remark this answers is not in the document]
pankajkoti: +1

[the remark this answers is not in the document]
pankajkoti: and also catching AirflowRuntimeError?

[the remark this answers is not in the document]
tatiana: Addressed in 6efaeed3c2b91fc423d1674d9edc2cc9040636e7

cosmos/operators/_watcher/base.py
```diff
@@ -583,6 +593,23 @@ def _cache_compiled_sql(self, ti: Any, context: Context) -> None:
...
+            if self.poke_retry_number > 0:
+                raise AirflowException(
+                    f"The dbt build command failed in producer task. Please check the log of task {self.producer_task_id} for details."
+                )
+            else:
+                # This handles the scenario of tasks that failed with `State.UPSTREAM_FAILED`
+                return self._fallback_to_non_watcher_run(try_number, context)
+
+        if producer_task_state == "skipped":
+            return self._fallback_to_non_watcher_run(try_number, context)
```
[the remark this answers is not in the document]
tatiana: Addressed in 73da5275c669dd43b73b25887020f3bda1f4217b

cosmos/operators/_watcher/xcom.py
tatiana: That's a great question, @pankajastro! ` _persist_backup` is called from `safe_xcom_push` which already holds `xcom_set_lock`.   So `_persist_backup` is already serialised — no additional lock needed. The `xcom_set_lock` protects both the XCom push and the backup buffer update and variable write as one atomic operation.

cosmos/operators/_watcher/xcom.py
tatiana: Addressed in 91ee5398362e25f9aef8ce6214da15d2124f755b

cosmos/operators/watcher_kubernetes.py
```diff
@@ -107,12 +115,11 @@ def execute(self, context: Context, **kwargs: Any) -> Any:
         try_number = getattr(task_instance, "try_number", 1)
 
         if try_number > 1:
-            self.log.info(
+            _restore_xcom_from_variable(context)
```
pankajkoti, line 116: We would need to call the `_init_xcom_backup(context)` in the `execute` method here, like the `watcher.py` `execute()` here, right? Not sure if the restore would have an effect here, otherwise. And also need to call `_delete_xcom_backup_variable(context)` on success?

cosmos/operators/_watcher/xcom.py
```diff
@@ -0,0 +1,126 @@
...
+
+    backup: dict[str, Any] = json.loads(zlib.decompress(base64.b64decode(compressed.encode("utf-8"))).decode("utf-8"))
+    for key, value in backup.items():
+        safe_xcom_push(task_instance=ti, key=key, value=value)
+    logger.info("Restored %d XCom entries from Variable '%s'", len(backup), var_key)
+
+    try:
+        Variable.delete(var_key)
+        logger.debug("Deleted XCom backup Variable '%s' after restore", var_key)
+    except KeyError:
```
pankajkoti, line 124: same question here about catching AirflowRuntimeError too

cosmos/operators/_watcher/xcom.py
```diff
@@ -0,0 +1,126 @@
...
+XCOM_BACKUP_VARIABLE_PREFIX = "cosmos_xcom_backup__"
+
+
+def _xcom_backup_variable_key(dag_id: str, task_group_id: str | None, run_id: str) -> str:
+    """Build a unique Airflow Variable key for the XCom backup of a watcher producer run."""
+    parts = [XCOM_BACKUP_VARIABLE_PREFIX, dag_id.replace(".", "___")]
+    if task_group_id:
+        parts.append(task_group_id.replace(".", "__"))
+    parts.append(run_id.replace(".", "_"))
+    return "__".join(parts)
```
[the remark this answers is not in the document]
tatiana: This is not a real use case. There will be only one producer in the top level.

cosmos/operators/_watcher/base.py
```diff
@@ -544,6 +551,16 @@ def execute_complete(self, context: Context, event: dict[str, Any]) -> None:
...
             )
 
+        if reason == WatcherEventReason.PRODUCER_SKIPPED:
+            logger.info(
+                "Watcher producer task '%s' was skipped. Falling back to running dbt for %s '%s'.",
+                self.producer_task_id,
+                self._resource_label.lower(),
+                self.model_unique_id,
+            )
+            self._fallback_to_non_watcher_run(try_number=2, context=context)
```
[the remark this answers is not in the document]
tatiana: Addressed in 074c37ec6db698b875efdc28e385f32270d108eb

pankajkoti (approved):
1. Now that we also handle variable access within the lock, do we see any performance degradation before and after the PR with a quick dag run WRT to the total dag run duration?

2. Can we sanity test this with the following combinations:
 i. subprocess vs dbt Runner
ii. deferrable vs non-deferrable (together with subprocess vs dbt Runner)
iii. threads count > 1 (maybe 8 or 12)
iv. Is the Kubernetes mode XCOM restore functional?

3. Also, this will certainly cause conflicts with PR https://github.com/astronomer/astronomer-cosmos/pull/2557 that we opened last week. How should we tackle it? I was able to do good testing on that PR. But I am not sure, I will have time this week to handle the conflicts and retest. If we want to get this PR in first, should we defer the other issue/PR to the next sprint? Or would it be possible to incorporate the changes in that PR into this PR, and we can close the other PR? 

LGTM, I am approving, and have some suggestions and questions inline as well.

cosmos/operators/_watcher/xcom.py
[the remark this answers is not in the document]
pankajkoti: Reopening the conversation to check on also catching AirflowRuntimeError?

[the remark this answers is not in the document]
pankajkoti: Reopening the conversation to check on also catching AirflowRuntimeError?

tatiana:
I managed to cut 1.14.1a1, so more users can test this change cc @johnhoran @michal-mrazek

cosmos/operators/_watcher/xcom.py
tatiana: I believe this is not needed, since the previous error happened during DAG-parsing in the Astro Executor, and this one is during task execution

cosmos/operators/watcher_kubernetes.py
tatiana: Great point, @pankajkoti , I just fixed the K8s behaviour in eb0aee932494b8113d229b11a79717812cf5886c and tested the DAG failure scenario as well - both via a unittest and manual test

cosmos/operators/_watcher/xcom.py
tatiana: I believe the same applies here: https://github.com/astronomer/astronomer-cosmos/pull/2559#discussion_r3116921777

cosmos/operators/_watcher/xcom.py
tatiana: I believe this is not needed, since the previous error happened during DAG-parsing in the Astro Executor, and this one is during task execution

cosmos/operators/_watcher/base.py
tatiana: @pankajkoti I made a few improvements in the logic in the previous commits, please, could you take a look and confirm if it is better?

cosmos/operators/_watcher/base.py
tatiana: I fully agree with this request. Since we have been using this pattern in other parts and my PR is already ginormous, could I address this in a follow-up PR?

michal-mrazek:
@tatiana I can confirm we no longer see this issue, thank you for such a quick fix ❤️ 

One thing (not related) that could be improved is the logging. I already assigned myself the issues for better logs, so I can help with that. Ideally, from user perspective, also seeing something like "5 out of 5 dbt tests suceeded" would be nice in the test nodes :)

michal-mrazek:
@tatiana but that can be done in 1.14.2 :)

cosmos/operators/watcher.py
```diff
@@ -176,6 +182,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         kwargs["should_store_compiled_sql"] = False
         kwargs.setdefault("priority_weight", PRODUCER_WATCHER_DEFAULT_PRIORITY_WEIGHT)
         kwargs.setdefault("weight_rule", WATCHER_TASK_WEIGHT_RULE)
+        existing_on_retry_callback = kwargs.get("on_retry_callback")
+        if existing_on_retry_callback is None:
+            kwargs["on_retry_callback"] = _backup_xcom_to_variable
+        elif isinstance(existing_on_retry_callback, list):
+            kwargs["on_retry_callback"] = [*existing_on_retry_callback, _backup_xcom_to_variable]
+        else:
+            kwargs["on_retry_callback"] = [existing_on_retry_callback, _backup_xcom_to_variable]
```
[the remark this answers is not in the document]
tatiana: Addressed in 5c5a6b75cbfa318a2ce2f3cea3f7b6b733589c59

cosmos/operators/_watcher/xcom.py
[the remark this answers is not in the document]
tatiana: I believe this is not needed, since the previous error occurred during DAG parsing in the Astro Remote Executor, and this one occurs during task execution.

cosmos/operators/_watcher/xcom.py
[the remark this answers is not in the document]
tatiana: Addressed in 98ffb047a810e398b3b4048101d90632ddd38a42

tests/utils.py
```diff
@@ -26,21 +26,22 @@
...
-        if expect_success:
-            return dag_run.state == DagRunState.SUCCESS
-        else:
-            return dag_run.state == DagRunState.FAILED
+        return dag_run.state == expected_dag_state
     return True
 
 
-def new_test_dag(dag: DAG) -> DagRun:
+def new_test_dag(dag: DAG, expected_dag_state: DagRunState = DagRunState.SUCCESS) -> DagRun:
```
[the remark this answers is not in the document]
tatiana: Addressed in 9821737d98871c63f872f9ad5e94f747d7215390

cosmos/operators/_watcher/state.py
```diff
@@ -68,6 +68,13 @@ def safe_xcom_push(task_instance: TaskInstance, key: str, value: Any) -> None:
     """
     with xcom_set_lock:
         task_instance.xcom_push(key=key, value=value)
+        var_key = getattr(task_instance, "_cosmos_xcom_backup_var_key", None)
+        if isinstance(var_key, str):
+            from cosmos.operators._watcher.xcom import _persist_backup
+
+            backup_buffer: dict[str, Any] = task_instance._cosmos_xcom_backup_buffer  # type: ignore[attr-defined]
+            backup_buffer[key] = value
+            _persist_backup(var_key, backup_buffer)
```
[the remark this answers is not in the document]
tatiana: This contention is a known trade-off for correctness. Moving _persist_backup() outside the lock introduces a real data loss race condition:
                                                                                  
Scenario with lock released before persist:                                     
* Thread A: lock => push key_a  =>  buffer={key_a} => snapshot_a={key_a}  =>  unlock 
* Thread B: lock =>  push key_b =>  buffer={key_a, key_b} => snapshot_b={key_a, key_b} => unlock
* Thread A: _persist_backup(snapshot_a)  => Variable = {key_a}
* Thread B: _persist_backup(snapshot_b)  => Variable = {key_a, key_b}              
                                                                                  
If Thread A's Variable.set() lands after Thread B's, the Variable ends up with  {key_a} only - key_b is lost. The last writer wins, and with concurrent threads, the ordering is non-deterministic.
                                                                                  
These are the reasons why I think the current approach is acceptable:                                       
- xcom_push() is already a DB/API call held under the same lock - the lock was specifically added because dbt's multi-threaded execution caused race conditions without it. Adding Variable.set() roughly doubles the hold time but doesn't fundamentally change the contention model.                                      
- If throughput became a real problem, the better solution would be to stop persisting on every push and only flush once at the end. The trade-off is that we would not enable retries if there were an OOM or a SigTerm.
_backup_xcom_to_variable — accepting the loss of crash-safety. But that's a design trade-off, not a lock scope issue.

cosmos/operators/_watcher/xcom.py
```diff
@@ -0,0 +1,126 @@
...
+    backup: dict[str, Any] = json.loads(zlib.decompress(base64.b64decode(compressed.encode("utf-8"))).decode("utf-8"))
+    for key, value in backup.items():
+        safe_xcom_push(task_instance=ti, key=key, value=value)
+    logger.info("Restored %d XCom entries from Variable '%s'", len(backup), var_key)
+
+    try:
+        Variable.delete(var_key)
+        logger.debug("Deleted XCom backup Variable '%s' after restore", var_key)
+    except KeyError:
+        logger.debug("XCom backup Variable '%s' already deleted", var_key)
```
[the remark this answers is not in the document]
tatiana: It is not Airflow generating that exception, it is Astro Remote Executor implementation

tests/operators/test_watcher_kubernetes_unit.py
```diff
@@ -92,13 +92,58 @@ def test_skips_retry_attempt(mock_execute, caplog):
     ti.try_number = 2
     context = {"ti": ti}
 
-    with caplog.at_level(logging.INFO):
-        result = op.execute(context=context)
+    with pytest.raises(AirflowSkipException, match="does not support Airflow retries"):
+        op.execute(context=context)
 
     mock_execute.assert_not_called()
```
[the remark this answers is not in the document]
tatiana: Addressed in 34eedcbccc29947dceb63c41f2ad8d768bdffbd4

cosmos/operators/_watcher/base.py
tatiana: Same as https://github.com/astronomer/astronomer-cosmos/pull/2559#discussion_r3117026015

tatiana:
Thanks @pankajastro, @pankajkoti, and @michal-mrazek for the thorough reviews and the quick turnaround - I really appreciate it.

**On performance (@pankajkoti)**

Good question. Since we're now handling more variables per model update and within the lock, some performance impact is expected. Rather than block this PR on ad-hoc measurements, I'd suggest we validate this through the [benchmarking task](https://github.com/astronomer/astronomer-cosmos/issues/2052) already scoped for this sprint - that'll give us a more rigorous signal than a single DAG run. The tradeoff feels right given the context: Cosmos 1.14.0 shipped with a critical bug that reports failed dbt models as successful, which is currently blocking adoption. We have reports from at least three customers who are waiting for this fix, as well as from one community member.

Assuming the benchmarking shows significant performance degradation, we could switch the XCom persistence from per-model to per-task. The trade-off is that per-task persistence wouldn't recover state from OOM, SIGKILL, SIGTERM, or other OS-level termination events, but it would meaningfully improve performance. This feels like a good candidate for a follow-up PR, informed by benchmark data.

**On test coverage (@pankajastro, @pankajkoti)**

I've validated the following combinations end-to-end:

- `example_watcher_failing_tests.py` - watcher + dbt_runner (default)
- `example_watcher_failing_tests_subprocess.py` - watcher + subprocess
- `example_watcher_failing_multi_folder.py` - watcher + group_nodes_by_folder
- `example_watcher_failing_tests_kubernetes.py` - watcher + Kubernetes (XCOM restore confirmed working)
- `example_watcher_failing_tests_non_deferrable.py` - watcher + dbt_runner + non-deferrable

Happy to add any combinations you feel are still missing - now, or after this PR is merged.

**On the conflict with PR #2557 (@pankajkoti)**

Given the severity of the bug this PR addresses, I'd propose we prioritise getting this one merged first. PR #2557 is a valuable improvement, but it can wait - there are still 3 days in the sprint (30%) for a rebase, and if it carries over to the next sprint, that seems like an acceptable cost. I'd prefer not to expand the scope of this PR further (it's already +1,230/-110), as that would slow down the fix for a critical issue.

**On logging (@michal-mrazek)**

Glad the fix resolves it on your end! Fully agree that "5 out of 5 dbt tests succeeded" style logging would be a nice UX improvement - thanks for picking up those issues. 1.14.2 or 1.15.0 sounds like a good target.

Thanks again, all.

cosmos/operators/_watcher/base.py
tatiana: https://github.com/astronomer/astronomer-cosmos/issues/2585

johnhoran:
Looks good to me.  I do still think we need to address the propagation of the skipped state to tasks downstream of the producer task, but perhaps that is out of scope for this PR.

tatiana:
@johnhoran thanks a lot for this feedback!

Please could you share some steps so we can reproduce this scenario, where we may be missing the propagation of the skipped state to downstream tasks?

Given how critical this change is, as it is blocking users from moving to Cosmos 1.14, the PR had one approval, and we ran several tests and had confirmation from an external contributor that this fixes the issue. I'm going to move forward with merging and creating the PR to release this change.

I'll work with @johnhoran so we can have a test case for the scenario he raised, and can address it as well.

tatiana:
Follow-up ticket that is blocking releasing this PR: https://github.com/astronomer/astronomer-cosmos/issues/2594https://github.com/astronomer/astronomer-cosmos/issues/2594 (solution is in progress)

PR #2563, opened 2026-04-16
Fix incorrectly skipped source downstream tasks in the watcher

pankajastro:
The DFS previously marked every transitive dependent of a stale source as skip, regardless of whether the node had additional clean upstream paths that would allow it to succeed.

For example, a model that depends on both a stale source and a clean upstream model was pre-emptively excluded even though dbt would have run it successfully (especially for warn-status sources where dbt continues execution normally).

A node is now added to the skip set only when all of its depends_on entries are either known-stale sources or already-skipped nodes. Nodes with at least one clean upstream path are left out of the skip set and allowed to run. The DFS still re-evaluates each candidate when a new node joins the skip set, so purely stale chains are handled correctly.

closes: https://github.com/astronomer/astronomer-cosmos/issues/2536

### DIFF
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -89,7 +89,7 @@ def _default_freshness_callback(
     # DFS from each stale source.  A dependent is added to the skip set only when every entry in
     # its depends_on is either a known-stale source or already in the skip set.  This preserves
     # nodes that have at least one clean upstream path: they may succeed and should not be
-    # pre-emptively excluded.  When a new node is added to visited its own dependents are queued
+    # preemptively excluded.  When a new node is added to visited its own dependents are queued
     # so they can be re-evaluated with the updated skip set.
     _excludable_resource_types = {DbtResourceType.MODEL, DbtResourceType.SEED, DbtResourceType.SNAPSHOT}
     visited: set[str] = set()
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -99,10 +99,10 @@ def _default_freshness_callback(
         for dependent_id in dependents.get(current, set()):
             if dependent_id in visited:
                 continue
-            node = nodes.get(dependent_id)
-            if node is None:
+            dependent_node = nodes.get(dependent_id)
+            if dependent_node is None:
                 continue
-            if all(dep in stale_source_ids or dep in visited for dep in node.depends_on):
+            if all(dep in stale_source_ids or dep in visited for dep in dependent_node.depends_on):
                 visited.add(dependent_id)
                 queue.append(dependent_id)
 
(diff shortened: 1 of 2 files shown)


cosmos/operators/watcher.py
```diff
@@ -80,14 +86,23 @@ def _default_freshness_callback(
...
     queue = list(stale_source_ids)
     while queue:
         current = queue.pop()
         for dependent_id in dependents.get(current, set()):
-            if dependent_id not in visited:
+            if dependent_id in visited:
+                continue
+            dependent_node = nodes.get(dependent_id)
+            if dependent_node is None:
+                continue
```
tatiana, line 104: Did we add tests to cover these scenarios as well?

pankajkoti (commented):
Off-topic, but do we know why we use source staleness to decide whether to run a model or not? I guess dbt does not do that, no?

pankajastro:
> Off-topic, but do we know why we use source staleness to decide whether to run a model or not? I guess dbt does not do that, no?

Good question. The main idea is that if upstream sources are stale, any downstream models built on top of them will also be stale, so running them doesn’t add much value. Instead, it just consumes compute and time.

Using source freshness (or staleness) as a signal lets us skip unnecessary runs and save resources.

You’re right that dbt doesn’t enforce this by default—its freshness checks are more for monitoring and alerting. This approach is more about optimizing pipeline efficiency rather than something dbt does out of the box. dbt has similar concept called [state-aware-orchestration](https://docs.getdbt.com/docs/deploy/state-aware-about?version=1.12#optimizing-builds-with-state-aware-orchestration)

cosmos/operators/watcher.py
pankajastro: updated tests to cover these cases

tatiana (approved):
Thanks for fixing this, @pankajastro !

Some questions: in the PR description, why did you mention this relates to https://github.com/astronomer/astronomer-cosmos/issues/2536 and not that it fixes it? What is missing for us to close #2536? I feel like this addresses it? Do we expect a follow up PR?

The title may explain how, but not necessarily the why we're making the change. It would be great if the title reflected the problem we're facing
- Fix incorrectly skipped source downstream tasks in the watcher
Or something along these lines

pankajastro:
> Thanks for fixing this, @pankajastro !
> 
> Some questions: in the PR description, why did you mention this relates to #2536 and not that it fixes it? What is missing for us to close #2536? I feel like this addresses it? Do we expect a follow up PR?
> 
> The title may explain how, but not necessarily the why we're making the change. It would be great if the title reflected the problem we're facing
> 
> * Fix incorrectly skipped source downstream tasks in the watcher
>   Or something along these lines

Wanted to discuss: https://github.com/astronomer/astronomer-cosmos/issues/2536#issuecomment-4257852826 and there is not any follow-up PR planned

PR #2570, opened 2026-04-16
Warn and normalize when source_rendering_behavior=None is passed

### LINKED ISSUE #2568: dbt source freshness still executes despite source_rendering_behavior=none after in Cosmos 1.14.0
After upgrading to Cosmos 1.14.0, users are experiencing performance degradation due to unexpected execution of dbt source freshness checks in producer tasks.

The configuration explicitly sets source_rendering_behavior=None, which should prevent sources (including freshness checks) from being rendered or executed. However, dbt source freshness is still being triggered as part of the producer task, leading to additional runtime overhead.

This behavior appears inconsistent with the intended functionality introduced in PR #2467, where disabling source rendering should skip related operations.

Expected Behavior:
When source_rendering_behavior is set to "none", dbt sources—including freshness checks—should not be rendered or executed.

Actual Behavior:
dbt source freshness checks are still executed within the producer task, causing performance degradation.

Additional Context:

This issue began after upgrading to Cosmos 1.14.0
Related PR: https://github.com/astronomer/astronomer-cosmos/pull/2467
Impact: Increased runtime and unnecessary workload in producer tasks

pankajastro:
Python does not enforce type annotations at runtime, so passing None for source_rendering_behavior bypasses the SourceRenderingBehavior.NONE default, causing sources to be rendered incorrectly and an AttributeError in converter.py when .value is called on None.

Add a guard in RenderConfig.__post_init__ that issues a UserWarning and normalizes None to SourceRenderingBehavior.NONE, consistent with the existing deprecation warning pattern in __post_init__.

DAG
```python
example_watcher = DbtDag(
        execution_mode=ExecutionMode.WATCHER,
    ),
    project_config=ProjectConfig(jaffle_shop_path),
    render_config=RenderConfig(source_rendering_behavior=None), 
    profile_config=default_profile,
    # normal dag parameters
    schedule="@daily",
    catchup=False,
    dag_id="example_watcher",
    default_args={"retries": 0},
    operator_args={
        "dbt_cmd_flags": ["--log-level", "debug"],
    },
)
```

Before change DAG topology 
Screenshot 2026-04-17 at 1 42 49 AM

After Change DAG topology 
Screenshot 2026-04-17 at 1 41 10 AM

closes: https://github.com/astronomer/astronomer-cosmos/issues/2568

### DIFF
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -118,6 +118,14 @@ class RenderConfig:
                 "RenderConfig.dbt_deps is deprecated since Cosmos 1.9 and will be removed in Cosmos 2.0. Use ProjectConfig.install_dbt_deps instead.",
                 DeprecationWarning,
             )
+        if self.source_rendering_behavior is None:
+            warnings.warn(
+                "Passing None for source_rendering_behavior is not supported. "
+                "Use SourceRenderingBehavior.NONE to disable source rendering. "
+                "Defaulting to SourceRenderingBehavior.NONE.",
+                UserWarning,
+            )
+            self.source_rendering_behavior = SourceRenderingBehavior.NONE
         self.project_path = Path(dbt_project_path) if dbt_project_path else None
         # allows us to initiate this attribute from Path objects and str
         self.dbt_ls_path = Path(self.dbt_ls_path) if self.dbt_ls_path else None
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -5,7 +5,7 @@ from unittest.mock import Mock, PropertyMock, call, patch
 import pytest
 
 from cosmos.config import CosmosConfigException, ExecutionConfig, ProfileConfig, ProjectConfig, RenderConfig
-from cosmos.constants import ExecutionMode, InvocationMode
+from cosmos.constants import ExecutionMode, InvocationMode, SourceRenderingBehavior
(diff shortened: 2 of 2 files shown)


cosmos/config.py
```diff
@@ -118,6 +118,14 @@ def __post_init__(self, dbt_project_path: str | Path | None) -> None:
                 "RenderConfig.dbt_deps is deprecated since Cosmos 1.9 and will be removed in Cosmos 2.0. Use ProjectConfig.install_dbt_deps instead.",
                 DeprecationWarning,
             )
+        if self.source_rendering_behavior is None:
```
tatiana, line 121: @pankajastro, what do you see as the tradeoffs of just raising an exception and fixing this on behalf of the user?

pankajastro: The issue is that Python does not enforce strict typing, which is why users are able to pass `source_rendering_behavior=None` without causing a crash.

If we raise an exception:

- It will break existing DAGs for users who are currently using source_rendering_behavior=None, which may not be a desirable experience.

If we raise a warning instead:

- It alerts users to use the parameter correctly.
- It gives them time to fix their usage.
- Later, we can enforce stricter behavior by raising an exception.

tatiana (approved):

PR #2572, opened 2026-04-17
Refactor `_default_freshness_callback` to return list of (id, state) tuple

pankajastro:
- Changed _default_freshness_callback return type from tuple[list[str], str] to list[tuple[str, str]] so each node carries its own state instead of      
  sharing a single status string
  - Renamed _skipped_node_token → _apply_node_state_tokens and _push_skipped_xcom_for_model → _push_node_state_xcom; both now accept an explicit state     
  argument, enabling custom callbacks to push states other than "skipped"                                                                                  
  - Nodes are added to self.exclude only when their state is "skipped"; other states push XCom but do not affect the dbt run
  - Updated all call sites and tests; added a new test for the non-"skipped" state path     

This set would set ground work for https://github.com/astronomer/astronomer-cosmos/issues/2053

Related-to: https://github.com/astronomer/astronomer-cosmos/issues/2053


tatiana (approved):
Thanks, @pankajastro , it looks good, let's ship this as part of Cosmos 1.15.0

PR #2576, opened 2026-04-19
`fix-issue-2393`: Support rendering of `DbtResourceType.EXPOSURE`

### LINKED ISSUE #2393: [Bug] `DbtResourceType.EXPOSURE` is not rendered since `v1.12.0`
### Astronomer Cosmos Version

1.13.0, 1.12.0, 1.12.1

### dbt-core version

1.11.6, 1.10.13

### Versions of dbt adapters

```
dbt-adapters==1.22.6
dbt-common==1.37.2
dbt-core==1.11.6
dbt-extractor==0.6.0
dbt-protos==1.0.431
dbt-semantic-interfaces==0.9.0
dbt-snowflake==1.11.2
```

### LoadMode

DBT_LS

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

2.10.5

### Operating System

_No response_

### Deployment

Docker-Compose

### Deployment details

_No response_

### What happened?

# Bug: `node_converters` not applied for `DbtResourceType.EXPOSURE` in v1.12.x and v1.13.0

## Description

`DbtResourceType.EXPOSURE` nodes defined in `render_config.node_converters` are silently ignored during DAG rendering, producing the following warning instead:

```
{graph.py:426} WARNING - Unavailable conversion function for <DbtResourceType.EXPOSURE>
(node <exposure.{project_name}.{node_name}>). Define a converter function using render_config.node_converters.
```

## Root Cause

The issue lies in `create_task_metadata()` in `cosmos/airflow/graph.py` ([[L421–L427](https://github.com/astronomer/astronomer-cosmos/blob/astronomer-cosmos-v1.13.0/cosmos/airflow/graph.py#L421C10-L427C20)](https://github.com/astronomer/astronomer-cosmos/blob/astronomer-cosmos-v1.13.0/cosmos/airflow/graph.py#L421C10-L427C20)):

```python
if DbtResourceType(node.resource_type) in DEFAULT_DBT_RESOURCES and node.resource_type in dbt_resource_to_class:
    ...
else:
    msg = (
        f"Unavailable conversion function for <{node.resource_type}> (node <{node.unique_id}>). "
        "Define a converter function using render_config.node_converters."
    )
    logger.warning(msg)
    return None
```

`DbtResourceType.EXPOSURE` satisfies the first condition (`in DEFAULT_DBT_RESOURCES`), but fails the second condition because `EXPOSURE` has no corresponding entry in `dbt_resource_to_class`. As a result, the function falls into the `else` branch and returns `None` — **before** `node_converters.get()` is ever called.

This regression was introduced in **v1.12.0**, when `create_task_metadata()` began evaluating the `dbt_resource_to_class` check prior to consulting `node_converters`.

http://github.com/astronomer/astronomer-cosmos/blob/astronomer-cosmos-v1.13.0/cosmos/airflow/graph.py#L524-L565

## Expected Behavior

When a user defines a converter for `DbtResourceType.EXPOSURE` in `render_config.node_converters`, it should be applied during rendering — consistent with the behavior of other node types.

## Proposed Fix

The simplest fix is to add a default class entry for `DbtResourceType.EXPOSURE` in `dbt_resource_to_class`, so that exposure nodes pass the gate condition and reach the `node_converters` lookup.

Alternatively, the condition could be restructured to check `node_converters` first, before falling back to `dbt_resource_to_class`.

## Steps to Reproduce

1. Define a `node_converters` entry for `DbtResourceType.EXPOSURE` in `RenderConfig`
2. Run DAG rendering with a dbt project that includes exposure nodes
3. Observe that the converter is never invoked, and the warning above is logged

### Relevant log output

```shell
{graph.py:426} WARNING - Unavailable conversion function for <DbtResourceType.EXPOSURE>
(node <exposure.{project_name}.{node_name}>). Define a converter function using render_config.node_converters.
```

### How to reproduce

Create DBT DAG with exposure and define `node_converters` on `RenderConfig`.

### Anything else :)?

_No response_

### Contact Details

_No response_

jroachgolf84:
## Description

Per issue #2393, `DbtResourceType.EXPOSURE` nodes defined in `render_config.node_converters` are silently ignored during DAG rendering, producing the following warning instead:

```
{graph.py:426} WARNING - Unavailable conversion function for <DbtResourceType.EXPOSURE>
(node <exposure.{project_name}.{node_name}>). Define a converter function using render_config.node_converters.
```

This PR ensures that exposures are properly rendered in the Airflow graph (as an `EmptyOperator`) with the appropriate metadata.

## Related Issue(s)

closes: #2393

## Breaking Change?

## Testing

These changes were tested via a unit-test (`test_create_task_metadata_exposure`) as well as E2E by running Airflow with Docker.

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -297,6 +297,7 @@ def create_dbt_resource_to_class(test_behavior: TestBehavior) -> dict[str, str]:
             DbtResourceType.SEED: "DbtBuild",
             DbtResourceType.TEST: "DbtTest",
             DbtResourceType.SOURCE: "DbtSource",
+            DbtResourceType.EXPOSURE: "DbtExposure"
         }
     else:
         dbt_resource_to_class = {
@@ -305,6 +306,7 @@ def create_dbt_resource_to_class(test_behavior: TestBehavior) -> dict[str, str]:
             DbtResourceType.SEED: "DbtSeed",
             DbtResourceType.TEST: "DbtTest",
             DbtResourceType.SOURCE: "DbtSource",
+            DbtResourceType.EXPOSURE: "DbtExposure"
         }
     return dbt_resource_to_class
 
@@ -413,6 +415,24 @@ def create_task_metadata(  # noqa: C901
                 else:
                     args = {}
                 return TaskMetadata(id=task_id, operator_class="airflow.operators.empty.EmptyOperator", arguments=args)
+
+        elif node.resource_type == DbtResourceType.EXPOSURE:
+            # Build the task_id and update args
+            task_id, args = _get_task_id_and_args(
+                node=node,
+                args=args,
+                use_task_group=use_task_group,
+                normalize_task_id=render_config.normalize_task_id,
+                normalize_task_display_name=render_config.normalize_task_display_name,
+                resource_suffix=r"exposure",
+                execution_mode=execution_mode,
+            )
+
+            # Strip down args to pass to EmptyOperator
+            args = {"task_display_name": args["task_display_name"]} if "task_display_name" in args else {}
+
+            return TaskMetadata(id=task_id, operator_class="airflow.operators.empty.EmptyOperator", arguments=args)
+
         else:  # DbtResourceType.MODEL, DbtResourceType.SEED and DbtResourceType.SNAPSHOT
             if node.fqn and len(node.fqn) > 0:
                 args[models_select_key] = f"fqn:{'.'.join(node.fqn)}"
--- a/tests/airflow/test_graph.py
+++ b/tests/airflow/test_graph.py
@@ -706,6 +706,31 @@ def test_create_task_metadata_snapshot(caplog):
     assert metadata.arguments == {"select": "my_snapshot"}
 
 
+def test_create_task_metadata_exposure(caplog):
+    sample_node = DbtNode(
+        unique_id=f"{DbtResourceType.EXPOSURE.value}.my_folder.my_exposure",
+        resource_type=DbtResourceType.EXPOSURE,
+        depends_on=[],
+        path_base=Path("."),
+        original_file_path=Path("."),
+        tags=[],
+        config={},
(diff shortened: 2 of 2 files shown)


tatiana (commented):
Hey @jroachgolf84 , thanks so much for working on this — really excited to see your contribution!

Since Cosmos v1.12.0, Cosmos has stopped rendering exposures when users try to run this DAG. Cosmos used to handle this dynamically, without requiring explicit logic for exposures via this feature:
https://astronomer.github.io/astronomer-cosmos/guides/translate_dbt_to_airflow/dag-customization.html

This is an example of a DAG that used to work in previous versions:
https://github.com/astronomer/astronomer-cosmos/blob/main/dev/dags/example_cosmos_sources.py

In my opinion, issue #2393 uncovers more than just the exposure node problem. My impression is that Cosmos has stopped supporting the previous level of pluggability it offered. This could potentially apply to `analysis` and `metrics` nodes, which may be available in the `manifest.json` or `dbt ls` and users would like to handle somehow. 

Ideally, the fix would not be specific to exposures but would instead restore users' ability to resolve dynamically resolved dbt node types, as before.

I believe the regression may have been introduced by one of these changes:
#2080
#1759

Would you be open to proposing an alternative change that addresses the root cause? It could live in a separate PR. With this we could discuss the approaches and decide on the solution.

jroachgolf84:
@tatiana - thanks for the feedback! I'd be open to either re-working this PR to include the changes you mentioned. OR, I'd also be open to merging this PR with this logic and opening up a new one to address the other points you mentioned. What would you like to do?

tatiana:
@jroachgolf84 my preference would be with another PR, so we can compare the approaches and make a decision on how to proceed

jroachgolf84:
@tatiana, sounds good - I'll open up another PR to address this broader scope.

jroachgolf84:
Closing this PR in favor of #2608.

PR #2577, opened 2026-04-20
YAML selector parsing should skip malformed selectors instead of failing entirely

### LINKED ISSUE #2565: YAML selector parsing should skip malformed selectors instead of failing entirely
## Description

When using `LoadMode.DBT_MANIFEST` with `RenderConfig(selector=...)`, Cosmos reads the `selectors` section from `manifest.json` and parses **all** selector entries via `YamlSelectors.parse()`. If **any** selector entry in the manifest is malformed (missing `name` or `definition` keys), parsing fails with a `CosmosValueError`, breaking the entire DAG — even if the user's requested selector is valid.

## How to reproduce

A customer reported this error:

```
cosmos.exceptions.CosmosValueError: Selector definition for 'Company_System' must contain 'name' and 'definition' keys.
```

The customer's `selectors.yml` is valid and works with dbt directly. The likely root cause is that the `manifest.json` was not regenerated after updating `selectors.yml`, so the manifest contains a stale/incomplete entry for `Company_System`.

Since dbt reads `selectors.yml` directly at runtime, `dbt run --selector Company_System` works. But Cosmos reads the pre-compiled manifest, which likely has the old/malformed entry.

## Current behavior

`YamlSelectors.parse()` raises `CosmosValueError` on the first malformed selector, preventing any selector from being used:

https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/dbt/selector.py#L1393-L1394

## Expected behavior

Two possible improvements to discuss:

1. **Skip malformed selectors with a warning** instead of raising an error, so that valid selectors in the same manifest still work.
2. **Only validate/parse the specific selector the user requested** (via `RenderConfig.selector`), rather than all selectors in the manifest.

Either approach would make Cosmos more resilient to stale manifests or manifests with partially valid selectors.

## Additional context

- Related PR: #2261 (original YAML selector support)
- cc @YourRoyalLinus I'd love to hear your thoughts on this - since you put lots of thoughts on the feature and the proposed behaviour may make things worse. I am also curious if it would make sense for us to fetch the selectors definition from the `selectors.yaml` instead from the `manifest.json`

YourRoyalLinus:
## Description

Skip malformed selectors with a warning instead of raising an error, so that valid selectors in the same manifest still work.

I reproduced the bug outlined in the related issue and verified these changes will resolve the problem.

## Related Issue(s)

Closes #2565

## Breaking Change?
N/A

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,6 +246,7 @@ jobs:
           PYTEST_SPLIT_GROUP: ${{ matrix.split-group }}
           AIRFLOW__COSMOS__ENABLE_CACHE_DBT_LS: 0
           AIRFLOW__COSMOS__ENABLE_CACHE_DBT_YAML_SELECTORS: 0
+          AIRFLOW__COSMOS__ENABLE_LAX_SELECTOR_PARSING: 0
           AIRFLOW_HOME: /home/runner/work/astronomer-cosmos/astronomer-cosmos/
           AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT: 120
           AIRFLOW__CORE__DAG_FILE_PROCESSOR_TIMEOUT: 120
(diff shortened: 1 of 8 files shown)


tatiana (commented):
@YourRoyalLinus, thanks a lot for further improving this feature.
I agree with the approach - it feels the safest, since the warnings could be hidden from users, while the error would popup to all users.
Please could you address the codecov feedback? I'm happy for us to approve this change!

tatiana (approved):

PR #2584, opened 2026-04-21
Add producer-internal dbt retry for ExecutionMode.WATCHER

### LINKED ISSUE #1978: Improve retries in ExecutionMode.WATCHER

pankajkoti:
When `dbt_retry_count > 0` is set on the producer via `setup_operator_args`, the `DbtProducerWatcherOperator` now runs `dbt retry` within the same temporary directory after a `dbt build` that has failures. This retries failed nodes using the existing `run_results.json` without requiring cross-attempt artefact persistence.

Key changes:
- Add `dbt_retry_count` parameter to `DbtProducerWatcherOperator`
- Add `_post_dbt_invoke` hook in `AbstractDbtLocalBase` so the producer can inject retry logic without duplicating `run_command`
- Defer XCom push for failed and skipped nodes when retries are configured, so consumers don't see intermediate states
- Add `flaky_model` test fixtures and example DAG for CI integration testing

closes: #1978 
closes: https://github.com/astronomer/oss-integrations-private/issues/374

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -25,6 +25,7 @@ from cosmos.operators._watcher.state import (
     _log_dbt_event,
     build_producer_state_fetcher,
     get_xcom_val,
+    is_dbt_node_status_failed,
     is_dbt_node_status_skipped,
     is_dbt_node_status_success,
     is_dbt_node_status_terminal,
@@ -237,6 +238,47 @@ def _ensure_subprocess_model_outlet_uris(
         model_outlet_uris[_MODEL_OUTLET_URIS_ATTEMPTED_KEY] = []  # type: ignore[assignment]
 
 
+def _push_or_defer_status(
+    task_instance: Any,
+    unique_id: str,
+    dbt_node_status: str,
+    status_value: dict[str, Any],
+    pending_failures: dict[str, dict[str, Any]] | None,
+) -> None:
+    """Push a node's status to XCom, or defer it into pending_failures for retry.
+
+    When ``pending_failures`` is provided:
+    - Failed/skipped nodes are stored in ``pending_failures`` instead of XCom,
+      because ``dbt retry`` re-runs both failed nodes and their skipped dependents.
+    - Nodes that were previously pending but now succeeded are removed from
+      ``pending_failures`` and pushed to XCom immediately.
+    """
+    should_defer = pending_failures is not None and (
+        is_dbt_node_status_failed(dbt_node_status) or is_dbt_node_status_skipped(dbt_node_status)
+    )
+    if should_defer:
+        assert pending_failures is not None  # narrowing for mypy
+        pending_failures[unique_id] = status_value
+        logger.info(
+            "Deferring XCom push for node '%s' (status='%s') — will retry",
+            unique_id,
+            dbt_node_status,
+        )
+    else:
+        if pending_failures is not None and unique_id in pending_failures:
+            del pending_failures[unique_id]
+            logger.info(
+                "Node '%s' succeeded on retry (status='%s') — pushing XCom now",
+                unique_id,
+                dbt_node_status,
+            )
+        safe_xcom_push(
+            task_instance=task_instance,
+            key=f"{unique_id.replace('.', '__')}_status",
+            value=status_value,
+        )
+
+
 def store_dbt_resource_status_from_log(
     line: str,
     extra_kwargs: Any,
@@ -245,6 +287,7 @@ def store_dbt_resource_status_from_log(
     test_results_per_model: dict[str, list[str]] | None = None,
     model_outlet_uris: dict[str, list[str]] | None = None,
     dataset_namespace: str | None = None,
+    pending_failures: dict[str, dict[str, Any]] | None = None,
 ) -> None:
     """
     Parses a single line from dbt JSON logs and stores node status to Airflow XCom.
@@ -265,6 +308,11 @@ def store_dbt_resource_status_from_log(
     :param model_outlet_uris: Mutable dict mapping unique_id to outlet URIs.
         Populated lazily from the manifest on first terminal status detection.
     :param dataset_namespace: The OL-compatible dataset namespace for URI construction.
+    :param pending_failures: When not None, failed model/seed/snapshot statuses are
+        stored in this dict instead of being pushed to XCom immediately. This allows
+        the producer to retry failed nodes via ``dbt retry`` before consumers see a
+        failure. Keyed by unique_id, values are the status dicts. Test results are
+        never deferred (they use a separate aggregation mechanism).
     """
     try:
         log_line = json.loads(line)
@@ -312,15 +360,11 @@ def store_dbt_resource_status_from_log(
                     _ensure_subprocess_model_outlet_uris(model_outlet_uris, dataset_namespace, project_dir)
                     outlet_uris = model_outlet_uris.get(unique_id, []) if model_outlet_uris else []
 
-                status_value: dict[str, Any] | str = {
+                status_value: dict[str, Any] = {
                     "status": dbt_node_status,
                     "outlet_uris": outlet_uris,
                 }
-                safe_xcom_push(
-                    task_instance=context["ti"],
-                    key=f"{unique_id.replace('.', '__')}_status",
-                    value=status_value,
-                )
+                _push_or_defer_status(context["ti"], unique_id, dbt_node_status, status_value, pending_failures)
 
             # Extract and push compiled_sql for models (centralised for both subprocess and node-event)
             # compiled_sql is available for both success and failed models - it's compiled before execution
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -43,6 +43,10 @@ except ImportError:  # pragma: no cover
 
 logger = get_logger(__name__)
 
+# Protects mutations of the ``pending_failures`` dict passed to
+# ``_push_or_defer_status`` from concurrent dbt runner callback threads.
+_pending_failures_lock = threading.Lock()
+
 # Subset of dbt event types that represent errors/failures.
 # Used (together with node status lifecycle events like NodeStart/NodeCompiling/
 # NodeExecuting/NodeFinished) to build _DBT_EVENT_ALLOWLIST, which controls which
@@ -252,26 +256,33 @@ def _push_or_defer_status(
       because ``dbt retry`` re-runs both failed nodes and their skipped dependents.
     - Nodes that were previously pending but now succeeded are removed from
       ``pending_failures`` and pushed to XCom immediately.
+
+    All mutations of ``pending_failures`` are performed under
+    ``_pending_failures_lock`` because dbt runner callbacks are invoked from
+    multiple threads.
     """
     should_defer = pending_failures is not None and (
         is_dbt_node_status_failed(dbt_node_status) or is_dbt_node_status_skipped(dbt_node_status)
     )
     if should_defer:
         assert pending_failures is not None  # narrowing for mypy
-        pending_failures[unique_id] = status_value
+        with _pending_failures_lock:
+            pending_failures[unique_id] = status_value
         logger.info(
             "Deferring XCom push for node '%s' (status='%s') — will retry",
             unique_id,
             dbt_node_status,
         )
     else:
-        if pending_failures is not None and unique_id in pending_failures:
-            del pending_failures[unique_id]
-            logger.info(
-                "Node '%s' succeeded on retry (status='%s') — pushing XCom now",
-                unique_id,
-                dbt_node_status,
-            )
+        if pending_failures is not None:
+            with _pending_failures_lock:
+                removed = pending_failures.pop(unique_id, None)
+            if removed is not None:
+                logger.info(
+                    "Node '%s' succeeded on retry (status='%s') — pushing XCom now",
+                    unique_id,
+                    dbt_node_status,
+                )
         safe_xcom_push(
             task_instance=task_instance,
             key=f"{unique_id.replace('.', '__')}_status",
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -636,6 +636,24 @@ class AbstractDbtLocalBase(AbstractDbtBase):
             async_op_configurator(self, sql=sql)
             async_context["async_operator"].execute(self, context)
 
+    def _post_dbt_invoke(
+        self,
+        result: Any,
+        context: Context,
+        env: dict[str, str | bytes | os.PathLike[Any]],
+        tmp_project_dir: str,
+        profile_path: Path,
+    ) -> Any:
+        """Hook called after the main dbt command finishes, before post-processing and exception handling.
+
+        The temporary project directory is still available at this point. Subclasses
+        can override this to perform additional work (e.g. run ``dbt retry`` for
+        failed nodes in the watcher producer). The returned value replaces ``result``
+        for downstream exception handling — return a successful result to suppress
+        the original failure.
+        """
+        return result
+
     def run_command(  # noqa: C901
         self,
         cmd: list[str],
@@ -689,6 +707,9 @@ class AbstractDbtLocalBase(AbstractDbtBase):
                     self._sources_json = _read_target_sources_json(tmp_dir_path)
                     self.handle_exception(result)
                     return result
+
+                result = self._post_dbt_invoke(result, context, env, tmp_project_dir, profile_path)
+
                 if is_openlineage_common_available:
                     self.calculate_openlineage_events_completes(env, tmp_dir_path, full_cmd)
                     if AIRFLOW_VERSION.major < _AIRFLOW3_MAJOR_VERSION:
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 import contextlib
 import functools
+import os
 from collections.abc import Callable
+from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 from airflow.exceptions import AirflowException, AirflowSkipException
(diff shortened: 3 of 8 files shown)


pankajkoti:
`dbt retry` in action. The model is written in such a way that it fails for the first time and succeeds upon retry.


Screenshot 2026-04-21 at 5 59 35 PM

Screenshot 2026-04-21 at 7 37 57 PM

Screenshot 2026-04-21 at 7 40 30 PM

cosmos/operators/_watcher/base.py
```diff
@@ -237,6 +238,47 @@ def _ensure_subprocess_model_outlet_uris(
...
+    - Failed/skipped nodes are stored in ``pending_failures`` instead of XCom,
+      because ``dbt retry`` re-runs both failed nodes and their skipped dependents.
+    - Nodes that were previously pending but now succeeded are removed from
+      ``pending_failures`` and pushed to XCom immediately.
+    """
+    should_defer = pending_failures is not None and (
+        is_dbt_node_status_failed(dbt_node_status) or is_dbt_node_status_skipped(dbt_node_status)
+    )
+    if should_defer:
+        assert pending_failures is not None  # narrowing for mypy
```
pankajastro: TODO

cosmos/operators/local.py
```diff
@@ -636,6 +636,24 @@ def _handle_async_execution(self, tmp_project_dir: str, context: Context, async_
...
+    ) -> Any:
+        """Hook called after the main dbt command finishes, before post-processing and exception handling.
+
+        The temporary project directory is still available at this point. Subclasses
+        can override this to perform additional work (e.g. run ``dbt retry`` for
+        failed nodes in the watcher producer). The returned value replaces ``result``
+        for downstream exception handling — return a successful result to suppress
+        the original failure.
+        """
+        return result
```
pankajastro, line 655: Do we need this method here? I see that you are returning the result without processing it.

cosmos/operators/watcher.py
```diff
@@ -377,10 +386,121 @@ def _apply_source_freshness(self, context: Context) -> None:
...
+        After each retry iteration the log-parsing callback updates
+        ``self._pending_failures``: nodes that now succeed are removed and their
+        XCom status is pushed immediately; nodes that still fail remain pending.
+
+        Returns the result of the last ``dbt retry`` invocation (or ``None`` if
+        no retry was executed), so the caller can decide whether to propagate the
+        original build failure.
+        """
+        last_result = None
+        for attempt in range(1, self.dbt_retry_count + 1):
```
pankajastro, line 450: Does dbt retry have a flag to specify how many times the user wants to retry? Running the retry command multiple times feels like a bit of extra work

cosmos/operators/watcher.py
```diff
@@ -193,6 +196,11 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self._dataset_namespace: str | None = None
         self._model_outlet_uris: dict[str, list[str]] = {}
 
+        # Holds failed model statuses during the retry loop so that consumers
+        # don't see intermediate "failed" XCom values that will be retried.
+        # Maps unique_id -> {"status": ..., "outlet_uris": ...}
+        self._pending_failures: dict[str, dict[str, Any]] = {}
```
pankajastro, line 202: where are you building this dict?

cosmos/operators/local.py
pankajkoti: Yes, this is hook that has a no-op in the local mode, but watcher overrides this method to trigger dbt retry command.

Otherwise, if we don't have this hook, we will have duplicate most of the execute method in the watcher.

cosmos/operators/watcher.py
pankajkoti: `dbt retry` is a singular command; it does not retry by itself upon failures. Every time a user wants to retry, they will run the `dbt retry` command. 
We give this as a configurable parameter to users. The default is `0`, meaning we won't execute a dbt retry by default; however, users can set it to values they would like to `1`, `2`, etc., similar to how they can set the retries value for Airflow DAGs

cosmos/operators/watcher.py
pankajkoti: It gets populated in the method `_push_or_defer_status` in the file `cosmos/operators/_watcher/base.py` which gets called when processing the logs from dbt command output.

https://github.com/astronomer/astronomer-cosmos/pull/2584/changes#diff-dc74b407f8befb3f2bdbd9bac3669ea8c2b5701d6f8e83a929bbe9d4b3bb5da3R261

cosmos/operators/watcher.py
pankajastro: Airflow’s retry mechanism has a native implementation for waiting between retries, along with several additional features. I’m a bit concerned that here we’re just running the command sequentially, which doesn’t feel very natural.

pankajastro (approved):

cosmos/operators/watcher.py
pankajkoti: Yes, we can add a configuration for delay in the next iteration. I believe the current implementation could still help get rid of the failures where we see the model fails on first run, but succeeds on second run (transient failures or a dependency on another model such that the other table needs to exist before this one but misconfigured dependencies)

cosmos/operators/watcher.py
```diff
@@ -377,10 +386,121 @@ def _apply_source_freshness(self, context: Context) -> None:
...
+            )
+
+            if not self._has_run_results(tmp_project_dir):
+                logger.warning(
+                    "run_results.json not found in %s/target/ — cannot run dbt retry. "
+                    "Flushing %d remaining failure(s) to XCom.",
+                    tmp_project_dir,
+                    len(self._pending_failures),
+                )
+                break
```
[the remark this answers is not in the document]
pankajkoti: addressed

tests/operators/test_watcher.py
```diff
@@ -2428,3 +2428,329 @@ def test_run_dbt_runner_skips_callback_during_source_freshness(self):
...
+    def test_build_retry_command(self):
+        op = self._make_producer(profile_config=profile_config)
+        op.dbt_executable_path = "dbt"
+        op.dbt_cmd_global_flags = []
+        op.partial_parse = False
+
+        cmd = op._build_retry_command("/tmp/project", Path("/tmp/profiles/profiles.yml"))
+        assert "retry" in cmd
+        assert "--no-partial-parse" in cmd
+        assert "--project-dir" in cmd
```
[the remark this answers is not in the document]
pankajkoti: addressed

cosmos/operators/_watcher/base.py
```diff
@@ -265,6 +308,11 @@ def store_dbt_resource_status_from_log(
     :param model_outlet_uris: Mutable dict mapping unique_id to outlet URIs.
         Populated lazily from the manifest on first terminal status detection.
     :param dataset_namespace: The OL-compatible dataset namespace for URI construction.
+    :param pending_failures: When not None, failed model/seed/snapshot statuses are
+        stored in this dict instead of being pushed to XCom immediately. This allows
+        the producer to retry failed nodes via ``dbt retry`` before consumers see a
+        failure. Keyed by unique_id, values are the status dicts. Test results are
+        never deferred (they use a separate aggregation mechanism).
```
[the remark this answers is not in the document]
pankajkoti: addressed. resetting the test_results_per_model

cosmos/operators/watcher.py
```diff
@@ -377,10 +386,136 @@ def _apply_source_freshness(self, context: Context) -> None:
...
+        ``--full-refresh``, ``--select``, ``--exclude`` are omitted because
+        ``dbt retry`` does not accept them — it derives the node set from
+        ``run_results.json``.
+        """
+        retry_cmd = [self.dbt_executable_path]
+        retry_cmd.extend(self.dbt_cmd_global_flags)
+        if not self.partial_parse:
+            retry_cmd.append("--no-partial-parse")
+        retry_cmd.append("retry")
+        retry_cmd.extend(self._generate_dbt_flags(tmp_project_dir, profile_path))
```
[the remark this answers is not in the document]
pankajkoti: addressed in a newer commit

cosmos/operators/watcher.py
```diff
@@ -377,10 +386,136 @@ def _apply_source_freshness(self, context: Context) -> None:
...
+        """Run ``dbt retry`` for failed nodes, then flush any remaining failures to XCom.
+
+        If all retries succeed, returns the last retry result (which is successful)
+        so that ``handle_exception`` does not raise. Otherwise returns the original
+        failed build result.
+        """
+        if self.dbt_retry_count > 0 and self._pending_failures:
+            last_retry_result = self._run_dbt_retry_loop(context, env, tmp_project_dir, profile_path)
+            all_retries_succeeded = not self._pending_failures
+            self._flush_pending_failures(context)
```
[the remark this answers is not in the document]
pankajkoti: I have made changes to trigger retries by also checking the result status; however, test retries are currently broken without this change. I have logged a follow-up ticket to fix test retries https://github.com/astronomer/astronomer-cosmos/issues/2598

cosmos/operators/_watcher/base.py
```diff
@@ -237,6 +238,47 @@ def _ensure_subprocess_model_outlet_uris(
...
+      ``pending_failures`` and pushed to XCom immediately.
+    """
+    should_defer = pending_failures is not None and (
+        is_dbt_node_status_failed(dbt_node_status) or is_dbt_node_status_skipped(dbt_node_status)
+    )
+    if should_defer:
+        assert pending_failures is not None  # narrowing for mypy
+        pending_failures[unique_id] = status_value
+        logger.info(
+            "Deferring XCom push for node '%s' (status='%s') — will retry",
```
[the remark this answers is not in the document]
pankajkoti: good callout, using a lock now.

cosmos/operators/watcher.py
```diff
@@ -377,10 +386,121 @@ def _apply_source_freshness(self, context: Context) -> None:
         freshness_results = self._freshness_callback(context, dag, task_group, nodes, self._sources_json)
         self._apply_node_state_tokens(context, freshness_results)
 
+    def _build_retry_command(self, tmp_project_dir: str, profile_path: Path) -> list[str]:
+        """Build a ``dbt retry`` command using the same project/profile flags as the original build."""
+        retry_cmd = [self.dbt_executable_path]
+        retry_cmd.extend(self.dbt_cmd_global_flags)
+        if not self.partial_parse:
+            retry_cmd.append("--no-partial-parse")
+        retry_cmd.append("retry")
```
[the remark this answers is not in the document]
pankajkoti: This has been addressed

cosmos/operators/watcher.py
```diff
@@ -377,10 +386,175 @@ def _apply_source_freshness(self, context: Context) -> None:
...
+
+        Returns the result of the last ``dbt retry`` invocation (or ``None`` if
+        no retry was executed), so the caller can decide whether to propagate the
+        original build failure.
+        """
+        last_result = None
+        for attempt in range(1, self.dbt_retry_count + 1):
+            if not self._pending_failures:
+                break
+
```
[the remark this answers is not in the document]
pankajkoti: we have a ticket to handle test failures: https://github.com/astronomer/astronomer-cosmos/issues/2598

dev/dags/dbt/jaffle_shop/models/flaky_model.sql
```diff
@@ -0,0 +1,30 @@
+{{
```
tatiana, line 1: The jaffle shop project should remain as close as possible to our original setup, without any additional changes (new models or scenarios).

In PR #2559 we introduced two dbt projects that simulate testing scenarios:
- `watcher_failing_tests`
- `multi_folder_failing`

WDYT if we reused one of those?

dev/dags/example_watcher_retry.py
```diff
@@ -0,0 +1,61 @@
+"""
```
tatiana, line 1: Same here, it would be great if we could reuse/adapt/extend one of the many DAGs that were introduced in #2559:

- example_watcher_failing_multi_folder.py
- example_watcher_failing_tests.py 
- example_watcher_failing_tests_kubernetes.py 
- example_watcher_failing_tests_non_deferrable.py 
- example_watcher_failing_tests_subprocess.py

We could rename, move, or adapt one/multiple of them as needed

tests/operators/test_watcher.py
```diff
@@ -2428,3 +2428,431 @@ def test_run_dbt_runner_skips_callback_during_source_freshness(self):
...
+    def test_execute_clears_pending_failures(self, mock_execute, mock_persist, mock_variable):
+        """execute() should clear _pending_failures at the start."""
+        op = self._make_producer(dbt_retry_count=1)
+        op._pending_failures = {"model.pkg.stale": {"status": "error", "outlet_uris": []}}
+        ti = _MockTI()
+        context = {"ti": ti, "run_id": "test_run"}
+
+        op.execute(context=context)
+
+        assert op._pending_failures == {}
```
tatiana, line 2858: Overall, I'm worried about the large amount of test code being introduced (400 lines) and that it's mostly unittests that mock interfaces and don't actually run the DAG or evaluate behaviour.

My vote would be to have one meaningful integration test, similar to https://github.com/astronomer/astronomer-cosmos/blob/34eedcbccc29947dceb63c41f2ad8d768bdffbd4/tests/operators/test_watcher.py#L1915, and to have unittests covering edge cases.

cosmos/operators/local.py
```diff
@@ -636,6 +636,24 @@ def _handle_async_execution(self, tmp_project_dir: str, context: Context, async_
             async_op_configurator(self, sql=sql)
             async_context["async_operator"].execute(self, context)
 
+    def _post_dbt_invoke(
```
tatiana, line 639: Why aren't we using Cosmos existing `callback` mechanisms?

The fact that this change is being done in `local.py` means it wouldn't work for `KUBERNETES_WATCHER` - have you considered this use-case?

cosmos/operators/watcher.py
```diff
@@ -377,10 +386,175 @@ def _apply_source_freshness(self, context: Context) -> None:
         freshness_results = self._freshness_callback(context, dag, task_group, nodes, self._sources_json)
         self._apply_node_state_tokens(context, freshness_results)
 
+    # Flags that ``dbt retry`` does not accept — it derives the node set from
+    # ``run_results.json`` and does not support selection or refresh overrides.
+    _RETRY_UNSUPPORTED_FLAGS = frozenset({"--select", "--exclude", "--selector", "--models", "--full-refresh"})
```
tatiana, line 391: Out of curiousity, which flags are actually supported and we would like to reuse? Is it just the project and profile ones?

cosmos/operators/watcher.py
```diff
@@ -377,10 +386,175 @@ def _apply_source_freshness(self, context: Context) -> None:
...
+
+            retry_cmd = self._build_retry_command(tmp_project_dir, profile_path)
+            try:
+                last_result = self.invoke_dbt(
+                    command=retry_cmd,
+                    env=env,
+                    cwd=tmp_project_dir,
+                    context=context,
+                )
+            except Exception:
```
tatiana, line 486: Let's capture specific exceptions to avoid hidding other issues

cosmos/operators/watcher.py
```diff
@@ -377,10 +386,175 @@ def _apply_source_freshness(self, context: Context) -> None:
...
+                logger.warning(
+                    "dbt retry attempt %d/%d raised an exception — %d failure(s) remain pending",
+                    attempt,
+                    self.dbt_retry_count,
+                    len(self._pending_failures),
+                    exc_info=True,
+                )
+        return last_result
```
tatiana, line 494: Do users have visibility of intermediate retry outputs? How does this look like in the UI..? Would it make sense to group by retry?

pankajkoti:
Thanks @tatiana for the review. Before working through the inline comments, I want to flag a correctness concern that affects whether this approach is viable.

**The problem:** `dbt retry` overwrites `run_results.json` with only the retried nodes' results. From [`dbt/task/retry.py`](https://github.com/dbt-labs/dbt-core/blob/v1.11.8/core/dbt/task/retry.py), [`RetryTask`](https://github.com/dbt-labs/dbt-core/blob/v1.11.8/core/dbt/task/retry.py#L73) wraps the original task class with a [subset graph](https://github.com/dbt-labs/dbt-core/blob/v1.11.8/core/dbt/task/retry.py#L159-L167) of just the failed `unique_ids`, and `GraphRunnableTask` then [writes](https://github.com/dbt-labs/dbt-core/blob/v1.11.8/core/dbt/task/runnable.py#L640-L649) only those results to `run_results.json`. No merge with the prior file. Same for `target/compiled/*`.

This means anything downstream that reads `target/` after the retry loop sees a partial view of the build. For example, a user `callback(tmp_project_dir, ...)` would see only the retried nodes' artifacts, not the full picture of the original build. Per-node XCom statuses (`<node>_status`) are unaffected, since this PR captures them incrementally via the log-stream callback during execution, not from `run_results.json`.

**Compared with task-level retry ([#2559](https://github.com/astronomer/astronomer-cosmos/pull/2559)):** each attempt starts with a fresh temp project, so within an attempt `target/` is consistent with what was actually executed. Cross-attempt XCom state is preserved via the Variable backup, and it works for `WATCHER_KUBERNETES`. The only thing this PR adds over task-level retry is speed on transient failures, paid for with the partial-target side-effect, no k8s support, and around 280 lines of new code (mostly in `watcher.py` and the `_watcher/` helpers, plus a hook in shared `local.py`).

**Where I net out:** the cost/benefit isn't there as currently scoped. I see two viable paths:

1. **Hold off.** Task-level retry already covers correctness.
2. **Snapshot-and-merge `target/` between attempts** so post-processing reads a complete view, and factor the retry orchestration into `cosmos/operators/_watcher/` so a follow-up can wire it into `WATCHER_KUBERNETES`. I don't yet know how snapshotting would work in practice, where we'd store those artifacts temporarily, etc., which adds further complexity.

Leaning toward (1) unless we think the speed win is meaningful enough to justify (2). WDYT?

pankajkoti:
Converted the PR back into draft mode based on https://github.com/astronomer/astronomer-cosmos/pull/2584#issuecomment-4407329907

pankajkoti:
Astronomer Team members @pankajastro, @tatiana, @yetudada and I met today, and after discussing further on https://github.com/astronomer/astronomer-cosmos/pull/2584#issuecomment-4407329907, we unanimously agreed that now may not be the right time yet to proceed and take this approach further until we deem it necessary to have faster retries and we are ready to invest in finding a reliable way of achieving those without side effects that could include breaking existing end-user flows. Our retry implementation in main already supports the retry story goals documented in https://astronomer.github.io/astronomer-cosmos/guides/run_dbt/airflow-worker/watcher-retry-history.html#goals.

Hence, I am closing this PR.

PR #2592, opened 2026-04-22
Keep watcher sensor polling when producer is still running

pankajkoti:
Previously, any retry of a watcher consumer sensor (try_number > 1) unconditionally fell back to a non-watcher dbt run. This meant that if the sensor timed out while the producer was still working, the retry would launch a duplicate dbt invocation instead of continuing to wait for the producer's XCom updates.

On retry, first check the producer task state. If the producer has reached a terminal state (success, failed, skipped, upstream_failed, removed), fall back to a non-watcher run as before — this covers both producer completion and manual task clears from the UI. Otherwise, keep polling so the sensor can pick up where it left off without spawning a redundant run.

closes: https://github.com/astronomer/oss-integrations-private/issues/359

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -28,6 +28,7 @@ from cosmos.operators._watcher.state import (
     is_dbt_node_status_skipped,
     is_dbt_node_status_success,
     is_dbt_node_status_terminal,
+    is_producer_task_terminated,
     safe_xcom_push,
     xcom_set_lock,
 )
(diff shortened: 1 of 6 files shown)


pankajkoti:
**Retry attempt on the sensor when the producer is still running**:
Screenshot 2026-04-22 at 5 36 39 PM

**Successful completion of DAG afterwards**:
Screenshot 2026-04-22 at 5 37 31 PM

dev/dags/cosmos_manifest_selectors_example.py
```diff
@@ -58,7 +58,7 @@
         profile_config=profile_config,
         render_config=RenderConfig(
             load_method=LoadMode.DBT_MANIFEST,
-            select=["+customers"],
+            select=["+customers", "raw_orders", "raw_payments"],
```
pankajkoti:
```suggestion
            select=["+customers"],
```
This change is outside the scope of the current PR. The DAG is flaky. And a fix is attempted in PR https://github.com/astronomer/astronomer-cosmos/pull/2593. Added temporarily to test the CI.

cosmos/operators/_watcher/base.py
```diff
@@ -600,6 +601,28 @@ def _cache_compiled_sql(self, ti: Any, context: Context) -> None:
...
+        # Producer is still active — the sensor likely timed out while the
+        # producer was still working.  Keep polling instead of launching a
+        # duplicate dbt run.
+        logger.info(
+            "Try #%s but producer '%s' is still %s — continuing to poll instead of fallback.",
+            try_number,
+            self.producer_task_id,
+            producer_task_state or "unknown",
+        )
+        return None
```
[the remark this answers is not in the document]
pankajkoti: I don't think any of these conditions need to be handled here. If state fetching returns `None` (`ImportError` on Airflow deps, `NameError` in the Airflow 3 supervisor path, or a missing producer `TI` row), that indicates a deployment or database state problem, not something the sensor should paper over with a duplicate dbt run. And it's not an infinite poll; the sensor is bounded by its timeout and will fail loudly, which is the signal we want in a broken environment.

dev/dags/cosmos_manifest_selectors_example.py
pankajkoti: rebased so no longer need this.

tatiana (approved):
Looks great, thanks a lot, @pankajkoti for creating a new PR and rebasing after the many changes introduced in the meantime.

PR #2488, opened 2026-03-21
feat: Add `GCP_GKE` and `WATCHER_GCP_GKE` execution modes

vricciardulli:
## Description

Added `GCP_GKE` and `WATCHER_GCP_GKE` execution modes.

It's the same as `KUBERNETES` and `WATCHER_KUBERNETES` modes, but uses GKE operators (i.e. `GKEStartPodOperator`).

This enables a user to use cosmos in Managed Airflow Gen 3 (previously Cloud Composer 3) with a self-managed custom GKE cluster.

**Tests**:
* Added unit tests
* **Did not add any integration test.**
* I tested this on my personal GCP in Managed Airflow Gen 3 (previously Cloud Composer 3) with custom GKE cluster.

**Implementation details**:
* Given the amount of duplication between the k8s modes and the new GKE ones, I've extracted all common code and placed it in helper functions in `cosmos/operators/_k8s_common.py`.
* I've attempted to type hint as best as possible, but I had to ignore some type-checker errors. **I've marked those with comments**.
* An alternative approach would have been to use a factory to generate the classes based on the pod operator provided, but I think that this approach would have been too opaque.

**Docs**:
* I did not add separate pages documenting these modes.
* I concisely added details of GKE modes in the k8s and watcher k8s pages.
* Also added to the mermaid graph.

Co-Authored-By: Claude Opus 4.6 (1M context) [noreply@anthropic.com](mailto:noreply@anthropic.com)

## Related Issue(s)

closes https://github.com/astronomer/astronomer-cosmos/issues/2487
closes https://github.com/astronomer/astronomer-cosmos/issues/2379

## Breaking Change?

There should not be a breaking change.

## Checklist

- [x] I have made corresponding changes to the documentation (if required)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #2595, opened 2026-04-22
Emit per-model outlets in `ExecutionMode.WATCHER_KUBERNETES`

MichaelRBlack:
## Summary

Cosmos 1.14.0 added per-model outlet URI emission to `ExecutionMode.WATCHER`
consumer sensors, but the `WATCHER_KUBERNETES` path never wired up the
producer's log-parsing callback with the `model_outlet_uris` /
`dataset_namespace` bindings that the SUBPROCESS and DBT_RUNNER paths use.
As a result, the per-node `*_status` XCom dict always carried
`"outlet_uris": []`, and the Kubernetes consumer sensors had no way to emit
Airflow Assets — so lineage catalogs that read from Airflow's outlet system
(OpenLineage, OpenMetadata, etc.) showed pipelines with empty inputs/outputs
and no table-level edges for WATCHER_KUBERNETES-mode runs.

### Producer side

`DbtProducerWatcherKubernetesOperator`:
- Accepts `manifest_filepath` (already threaded through `task_args` from
  `ProjectConfig.manifest_path` by `cosmos.converter`).
- At `execute` time, resolves the dataset namespace from the profile and
  populates `self._model_outlet_uris` by reading the scheduler-side
  manifest. The pod's own `target/manifest.json` isn't reachable from the
  scheduler, so the SUBPROCESS watcher's `project_dir`-based lookup can't
  be reused here — we read the manifest Cosmos already parses for DAG
  rendering instead.
- Publishes both values into module-level globals alongside the existing
  `producer_task_context`, so the static `WatcherKubernetesCallback` can
  thread them through to `store_dbt_resource_status_from_log`.

### Consumer side

`DbtConsumerWatcherKubernetesSensor` now overrides `execute` and
`execute_complete` to call a new `_emit_datasets` method, mirroring
`DbtConsumerWatcherSensor`. `execute_complete` extracts `outlet_uris` from
the trigger event before emitting.

### Shared helper

`LocalDbtBaseOperator.register_dataset`'s body is extracted into a new
`cosmos.dataset.register_dataset_on_task` free function so both execution
modes can use it without inheriting from the local operator. The local
method becomes a thin delegate; its public signature and behaviour are
preserved.

## Backwards compatibility

No breaking changes. When `manifest_filepath` is not supplied or the
manifest file isn't present on the scheduler, the producer logs a warning
and skips outlet population — same graceful degradation as the SUBPROCESS
watcher when `{project_dir}/target/manifest.json` is absent. dbt execution
and XCom status reporting continue to work in that case; only dataset
emission is lost.

## Test plan

New unit tests under `tests/operators/test_watcher_kubernetes_unit.py`:

- [x] Producer `execute` resolves namespace + populates `_model_outlet_uris` from the manifest
- [x] Producer `execute` publishes outlets/namespace/context into module globals
- [x] Producer `execute` silently skips outlet population when manifest is missing
- [x] `WatcherKubernetesCallback.progress_callback` passes outlet state into `store_dbt_resource_status_from_log`
- [x] Consumer `_emit_datasets` registers one Asset per outlet URI
- [x] Consumer `_emit_datasets` no-ops when `emit_datasets=False` or `_outlet_uris=[]`
- [x] Consumer `execute_complete` stashes `outlet_uris` from the trigger event before emitting

An `autouse` fixture resets the module-level producer globals between tests so state doesn't leak.

## Context for review

Found this gap while setting up OpenMetadata ingestion for a `WATCHER_KUBERNETES` deployment: Airflow pipelines showed up in OM but had empty inputs/outputs, so no table-level lineage edges were drawn. Traced it to the producer callback not threading outlet bindings through — a regression-adjacent gap from Cosmos 1.14.0, where WATCHER got outlet support but WATCHER_KUBERNETES didn't. Code change above mirrors the SUBPROCESS path as closely as possible.


tatiana (commented):
@MichaelRBlack, sorry for the delay in reviewing your ocntribution, I'd love to see it shipped in Cosmos!

Please, could you rebase and fix the checks?

tatiana:
@MichaelRBlack, we're very close - I believe unit tests started failing after a rebase https://github.com/astronomer/astronomer-cosmos/pull/2488.

Please, could you take a look?

MichaelRBlack:
@tatiana I've updated my PR so it can be ready for review again

cosmos/operators/_k8s_common.py
```diff
@@ -485,10 +497,58 @@ def build_watcher_pod_manager(operator: K8sWatcherProducerProtocol) -> CosmosKub
...
+        return
+    if not operator.manifest_filepath:
+        operator.log.warning(
+            "manifest_filepath not supplied to %s; per-model dataset emission is disabled for this run. "
+            "Pass ProjectConfig.manifest_path to enable it.",
+            type(operator).__name__,
+        )
+        return
+    manifest_path = Path(operator.manifest_filepath)
+    if not manifest_path.exists():
```
tatiana: I think this will break remote `ProjectConfig.manifest_path` values. `ProjectConfig` stores `manifest_path` as an Airflow `ObjectStoragePath`, and Cosmos supports remote manifests such as `s3://...`. In Airflow 3.1, changing that to `Path(...)` turns it into a local path like `s3:/...`, so I believe this `exists()` check will fail, and the watcher producer will silently skip generating outlet URIs for remote manifests.

Could we keep the original `ObjectStoragePath` here and make `compute_model_outlet_uris()` read via `manifest_path.open(...)`, similar to `DbtGraph._load_manifest_from_file()`? It would also be good to add a unit test using an `ObjectStoragePath`/remote-style manifest path to ensure we don't see any regressions.

tatiana (commented):
Hi @MichaelRBlack , thanks a lot for addressing all the feedback.

I left some inline comment that I believe is a quick fix, and that will help us guaranteeing this feature is robust for users leveraging Cosmos remote manifests. I believe this is the last blocker for us to merge this change.

Once this feedback is addresed, I'm happy to approve and merge the change.

Some non-blocker feedback, which I'd really appreciate if you could do in a follow-up PR: I noticed that  `_emit_datasets`, `execute`, `execute_complete` are now duplicated in 3 places: the SUBPROCESS `DbtConsumerWatcherSensor` (already existed) and the K8s + GCP-GKE consumers added here. All of them inherit from `BaseConsumerSensor`, and `register_dataset_on_task(self, ...)` no longer needs the local operator. Could we move these 3 methods to `BaseConsumerSensor` and drop the copies (SUBPROCESS included)? I really liked your refactor and creation of `register_dataset_on_task` and I think this could help significantly on this minor refactor.

There is also some duplication on the producer `__init__` (`_should_generate_model_uris`, `manifest_filepath` and `_dataset_namespace` / `_model_outlet_uris`) - they seem duplicated between the 2 producers - if we could refactor this in the same follow up PR, it would be great.

cosmos/operators/_k8s_common.py
MichaelRBlack: Good catch — fixed in 78f7b3ad.

`compute_model_outlet_uris` now accepts `str | Path | ObjectStoragePath` and reads via `path.open()` (normalising a plain `str` to `Path` first), mirroring `DbtGraph._load_manifest_from_file`. The watcher producer no longer wraps `manifest_filepath` in `Path()` or calls `.exists()` — it forwards the `ObjectStoragePath` through unchanged, so remote schemes (`s3://`, `gs://`, …) are preserved; missing/unreadable manifests are still handled gracefully inside `compute_model_outlet_uris` (returns `{}` and logs).

Added tests: one reading a manifest through an `ObjectStoragePath`, and one asserting the producer forwards a remote manifest path to `compute_model_outlet_uris` unchanged (regression guard).

MichaelRBlack: Update: rather than a follow-up, I've addressed both refactors in this PR (commit e4df60cb):

- Hoisted `_emit_datasets` and the `execute` / `execute_complete` emission hooks into `BaseConsumerSensor` (using `register_dataset_on_task`), and dropped the SUBPROCESS, K8s, and GCP-GKE copies — all three now inherit the shared behaviour.
- Extracted the duplicated producer `__init__` state (`_should_generate_model_uris`, `manifest_filepath`, `_dataset_namespace` / `_model_outlet_uris`, plus the tests/context/callback wiring) into `_k8s_common.init_watcher_producer` / `finalize_watcher_producer`, shared by both producers.

Net ~125 fewer lines. Added canonical `BaseConsumerSensor` emit/execute tests under `tests/operators/_watcher` and reduced the K8s/GCP-GKE consumer suites to one inheritance smoke test each.

tatiana (approved):
@MichaelRBlack Thank you so much for all the hard-work and iterations on this PR. It is a  very important feature, and I really appreciate you taking it to the finishing line. Since it is - to an extent - a new feature, we'll release it as part of Cosmos 1.16.0.

We'll make sure we cut an alpha sooner than later, so you can validate it.

Issue #2596, opened 2026-04-22
`ExecutionMode.WATCHER` and `depends_on_past`


tatiana:
There is a problem that is inherent to the Watcher implementation (from conception, until now - 1.14.x - in all existing versions of Airflow to date - latest is 3.2.0):

**DAG Run 1:**
- Producer attempt 1: runs dbt build => fails
- Producer attempt 2: skips (AirflowSkipException)
- Consumer model_f: detects error => fails => retries => falls back to _fallback_to_non_watcher_run => running dbt run `--select model_f`

**DAG Run 2 starts (with depends_on_past=True):**
- `depends_on_past` checks the same task in the previous run. The producer in Run 1 is "skipped". Airflow treats skipped as satisfying `depends_on_past`. So Run 2's producer can start immediately.
- Run 2's producer runs dbt build (which includes model_f) while Run 1's consumer is still running `dbt run --select model_f` from the fallback.

This creates a race: two dbt processes hitting the same tables concurrently.

This behaviour is inherent to the watcher design, where producers and consumers run independently.

There are two workarounds for this issue:
1. `max_active_runs=1` on the DAG - prevents Run 2 from starting until Run 1 fully completes
2. Check if dbt implements its own concurrency control in a way that it would acquire locks on models being processed

johnhoran:
A third solution, assuming https://github.com/astronomer/astronomer-cosmos/pull/2597 gets merged would be to use `wait_for_downstream`.  If `depends_on_past` is true then the `dbt_producer_watcher_done` task could be placed so it is downstream of the leaves of the other tasks in the dbt task group.  Then `wait_for_downstream` could be forced true for all tasks in the group except `dbt_producer_watcher_done`, which would force airflow to wait before scheduling any tasks in the task group until the previous `dbt_producer_watcher_done` is complete.

tatiana:
@johnhoran The solution seems to work for `DbtTaskGroup`, how do you see it working for `DbtDag`?

Also, how do you suggest we create an integration test simulating this scenario? I feel like it may be beyond `dag.test()` capabilities

johnhoran:
Well for DbtDag you have two options.  `max_active_runs` would work most of the time, but I'm not a huge fan of that parameter because I see a lot of cases where a dag run fails and then the next dagrun starts, but is blocked by a depends_on_past.  So to unblock we have to clear the previous dagrun and then manually fail the new dagrun so the previous one can restart.

[...]

PR #2597, opened 2026-04-22
Prevent watcher producer skip propagating to downstream tasks via gateway task

### LINKED ISSUE #2594: DbtTaskGroup watcher: producer skip propagates to downstream tasks
## Problem

When using `ExecutionMode.WATCHER` with `DbtTaskGroup`, if the producer task is skipped on retry (via `AirflowSkipException`), the skip state propagates to all tasks downstream of the task group — even when all consumer tasks inside the group succeeded. This behaviour is a side-effect of the PR https://github.com/astronomer/astronomer-cosmos/pull/2559.

This means any non-Cosmos tasks wired after the `DbtTaskGroup` (e.g., reporting, notifications, cleanup) are silently skipped, breaking the expected DAG execution flow.

**This makes `ExecutionMode.WATCHER` behave differently from `ExecutionMode.LOCAL`** when used with `DbtTaskGroup`: in `LOCAL` mode, each model runs as an independent task and a single model failure does not cause downstream tasks to be skipped. Users switching from `LOCAL` to `WATCHER` may not expect this change in behaviour.

## Video demonstrating the problem

https://github.com/user-attachments/assets/3cd03975-3c97-44c5-bf6f-8f56796815a9

## Steps to reproduce

1. Create a `DbtTaskGroup` with `ExecutionMode.WATCHER`
2. Wire an `EmptyOperator` (or any task) downstream of the group: `dbt_group >> post_dbt`
3. Configure `retries >= 1` on the DAG
4. Run a dbt project where at least one model fails on the first attempt

An example DAG and dbt project to reproduce this issue are available in #2591:
- DAG: [`dev/failed_dags/example_watcher_downstream_not_skipped.py`](https://github.com/astronomer/astronomer-cosmos/pull/2591)
- dbt project: [`dev/dags/dbt/watcher_downstream_not_skipped/`](https://github.com/astronomer/astronomer-cosmos/pull/2591)

## Expected behaviour

- The producer fails on the first attempt, retries, and skips
- Consumer tasks read model statuses from XCom and complete (some succeed, some may retry individually)
- `post_dbt` runs after the task group completes

## Actual behaviour

- `post_dbt` is skipped because Airflow's default `trigger_rule="all_success"` sees the producer's skipped state and skips downstream tasks

## Context

This issue does not affect `DbtDag`, where the producer-to-consumer dependency is handled differently (`producer >> consumers` with `trigger_rule="always"`).

In `DbtTaskGroup`, setting `trigger_rule="always"` on consumers is not an option because users may have upstream tasks before the group, and `trigger_rule="always"` would cause consumers to run before those upstream tasks complete.

tatiana:
When using `ExecutionMode.WATCHER` with `DbtTaskGroup`, the producer task may be skipped on retry (via `AirflowSkipException`) since https://github.com/astronomer/astronomer-cosmos/pull/2559. Because the producer is a leaf task inside the `TaskGroup`, Airflow's default `trigger_rule="all_success"` causes any tasks downstream of the group to be skipped as well - even when all consumer tasks succeeded. **This makes `ExecutionMode.WATCHER` behave differently from `ExecutionMode.LOCAL`** when used with `DbtTaskGroup`. This issue does not happen in `DbtDag`.

This PR introduces a downstream gateway task (`dbt_producer_watcher_done`) for the producer when in watcher mode. This ensures the producer's skip state does not propagate outside the task group, since the gateway task has `trigger_rule=none_failed`.

This solution was originally proposed by @johnhoran in https://github.com/astronomer/astronomer-cosmos/pull/2430. The code has changed significantly since, so we're opening a new PR and adding him as a co-author. It is an alternative approach to https://github.com/astronomer/astronomer-cosmos/pull/2591.

Co-authored-by: John Horan <jhoran@zendesk.com>

**Behaviour before and after**

The newly introduced DAG `example_watcher_downstream_not_skipped` illustrates the problematic behaviour this PR aims to address.

Video before this change:

https://github.com/user-attachments/assets/3cd03975-3c97-44c5-bf6f-8f56796815a9

Video after this change:

https://github.com/user-attachments/assets/3a50b882-9ed0-4d16-b5ae-662bec76b4e3

**Alternative approaches**

- Cosmos changing the `trigger_rule` of the downstream tasks as implemented in https://github.com/astronomer/astronomer-cosmos/pull/2591
- Accept that `ExecutionMode.LOCAL` and `ExecutionMode.WATCHER` are not fully compatible and document that users should set `trigger_rule="none_failed_min_one_success"` manually

**Implications**

- The DAG topology is changed to include a new `EmptyOperator` task (`dbt_producer_watcher_done`)

**Unsolved problem: `depends_on_past`**

This PR does not solve the problem: https://github.com/astronomer/astronomer-cosmos/issues/2596.

PR #2430 also doesn't solve the `depends_on_past` race condition. The PR fixes the retry/skip mechanism (backup XCom, restore on retry, skip producer on second attempt), but it doesn't prevent Run 2's producer from starting while Run 1's consumers are still running their fallback dbt commands.

Closes #2594

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -716,6 +716,20 @@ def _add_watcher_producer_task(
     )
     producer_airflow_task = create_airflow_task(producer_task_metadata, dag, task_group=task_group)
     tasks_map[PRODUCER_WATCHER_TASK_ID] = producer_airflow_task
+
+    # For DbtTaskGroup, add a gate task that absorbs the producer's skip state
+    # so it does not propagate to tasks downstream of the group.
+    # Not needed for DbtDag where producer >> consumers with trigger_rule="always" handles this.
+    if task_group is not None:
+        from cosmos.operators._watcher.base import create_producer_done_task
+
+        producer_done_task = create_producer_done_task(
+            dag=dag,
+            task_group=task_group,
+            task_id=f"{PRODUCER_WATCHER_TASK_ID}_done",
+        )
+        producer_airflow_task >> producer_done_task
+
     return producer_airflow_task
 
 
@@ -732,8 +746,8 @@ def _add_watcher_dependencies(
     - make the producer task to be the parent of the root dbt nodes, without blocking them from sensing XCom
     """
     for node_id, task_or_taskgroup in tasks_map.items():
-        # We do not want to set a dependency between the producer task and itself
-        if node_id == PRODUCER_WATCHER_TASK_ID:
+        # We do not want to set a dependency between the producer task (or its gate) and itself
+        if node_id in (PRODUCER_WATCHER_TASK_ID, f"{PRODUCER_WATCHER_TASK_ID}_done"):
             continue
 
         node_tasks = (
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -25,6 +25,7 @@ from cosmos.constants import (
     DBT_SETUP_ASYNC_TASK_ID,
     DBT_TEARDOWN_ASYNC_TASK_ID,
     DEFAULT_DBT_RESOURCES,
+    PRODUCER_WATCHER_DONE_TASK_ID,
     PRODUCER_WATCHER_TASK_ID,
     SUPPORTED_BUILD_RESOURCES,
     TESTABLE_DBT_RESOURCES,
@@ -726,7 +727,7 @@ def _add_watcher_producer_task(
         producer_done_task = create_producer_done_task(
             dag=dag,
             task_group=task_group,
-            task_id=f"{PRODUCER_WATCHER_TASK_ID}_done",
+            task_id=PRODUCER_WATCHER_DONE_TASK_ID,
         )
         producer_airflow_task >> producer_done_task
 
@@ -747,7 +748,7 @@ def _add_watcher_dependencies(
     """
     for node_id, task_or_taskgroup in tasks_map.items():
         # We do not want to set a dependency between the producer task (or its gate) and itself
-        if node_id in (PRODUCER_WATCHER_TASK_ID, f"{PRODUCER_WATCHER_TASK_ID}_done"):
+        if node_id in (PRODUCER_WATCHER_TASK_ID, PRODUCER_WATCHER_DONE_TASK_ID):
             continue
 
         node_tasks = (
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -668,3 +668,29 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
             return True
         else:
             raise AirflowException(f"{self._resource_label} '{self.model_unique_id}' finished with status '{status}'")
+
+
+def create_producer_done_task(dag: Any, task_group: Any, task_id: str) -> Any:
+    """Create an EmptyOperator that absorbs the producer's skip state on retry.
+
+    This task sits downstream of the producer inside the DbtTaskGroup. When the producer
+    is skipped on retry, this task still succeeds (trigger_rule=NONE_FAILED), preventing
+    the skip from propagating to tasks downstream of the group.
+    """
+    try:
+        from airflow.providers.standard.operators.empty import EmptyOperator
+    except ImportError:
+        from airflow.operators.empty import EmptyOperator  # type: ignore[no-redef]
+
+    from airflow.utils.trigger_rule import TriggerRule
+
+    return EmptyOperator(  # type: ignore[no-untyped-call]
+        task_id=task_id,
+        dag=dag,
+        task_group=task_group,
+        trigger_rule=TriggerRule.NONE_FAILED,
+        doc_md=(
+            "**Cosmos internal task.** Absorbs the producer's skip state on retry "
+            "so it does not propagate to tasks downstream of this DbtTaskGroup."
+        ),
+    )
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -682,7 +682,10 @@ def create_producer_done_task(dag: Any, task_group: Any, task_id: str) -> Any:
     except ImportError:
         from airflow.operators.empty import EmptyOperator  # type: ignore[no-redef]
 
-    from airflow.utils.trigger_rule import TriggerRule
+    try:
+        from airflow.task.trigger_rule import TriggerRule
+    except ImportError:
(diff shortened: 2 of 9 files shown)


cosmos/airflow/graph.py
```diff
@@ -732,8 +746,8 @@ def _add_watcher_dependencies(
     - make the producer task to be the parent of the root dbt nodes, without blocking them from sensing XCom
     """
     for node_id, task_or_taskgroup in tasks_map.items():
-        # We do not want to set a dependency between the producer task and itself
-        if node_id == PRODUCER_WATCHER_TASK_ID:
+        # We do not want to set a dependency between the producer task (or its gate) and itself
+        if node_id in (PRODUCER_WATCHER_TASK_ID, f"{PRODUCER_WATCHER_TASK_ID}_done"):
```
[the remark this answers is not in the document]
tatiana: The only goal of the gateway task is to avoid unintentionally skipping `DbtTaskGroup` downstream tasks from being run - it is not relevant in other parts of the code, and for this reason, we're skipping adding it to this data structure. The comment explains the motivation.

tests/operators/test_watcher.py
```diff
@@ -1746,7 +1747,9 @@ def test_dbt_task_group_with_watcher():
     assert isinstance(dag_dbt_task_group_watcher.task_dict["dbt_task_group.customers_run"], DbtRunWatcherOperator)
     assert isinstance(dag_dbt_task_group_watcher.task_dict["dbt_task_group.orders_run"], DbtRunWatcherOperator)
 
-    assert dag_dbt_task_group_watcher.task_dict["dbt_task_group.dbt_producer_watcher"].downstream_task_ids == set()
+    assert dag_dbt_task_group_watcher.task_dict["dbt_task_group.dbt_producer_watcher"].downstream_task_ids == {
+        "dbt_task_group.dbt_producer_watcher_done",
+    }
```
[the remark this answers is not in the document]
tatiana: Addressed in 6a1587dc9bbdcfcc8fd03e8bf665d25b0b3f29b9

cosmos/airflow/graph.py
```diff
@@ -716,6 +716,20 @@ def _add_watcher_producer_task(
...
+    # so it does not propagate to tasks downstream of the group.
+    # Not needed for DbtDag where producer >> consumers with trigger_rule="always" handles this.
+    if task_group is not None:
+        from cosmos.operators._watcher.base import create_producer_done_task
+
+        producer_done_task = create_producer_done_task(
+            dag=dag,
+            task_group=task_group,
+            task_id=f"{PRODUCER_WATCHER_TASK_ID}_done",
+        )
```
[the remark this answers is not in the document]
tatiana: Addressed in 6a1587dc9bbdcfcc8fd03e8bf665d25b0b3f29b9

cosmos/operators/_watcher/base.py
```diff
@@ -668,3 +668,29 @@ def poke(self, context: Context) -> bool:
...
+    This task sits downstream of the producer inside the DbtTaskGroup. When the producer
+    is skipped on retry, this task still succeeds (trigger_rule=NONE_FAILED), preventing
+    the skip from propagating to tasks downstream of the group.
+    """
+    try:
+        from airflow.providers.standard.operators.empty import EmptyOperator
+    except ImportError:
+        from airflow.operators.empty import EmptyOperator  # type: ignore[no-redef]
+
+    from airflow.utils.trigger_rule import TriggerRule
```
pankajastro: shall we also try/except here 
```python
try:
    from airflow.sdk import TriggerRule
except ImportError:
    from airflow.utils.trigger_rule import TriggerRule
```

cosmos/airflow/graph.py
```diff
@@ -716,6 +716,20 @@ def _add_watcher_producer_task(
...
+    # For DbtTaskGroup, add a gate task that absorbs the producer's skip state
+    # so it does not propagate to tasks downstream of the group.
+    # Not needed for DbtDag where producer >> consumers with trigger_rule="always" handles this.
+    if task_group is not None:
+        from cosmos.operators._watcher.base import create_producer_done_task
+
+        producer_done_task = create_producer_done_task(
+            dag=dag,
+            task_group=task_group,
+            task_id=f"{PRODUCER_WATCHER_TASK_ID}_done",
```
pankajkoti: I think it would be nice to define a constant for this task ID, as I see it getting reused below.

pankajastro (approved):
LGTM

cosmos/operators/_watcher/base.py
```diff
@@ -668,3 +668,29 @@ def poke(self, context: Context) -> bool:
             return True
         else:
             raise AirflowException(f"{self._resource_label} '{self.model_unique_id}' finished with status '{status}'")
+
+
+def create_producer_done_task(dag: Any, task_group: Any, task_id: str) -> Any:
```
pankajkoti:
```suggestion
def create_producer_done_task(dag: DAG, task_group: TaskGroup, task_id: str) -> EmptyOperator:
```
would that be possible?

pankajkoti (approved):
LGTM. I believe this is the right direction. Thanks a lot for finding a simpler approach than https://github.com/astronomer/astronomer-cosmos/pull/2591

cosmos/operators/_watcher/base.py
tatiana: Addressed in 456671f257c64369b33100170992faabf626428c

cosmos/airflow/graph.py
tatiana: Addressed in 456671f257c64369b33100170992faabf626428c

cosmos/operators/_watcher/base.py
tatiana: Addressed in 13c15e233d1fa9b593800dc224aeefdcb1925c43

cosmos/airflow/graph.py
```diff
@@ -716,6 +717,20 @@ def _add_watcher_producer_task(
...
+    if task_group is not None:
+        from cosmos.operators._watcher.base import create_producer_done_task
+
+        producer_done_task = create_producer_done_task(
+            dag=dag,
+            task_group=task_group,
+            task_id=PRODUCER_WATCHER_DONE_TASK_ID,
+        )
+        producer_airflow_task >> producer_done_task
+
```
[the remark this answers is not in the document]
tatiana: This is an edge case (no dbt nodes being selected) that is not expected in practice, and it would apply to other `ExecutionModes` to an extent. It is outside of the scope of this ticket to resolve it, and the changes introduced here do not make it worse.

PR #2600, opened 2026-04-23
Add watcher retry behaviour history documentation

tatiana:
Adds a dedicated documentation page that traces the evolution of retry behaviour in `ExecutionMode.WATCHER` across Cosmos releases (1.11.0 through 1.14.1 - the last one to be released).                                                                       
                                                               
Understanding how retries work in watcher mode has been difficult because the behaviour changed significantly across releases - from dangerous duplicate dbt build runs, to forced retries=0, to silently marking failed models as successful. This page provides a single reference that documents each version's behaviour, the bugs introduced, and the PRs that addressed them.

The page is linked from the main watcher execution mode documentation, but is not listed in the navigation sidebar (`:orphan:`) to avoid cluttering the guides index, since it is not a guide.

PR #2604, opened 2026-04-23
Restore memory-optimised imports docs for Cosmos < 1.14.0

pankajkoti:
PR #2531 removed the "Enable Memory-Optimized Imports" section from the memory optimization guide and the detailed explanation/examples from the cosmos-conf reference, because the switch to lazy imports in cosmos/__init__.py made enable_memory_optimised_imports a no-op.

However, that change only shipped in Cosmos 1.14.0. Users on earlier versions still need the flag (and the full-module-path import pattern it requires) to reduce the DAG processor and worker memory footprint, and removing the docs left them with no guidance.

Bring the content back, clearly scoped to Cosmos < 1.14.0:

- Add it to memory_optimization.rst as section 6 (just before the debug-mode profiling section), with an "Applies to" preamble so readers on 1.14.0+ know to skip it.
- Restore the long-form explanation and both literalinclude example blocks in cosmos-conf.rst, with a note that the setting is a no-op on 1.14.0 and newer.

related: #2531

### DIFF
--- a/docs/optimize_performance/memory_optimization.rst
+++ b/docs/optimize_performance/memory_optimization.rst
@@ -179,7 +179,7 @@ Cosmos provides various configuration options and execution modes to optimize me
 
 **Applies to**: Cosmos versions **earlier than 1.14.0**. Since Cosmos 1.14.0, ``cosmos/__init__.py`` uses lazy imports by default and this setting is a no-op. If you are on 1.14.0 or newer, skip this section.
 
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Great catch, @pankajkoti !

pankajastro (approved):

pankajkoti:
> Great catch, @pankajkoti !

yes thanks. Came to mind after having a sneak peek of https://github.com/astronomer/astronomer-cosmos/pull/2600 and that we don't have version-aware docs :), yet.

PR #2615, opened 2026-04-27
Watcher depends on past

### LINKED ISSUE #2596: `ExecutionMode.WATCHER` and `depends_on_past`

johnhoran:
## Description
Adds a `wait_for_downstream` to all tasks in the DbtTaskGroup if `depends_on_past` is set to true.  If `depends_on_past` is set to true, then the aim is to consider the DbtTaskGroup as a single atomic step.  Most of the time this is fine, but as outlined in #2596, if a run fails and is marked as skipped, then the producer task from a  subsequent run can start, even if the watcher consumer tasks from the initial run haven't completed.

This PR aims to fix that by moving the leaves of the task group so they are upstream of the watcher done task, and have `wait_for_downstream` set to true, in which case airflow will only start a task group if the previous run watcher done task has completed.
Before:
Screenshot 2026-04-27 at 11 25 17

After this change when using `TestBehavior.AFTER_ALL`
Screenshot 2026-04-27 at 11 27 49

After this change when using `TestBehavior.AFTER_EACH`
Screenshot 2026-04-27 at 11 26 21

---
As an aside, I also put the test added for `TestBehavior.AFTER_ALL` task_map, mostly so I'm only modifying the `wait_for_downstream` in a single method, but also I suspect there was a bug when using it with `ExecutionConfig.AIRFLOW_ASYNC`.
https://github.com/astronomer/astronomer-cosmos/blob/c9191045a609325882824a417e984bf4d8c40b75/cosmos/airflow/graph.py#L881-L883
I haven't used that mode, so I can't really test that.

## Related Issue(s)

closes #2596 

## Checklist


tests/airflow/test_graph.py
```diff
@@ -2228,3 +2228,78 @@ def test_add_watcher_producer_task_passes_freshness_callback_via_setup_operator_
...
+            "tg.child_run",
+            "tg.dbt_producer_watcher",
+            "tg.child2_v2_run",
+            "tg.child2_v2_b.test",
+        }
+    if test_behavior == TestBehavior.AFTER_ALL:
+        assert dag.task_dict["tg.dbt_producer_watcher_done"].upstream_task_ids == {
+            "tg.dbt_producer_watcher",
+            "tg.astro_shop_test",
+        }
```
tatiana, line 2305: The new test covers the dependency topology well. A small extra assertion that `tg.dbt_producer_watcher_done.wait_for_downstream is False` would make the intended gate behaviour explicit.

tatiana (approved):
@johnhoran Thanks a lot for working on this, and I'm sorry for the delay in reviewing. The approach makes sense to me, and I’m comfortable moving forward so we can release the fix.

A few non-blocking follow-ups that would be helpful, in a separate PR after this lands:

1. As we agreed, this only addresses the race for `DbtTaskGroup`, since the `dbt_producer_watcher_done` gate only exists when `task_group is None`. Could we explicitly:
- Document/call out that `DbtDag` is still not covered, and
- Create a follow-up issue or PR to either raise an exception on `DbtDag` or properly handle it there as well?

2. The watcher docs still describe `max_active_runs=1` as the workaround for the `depends_on_past=True` race. Once this lands, that section should be updated to explain the new `DbtTaskGroup` behaviour and any remaining limitations for `DbtDag`.

3. Adding the `AFTER_ALL` test task to `tasks_map` looks useful, but it is also a behavioural side-change: `tasks_map` now includes an additional entry. In the future, it would be great to split this up into a separate PR. If, for now, we could document this change of behaviour, it would be great.

4. Minor inline comment (https://github.com/astronomer/astronomer-cosmos/pull/2615#discussion_r3266242337)

I’m okay with the current test staying as wiring/topology coverage rather than a full cross-run scheduler integration test, since reproducing the actual concurrent-run race cleanly is much harder and probably not worth blocking this fix.

tatiana:

PR #2616, opened 2026-04-28
Fix duplicate `deferrable` kwarg in `DbtRunAirflowAsyncBigqueryOperator`

pankajastro:
Pop 'deferrable' from kwargs before passing to
BigQueryInsertJobOperator.__init__ to avoid a TypeError when the caller explicitly supplies the argument. Defaults to True when not provided, preserving existing behaviour.

### DIFF
--- a/cosmos/operators/_asynchronous/bigquery.py
+++ b/cosmos/operators/_asynchronous/bigquery.py
@@ -33,8 +33,11 @@ from cosmos.config import ProfileConfig
 from cosmos.constants import AIRFLOW_VERSION
 from cosmos.dataset import get_dataset_alias_name
 from cosmos.exceptions import CosmosValueError
+from cosmos.log import get_logger
 from cosmos.operators.local import AbstractDbtLocalBase
 from cosmos.settings import remote_target_path, remote_target_path_conn_id
+
+logger = get_logger(__name__)
 
 DEFAULT_PRODUCER_ASYNC_TASK_ID = "dbt_setup_async"
 
@@ -118,7 +121,7 @@ class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtL
         # To fix this, we temporarily set the base class to only BigQueryInsertJobOperator during initialization,
         # then restore the full inheritance chain afterward.
         if kwargs.pop("deferrable", True) is False:
-            self.log.warning(
+            logger.warning(
                 "DbtRunAirflowAsyncBigqueryOperator requires deferrable=True. "
                 "The provided value of False has been ignored."
             )
--- a/cosmos/operators/_asynchronous/bigquery.py
+++ b/cosmos/operators/_asynchronous/bigquery.py
@@ -117,11 +117,12 @@ class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtL
         # BigQueryInsertJobOperator parameters and hence fails to initialise the operator due to missing arguments.
         # To fix this, we temporarily set the base class to only BigQueryInsertJobOperator during initialization,
         # then restore the full inheritance chain afterward.
+        deferrable = kwargs.pop("deferrable", True)
         DbtRunAirflowAsyncBigqueryOperator.__bases__ = (BigQueryInsertJobOperator,)
         super().__init__(
             gcp_conn_id=self.gcp_conn_id,
             configuration=self.configuration,
-            deferrable=True,
+            deferrable=deferrable,
             **kwargs,
         )
         self.async_context = extra_context or {}
--- a/cosmos/operators/_asynchronous/bigquery.py
+++ b/cosmos/operators/_asynchronous/bigquery.py
@@ -117,12 +117,16 @@ class DbtRunAirflowAsyncBigqueryOperator(BigQueryInsertJobOperator, AbstractDbtL
         # BigQueryInsertJobOperator parameters and hence fails to initialise the operator due to missing arguments.
         # To fix this, we temporarily set the base class to only BigQueryInsertJobOperator during initialization,
         # then restore the full inheritance chain afterward.
-        deferrable = kwargs.pop("deferrable", True)
+        if kwargs.pop("deferrable", True) is False:
+            self.log.warning(
+                "DbtRunAirflowAsyncBigqueryOperator requires deferrable=True. "
+                "The provided value of False has been ignored."
+            )
         DbtRunAirflowAsyncBigqueryOperator.__bases__ = (BigQueryInsertJobOperator,)
         super().__init__(
             gcp_conn_id=self.gcp_conn_id,
             configuration=self.configuration,
-            deferrable=deferrable,
+            deferrable=True,
             **kwargs,
         )
         self.async_context = extra_context or {}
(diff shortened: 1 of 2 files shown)


tatiana (commented):
@pankajastro Thanks for the PR! I have some concerns about this change worth discussing before we move forward.

Could you give more details on the `TypeError` issue you mentioned? Would love to understand the motivation - is there a concrete use case where a user would want `AIRFLOW_ASYNC` without deferrable behaviour?

The execution mode is named `AIRFLOW_ASYNC`, and one of the core design decisions behind it was to *enforce* deferrable behaviour by default — not to make it optional or configurable. Parametrising this feels at odds with the original intent and could lead to user confusion: if someone selects `AIRFLOW_ASYNC` but disables deferrable behaviour, the mode name becomes misleading.

Happy to be convinced otherwise, but I'd want us to be deliberate about relaxing a design constraint that was added intentionally.

pankajastro:
> @pankajastro Thanks for the PR! I have some concerns about this change worth discussing before we move forward.
> 
> Could you give more details on the `TypeError` issue you mentioned? Would love to understand the motivation - is there a concrete use case where a user would want `AIRFLOW_ASYNC` without deferrable behaviour?
> 
> The execution mode is named `AIRFLOW_ASYNC`, and one of the core design decisions behind it was to _enforce_ deferrable behaviour by default — not to make it optional or configurable. Parametrising this feels at odds with the original intent and could lead to user confusion: if someone selects `AIRFLOW_ASYNC` but disables deferrable behaviour, the mode name becomes misleading.
> 
> Happy to be convinced otherwise, but I'd want us to be deliberate about relaxing a design constraint that was added intentionally.

We get error
```
Broken DAG: [/usr/local/airflow/dags/cosmos_sync_dag.py]
  File "/usr/local/lib/python3.12/site-packages/airflow/models/baseoperator.py", line 508, in apply_defaults
    result = func(self, **kwargs, default_args=default_args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/cosmos/operators/_asynchronous/bigquery.py", line 121, in __init__
    super().__init__(
TypeError: airflow.providers.google.cloud.operators.bigquery.BigQueryInsertJobOperator.__init__() got multiple values for keyword argument 'deferrable'
```
When `operator_args` is 
```python
operator_args={ "deferrable": True/False }
```

At a minimum, we should guard against cases where kwargs contains deferrable so the workflow doesn’t break entirely. Maybe allowing an acceptable value only True.

This flexibility is especially useful for testing. For example, in our benchmarking work (https://github.com/astronomer/oss-integrations-private/issues/176), we want a very similar setup to compare async and sync performance, and this approach feels both practical and maintainable.

During the initial work on that issue, we implemented a temporary workaround and published an alpha for testing:
https://github.com/astronomer/oss-integrations-private/issues/176#issuecomment-3250664816

The changes required here are relatively small, so the maintenance overhead should be minimal. Overall, this would help in testing and also safty handling for `deferrable` in kwargs.

cosmos/operators/_asynchronous/bigquery.py
```diff
@@ -117,11 +117,12 @@ def __init__(
         # BigQueryInsertJobOperator parameters and hence fails to initialise the operator due to missing arguments.
         # To fix this, we temporarily set the base class to only BigQueryInsertJobOperator during initialization,
         # then restore the full inheritance chain afterward.
+        deferrable = kwargs.pop("deferrable", True)
```
pankajkoti:
```suggestion
        user_deferrable = kwargs.pop("deferrable", True)
        if user_deferrable is False:
            self.log.warning(
                "deferrable=False is ignored for DbtRunAirflowAsyncBigqueryOperator; "
                "AIRFLOW_ASYNC execution mode is always deferrable."
            )
```

cosmos/operators/_asynchronous/bigquery.py
```diff
@@ -117,11 +117,12 @@ def __init__(
...
         DbtRunAirflowAsyncBigqueryOperator.__bases__ = (BigQueryInsertJobOperator,)
         super().__init__(
             gcp_conn_id=self.gcp_conn_id,
             configuration=self.configuration,
-            deferrable=True,
+            deferrable=deferrable,
```
pankajkoti:
```suggestion
            deferrable=True,
```

pankajkoti (commented):
+1 to what @tatiana is saying. My opinion on this one is to log a warning if the user sends `deferrable=False`, and that we keep the runtime behaviour consistent with the name of the execution mode itself.

Another alternative, if we want to be upfront, is to raise an error if the user sends deferrable=False to avoid misconfiguration/unexpected behaviour, but I don't think we should allow executing synchronously in the `AIRFLOW_ASYNC` execution mode.

pankajastro:
> 1 to what @tatiana is saying. My opinion on this one is to log a warning if the user sends `deferrable=False`, and that we keep the runtime behaviour consistent with the name of the execution mode itself.
> 
> Another alternative, if we want to be upfront, is to raise an error if the user sends deferrable=False to avoid misconfiguration/unexpected behaviour, but I don't think we should allow executing synchronously in the `AIRFLOW_ASYNC` execution mode.

The intention of these changes is to enable testing of the async execution path in non-deferrable mode particular to support issue: https://github.com/astronomer/oss-integrations-private/issues/176, as well as, of course, fixing the `multiple values for kwargs` error.

pankajkoti:
> The intention of these changes is to enable testing of the async execution path in non-deferrable mode particular to support issue: [astronomer/oss-integrations-private#176](https://github.com/astronomer/oss-integrations-private/issues/176), as well as, of course, fixing the `multiple values for kwargs` error.

The multiple values for kwargs will go away, hopefully, with the suggestions I proposed. And if the intention is only to test internally for benchmarking, we should not have it allowed and exposed for Async execution mode interface IMO. Would there be any other way to test it, like by extending to have an internal class in the benchmark repo that has the same parents but setting deferrable=False or mocking the class to have deferrable=False during the test experiment or another approach?

pankajkoti (approved):
Based on the discussion in the stand-up for the ease-of-testing purposes mentioned by @pankajastro, I am okay to get this in if @tatiana is also convinced to.

pankajastro:
> The multiple values for kwargs will go away, hopefully, with the suggestions I proposed. And if the intention is only to test internally for benchmarking, we should not have it allowed and exposed for Async execution mode interface IMO. Would there be any other way to test it, like by extending to have an internal class in the benchmark repo that has the same parents but setting deferrable=False or mocking the class to have deferrable=False during the test experiment or another approach?

Updated PR: https://github.com/astronomer/cosmos-benchmark/pull/12 to use mocking for the `deferrable` kwarg. There might still be some questions around maintainability with this approach—would appreciate your thoughts (PTAL). If we’re good with mocking, I’ll update this PR accordingly.

cosmos/operators/_asynchronous/bigquery.py
pankajastro: Fixed

pankajastro:
Since we agreed to use mocking in benchmarking, I removed the changes from this PR that allowed the user to set the deferrable value to false.

tatiana:

PR #2462, opened 2026-03-14
[Draft] Style guide edits

lzdanski:
## Description

This PR updates all the docs following the lightweight style guide proposed in https://github.com/astronomer/astronomer-cosmos/pull/2460. 

The main changes were:

- Adding Apache Airflow registered trademark
- Making header styles consistent across docs
- Make capitalization of Docker/docker consistent across docs
- Updating *Dag* spelling to match Airflow community
- Fixed currently published duplicated content on https://astronomer.github.io/astronomer-cosmos/policy/contributing.html#setting-up-the-cosmos-development-environment.

Even if this PR is not merged due to its size, in future you can use it as context for AI agents to create and apply style guide changes with higher accuracy.

## Related Issue(s)

N/A

## Breaking Change?

None found locally.

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

Build checked locally for page rendering issues.

PR #2460, opened 2026-03-13
[Docs] Contributing docs - additional resources

lzdanski:
## Description

This PR provides the following:

- Documentation about the Cosmos Docs-specific Sphinx styling and architecture conventions.
- Docs templates that can be adapted for helping get started with writing new content, or, adapted into AI context for facilitating LLM drafting or review (Index, Procedure, and Reference templates).

## Breaking Change?

None expected

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

Checked build and links locally.

PR #2623, opened 2026-04-30
Docs: capitalize Docker as a product name in prose

pankajkoti:
## Description

Per the lightweight style guide proposed in #2460, capitalize `Docker` when it
refers to the product (Docker image, Docker daemon, Docker container, Docker
socket, Docker operators) in docs prose. Literal CLI invocations such as
`docker build` are wrapped in double backticks and kept lowercase.

This re-applies the `docker → Docker` portion of #2462 against current `main` as
a focused, standalone change.

## Related Issue(s)

Refs #2460

## Breaking Change?

No. Documentation only.

## Checklist

### DIFF
--- a/docs/guides/run_dbt/container/azure-container-instance.rst
+++ b/docs/guides/run_dbt/container/azure-container-instance.rst
@@ -74,7 +74,7 @@ In order to run a container in Azure Container Instance, it needs access to the 
 
 **Build the dbt Docker image**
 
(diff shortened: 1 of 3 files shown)


tatiana (approved):
Thanks for improving these, @pankajkoti ! It would be great if we could have this type of improvement via a pre-commit check - so we avoid them in the first place.

PR #2626, opened 2026-05-03
`fix/issue-1206`: Adding support for `extra__google_cloud_platform__project` in `GoogleCloudServiceAccountDictProfileMapping`

### LINKED ISSUE #1206: [Bug] Bad behavior for `GoogleCloudServiceAccountDictProfileMapping` on Astronomer when using a keyfile
### Astronomer Cosmos Version

Other Astronomer Cosmos version (please specify below)

### If "Other Astronomer Cosmos version" selected, which one?

1.6.0

### dbt-core version

1.8.6

### Versions of dbt adapters

```
dbt-adapters==1.4.1
dbt-bigquery==1.8.2
```

### LoadMode

DBT_LS

### ExecutionMode

LOCAL

### InvocationMode

None

### airflow version

2.10.1

### Operating System

macOS Sonoma (Version 14.5)

_No response_

### Deployment

Astronomer

### Deployment details

_No response_

### What happened?

The `GoogleCloudServiceAccountDictProfileMapping` performs the following mapping, from the Airflow connection field name to the dbt field name, according to the [source code](https://github.com/astronomer/astronomer-cosmos/blob/b80ab6a2b7a7bb329dbbd777923584640047616a/cosmos/profiles/bigquery/service_account_keyfile_dict.py#L34-L40):
* `extra.project` to `project`
* `extra.dataset` to `dataset`
* `extra.keyfile_dict` or `keyfile_dict` or `extra__google_cloud_platform__keyfile_dict` to `keyfile_json`.

The problem is that on Astronomer, when we create a Google Cloud Platform connection using the keyfile variant, the connection fields are `extra__google_cloud_platform__project` and `extra__google_cloud_platform__keyfile_dict`. Causing a dbt run to fail as the generate profile does not have a `project` field. Here are the logs printing the Airflow connection fields as a YAML object:

```
[2024-09-17, 02:16:32 UTC] {logging_mixin.py:190} INFO - conn_type: google_cloud_platform
extra:
  extra__google_cloud_platform__keyfile_dict: '{   "type": "service_account",   "project_id":
    "my-project-id",   "private_key_id": ...,   "private_key": ...,   "client_email": "prj-cross-brand-test-bq@prj-cross-brand-test.iam.gserviceaccount.com",   "client_id":
    "116865537715351849116",   "auth_uri": "https://accounts.google.com/o/oauth2/auth",   "token_uri":
    "https://oauth2.googleapis.com/token",   "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",   "client_x509_cert_url": ...,   "universe_domain":
    "googleapis.com" }'
  extra__google_cloud_platform__project: my-project-id
```

I suggest to add the `extra__google_cloud_platform__project` connection field to the mapping for the `project` field in the dbt profile (I can submit a pull request for this change).

Note that this only applies to a Google Cloud Platform connection created from the Astro Environment Manager, as the `extra.project` field is present when creating a connection directly from the Airflow UI using the `google_cloud_platform` connection type. Here are the logs printing the Airflow connection fields as a YAML object when the connection is created directly from the Airflow UI:

```
[2024-09-17, 02:11:15 UTC] {logging_mixin.py:190} INFO - conn_type: google_cloud_platform
description: ''
extra:
  is_anonymous: false
  keyfile_dict: '{   "type": "service_account",   "project_id": "my-project-id",   "private_key_id":
    ...,   "client_email": ...,   "client_id":
    ...,   "auth_uri": "https://accounts.google.com/o/oauth2/auth",   "token_uri":
    "https://oauth2.googleapis.com/token",   "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",   "client_x509_cert_url":
    ...,   "universe_domain":
    "googleapis.com" }'
  num_retries: 5
  project: my-project-id
host: ''
login: ''
schema: ''
```

### Relevant log output

_No response_

### How to reproduce

1. Create a Google Cloud Platform connection (named `google_cloud_default`) using the keyfile variant on Astronomer, and link a Deployment to it:
   image
   image
2. Create a `DbtDag` using the following `profile_config`:

   ```python
       profile_name="your_profile_name",
       target_name="dev",
           profile_args={
               "dataset": "airflow",
           },
       ),
   )
   ```

### Anything else :)?

_No response_

### Contact Details

_No response_

jroachgolf84:
## Description

Per #1206, the following exception would occur when attempting to use the `GoogleCloudServiceAccountDictProfileMapping` Connection:

> The problem is that on Astronomer, when we create a Google Cloud Platform connection using the keyfile variant, the connection fields are `extra__google_cloud_platform__project` and `extra__google_cloud_platform__keyfile_dict`, causing a dbt run to fail as the generated profile does not have a project field.

This PR addresses this by adding support for the `extra__google_cloud_platform__project` parameter.

## Testing

A single unit test has been added to test the handling of this new parameter. This can be run with the following command:

```bash
hatch run tests.py3.11-2.10-1.8:test-cov
```

## Related Issue(s)

closes: #1206

## Breaking Change?

## Checklist


pankajastro (commented):
@jroachgolf84 thanks for the PR. Do we need add similar changes on other Google profile mapping?

jroachgolf84:
@pankajastro - you're right, I updated to include the `dataset` mapping (adding `extra__google_cloud_platform__dataset` parameter).

cosmos/profiles/bigquery/service_account_keyfile_dict.py
```diff
@@ -32,9 +32,9 @@ class GoogleCloudServiceAccountDictProfileMapping(BaseProfileMapping):
     secret_fields = ["private_key_id", "private_key"]
 
     airflow_param_mapping = {
-        "project": "extra.project",
+        "project": ["extra.project", "extra__google_cloud_platform__project"],
         # multiple options for dataset because of older Airflow versions
```
pankajastro, line 36: **Scope question — sibling mappings have the same latent bug.**

`cosmos/profiles/bigquery/oauth.py:29-30` and `cosmos/profiles/bigquery/service_account_file.py:29-30` still hardcode `"project": "extra.project"` / `"dataset": "extra.dataset"`. Any Astronomer-style GCP connection used with those mappings will hit the same `extra__google_cloud_platform__*` issue this PR fixes here.

Worth either extending the same list-of-paths fix to those two files now (one-line change each, very low risk) or filing a follow-up issue so this doesn't get lost. The `keyfile_json` field on the next line already established this pattern; applying it consistently across all three BigQuery mappings would be the cleanest end state.

tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
```diff
@@ -50,6 +50,25 @@ def mock_bigquery_conn_with_dict(request):  # type: ignore
         yield conn
 
 
+@pytest.fixture(params=get_fixture_params())
+def mock_bigquery_conn_extra__gcp__with_dict(request):  # type: ignore
```
pankajastro: **Naming nit (cosmetic).** Double-underscore identifiers (`mock_bigquery_conn_extra__gcp__with_dict`, and `test_extra__gcp__project` / `test_extra__gcp__dataset` below) mirror the Airflow connection key shape, but they read awkwardly and brush against Python's name-mangling convention (`__name` inside a class).

Suggest: `mock_bigquery_conn_legacy_extra_keys` for the fixture and `test_profile_with_legacy_extra_keys` for the test. Same readability, no double underscores.

tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
```diff
@@ -50,6 +50,25 @@ def mock_bigquery_conn_with_dict(request):  # type: ignore
...
+    """
+    Mocks and returns an Airflow BigQuery connection with extra__google_cloud_platform__project.
```
pankajastro: Minor — the docstring says "with `extra__google_cloud_platform__project`" but the fixture sets both `…__project` *and* `…__dataset`. Consider: "Mocks an Airflow BigQuery connection that uses the legacy `extra__google_cloud_platform__*` extra keys for project and dataset."

tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
```diff
@@ -50,6 +50,25 @@ def mock_bigquery_conn_with_dict(request):  # type: ignore
...
+    """
+    extra = {
+        "extra__google_cloud_platform__dataset": "my_dataset",
+        "extra__google_cloud_platform__project": "my_project",  # Update key
```
pankajastro: **Optional — coverage gap on lookup precedence.**

The new tests prove the fallback works when *only* the legacy keys are present. They don't pin down what happens when *both* `extra.project` and `extra__google_cloud_platform__project` are set in `extra` — which is the regression risk if anyone ever reorders the list in `airflow_param_mapping`.

A cheap way to add this is to parametrize the fixture with two `extra` variants (legacy-only, both-present) and assert `project == "my_project"` in both cases. The current ordering means `extra.project` should win when both are present; locking that in via a test prevents an accidental reorder from silently flipping the precedence.

tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
```diff
@@ -87,6 +106,46 @@ def test_profile(mock_bigquery_conn_with_dict: Connection):
...
+        "keyfile_json": {
+            "type": "service_account",
+            "private_key_id": "{{ env_var('COSMOS_CONN_GOOGLE_CLOUD_PLATFORM_PRIVATE_KEY_ID') }}",
+            "private_key": "{{ env_var('COSMOS_CONN_GOOGLE_CLOUD_PLATFORM_PRIVATE_KEY') }}",
+        },
+    }
+    assert profile_mapping.profile == expected
+
+
+def test_extra__gcp__dataset(mock_bigquery_conn_extra__gcp__with_dict: Connection):
```
pankajastro: **Suggestion: merge with the test above.**

`test_extra__gcp__project` (line 109) and `test_extra__gcp__dataset` (this one) are functionally identical:
- same fixture
- same `GoogleCloudServiceAccountDictProfileMapping(...)` instantiation
- same `expected` dict
- same full-`profile`-equality assertion

Either test alone proves both `project` and `dataset` resolve correctly, because both fields appear in the `expected` dict and both assertions compare the *entire* profile. Collapsing into a single `test_profile_with_legacy_extra_keys` removes ~20 duplicated lines without losing any coverage.

Not a blocker — extra duplication isn't harmful — but it's a free simplification.

pankajastro (commented):
Review of the fix for #1206. The core change is correct and reuses the existing list-of-paths resolver pattern (see `cosmos/profiles/base.py:300-314`) that `keyfile_json` already uses. Backward compatibility is preserved by ordering `extra.project` before the legacy key.

A few suggestions inline — mostly test cleanup and a scope question about the sibling profile mappings. None are blockers; the production change itself is approvable.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of @pankajastro

jroachgolf84:
Apologies, just catching up on this one. I need to go back and address @pankajastro's feedback.

cosmos/profiles/bigquery/service_account_keyfile_dict.py
jroachgolf84: Addressed for `.../oauth.py` and `.../service_account_file.py`.

tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
jroachgolf84: Updated.

tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
jroachgolf84: Resolved.

tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
jroachgolf84: Resolved.

tests/profiles/bigquery/test_bq_service_account_keyfile_dict.py
jroachgolf84: NA for right now, the order of those lists are not going to change (as it stands).

jroachgolf84 (commented):
Feedback addressed!

pankajastro (approved):

PR #2629, opened 2026-05-04
Sanitize disallowed chars from xcom backup variable key

### LINKED ISSUE #2619: [Bug] Watcher XCom backup Variable key not sanitized for ':' and '+' in run_id
### Astronomer Cosmos Version
1.14.1

### Apache Airflow Version
3.x (reproduced on Astro Runtime 3.1-13)

### Python Version
3.12

### ExecutionMode
WATCHER

### Deployment
Astronomer

### What happened?

When `ExecutionMode.WATCHER` is used on Airflow 3, the new XCom backup mechanism (introduced in #2559, shipped in 1.14.1) writes an Airflow Variable per producer task using the helper `_xcom_backup_variable_key` in [`cosmos/operators/_watcher/xcom.py`](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/_watcher/xcom.py#L24-L30):

```python
def _xcom_backup_variable_key(dag_id, task_group_id, run_id):
    parts = [XCOM_BACKUP_VARIABLE_PREFIX, dag_id.replace(".", "___")]
    if task_group_id:
        parts.append(task_group_id.replace(".", "__"))
    parts.append(run_id.replace(".", "_"))
    return "__".join(parts)
```

The sanitizer only strips `.`. Airflow 3 default run_ids look like `scheduled__2026-04-27T11:00:00+00:00` (or `manual__2026-04-27T11:00:00+00:00` for manual triggers), so the resulting Variable name ends up with `:` and `+` in it.

Any secrets backend with strict naming rules then rejects the call. With GCP `CloudSecretManagerBackend` (Secret Manager IDs only allow `[A-Za-z0-9_-]`), every `Variable.get` / `Variable.set` from the watcher producer logs:

```
Google Cloud API Call Error (InvalidArgument):
Invalid secret ID
airflow-variables-<env>-cosmos_xcom_backup__<dag_id>__scheduled__2026-04-27T11:00:00+00:00
```

The same risk exists for any other strict-naming backend (e.g., AWS Secrets Manager has a similar restriction).

### Relevant log output

```
Google Cloud API Call Error (InvalidArgument): Invalid secret ID
airflow-variables-<env>-cosmos_xcom_backup__<dag_id>__scheduled__2026-04-27T11:00:00+00:00
```

### How to reproduce

1. Use Cosmos 1.14.1 on Airflow 3 with `ExecutionMode.WATCHER`.
2. Configure GCP Secret Manager (or any backend that restricts variable names to `[A-Za-z0-9_-]`) as the Airflow Variables backend.
3. Trigger the DAG (either on its schedule or manually). Both default run_id formats, `scheduled__<iso8601>+00:00` and `manual__<iso8601>+00:00`, contain `:` and `+`. The only way to dodge the bug is to pass an explicit `[A-Za-z0-9_-]`-safe `run_id` when triggering.
4. Observe the watcher producer logs flooded with `InvalidArgument` errors on every `Variable.get` / `Variable.set`.

### Suggested fix

Sanitize the Variable key against the strictest common backend constraint, `[A-Za-z0-9_-]`, by replacing any other character with `_` or another predefined replacement. This should be applied in:

1. `_xcom_backup_variable_key` in `cosmos/operators/_watcher/xcom.py` (the actual reported bug).
2. As defense-in-depth, `_create_cache_identifier` in `cosmos/cache.py` (used to build `cosmos_cache_*` Variable names). It does not embed `run_id` today, but applies the same `replace(".", "_")` pattern and would regress identically if a future change widened its inputs.

A shared helper would keep the two call sites consistent.

### Anything else

The default Airflow run_id format for both scheduled (`scheduled__<iso8601>+00:00`) and manual (`manual__<iso8601>+00:00`) runs contains `:` and `+`, so any deployment using a strict-naming secrets backend as the Variables backend will hit this on every watcher producer execution.

MichaelRBlack:
## Summary

Cosmos's `WATCHER_KUBERNETES` execution mode persists XCom backups under Airflow Variables keyed `cosmos_xcom_backup__<dag>__<run_id>` ([cosmos/operators/_watcher/xcom.py#L24-L30](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/_watcher/xcom.py#L24-L30)). The existing key generator only replaces `.` with underscores. Airflow run IDs routinely contain `:` (timestamps) and `+` (timezone offsets) — e.g. `scheduled__2026-05-04T10:15:00+00:00` — which AWS Secrets Manager rejects with:

```
ClientError: An error occurred (ValidationException) when calling the GetSecretValue operation:
Invalid name. Must be a valid name containing alphanumeric characters,
or any of the following: -/_+=.@!
```

(GCP Secret Manager is even stricter — `[A-Za-z0-9_-]` only.)

When an external secrets backend is configured, Airflow's SDK walks the backend chain on `Variable.set` as well as `Variable.get` (`airflow.sdk.execution_time.context._set_variable` invokes `get_variable` on each backend). So every `safe_xcom_push` triggered from `_persist_backup` emits a `ValidationException` from the secrets backend. The Variable does ultimately get persisted via the metadata DB backend, but the noise spams task logs once per dbt model state transition. For a large model run that's hundreds of false-positive `ERROR Unable to retrieve variable from secrets backend (SecretsManagerBackend)` lines, which obscures real failures.

## Fix

Normalize all key components to `[A-Za-z0-9_-]` — the most restrictive common subset across the secrets backends Airflow ships with — so the resulting key round-trips through any backend without raising. A small `_sanitize_key_component` helper applies a `re.sub(r"[^A-Za-z0-9_-]", "_", ...)` on top of the existing component-specific period-replacement counts (3/2/1 underscores), preserving the prior visual format for keys that didn't contain disallowed chars.

## Reproducer

```python
from cosmos.operators._watcher.xcom import _xcom_backup_variable_key

# Before:
_xcom_backup_variable_key("dbt_daily", None, "scheduled__2026-05-04T10:15:00+00:00")
# 'cosmos_xcom_backup__dbt_daily__scheduled__2026-05-04T10:15:00+00:00'  ← AWS rejects

# After:
# 'cosmos_xcom_backup__dbt_daily__scheduled__2026-05-04T10_15_00_00_00'  ← accepted
```

## Tests

- New `TestXcomBackupVariableKey` class in `tests/operators/_watcher/test_state.py` with three cases:
  - `:` and `+` from a typical scheduled run_id are removed.
  - The existing `.` → `___`/`__`/`_` replacement is preserved (backwards-compat).
  - Parametrized check that the produced key contains only AWS-Secrets-Manager-safe characters across a range of dag_id / task_group_id / run_id combinations.
- Full file passes locally: `pytest tests/operators/_watcher/test_state.py` → 77 passed.

Closes: https://github.com/astronomer/astronomer-cosmos/issues/2619

### DIFF
--- a/cosmos/operators/_watcher/xcom.py
+++ b/cosmos/operators/_watcher/xcom.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import base64
 import json
+import re
 import zlib
 from typing import Any
 
@@ -20,14 +21,35 @@ logger = get_logger(__name__)
 
 XCOM_BACKUP_VARIABLE_PREFIX = "cosmos_xcom_backup"
 
+# Characters that secrets backends commonly reject in Variable keys. AWS
+# Secrets Manager allows alphanumerics + ``-/_+=.@!``; GCP Secret Manager is
+# stricter at ``[A-Za-z0-9_-]``. Run IDs routinely contain ``:`` and ``+``
+# (timestamps and timezone offsets, e.g. ``scheduled__2026-05-04T10:15:00+00:00``)
+# and dag/task-group IDs can contain ``.``. Sanitize all components down to
+# ``[A-Za-z0-9_-]`` so the resulting key is portable across backends and does
+# not log a ``ValidationException`` on every ``Variable.set`` call when an
+# external secrets backend is configured (Airflow walks the backend chain on
+# set as well as get).
+_DISALLOWED_VARIABLE_KEY_CHAR_RE = re.compile(r"[^A-Za-z0-9_-]")
+
 
 def _xcom_backup_variable_key(dag_id: str, task_group_id: str | None, run_id: str) -> str:
-    """Build a unique Airflow Variable key for the XCom backup of a watcher producer run."""
-    parts = [XCOM_BACKUP_VARIABLE_PREFIX, dag_id.replace(".", "___")]
+    """Build a unique Airflow Variable key for the XCom backup of a watcher producer run.
+
+    The component-specific period-replacement counts (3 underscores for dag_id,
+    2 for task_group_id, 1 for run_id) are preserved so keys remain visually
+    parseable, then any remaining disallowed character is normalized to ``_``.
+    """
+    parts = [XCOM_BACKUP_VARIABLE_PREFIX, _sanitize_key_component(dag_id.replace(".", "___"))]
     if task_group_id:
-        parts.append(task_group_id.replace(".", "__"))
-    parts.append(run_id.replace(".", "_"))
+        parts.append(_sanitize_key_component(task_group_id.replace(".", "__")))
+    parts.append(_sanitize_key_component(run_id.replace(".", "_")))
     return "__".join(parts)
+
+
+def _sanitize_key_component(value: str) -> str:
+    """Replace characters that secrets backends reject in Variable keys with ``_``."""
+    return _DISALLOWED_VARIABLE_KEY_CHAR_RE.sub("_", value)
 
 
 def _get_task_group_id(ti: Any) -> str | None:
--- a/tests/operators/_watcher/test_state.py
+++ b/tests/operators/_watcher/test_state.py
@@ -25,6 +25,7 @@ from cosmos.operators._watcher.xcom import (
     _init_xcom_backup,
     _persist_backup,
     _restore_xcom_from_variable,
+    _xcom_backup_variable_key,
 )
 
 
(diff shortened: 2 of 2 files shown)


tatiana (approved):
This looks great, @MichaelRBlack. Thank you very much for fixing https://github.com/astronomer/astronomer-cosmos/issues/2619!

We'll ship this as part of Cosmos 1.14.2.

tatiana:

Issue #1425, opened 2024-12-24
DBT


AkashSonu14:
def dag_dbt_sales_dw_cosmos_specific_DBTDag():
    start = EmptyOperator(task_id='start')

    transformed_data = DbtTask(
        task_id="transform",
        project_config=ProjectConfig(DBT_PROJECT_PATH),
        profile_config=profile_config,
        default_args={"retries": 2},     
        render_config=RenderConfig(
            select=["City_data","Set_data"],  
            exclude=["Set_data"],  
        ),
    )
    transformed_data_1 = DbtDag(
        task_id="transform_1",
        project_config=ProjectConfig(DBT_PROJECT_PATH),
        profile_config=profile_config,
        default_args={"retries": 2},     
        render_config=RenderConfig(
            select=["Order_status_data","Set_data"],  
            exclude=["Set_data"],  
        ),
        
    )

correct it like transform_1 should run only if transfrom task is passes

pankajkoti:
Hi @AkashSonu14, could you please update the title of this issue to something more descriptive? A title like “DBT” provides little context for the community to understand the issue at a glance. You can refer to the open issues here: https://github.com/astronomer/astronomer-cosmos/issues to see examples of how the community reports issues with clear titles and descriptions. Thank you!

PR #2636, opened 2026-05-05
feat: Fail when dbt command returns 'Nothing to do ...' 

### LINKED ISSUE #1435: [Feature] Fail DAG Execution When dbt Nodes or Paths Don’t Exist
### Description

Currently, when dbt-core is executed with a node name or path that doesn’t exist, it prints "nothing to do" and completes successfully. While this behavior makes sense from a program perspective, it can lead to issues in Cosmos workflows when nodes or paths are misspelled or incorrectly configured.

For instance, consider a dbt DAG structured as follows:
```
A
 \
  C
  /
 B
```
If there is a typo in the select command for B, the dbt task for B does nothing, but C continues to run and succeeds. This behavior makes it difficult to identify configuration errors, especially in scenarios where node or tag names change but are not updated consistently across the workflow.

We propose a feature in Cosmos to treat dbt nodes or paths that result in "nothing to do" as a failure. This would ensure that:
1. Misspelled or missing node names/paths are caught early
2. DAG executions fail explicitly when critical dependencies are not met, preventing downstream nodes from running

### Use case/motivation

This feature would help users:
- Catch errors arising from renaming or misconfiguring nodes/tags in dbt DAGs
- Improve reliability and accuracy in workflows where dbt nodes depend on specific upstream tasks. This feature would add robustness to Cosmos-managed dbt workflows by preventing silent errors that could lead to incorrect data processing or reporting

**Proposed Solution**
- Introduce an option in Cosmos to detect and fail tasks where dbt prints "nothing to do"
- Provide detailed error messages in the logs indicating the problematic node or path

### Related issues

_No response_

jroachgolf84:
## Description

Per issue #1425:

> Currently, when dbt-core is executed with a node name or path that doesn’t exist, it prints "nothing to do" and completes successfully. While this behavior makes sense from a program perspective, it can lead to issues in Cosmos workflows when nodes or paths are misspelled or incorrectly configured.
> 
> For instance, consider a dbt DAG structured as follows:
> 
> ```
> A
>  \
>   C
>   /
>  B
> ```
> 
> If there is a typo in the select command for B, the dbt task for B does nothing, but C continues to run and succeeds. This behavior makes it difficult to identify configuration errors, especially in scenarios where node or tag names change but are not updated consistently across the workflow.

This PR introduces the `fail_on_nothing_to_do` parameter in the `AbstractDbtLocalBase`. This value takes a default value of `False`, but if `True`, will fail execution of the Task if `Nothing to do` is present in the output.

### Unit Tests

There were five new unit-tests added to test the functionality implemented as part of this PR. These were:

- `test_handle_exception_dbt_runner_nothing_to_do`
- `test_handle_exception_dbt_runner_nothing_to_do_flag_off`
- `test_run_dbt_runner_nothing_to_do_callback`
- `test_handle_exception_subprocess_nothing_to_do`
- `test_handle_exception_subprocess_nothing_to_do_flag_off`

### E2E Testing

These changes were also tested E2E for both Airflow 2 and Airflow 3. A new DAG was added to `dev/failed` DAGs. The code snippet below points to `stage_customers` rather than `stg-customer`. This results in a "Nothing to do ..." message. Since `fail_on_nothing_to_do=True`, the Task and DAG fail.

This was also tested with `fail_on_nothing_to_do=False`. When this was the case, the Task and DAG succeeded.

```python

...

    stage_customers = DbtRunLocalOperator(
        project_dir=DBT_PROJ_DIR,
        task_id="run",
        dbt_cmd_flags=["--models", "stage_customers"],  # Should really be stg_customers
        install_deps=True,
        append_env=True,
        fail_on_nothing_to_do=True,
    )

...

```

## Related Issue(s)

closes: #1435

## Breaking Change?

## Checklist


jroachgolf84:
@tatiana - ready for review!

jroachgolf84:
cc: @tatiana - just gently bumping this one :)

cosmos/operators/local.py
```diff
@@ -204,6 +206,7 @@ def __init__(
         callback_args: dict[str, Any] | None = None,
         should_store_compiled_sql: bool = True,
         should_upload_compiled_sql: bool = False,
+        fail_on_nothing_to_do: bool = False,
```
tatiana, line 209: Could we move `fail_on_nothing_to_do` to the end of the existing parameters, after `dbt_runner_callbacks` and before `**kwargs`? Without this, the existing code that passes
  `append_env` positionally would now bind that value to `fail_on_nothing_to_do`

cosmos/operators/local.py
```diff
@@ -178,6 +179,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
     :param target_name: A name to use for the dbt target. If not provided, and no target is found
         in your project's dbt_project.yml, "cosmos_target" is used.
     :param should_store_compiled_sql: If true, store the compiled SQL in the compiled_sql rendered template.
+    :param fail_on_nothing_to_do: If true, exception will be thrown when dbt output includes "Nothing to do ...".
```
tatiana, line 182: Please update the docs `docs/guides/run_dbt/operators/operator-args.rst`, similar to how we documented `append_env` and other operator args.

tatiana (commented):
@jroachgolf84 this looks great, thank you!

Could we also add a Cosmos config setting in `settings.py` for the default value of this option? For example, `[cosmos] fail_on_nothing_to_do` / `AIRFLOW__COSMOS__FAIL_ON_NOTHING_TO_DO`, defaulting to `False`.

That would let users opt into this behavior across all applicable Cosmos local operators in one place, while still allowing an explicit operator arg like `fail_on_nothing_to_do=False` or `True` to override the global default per task.

Could we also update the docs?

if you can address this feedback today, we can make this part of Cosmo 1.15.0 release, otherwise we'll ship in Cosmos 1.16.0

PR #2641, opened 2026-05-06
Docs: normalize heading underlines in getting_started/

pankajkoti:
## Description

Per the lightweight style guide proposed in #2460, RST heading underlines
should follow the canonical hierarchy:

- Page title: `=`
- H1: `~`
- H2: `+`
- H3: `^`

The pages under `docs/getting_started/` used a mix of `-`/`+`/`'` for those
levels. Remap them positionally per file (first level encountered becomes `=`,
second becomes `~`, etc.) so the rendered hierarchy matches the convention.
Underline lengths and heading text are unchanged; only the underline character
is swapped.

This is the first of several focused PRs covering the underline-character
slice of the header commit in #2462. Subsequent PRs will cover other
subdirectories under `docs/`.

## Related Issue(s)

Refs #2460

## Breaking Change?

No. Documentation only.

## Checklist

### DIFF
--- a/docs/getting_started/astro-cli-quickstart.rst
+++ b/docs/getting_started/astro-cli-quickstart.rst
@@ -3,7 +3,7 @@
 .. title:: Run Cosmos with the Astro CLI
 
 Run Cosmos with the Astro CLI
------------------------------
+=============================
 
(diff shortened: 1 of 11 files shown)


pankajkoti:
I have manually compared and confirmed the before-and-after and see no regression in the built docs.

pankajastro (approved):

PR #2648, opened 2026-05-07
Import DAG from airflow.sdk in cosmos/airflow/dag.py

pankajkoti:
Follow-up to #2644. That PR converted the DAG import in seven cosmos modules to prefer `airflow.sdk` and fall back to `airflow.models.dag`, but the module defining `DbtDag` (`cosmos/airflow/dag.py`) was missed and still triggers the Airflow 3.2 deprecation warning at import time.

This applies the same try/except pattern there.

### DIFF
--- a/cosmos/airflow/dag.py
+++ b/cosmos/airflow/dag.py
@@ -6,7 +6,10 @@ from __future__ import annotations
 
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


pankajkoti:
CI does not look happy with this, need to investigate later.

Issue #2598, opened 2026-04-23
Fix test retry handling in ExecutionMode.WATCHER


pankajkoti:
Test retries in ExecutionMode.WATCHER are broken in two ways:

**1. Manual retry (clearing the test sensor from the Airflow UI) or Airflow retries raises an exception**
When a user clears a DbtTestWatcherOperator task from the Airflow UI or when retries are set at the Airflow level, `try_number` increments, and the sensor enters `_fallback_to_non_watcher_run`, which explicitly raises
 `AirflowException`:
       Test re-execution is not yet supported in watcher mode.
       Tests for model '<model_unique_id>' cannot be retried.

 Unlike model/seed/snapshot consumers that fall back to local dbt run --select <model>, test sensors have no fallback path.

**2. Producer-internal dbt retry (#1978) does not fully cover test failures**
The producer-internal dbt retry mechanism added in #2584 retries failed nodes automatically within the same task execution. However, test failures are not fully handled:
- When a test fails during dbt build, the aggregated test status ("fail") is pushed to XCom
   immediately via `push_test_result_or_aggregate`. Unlike model/seed/snapshot failures, test
  failures bypass the `_pending_failures` deferred-push mechanism.
- The test consumer sensor (`DbtTestWatcherOperator`) may read this intermediate "fail" XCom
  and raise `AirflowException` before the producer's retry loop has a chance to run dbt retry
  and overwrite the status.
- `dbt retry` does re-run failed tests and `test_results_per_model` is cleared before each
  retry attempt, so the retry itself works correctly at the dbt level. But the consumer may
  have already failed by then.

**Suggested approach**
- Manual retry: Implement dbt test --select <model> fallback in _fallback_to_non_watcher_run for test sensors (remove the AirflowException guard).
- Internal retry: Pass the deferred-push mechanism into push_test_result_or_aggregate so that "fail" aggregations are held back when retries are configured, similar to how _push_or_defer_status defers model failures.

pankajkoti:
We addressed part 1 of this ticket via PR #2658 and part 2 is no longer needed since we decided to close and not introduce #2584

PR #2658, opened 2026-05-11
Fix test sensor retry in ExecutionMode.WATCHER

pankajkoti:
## Summary

- Removes the `AirflowException` guard in `BaseConsumerSensor._fallback_to_non_watcher_run` that blocked manual clears and Airflow-level retries on `DbtTestWatcherOperator`.
- Adds a test-sensor fallback that runs `dbt test --select <model>` locally for the watched model: `DbtTestWatcherOperator` now sets `base_cmd = ["test"]` and overrides `_fallback_to_non_watcher_run`.

Addresses part 1 of #2598. The producer-internal test-retry race condition (part 2) is unchanged and remains tracked there.

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -417,15 +417,9 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
         Reconstructs the dbt command by cloning the project and re-running the model
         with appropriate flags, while ensuring flags like `--select` or `--exclude` are excluded.
 
-        For test sensors, re-execution is not supported in watcher mode; retries are skipped.
+        Subclasses may override to issue a different dbt subcommand (e.g. ``dbt test``
+        for test sensors or ``dbt source freshness`` for source sensors).
         """
-        if self.is_test_sensor:
-            raise AirflowException(
-                f"Test re-execution is not yet supported in watcher mode. "
-                f"{self._resource_label} '{self.model_unique_id}' cannot be retried. "
-                f"A future release will add fallback to local test execution."
-            )
-
         if try_number > 1:
             logger.info(
                 f"Retry attempt #%s – Running model '%s' from project '%s' using {self.__class__.__name__}",
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -608,8 +608,8 @@ class DbtTestWatcherOperator(DbtConsumerWatcherSensor):  # type: ignore[misc]
                 self.project_dir,
             )
 
-        model_selector = self.model_unique_id.split(".")[-1]
+        model_selector = self.model_unique_id.split(".", 2)[2]
         cmd_flags = ["--select", model_selector]
         self.build_and_run_cmd(context, cmd_flags=cmd_flags)
-        logger.info("dbt test completed successfully on retry for model '%s'", self.model_unique_id)
+        logger.info("dbt test completed successfully for model '%s'", self.model_unique_id)
         return True
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -571,10 +571,45 @@ class DbtTestWatcherOperator(DbtConsumerWatcherSensor):  # type: ignore[misc]
 
     Deferral is fully supported: the ``WatcherTrigger`` receives
     ``is_test_sensor=True`` and polls the correct aggregated key.
+
+    On manual clear from the Airflow UI or Airflow-level retry, the sensor falls
+    back to running ``dbt test --select <model>`` locally for this specific model.
     """
 
     template_fields: tuple[str, ...] = DbtConsumerWatcherSensor.template_fields  # type: ignore[operator]
 
+    # Override DbtRunMixin.base_cmd so build_and_run_cmd issues `dbt test` in the
+    # fallback path. The normal sensor poll path does not shell out, so this is
+    # only consulted when _fallback_to_non_watcher_run runs.
+    base_cmd = ["test"]
+
     @property
     def is_test_sensor(self) -> bool:
+        return True
+
+    def _fallback_to_non_watcher_run(self, try_number: int, context: Context) -> bool:
+        """Run ``dbt test --select <model>`` locally to retry tests for this model.
+
+        Used when the test sensor is manually cleared from the Airflow UI or when
+        Airflow-level retries fire. Producer flags are intentionally not forwarded
+        because some of them (e.g. ``--full-refresh``) are not valid for ``dbt test``.
+        """
+        if try_number > 1:
+            logger.info(
+                "Retry attempt #%s – Running tests for model '%s' from project '%s'",
+                try_number - 1,
+                self.model_unique_id,
+                self.project_dir,
+            )
+        else:
+            logger.info(
+                "Falling back to running tests for model '%s' from project '%s'",
+                self.model_unique_id,
+                self.project_dir,
+            )
+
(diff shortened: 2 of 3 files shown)


pankajkoti:
Screenshot 2026-05-11 at 6 14 58 PM

cosmos/operators/watcher.py
```diff
@@ -571,10 +571,45 @@ class DbtTestWatcherOperator(DbtConsumerWatcherSensor):  # type: ignore[misc]
...
+                self.project_dir,
+            )
+        else:
+            logger.info(
+                "Falling back to running tests for model '%s' from project '%s'",
+                self.model_unique_id,
+                self.project_dir,
+            )
+
+        model_selector = self.model_unique_id.split(".")[-1]
```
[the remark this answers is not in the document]
pankajkoti: addressed

tests/operators/test_watcher.py
```diff
@@ -2234,16 +2234,44 @@ def test_poke_reads_correct_xcom_key(self):
...
 
-        with pytest.raises(AirflowException, match="Test re-execution is not yet supported"):
-            sensor.poke(context)
+        result = sensor.poke(context)
+
+        assert result is True
+        sensor.build_and_run_cmd.assert_called_once()
+        _, kwargs = sensor.build_and_run_cmd.call_args
+        assert kwargs["cmd_flags"] == ["--select", self.MODEL_UID.split(".")[-1]]
+
```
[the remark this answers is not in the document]
pankajkoti: addressed

cosmos/operators/watcher.py
```diff
@@ -571,10 +571,45 @@ class DbtTestWatcherOperator(DbtConsumerWatcherSensor):  # type: ignore[misc]
...
+        cmd_flags = ["--select", model_selector]
+        self.build_and_run_cmd(context, cmd_flags=cmd_flags)
+        logger.info("dbt test completed successfully on retry for model '%s'", self.model_unique_id)
```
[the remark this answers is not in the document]
pankajkoti: addressed

cosmos/operators/_watcher/base.py
```diff
@@ -417,15 +417,9 @@ def _fallback_to_non_watcher_run(self, try_number: int, context: Context) -> boo
         Reconstructs the dbt command by cloning the project and re-running the model
         with appropriate flags, while ensuring flags like `--select` or `--exclude` are excluded.
 
-        For test sensors, re-execution is not supported in watcher mode; retries are skipped.
+        Subclasses may override to issue a different dbt subcommand (e.g. ``dbt test``
+        for test sensors or ``dbt source freshness`` for source sensors).
         """
```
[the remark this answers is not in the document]
pankajkoti: Tracked separately in #2659 to keep this PR focused on the test sensor retry fallback. The base.py selector bug affects run/seed/snapshot fallback paths as well, so it's worth its own PR with a versioned-model regression test.

cosmos/operators/watcher.py
```diff
@@ -571,10 +571,45 @@ class DbtTestWatcherOperator(DbtConsumerWatcherSensor):  # type: ignore[misc]
...
+    On manual clear from the Airflow UI or Airflow-level retry, the sensor falls
+    back to running ``dbt test --select <model>`` locally for this specific model.
     """
 
     template_fields: tuple[str, ...] = DbtConsumerWatcherSensor.template_fields  # type: ignore[operator]
 
+    # Override DbtRunMixin.base_cmd so build_and_run_cmd issues `dbt test` in the
+    # fallback path. The normal sensor poll path does not shell out, so this is
+    # only consulted when _fallback_to_non_watcher_run runs.
+    base_cmd = ["test"]
```
tatiana, line 589: @pankajkoti, did you consider inheriting from `DbtTestMixin`, instead of hard-coding this here?

cosmos/operators/watcher.py
```diff
@@ -571,10 +571,45 @@ class DbtTestWatcherOperator(DbtConsumerWatcherSensor):  # type: ignore[misc]
...
+                try_number - 1,
+                self.model_unique_id,
+                self.project_dir,
+            )
+        else:
+            logger.info(
+                "Falling back to running tests for model '%s' from project '%s'",
+                self.model_unique_id,
+                self.project_dir,
+            )
```
tatiana: Why is it relevant to log a different log message if `try_number > 1`? Could we have a single log message that includes the try-number and handles the rendering of this information by itself, without the need for an `if`?

tests/operators/test_watcher.py
```diff
@@ -2234,16 +2234,83 @@ def test_poke_reads_correct_xcom_key(self):
...
+        sensor.build_and_run_cmd = MagicMock()
+        ti = MagicMock()
+        context = self.make_context(ti)
+
+        with caplog.at_level("INFO"):
+            result = sensor._fallback_to_non_watcher_run(try_number=1, context=context)
+
+        assert result is True
+        assert any("Falling back to running tests for model" in r.message for r in caplog.records)
+        assert not any("Retry attempt" in r.message for r in caplog.records)
```
tatiana: How representative are these tests? They are relying on mocking and manually calling `_fallback_to_non_watcher_run`. Ideally, we'd have a test that confirms that during test retries, we're invoking `_fallback_to_non_watcher_run` - without us having to call it.

tatiana (approved):
Hi @pankajkoti , thanks for fixing this behaviour. Retries are the most complicated aspect of the watcher mode, and this helps stabilise it. I left some inline comments. Please address them before merging this PR.

cosmos/operators/watcher.py
pankajkoti: Yes, the problem is that the DbtTestMixin https://github.com/astronomer/astronomer-cosmos/blob/c23411b2b8a198849e26daf7abf7d338f1dc5733/cosmos/operators/base.py#L490 sends those arguments select, exclude, selector in super.init, which are not accepted by the parent class DbtConsumerWatcherSensor. So if we have to extend the Mixin, it would mean we would have to overwrite the init method to only call init DbtConsumerWatcherSensor. I felt it was simpler to have this than having to deal with multiple Mixins and the sensor base class.

tatiana: That's fine, it may be worth adding a comment that, due to the multiple class inheritance and incompatible arguments, it was simpler to add this than to inherit from `DbtTestMixin`

cosmos/operators/watcher.py
pankajkoti: This logging is consistent with what we have in the `BaseConsumerSensor`'s `_fallback_to_non_watcher_run method`  https://github.com/astronomer/astronomer-cosmos/blame/c23411b2b8a198849e26daf7abf7d338f1dc5733/cosmos/operators/_watcher/base.py#L429 added via PR https://github.com/astronomer/astronomer-cosmos/pull/2559.

pankajkoti: But I think it may be worth having just a single log, so I will make this change here. Not sure why we had that in the base sensor, maybe it's better that we make it consistent there too to have a single log line.

tests/operators/test_watcher.py
pankajkoti: Yes, makes sense, refactored the tests.

cosmos/operators/watcher.py
pankajkoti: done so, yes.

pankajkoti:
Going ahead with the merge here @tatiana . I have addressed the comments. Please let me know if you have further comments, and happy to address those in a follow-up PR.

tatiana:

PR #2659, opened 2026-05-11
Fix watcher fallback selector for versioned dbt models

pankajkoti:
## Summary

- `BaseConsumerSensor._fallback_to_non_watcher_run` used `model_unique_id.split(".")[-1]` to build the dbt `--select` value, which strips the version segment for versioned models (e.g. `model.pkg.my_model.v1` became `v1`).
- Switched to `split(".", 2)[2]`, matching `DbtNode.resource_name` in `cosmos/dbt/graph.py` — the canonical parsing used throughout Cosmos for dbt unique_ids (per the [dbt manifest spec](https://docs.getdbt.com/reference/artifacts/manifest-json#resource-details), `resource_type` and `package` cannot contain dots, so everything after the second dot is the full resource name including any version suffix).
- Added a regression test (`test_fallback_selector_preserves_version_suffix`) covering a versioned `unique_id`.

Surfaced by Copilot review on #2658.

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -447,7 +447,7 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
             raw_flags = upstream_task.add_cmd_flags()
             extra_flags = self._filter_flags(raw_flags)
 
-        model_selector = self.model_unique_id.split(".")[-1]
+        model_selector = self.model_unique_id.split(".", 2)[2]
         cmd_flags = extra_flags + ["--select", model_selector]
 
         self.build_and_run_cmd(context, cmd_flags=cmd_flags)  # type: ignore[attr-defined]
--- a/tests/operators/test_watcher.py
+++ b/tests/operators/test_watcher.py
@@ -1118,7 +1118,21 @@ class TestDbtConsumerWatcherSensor:
         sensor.build_and_run_cmd.assert_called_once()
         args, kwargs = sensor.build_and_run_cmd.call_args
         assert "--select" in kwargs["cmd_flags"]
-        assert MODEL_UNIQUE_ID.split(".")[-1] in kwargs["cmd_flags"]
+        assert MODEL_UNIQUE_ID.split(".", 2)[2] in kwargs["cmd_flags"]
+
+    def test_fallback_selector_preserves_version_suffix(self):
+        """For versioned dbt models (e.g. ``model.pkg.my_model.v1``) the selector must keep the version suffix."""
+        sensor = self.make_sensor()
+        sensor.model_unique_id = "model.jaffle_shop.stg_orders.v1"
+        ti = MagicMock()
+        ti.task.dag.get_task.return_value.add_cmd_flags.return_value = []
+        context = self.make_context(ti)
+        sensor.build_and_run_cmd = MagicMock()
+
+        sensor._fallback_to_non_watcher_run(2, context)
+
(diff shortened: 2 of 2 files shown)


cosmos/operators/_watcher/base.py
```diff
@@ -447,7 +447,7 @@ def _fallback_to_non_watcher_run(self, try_number: int, context: Context) -> boo
             raw_flags = upstream_task.add_cmd_flags()
             extra_flags = self._filter_flags(raw_flags)
 
-        model_selector = self.model_unique_id.split(".")[-1]
+        model_selector = self.model_unique_id.split(".", 2)[2]
```
tatiana, line 450: @pankajkoti, maybe instead of hard-coding index 2, we could retrieve the last index, similar to what we were doing:

```suggestion
        model_selector = self.model_unique_id.split(".", 2)[-1]
```

Do we have any official docs on how dbt forms `model_unique_id`? It could be worth mentioning what the variations are, that normally there would be two dots, but in some cases there are more, and why it is safe to group the last dots

tatiana (approved):
Thanks, @pankajkoti , this looks good, minor comment inline. The change doesn't block merging this PR, but I recommend adding a comment explaining the rationale. If you could double-check whether there are any other parts of the code where we're doing this type of handling, that would be great.

cosmos/operators/_watcher/base.py
pankajkoti: **On `[2]` vs `[-1]`**: I'd lean toward keeping `[2]`. For a well-formed `unique_id`, they return the same value; the difference is the failure mode. `[2]` raises `IndexError` and surfaces malformed input loudly, while `[-1]` would silently fall back to the `package` or `resource_type` and mis-issue the dbt command. It also matches the canonical parse already used in `DbtNode.resource_name` ([cosmos/dbt/graph.py](https://github.com/astronomer/astronomer-cosmos/blob/ad5e11c07797573b584cd6372846801c590912c6/cosmos/dbt/graph.py#L169)) and in cosmos/operators/watcher.py at lines [340](https://github.com/astronomer/astronomer-cosmos/blob/ad5e11c07797573b584cd6372846801c590912c6/cosmos/operators/watcher.py#L340), [546](https://github.com/astronomer/astronomer-cosmos/blob/ad5e11c07797573b584cd6372846801c590912c6/cosmos/operators/watcher.py#L546), and [613](https://github.com/astronomer/astronomer-cosmos/blob/ad5e11c07797573b584cd6372846801c590912c6/cosmos/operators/watcher.py#L613).

**On the docs ask**: per the [dbt manifest spec](https://docs.getdbt.com/reference/artifacts/manifest-json?version=1.12#resource-details), node unique_ids are `<resource_type>.<package>.<resource_name>`. Both `resource_type` and `package` are constrained identifiers that cannot contain dots, so the first two dots are unambiguous separators and everything after the second dot is the full resource name. For versioned models, dbt appends a fourth segment: `model.<package>.<resource_name>.<version>` (see [node_args.py](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/contracts/graph/node_args.py#L26C3-L31)).
`split(".", 2)[2]` returns the whole remainder (`my_model.v1`); `split(".")[-1]` returns just `v1`, which is the bug this PR fixes as mentioned in the PR description.

I would like to skip adding an inline docstring at this site because the parse is open-coded in multiple places across the codebase that are linked above. 

As we agreed, I am following up with a refactor [PR](https://github.com/astronomer/astronomer-cosmos/pull/2687/) that reuses a common method for these sites.

tatiana:

PR #2660, opened 2026-05-11
Add Snowflake profile mapping for non-encrypted private key file

### LINKED ISSUE #1281: [Feature] Support Snowflake Non-Encrypted Private Key Path
### Description

Using Cosmo's profiles, we have SnowflakeEncryptedPrivateKeyFilePem to support Private Key Files with a passphrase - but there is no option to support private keys without passphrase. 

### Use case/motivation

Create SnowflakePrivateKeyFilePem to support private key files without a passphrase.

### Related issues

_No response_

pankajastro:
## Summary
- Add `SnowflakePrivateKeyFilePemProfileMapping` for connections that authenticate with a private key file on disk **without** a passphrase. This was the one combination not covered by the existing four Snowflake profile mappings.
- The new mapping only claims a connection when neither `password` (the passphrase) nor `extra.private_key_content` is set, so the existing encrypted-file and content-based mappings continue to own their cases.
- Register the new class in `cosmos/profiles/__init__.py` (auto-discovery list + `__all__`) and re-export it from `cosmos/profiles/snowflake/__init__.py`. The docs page (`docs/reference/profiles/SnowflakePrivateKeyFilePem.rst`) is auto-generated by `docs/generate_mappings.py` at build time.

Closes #1281

## Test plan
- [x] New unit test file `tests/profiles/snowflake/test_snowflake_user_privatekey_file.py` covers:
  - Each required field missing → mapping does not claim
  - All required fields present, no passphrase, no content → mapping claims
  - Passphrase present → mapping defers (encrypted-file owns it)
  - `private_key_content` present → mapping defers (content mapping owns it)
  - `get_automatic_profile_mapping` dispatches to the new class for the unencrypted-file case
  - `get_automatic_profile_mapping` still dispatches to `SnowflakeEncryptedPrivateKeyFilePemProfileMapping` when a passphrase is set (regression guard)
  - `query_tag` propagation and the legacy `extra__snowflake__*` extras format
- [x] All 228 tests in `tests/profiles/` pass locally
- [x] `hatch run docs:build` succeeds with no warnings; `SnowflakePrivateKeyFilePem.html` is generated

### DIFF
--- a/cosmos/profiles/snowflake/user_privatekey_file.py
+++ b/cosmos/profiles/snowflake/user_privatekey_file.py
@@ -2,13 +2,12 @@
 
 from __future__ import annotations
 
-import json
-from typing import TYPE_CHECKING, Any
+import logging
+from typing import Any
 
 from cosmos.profiles.snowflake.base import SnowflakeBaseProfileMapping
 
-if TYPE_CHECKING:
-    from airflow.models import Connection
+logger = logging.getLogger(__name__)
 
 
 class SnowflakePrivateKeyFilePemProfileMapping(SnowflakeBaseProfileMapping):
@@ -20,7 +19,7 @@ class SnowflakePrivateKeyFilePemProfileMapping(SnowflakeBaseProfileMapping):
 
     airflow_connection_type: str = "snowflake"
     dbt_profile_type: str = "snowflake"
-    is_community: bool = True
+    is_community: bool = False
 
     required_fields = [
         "account",
@@ -45,32 +44,19 @@ class SnowflakePrivateKeyFilePemProfileMapping(SnowflakeBaseProfileMapping):
         result = super().can_claim_connection()
         if not result:
             return False
-        # Defer to the encrypted-file mapping when a passphrase is supplied.
+        # A passphrase is set: the connection is for an encrypted key. This mapping does not
+        # forward the passphrase to dbt, so claiming would yield a profile that fails at runtime.
+        # Refuse to claim and surface a warning so the user can switch to
+        # SnowflakeEncryptedPrivateKeyFilePemProfileMapping.
         if self.conn.password:
-            return False
-        # Defer to the content mappings when private_key_content is supplied.
-        if self.conn.extra_dejson.get("private_key_content") is not None:
+            logger.warning(
+                "%s will not claim connection %s: a passphrase is set on the Airflow connection. "
+                "Use SnowflakeEncryptedPrivateKeyFilePemProfileMapping for passphrase-protected keys.",
+                self.__class__.__name__,
+                self.conn_id,
+            )
             return False
         return True
-
-    @property
-    def conn(self) -> Connection:
-        """
-        Snowflake can be odd because the fields used to be stored with keys in the format
-        'extra__snowflake__account', but now are stored as 'account'.
-
-        This standardizes the keys to be 'account', 'database', etc.
-        """
-        conn = super().conn
-
-        conn_dejson = conn.extra_dejson
-
-        if conn_dejson.get("extra__snowflake__account"):
-            conn_dejson = {key.replace("extra__snowflake__", ""): value for key, value in conn_dejson.items()}
-
-        conn.extra = json.dumps(conn_dejson)
-
-        return conn
 
     @property
     def profile(self) -> dict[str, Any | None]:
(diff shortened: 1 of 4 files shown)


cosmos/profiles/snowflake/user_privatekey_file.py
```diff
@@ -0,0 +1,79 @@
...
+class SnowflakePrivateKeyFilePemProfileMapping(SnowflakeBaseProfileMapping):
+    """
+    Maps Airflow Snowflake connections to dbt profiles if they use a user/private key path without a passphrase.
+    https://docs.getdbt.com/docs/core/connect-data-platform/snowflake-setup#key-pair-authentication
+    https://airflow.apache.org/docs/apache-airflow-providers-snowflake/stable/connections/snowflake.html
+    """
+
+    airflow_connection_type: str = "snowflake"
+    dbt_profile_type: str = "snowflake"
+    is_community: bool = True
```
pankajkoti: This has to be false?

cosmos/profiles/snowflake/user_privatekey_file.py
```diff
@@ -0,0 +1,79 @@
...
+        "role": "extra.role",
+        "private_key_path": "extra.private_key_file",
+        "query_tag": "extra.query_tag",
+    }
+
+    def can_claim_connection(self) -> bool:
+        result = super().can_claim_connection()
+        if not result:
+            return False
+        # Defer to the encrypted-file mapping when a passphrase is supplied.
```
pankajkoti: What do we mean by defer here? it won't automatically use the encrypted-file mapping when the passphrase is supplied, right? User has to take an action, right? Maybe instead good to log a warning message suggesting the same before returning False?

cosmos/profiles/snowflake/user_privatekey_file.py
```diff
@@ -0,0 +1,79 @@
...
+        if self.conn.password:
+            return False
+        # Defer to the content mappings when private_key_content is supplied.
```
pankajkoti: same question as above

cosmos/profiles/snowflake/user_privatekey_file.py
```diff
@@ -0,0 +1,79 @@
...
+            return False
+        return True
+
+    @property
+    def conn(self) -> Connection:
+        """
+        Snowflake can be odd because the fields used to be stored with keys in the format
+        'extra__snowflake__account', but now are stored as 'account'.
+
+        This standardizes the keys to be 'account', 'database', etc.
```
pankajkoti: Don't see how we standardise the database key here. Can we correct the docstring?

pankajkoti (approved):
Approving, but hoping the comments are considered before merging.

cosmos/profiles/snowflake/user_privatekey_file.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def can_claim_connection(self) -> bool:
+        result = super().can_claim_connection()
+        if not result:
+            return False
+        # Defer to the encrypted-file mapping when a passphrase is supplied.
+        if self.conn.password:
+            return False
+        # Defer to the content mappings when private_key_content is supplied.
+        if self.conn.extra_dejson.get("private_key_content") is not None:
```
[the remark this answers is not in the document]
pankajastro: Fixed

cosmos/profiles/snowflake/user_privatekey_file.py
pankajastro: yes, fixed

cosmos/profiles/snowflake/user_privatekey_file.py
pankajastro: > What do we mean by defer here? it won't automatically use the encrypted-file mapping when the passphrase is supplied, right?

when passphrase given you can't use this profile. added log

cosmos/profiles/snowflake/user_privatekey_file.py
pankajastro: clean it. this was to handle old provider when extra key prefix with `extra`

cosmos/profiles/snowflake/user_privatekey_file.py
pankajastro: clean this

PR #2668, opened 2026-05-12
Export Clickhouse and encrypted Snowflake mappings from cosmos.profiles

pankajastro:
ClickhouseUserPasswordProfileMapping and SnowflakeEncryptedPrivateKeyPemProfileMapping were imported at the module level and registered in profile_mappings, but missing from` __all__.` As a result, "from cosmos.profiles import *" did not pick them up


pankajkoti (approved):
Good for now, but maybe we could explore in a follow-up PR if these can be lazily imported like we do for the root `__init__.py` file

PR #2670, opened 2026-05-13
Use `cosmos.log.get_logger` consistently

### LINKED ISSUE #1157: Make logging consistent from Cosmos operators
While implementing #1079, we ended up using `logger.info` as opposed to `self.log.info` in a few parts of the code to overcome a testing issue where `caplog` was not able to capture logs that were generated using  `self.log.info`.

This ticket aims to make logging consistent from Cosmos operators. It may be solved by #1108

pankajastro:
The rest of the codebase (31 modules) instantiates loggers via cosmos.log.get_logger so messages carry the (astronomer-cosmos) prefix when rich_logging is enabled. A few outliers still used the stdlib logger directly, breaking that branding and bypassing scoped log-level configuration:

- cosmos/operators/_asynchronous/base.py: dropped an unused `log = logging.getLogger(__name__)` (no callers in-module or via imports).
- cosmos/profiles/snowflake/user_privatekey_file.py: switched `logger` to get_logger; it is used for the passphrase warning.
- cosmos/dbt/parser/output.py: replaced a module-level `logging.error(...)` call (which hit the root logger) with a proper `logger.error(...)` via a module-level get_logger instance.
- cosmos/plugin/airflow3.py: replaced four module-level `logging.exception(...)` root-logger calls similarly.

closes: https://github.com/astronomer/astronomer-cosmos/issues/1157

### DIFF
--- a/cosmos/dbt/parser/output.py
+++ b/cosmos/dbt/parser/output.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import logging
 import re
 from typing import TYPE_CHECKING
 
@@ -11,6 +10,9 @@ if TYPE_CHECKING:
 
 from cosmos import __version__ as cosmos_version  # type: ignore[attr-defined]
 from cosmos.hooks.subprocess import FullOutputSubprocessResult
+from cosmos.log import get_logger
+
+logger = get_logger(__name__)
 
 DBT_NO_TESTS_MSG = "Nothing to do"
 DBT_WARN_MSG = "WARN"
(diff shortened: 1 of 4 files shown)


tatiana (approved):
Thanks for improving this, @pankajastro ! Would it be worth updating our agents' `CLAUDE.md` with recommendations on how to instantiate log, so we avoid similar patterns in the future? Happy for this to be addressed in a follow-up PR.

pankajastro:
> Thanks for improving this, @pankajastro ! Would it be worth updating our agents' `CLAUDE.md` with recommendations on how to instantiate log, so we avoid similar patterns in the future? Happy for this to be addressed in a follow-up PR.

Created PR: https://github.com/astronomer/astronomer-cosmos/pull/2679

PR #2671, opened 2026-05-13
Promote inline watcher status collections to named constants

pankajastro:
Watcher mode already defines frozenset constants for dbt and Airflow state classifications (DBT_SUCCESS_STATUSES, DBT_FAILED_STATUSES, PRODUCER_TERMINAL_STATES, etc.) but a few hot paths still inlined tuples or set literals for the same idea, making the membership checks visually inconsistent and harder to reuse.

Add three named frozensets in cosmos/operators/_watcher/state.py and replace the inline literals at the three call sites:

- DBT_SOURCE_FRESHNESS_STALE_STATUSES = {"error", "warn"} Replaces the tuple at cosmos/operators/watcher.py:86 used to detect stale dbt sources for downstream skip propagation.

- DATASET_EMITTING_RESOURCE_TYPES = {"model", "seed", "snapshot"} Replaces the bare-string set at cosmos/operators/_watcher/base.py where the watcher decides whether to look up outlet URIs from the manifest for a node.

- PRODUCER_FINAL_STATES = {"failed", "success", "skipped"} Replaces the tuple in cosmos/operators/_watcher/triggerer.py used to early-exit the async poll loop when the producer task has finished. Kept as a proper subset of PRODUCER_TERMINAL_STATES to preserve existing behaviour (this commit does not change which states the trigger exits on).

### DIFF
--- a/cosmos/operators/_watcher/state.py
+++ b/cosmos/operators/_watcher/state.py
@@ -25,9 +25,19 @@ DBT_SUCCESS_STATUSES = frozenset({"success", "pass", "warn"})
 DBT_FAILED_STATUSES = frozenset({"failed", "fail", "error", "runtime error"})
 DBT_SKIPPED_STATUSES = frozenset({"skipped"})
 
+# dbt source freshness statuses that mark a source as stale and propagate skips downstream.
+DBT_SOURCE_FRESHNESS_STALE_STATUSES = frozenset({"error", "warn"})
+
+# dbt resource types whose nodes can emit datasets (i.e. produce outputs with outlet URIs).
+DATASET_EMITTING_RESOURCE_TYPES = frozenset({"model", "seed", "snapshot"})
+
 # Airflow task states that indicate the producer has finished and will not deliver any more XCom updates.
 # Used to decide whether a sensor retry should fall back to a non-watcher run or keep polling.
 PRODUCER_TERMINAL_STATES = frozenset({"success", "failed", "skipped", "upstream_failed", "removed"})
+
+# Airflow task states checked by the watcher trigger to know the producer task has reached its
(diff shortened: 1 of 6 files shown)


cosmos/operators/_watcher/state.py
```diff
@@ -25,10 +25,20 @@
 DBT_FAILED_STATUSES = frozenset({"failed", "fail", "error", "runtime error"})
 DBT_SKIPPED_STATUSES = frozenset({"skipped"})
 
+# dbt source freshness statuses that mark a source as stale and propagate skips downstream.
+DBT_SOURCE_FRESHNESS_STALE_STATUSES = frozenset({"error", "warn"})
+
+# dbt resource types whose nodes can emit datasets (i.e. produce outputs with outlet URIs).
+DATASET_EMITTING_RESOURCE_TYPES = frozenset({"model", "seed", "snapshot"})
```
pankajkoti: Are these 2 specific to Watcher mode, or can they be moved to the root constants.py file? Or is it that based on execution mode the values could be different? If the values are specific to Watcher should be include Watcher prefix for those? And even with it what is a good place for these constants? It is this or constants.py

pankajkoti (approved):
Like the idea and thanks for the refactor, but have a question in-line.

cosmos/operators/_watcher/state.py
pankajastro: There are redundancy for `{"model", "seed", "snapshot"}`. Fixed it.

pankajastro: > If the values are specific to Watcher should be include Watcher prefix for those? And even with it what is a good place for these constants? It is this or constants.py

There are other constant in state.py which does not prefix watcher so i think we may need to discuss this further.

tatiana (approved):
Nice cleanup, @pankajastro !

PR #2677, opened 2026-05-14
Add store-failures to after-all tests

### LINKED ISSUE #2243: [Feature] How to add the `--store-failures` parameter in the `TestBehavior.AFTER_ALL` scenario.
### Description

_No response_

### Use case/motivation

_No response_

### Related issues

_No response_

mzl2233:
This updates the generated `TestBehavior.AFTER_ALL` dbt test task so it includes `--store-failures` by default while preserving any existing dbt command flags supplied by the caller. I added focused coverage for both the default flag injection and the existing-flags case.

Verification: `python3 -m compileall cosmos/airflow/graph.py tests/airflow/test_graph.py` and `git diff --check` passed. The targeted pytest command could not run in this environment because `hatch` is not installed and the fallback pytest run is missing the Airflow dependency.

Closes #2243

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -217,6 +217,11 @@ def create_test_task_metadata(
         task_args["selector"] = render_config.selector
         task_args["exclude"] = _convert_list_to_str(render_config.exclude)
 
+        dbt_cmd_flags = list(task_args.get("dbt_cmd_flags") or [])
+        if render_config.test_behavior == TestBehavior.AFTER_ALL and "--store-failures" not in dbt_cmd_flags:
+            dbt_cmd_flags.append("--store-failures")
+            task_args["dbt_cmd_flags"] = dbt_cmd_flags
+
     if node:
         exclude_detached_tests_if_needed(node, task_args, detached_from_parent)
         _override_profile_if_needed(task_args, node.profile_config_to_override)
--- a/tests/airflow/test_graph.py
+++ b/tests/airflow/test_graph.py
@@ -2181,6 +2181,38 @@ class TestSelectExcludeAsStringsInOperators:
         assert metadata.arguments["exclude"] is None
         assert metadata.arguments["selector"] is None
 
+    def test_create_test_task_metadata_after_all_adds_store_failures(self):
+        """AFTER_ALL test tasks should pass --store-failures to dbt test."""
+        render_config = RenderConfig(
+            test_behavior=TestBehavior.AFTER_ALL,
+        )
+        metadata = create_test_task_metadata(
+            test_task_name="test_all",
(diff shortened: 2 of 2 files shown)


tatiana (commented):
Thanks for working on this, @mzl2233!
  
A couple of concerns with the current approach:
 
1. It overrides user intent. dbt offers multiple ways to control store_failures - the store_failures test config (https://docs.getdbt.com/reference/resource-configs/store_failures) in `dbt_project.yml`, per-test YAML, or a `store_failures_as` config. Unconditionally appending `--store-failures` whenever it's absent from `dbt_cmd_flags` will flip the behaviour on for users who deliberately didn't enable it (or who configured it selectively at the project level). The CLI flag takes precedence over the project config, so this isn't just additive.

2. It's scoped to `TestBehavior.AFTER_ALL` only. Users running `TestBehavior.AFTER_EACH` or `TestBehavior.BUILD` may have the same need, and there's no technical reason to limit the feature to a single test behaviour.

3. There's no off switch, and it can't be templated. Once it's hardcoded into `dbt_cmd_flags`, callers can't disable it short of post-processing the task args.

A safer design would be to make store_failures opt-in, mirroring how we already handle `full_refresh`:
- Add a `store_failures: bool | str = False` parameter to [DbtTestMixin](https://github.com/astronomer/astronomer-cosmos/blob/4867faff96404658de69515b69d515acd52cb1a9/cosmos/operators/base.py#L474) and [DbtBuildMixin](https://github.com/astronomer/astronomer-cosmos/blob/4867faff96404658de69515b69d515acd52cb1a9/cosmos/operators/base.py#L349) in `cosmos/operators/base.py`, consumed before `super().__init__(**kwargs)`.
- Add `"store_failures"` to each mixin's template_fields.
- Emit `--store-failures` from each mixin's `add_cmd_flags()` when truthy (using the same `to_boolean` handling as `full_refresh` for templated string values).
- The existing `kwargs.pop("full_refresh", None)` in `AbstractDbtBase.__init__`, similar to what is done for [full_refresh](https://github.com/astronomer/astronomer-cosmos/blob/4867faff96404658de69515b69d515acd52cb1a9/cosmos/operators/base.py#L169), so the kwarg is silently dropped for run/seed/snapshot/source/etc. where dbt doesn't accept it.

Users would then enable it via `operator_args={"store_failures": True}`, and it would apply cleanly across `AFTER_EACH`, `AFTER_ALL`, and `BUILD` without overriding their dbt project configuration.

Would you be open to reworking the PR along these lines? Happy to help if any of the wiring isn't obvious.

pankajkoti:
Hi @mzl2233, thanks for the PR. Following up to see if you'd have a chance to take a look at the suggestions from @tatiana above.

PR #2679, opened 2026-05-14
Document lazy-logging standard + consistent in CLAUDE.md

pankajastro:
Add a Python Coding Standards section so future contributors (human and Claude) follow the lazy logging convention recently applied across cosmos/. Eager formatting via f-strings, .format(), or % builds the message even when the level filter would drop the record, and breaks structured handlers that key on record.msg.


tatiana (approved):
Thanks, @pankajastro, for addressing the feedback so promptly!

tatiana:

PR #2680, opened 2026-05-14
Convert remaining f-string logger calls to lazy form

pankajastro:
Follow-up to #2672. The original sweep caught ~47 single-line logger.X(f"...") sites but missed two patterns:

1. Multi-line calls of the form

       logger.exception(
           f"..."
       )

   in cosmos/plugin/airflow3.py (three sites) and cosmos/dbt/parser/output.py (one site). The original regex required the f-string on the same line as the logger call.

2. self.log.X(f"...") calls in operators that inherit Airflow's LoggingMixin. These never matched the original logger.X(f"...") pattern even though they violate the lazy logging rule identically. Affects azure_container_instance, docker, kubernetes, aws_ecs, gcp_cloud_run_job, and virtualenv operators.

Also converts one eager percent-interpolation call in cosmos/cache.py ("%s" % repr(e)) that used %-formatting instead of an f-string and so was outside the scope of #2672.

No behaviour change.

### DIFF
--- a/cosmos/cache.py
+++ b/cosmos/cache.py
@@ -236,7 +236,7 @@ def patch_partial_parse_content(partial_parse_filepath: Path, project_path: Path
             # it may be due a race condition of multiple processes trying to read/write this file
             data = msgpack.unpack(f)
     except ValueError as e:
-        logger.info("Unable to patch the partial_parse.msgpack file due to %s" % repr(e))
+        logger.info("Unable to patch the partial_parse.msgpack file due to %r", e)
     else:
         for node in data["nodes"].values():
             expected_filepath = node.get("root_path")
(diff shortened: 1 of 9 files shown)


tatiana (approved):
@pankajastro thanks for addressing this. Please, could you adapt our `CLAUDE.md` or `AGENTS.md` to instruct agents to use logging with lazy form instead of f-string?

pankajastro:
> @pankajastro thanks for addressing this. Please, could you adapt our `CLAUDE.md` or `AGENTS.md` to instruct agents to use logging with lazy form instead of f-string?

We have added this claude.md in PR https://github.com/astronomer/astronomer-cosmos/pull/2679

PR #2681, opened 2026-05-14
Unify operator logging on self.log

### LINKED ISSUE #1157: Make logging consistent from Cosmos operators

pankajastro:
PR #1108 established the convention that operator code uses self.log (Airflow's LoggingMixin) so messages land in the per-task-instance log shown in the Airflow UI, while library code uses cosmos.log.get_logger for branding and scoped log levels. PR #1079 deviated from that convention in three call sites because pytest's caplog could not capture self.log at the time; #1157 was filed to track cleaning up the workaround.

Current pytest/Airflow does capture self.log -- the neighbouring assertions in tests/operators/test_virtualenv.py already match self.log calls in the same file. Convert the three holdover sites to self.log so that:

- aws_ecs.py and gcp_cloud_run_job.py: the ECS/Cloud Run task result is emitted into the same task log as the "Running command" line above it instead of disappearing to worker stdout.
- virtualenv.py: the "Waiting for virtualenv lock" message routes to the task log alongside the other lock messages in the same method.

With the conversion the module-level `logger = get_logger(__name__)` and its import become dead in all three files; drop them.

test_run_command_with_virtualenv_dir, which asserts caplog.text.count("Waiting for virtualenv lock to be released") == 2, still passes after the change, confirming caplog captures self.log here.

closes #1157

### DIFF
--- a/cosmos/operators/aws_ecs.py
+++ b/cosmos/operators/aws_ecs.py
@@ -11,7 +11,6 @@ if TYPE_CHECKING:  # pragma: no cover
         from airflow.utils.context import Context  # type: ignore[attr-defined]
 
 from cosmos.config import ProfileConfig
-from cosmos.log import get_logger
 from cosmos.operators.base import (
     AbstractDbtBase,
     DbtBuildMixin,
@@ -23,8 +22,6 @@ from cosmos.operators.base import (
     DbtSourceMixin,
     DbtTestMixin,
 )
-
-logger = get_logger(__name__)
 
 DEFAULT_CONN_ID = "aws_default"
 DEFAULT_CONTAINER_NAME = "dbt"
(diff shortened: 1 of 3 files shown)


pankajkoti (approved):
Do we feel confident enough about the changes in aws_ecs.py and gcp_cloud_run_job.py? I guess we don't have dedicated tests for those, right? Happy to merge if you feel confident/were able to test it.

pankajastro:
Confident, yes. No log-asserting tests like the virtualenv change has, but `test_dbt_aws_ecs_build_and_run_cmd` and `test_dbt_gcp_cloud_run_job_build_and_run_cmd` (+ interceptor tests) do exercise the changed `build_and_run_cmd` path — re-ran both files, 16 passed. It's just swapping the logging sink (`logger` → `self.log`, the standard LoggingMixin already used a line above), so only the routing changes. Good to merge.

PR #2683, opened 2026-05-14
Prevent watcher producers from colliding on one XCom-backup key

### LINKED ISSUE #2625: [Bug] _get_task_group_id() always returns None in Airflow 2, causing all producer watcher tasks to share one xcom-backup key → UniqueViolation on every model completion
### Astronomer Cosmos Version

1.14.1

1.11.2

### Versions of dbt adapters

dbt-adapters==1.22.10
dbt-bigquery==1.11.0

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

WATCHER

### InvocationMode

DBT_RUNNER

### airflow version

2.10.5

### Operating System

Debian GNU/Linux

_No response_

### Deployment

Google Cloud Composer

### Deployment details

_No response_

### What happened?

When a DAG contains multiple DbtTaskGroup instances using ExecutionMode.WATCHER (e.g. one per source system), all dbt_producer_watcher tasks running concurrently in the same DAG run generate the same xcom-backup variable key. They race on INSERT → UniqueViolation logged as an error on every model completion.

**Root cause**

_get_task_group_id() in _watcher/xcom.py:33-36:

```
def _get_task_group_id(ti: Any) -> str | None:
    task = getattr(ti, "task", None)
    return getattr(task, "task_group_id", None) if task else None  # always None in Airflow 2
```

In Airflow 2, tasks have no task_group_id attribute — the correct path is task.task_group.group_id. Because this always returns None, the discriminating segment is omitted from the key, and every producer task in the same DAG run shares the same key:

`cosmos_xcom_backup__{dag_id}__{run_id}   ← no task_group_id segment`

### Relevant log output

```shell
{watcher.py:268} ERROR - Error in dbt event callback:
(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "variable_key_uq"
DETAIL: Key (key)=(cosmos_xcom_backup__my_dag__scheduled__2026-05-01T15:00:00+00:00) already exists.
```

### How to reproduce

Minimal repro — two DbtTaskGroups with ExecutionMode.WATCHER in the same DAG, both running concurrently:

```
from airflow.models.dag import DAG
from cosmos import DbtTaskGroup
from cosmos.config import RenderConfig
from cosmos.constants import ExecutionMode
import pendulum

with DAG(dag_id="repro_cosmos_xcom_collision", schedule="@hourly",
         start_date=pendulum.datetime(2025, 1, 1), catchup=False) as dag:

    group_a = DbtTaskGroup(
        group_id="group_a",
        execution_config=ExecutionConfig(execution_mode=ExecutionMode.WATCHER),
        render_config=RenderConfig(select=["path:models/some_path_a"]),
        # ... project_config, profile_config
    )

    group_b = DbtTaskGroup(
        group_id="group_b",
        execution_config=ExecutionConfig(execution_mode=ExecutionMode.WATCHER),
        render_config=RenderConfig(select=["path:models/some_path_b"]),
        # ... project_config, profile_config
    )
```
Trigger conditions:

Both groups must start in the same DAG run (no dependency between them → parallel)
Each group must have at least one model complete while both producers are still running
Airflow must be running with multiple workers (not LocalExecutor with parallelism=1)
To observe the error without waiting for a scheduled run:

Trigger the DAG manually
Watch logs of either dbt_producer_watcher task
Error appears on the first model completion in whichever producer loses the INSERT race

### Anything else :)?

_No response_

### Contact Details

_No response_

tatiana:
Closes #2625.
Closes BOSS-83.
Closes BOSS-157.

## The problem

When a DAG contained multiple `DbtTaskGroup`s using `ExecutionMode.WATCHER`, every `dbt_producer_watcher` in the same DAG run computed the **same** Airflow Variable key for its XCom backup:

```
cosmos_xcom_backup__{dag_id}__{run_id}
```

The task-group segment was supposed to discriminate concurrent producers, but the helper that read it (`_get_task_group_id` in `cosmos/operators/_watcher/xcom.py`) was reading `task.task_group_id` — an attribute that does not exist on Airflow operators in either Airflow 2 or Airflow 3 (the canonical path is `task.task_group.group_id`). Because the read went through `getattr(task, "task_group_id", None)`, it silently returned `None`; the segment was omitted from the key, and every producer in the DAG run wrote to the same Variable row.

The consequence depends on the Airflow version:

- **Airflow 2:** the second producer to insert raises `UniqueViolation` on `variable_key_uq` — a noisy `ERROR` on every model completion (this is the original symptom from the issue report).
- **Airflow 3:** `Variable.set` is an upsert (`ON CONFLICT DO UPDATE`), so the conflict is silently absorbed and the last writer wins. If one producer fails and retries, `_restore_xcom_from_variable` may load the **other** group's XCom snapshot — an undetected correctness bug.

## How to reproduce

`dev/failed_dags/example_watcher_xcom_collision.py` reproduces the issue end-to-end against the `jaffle_shop` project: two `DbtTaskGroup`s (`group_a` selecting `+customers`, `group_b` selecting `+orders`), no inter-group dependency, both running in parallel.

Reproduction conditions (from the issue report):
- both groups start in the same DAG run (no dependency between them → parallel)
- each group must have at least one model complete while both producers run
- Airflow must be running with multiple workers (not LocalExecutor with `parallelism=1`)

Trigger the DAG manually; on the first node-status XCom in whichever producer loses the INSERT race, you see — before this fix:

- on Airflow 2:
  ```
  {watcher.py:268} ERROR - Error in dbt event callback:
  (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "variable_key_uq"
  DETAIL: Key (key)=(cosmos_xcom_backup__example_watcher_xcom_collision__manual__...) already exists.
  ```
- on Airflow 3: nothing in the logs (silent overwrite of the backup row).

After this fix the two producers compute distinct keys (`…__group_a__…` and `…__group_b__…`) and neither symptom occurs.

## The fix

`_get_task_group_id` now reads `task.task_group.group_id` — the same path that `_get_dbt_dag_task_group_identifier` in `cosmos/airflow/graph.py` already uses. `.task_group`/`.group_id` are accessed directly (not via `getattr` with a `None` default), so a future rename or typo at this site raises `AttributeError` at first invocation instead of silently degrading to the colliding-key behaviour. The outer `getattr(ti, "task", None)` is kept because the runtime-TI shape varies legitimately between `TaskInstance`, `RuntimeTaskInstance`, and test stubs.

## How do we prevent this from happening again

Three independent reasons let the bug slip through, addressed in three independent ways:

1. **The unit test masked the bug with `MagicMock`.** The previous test did `ti.task = MagicMock(task_group_id="my_group")`. Because `MagicMock` fabricates any attribute access, `getattr(mock, "task_group_id", None)` returned `"my_group"` and the test passed — even though the same call against a real operator always returned `None`. The test (`tests/operators/_watcher/test_state.py::TestInitXcomBackup`) is reworked to mock `task.task_group.group_id`, and a new test `test_rejects_task_without_task_group_attribute` uses `MagicMock(spec=[...])` to forbid fabrication and asserts the helper raises loudly when fed a malformed task.

2. **No test asserted uniqueness across task groups.** A new integration test `test_dbt_task_groups_with_watcher_produce_distinct_xcom_backup_keys` in `tests/operators/test_watcher.py` builds two `DbtTaskGroup`s in one DAG, drives `_init_xcom_backup` against each real producer operator (exercising the actual code path the bug lived in), and asserts the resulting backup keys differ — and contain `group_a`/`group_b`. This test fails on `main` with the same colliding key the user sees in production, and passes after the helper fix.

3. **Direct attribute access replaces silent-fallback `getattr`** on the contract attributes (`task.task_group`, `.group_id`). A future typo or rename at the failure site now surfaces as `AttributeError` at first invocation instead of degrading to a colliding key. This is the safer failure mode the original `getattr` chain lacked.


cosmos/operators/_watcher/xcom.py
```diff
@@ -53,9 +53,27 @@ def _sanitize_key_component(value: str) -> str:
...
+    """
     task = getattr(ti, "task", None)
-    return getattr(task, "task_group_id", None) if task else None
+    if task is None:
+        return None
+    # ``task.task_group`` is part of every Airflow operator's contract (it is
+    # ``None`` for root-level tasks, a ``TaskGroup`` otherwise). Using direct
+    # attribute access here means a future rename or typo raises
+    # ``AttributeError`` at first invocation rather than silently degrading to
+    # the colliding-key behaviour we just fixed.
```
pankajastro: The docstring already covers the "direct access is intentional" rationale really well, so this inline comment slightly doubles up. The one piece that's worth keeping is the `task.task_group` contract for the next line. Maybe just:

```python
    # task.task_group is None for root-level tasks, a TaskGroup otherwise.
    task_group = task.task_group
    return task_group.group_id if task_group is not None else None
```

Pure cosmetic, take it or leave it.

pankajastro (approved):
Really clean fix — the root-cause analysis (silent `getattr` masking a non-existent attribute) and the three-layer regression strategy (real-attribute fix, `MagicMock(spec=…)` hardening, cross-group integration test) are excellent. Thanks for the thorough writeup!

One tiny doc nit: the PR description references `example_watcher_xcom_collision.py` as new, but it was removed in a follow-up commit and isn't in the final diff. Mind dropping the reference from the body so future readers don't go looking for it?

One cosmetic suggestion inline — nothing blocking.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) on behalf of @pankajastro

cosmos/operators/_watcher/xcom.py
```diff
@@ -53,9 +53,27 @@ def _sanitize_key_component(value: str) -> str:
...
+    the task-group segment from the XCom backup Variable key and causing every
+    watcher producer in the same DAG run to share one key. See
+    https://github.com/astronomer/astronomer-cosmos/issues/2625.
+    """
     task = getattr(ti, "task", None)
-    return getattr(task, "task_group_id", None) if task else None
+    if task is None:
+        return None
+    # ``task.task_group`` is part of every Airflow operator's contract (it is
+    # ``None`` for root-level tasks, a ``TaskGroup`` otherwise). Using direct
```
pankajkoti:
```suggestion
    # ``task.task_group`` is part of every Airflow operator's contract. Using direct
```
Can we double-check this? I guess it's always a `TaskGroup` instance when the operator is part of a DAG with/without a taskgroup? And I guess the `task.task_group.group_id` may be `None` for non-taskgroup tasks.

pankajkoti: I tried the below in Python 3.12 - Airflow 2.10 environment

```
from airflow import DAG
from datetime import datetime

with DAG('probe_dag', start_date=datetime(2026, 1, 1)) as dag:
    t = EmptyOperator(task_id='root_task')

print('task.task_group: ', t.task_group)
print('task.task_group is None: ', t.task_group is None)
print('task.task_group.group_id is None: ', t.task_group.group_id is None)
```


and see the output as:
```
task.task_group:  <airflow.utils.task_group.TaskGroup object at 0x10cd1c650>
task.task_group is None:  False
task.task_group.group_id is None:  True
```

cosmos/operators/_watcher/xcom.py
```diff
@@ -53,9 +53,27 @@ def _sanitize_key_component(value: str) -> str:
...
+    # attribute access here means a future rename or typo raises
+    # ``AttributeError`` at first invocation rather than silently degrading to
+    # the colliding-key behaviour we just fixed.
+    task_group = task.task_group
+    return task_group.group_id if task_group is not None else None
```
pankajkoti: Curious here in which cases `task_group` could be None

pankajkoti (approved):
Have a couple of questions in-line, but the fix for correcting the non-existent `task_group_id` -> `task.task_group.group_id` is a must, I agree.

cosmos/operators/_watcher/xcom.py
tatiana: Thanks, @pankajastro , addressed in 03cb3027e007e478db01e86159ebe6d1e7f73e6e

cosmos/operators/_watcher/xcom.py
tatiana: Thanks a lot, addressed in 03cb3027e007e478db01e86159ebe6d1e7f73e6e

cosmos/operators/_watcher/xcom.py
tatiana: Simplified and addressed in 03cb3027e007e478db01e86159ebe6d1e7f73e6e

tatiana:

PR #2684, opened 2026-05-14
Retry watcher downstream models on upstream-failure recovery

### LINKED ISSUE #2698: Watcher mode should retry downstream models skipped after upstream recovery
## Summary

In watcher mode, a producer dbt build can fail on an upstream model while dbt marks transitive downstream models as skipped. Cosmos currently stores those skipped statuses in XCom for the downstream consumer sensors. Because Airflow does not retry skipped tasks, a downstream model can remain unmaterialized even if the failed upstream model later succeeds through the existing consumer retry fallback. The DAG can then finish successfully while some downstream tables were never rebuilt.

## Reproduction shape

A minimal graph is enough to trigger the behavior:

- model_a succeeds.
- model_flaky depends on model_a and fails on the first producer attempt, then succeeds on retry or fallback.
- model_downstream depends on model_flaky.

On the first producer attempt, dbt marks model_downstream as skipped because its upstream failed. If Cosmos propagates that skipped status unchanged, the downstream consumer sensor raises AirflowSkipException and Airflow does not retry it.

## Expected behavior

Downstream models skipped specifically because of upstream-node failure should be eligible for retry after the upstream model recovers, so the DAG does not report success with missing downstream materializations.

## Proposed fix direction

When parsing dbt JSON logs in watcher mode, distinguish upstream-failure skip events from ordinary skip outcomes. For upstream-failure skips, store the downstream node status as failed rather than skipped so the Airflow consumer task can retry and use the existing fallback path.

The implementation should avoid changing behavior for skips caused by other dbt semantics, such as selector exclusions or empty models.

tatiana:
Closes #2698
Closes [BOSS-401](https://linear.app/astronomer/issue/BOSS-401/cosmos-watcher-mode-does-not-retry-downstream-models-upon-successful) (Customer issue).

> **Note:** This PR is stacked on top of #2682 (forward extra args to `test-integration`). The first commit on this branch is a cherry-pick of that change — drop it on rebase once #2682 merges.

## The problem

When a watcher producer's first `dbt build` attempt fails, dbt marks the failing model and every transitive downstream node with `node_status="skipped"`. The producer log parser pushed those `"skipped"` statuses straight to XCom, so each downstream consumer sensor raised `AirflowSkipException` and ended in SKIPPED. Airflow does not retry SKIPPED tasks — so even after the failing upstream recovered via its own consumer-retry fallback, the downstream model was never re-run.

The DAG completed `success` (Airflow treats SKIPPED as non-failure) with the downstream tables silently un-materialized. Luis Carbonell (Prodege) reported the "false green" symptom on `astronomer-cosmos==1.14.1` and `Airflow 3.2.1+astro.2`.

Reproduction setup:
- `model_a` → succeeds.
- `model_flaky` → fails on first attempt via a Postgres sequence pre-hook, succeeds on subsequent attempts.
- `model_downstream` → depends on `model_flaky`.

Without the fix, on retry:
- ✅ `model_flaky` recovers via the consumer's `_fallback_to_non_watcher_run` (existing retry path for FAILED consumers).
- ❌ `model_downstream` stays SKIPPED — Airflow never retries it.
- DAG: `success`.

## The fix

`store_dbt_resource_status_from_log` in `cosmos/operators/_watcher/base.py` now tracks unique_ids that dbt emits `SkippingDetails` (non-ephemeral upstream) or `LogSkipBecauseError` (ephemeral upstream) events for. Later `"skipped"` terminal events for those unique_ids are rewritten to `"failed"` so the downstream consumer fails on attempt 1, Airflow retries it, and the existing `_fallback_to_non_watcher_run` path runs the model locally once its upstream has recovered.

Why these two events are the right discriminator (traced through dbt source on `dbt-core==1.11`):

- `SkippingDetails` and `LogSkipBecauseError` are fired only from `BaseRunner.on_skip()` (`dbt/task/base.py:420-471`).
- `on_skip()` is reached only when `self.skip = True`.
- `self.skip = True` is set only inside `do_skip(cause=...)` (`dbt/task/base.py:474-476`) — verified by `grep -rn '\.skip = True'` returning exactly one site.
- `do_skip(cause=cause)` is called only from `runnable.py:358-360` when `runner.node.unique_id in self._skipped_children` — i.e. exclusively on upstream-node failure.

Empty, ephemeral, and selector-excluded skips don't go through `do_skip()` and don't fire these events, so they keep their existing `"skipped"` handling.

dbt also emits a later `NodeFinished` with `node_status="skipped"` for the same node from the runner's `finally` block (`dbt/task/runnable.py:266`), which would otherwise overwrite the rewritten XCom back to `"skipped"`. The per-execution accumulator set ensures both events are rewritten consistently regardless of arrival order.

The fix lives in the shared log parser, so it applies uniformly to both `InvocationMode.DBT_RUNNER` (protobuf events serialised via `MessageToJson`) and `InvocationMode.SUBPROCESS` (`--log-format json` stdout lines).

## Behaviour after this PR

| Attempt | producer | model_a | model_flaky | model_downstream |
|---|---|---|---|---|
| 1 | failed | success | failed | **failed** (was SKIPPED) |
| 2 (retry) | skipped (by design) | (done) | success via fallback | **success via fallback** |

## How we prevent this from regressing

- **New integration test** `test_dbt_task_group_watcher_retry_recovers_skipped_downstream` in `tests/operators/test_watcher.py` runs the new DAG via `dag.test()` and asserts:
  - `model_downstream_run.state == "success"` (was `skipped` before the fix).
  - `model_downstream_run.try_number > 1` — guards against a future refactor that happens to leave the task in `success` for the wrong reason.
- **New dbt project** `dev/dags/dbt/watcher_upstream_failure_recovery/` — kept separate from `watcher_downstream_not_skipped` so the two test scenarios don't conflate. Uses a distinct sequence name `_cosmos_recovery_fail_once_seq` so parallel CI runs can't share state.
- **Standalone repro DAG** `dev/failed_dags/example_watcher_recovers_skipped_downstream.py` for visual verification in Airflow standalone.
- **Refactor**: the fail-once-sequence cleanup is extracted into a `reset_fail_once_sequence` pytest fixture; both the new test and the existing `test_dbt_task_group_watcher_gateway_prevents_downstream_skip` now use it.

## Test plan

- [x] New integration test passes on Airflow 3.2 (`tests.py3.10-3.2-1.11:test-integration`, 109.65s call duration).
- [x] Existing `test_dbt_task_group_watcher_gateway_prevents_downstream_skip` still passes after the fixture refactor.
- [x] Manual repro via `example_watcher_recovers_skipped_downstream` in Airflow standalone: confirmed `model_downstream_run` lands in `success` (was `skipped`).
- [ ] CI passes on the full matrix.


cosmos/operators/_watcher/base.py
```diff
@@ -253,6 +254,33 @@ def _ensure_subprocess_model_outlet_uris(
...
+
+    dbt emits two events for a node that it skipped because of an upstream
+    failure: ``SkippingDetails``/``LogSkipBecauseError`` during ``on_skip()``,
+    and a later ``NodeFinished`` from the runnable ``finally``-block -- both
+    carry ``node_status="skipped"``. The first identifies the upstream-failure
+    cause (those events are reached only via ``do_skip(cause=...)``, exclusively
+    on upstream-node failure); the second would otherwise overwrite the
+    rewritten XCom with the original ``"skipped"``. Tracking affected
+    ``unique_id``\\ s in a per-execution set lets the parser rewrite both. See
+    Linear BOSS-401.
```
pankajkoti: How do we feel about linking private issues in the code that is OSS? Would it be a good idea to file a representative public issue and link here instead? If we agree would be nice to do it across the multiple occurrences of this identifier in the PR (I see there are about 9 of those)

cosmos/operators/_watcher/base.py
```diff
@@ -253,6 +254,33 @@ def _ensure_subprocess_model_outlet_uris(
...
+def _rewrite_upstream_failure_skip_status(
+    log_line: dict[str, Any],
+    unique_id: str | None,
+    dbt_node_status: Any,
+    upstream_failure_skipped_ids: set[str] | None,
+) -> Any:
+    """Track upstream-failure-skipped nodes and rewrite their status from "skipped" to "failed".
+
+    dbt emits two events for a node that it skipped because of an upstream
+    failure: ``SkippingDetails``/``LogSkipBecauseError`` during ``on_skip()``,
```
pankajkoti, line 266: Are these public dbt contracts? If internal, would it be possible to do a quick sanity check that these are consistent across the dbt versions we support? I guess for future releases we might be able to catch if these change when we test against newly released versions, but it would be nice to check for earlier versions once.

dev/dags/dbt/watcher_upstream_failure_recovery/dbt_project.yml
pankajkoti, line 1: Looks like we already have multiple watcher-related dbt projects: 
1. dev/dags/dbt/watcher_failing_tests
2. dev/dags/dbt/watcher_downstream_not_skipped


Would we instead prefer it be consolidated into an existing one? Happy for this to be considered in a subsequent PR if we agree to consolidate.

dev/failed_dags/example_watcher_recovers_skipped_downstream.py
```diff
@@ -0,0 +1,113 @@
+"""
+Demonstrate watcher-mode recovery of a downstream model that dbt skipped
+because its upstream failed on the producer's first attempt (BOSS-401).
+
+Without the fix in ``cosmos/operators/_watcher/base.py``:
+- A dbt model fails on the first producer attempt.
+- dbt marks every downstream node ``skipped`` with the upstream-failure cause.
```
pankajkoti, line 7: This applies to nested downstream nodes too, and the fix implementation covers that, right? 

Meaning for nodes A -> B -> C -> D, if A fails, all of B, C and D will get the upstream-failure cause, right?

pankajkoti (approved):
I was able to reproduce and validate that the fix solves the case with a DAG run with astro cli (I had only one downstream task for a flaky model, though, I didn't test further nested downstream tasks). Nice find on narrowing the event types for skip statuses tied to upstream failures.

Have some clarifying questions and a suggestion inline,

cosmos/operators/_watcher/base.py
tatiana: Addressed in https://github.com/astronomer/astronomer-cosmos/pull/2700/changes

cosmos/operators/_watcher/base.py
tatiana: Commit 025ed177 — every BOSS-401 reference (8 sites in source) replaced with #2698 (real public issue you filed).

dev/failed_dags/example_watcher_recovers_skipped_downstream.py
tatiana: Commit 4017c3ed — added model_downstream_2 (depends on model_downstream) to the dbt project,
    parametrized the integration test over both downstream task IDs so the regression catches every node along the
    chain. Verified passing on AF 3.2 × dbt 1.8/1.9/1.10/1.11.

Example of DAG run of the DAG included in the PR with a longer sequence of downstream tasks:
Screenshot 2026-05-19 at 12 09 42

dev/dags/dbt/watcher_upstream_failure_recovery/dbt_project.yml
tatiana: Created a follow up ticket in: https://github.com/astronomer/astronomer-cosmos/issues/2701

I believe we may want to keep a few variations, but I agree this deserves more thought

cosmos/operators/_watcher/base.py
```diff
@@ -253,6 +254,33 @@ def _ensure_subprocess_model_outlet_uris(
...
+    rewritten XCom with the original ``"skipped"``. Tracking affected
+    ``unique_id``\\ s in a per-execution set lets the parser rewrite both. See
+    #2698.
+    """
+    if not unique_id or upstream_failure_skipped_ids is None:
+        return dbt_node_status
+    if is_dbt_upstream_failure_skip_event(log_line.get("info", {}).get("name")):
+        upstream_failure_skipped_ids.add(unique_id)
+    if dbt_node_status == "skipped" and unique_id in upstream_failure_skipped_ids:
+        return "failed"
```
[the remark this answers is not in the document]
tatiana: In practice CPython's GIL makes both atomic, the only mutation is
  add (never remove), and a missed-read just degrades to the pre-fix behavior for one event — which the next event for that unique_id corrects. I'd reply
  inline that the race is benign and move on

cosmos/operators/_watcher/base.py
```diff
@@ -269,6 +297,7 @@ def store_dbt_resource_status_from_log(
     test_results_per_model: dict[str, list[str]] | None = None,
     model_outlet_uris: dict[str, list[str]] | None = None,
     dataset_namespace: str | None = None,
+    upstream_failure_skipped_ids: set[str] | None = None,
 ) -> None:
```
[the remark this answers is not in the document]
tatiana: Fixed in commit 2fdffcf8 — `DbtProducerWatcherKubernetesOperator` now holds a per-operator `_upstream_failure_skipped_ids` set, clears it in `execute`, and exposes it to `WatcherKubernetesCallback.progress_callback` via the same module-level-global pattern already used for `producer_task_context`. The callback forwards it to `store_dbt_resource_status_from_log`, so the #2698 rewrite applies uniformly to both watcher backends.

cosmos/operators/_watcher/base.py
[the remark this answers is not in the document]
tatiana: Deferring — the race here is benign under CPython:

- `set.add` and `in` are individually atomic at the bytecode level under the GIL.
- The only mutation is `add` (never remove), so a missed read just degrades one event back to pre-fix behaviour; the *next* event for that same `unique_id` corrects it (dbt always emits `SkippingDetails`/`LogSkipBecauseError` *and* a later `NodeFinished`, and both are rewritten consistently once the id is in the set).
- The rest of the log-parsing flow (XCom push, `tests_per_model` accumulator, etc.) is already invoked from those same dbt callback threads without explicit locking.

If we ever switch to free-threaded Python or move state to a structure where a partial read could corrupt invariants, we'll need a `threading.Lock` — captured for follow-up rather than added here.

tests/operators/test_watcher.py
```diff
@@ -85,6 +88,38 @@ class _MockContext(dict):
...
+            host=airflow_conn.host,
+            port=airflow_conn.port or 5432,
+            dbname=airflow_conn.schema or "postgres",
+            user=airflow_conn.login,
+            password=airflow_conn.password,
+        ) as conn:
+            conn.autocommit = True
+            with conn.cursor() as cur:
+                cur.execute(f"DROP SEQUENCE IF EXISTS public.{sequence_name}")
+
```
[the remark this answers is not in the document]
tatiana: Deferring — `sequence_name` is interpolated from hard-coded test constants (`_cosmos_fail_once_seq`, `_cosmos_recovery_fail_once_seq`), never user input, so there is no realistic injection surface here. Worth switching to `psycopg2.sql.Identifier` if we ever templatise the name; happy to do so in a follow-up.

cosmos/operators/_watcher/base.py
pankajkoti: Thanks @tatiana. I see the PR touching on the `--log-format` key. What I meant here was that the docstring talks about upstream failures: `SkippingDetails` / `LogSkipBecauseError` and whether they are the same for the dbt versions we support.

tatiana:

PR #2689, opened 2026-05-16
Refactor build_airflow_graph to satisfy the C901 limit

### LINKED ISSUE #1943: Refactor function `build_airflow_graph` to reduce code complexity
The pre-commit is failing with an error in PR: https://github.com/astronomer/astronomer-cosmos/pull/1939
```
ruff (legacy alias)...........................................................Failed
- hook id: ruff
- exit code: 1

C901 `build_airflow_graph` is too complex (11 > 10)
   --> cosmos/airflow/graph.py:595:5
    |
595 | def build_airflow_graph(
    |     ^^^^^^^^^^^^^^^^^^^
596 |     nodes: dict[str, DbtNode],
597 |     dag: DAG,  # Airflow-specific - parent DAG where to associate tasks and (optional) task groups
    |

Found 1 error.
```

00yhj22-debug:
## Description

`build_airflow_graph` carried a `# noqa: C901` suppression because its cyclomatic complexity was 11, one over the project's configured limit of 10.

This PR extracts the dbt test-task handling — the `AFTER_ALL`, `BUILD` and `AFTER_EACH` branches — into a new `_add_test_tasks` helper, mirroring the existing `_add_*` helpers already in `cosmos/airflow/graph.py`. Removing those four branches brings the complexity of `build_airflow_graph` down to 7, and the `# noqa: C901` suppression has been removed.

The change is a behavior-preserving refactor:

- The extracted statements are identical to the originals; only their location changed.
- The helper receives exactly the same values the inline code used.
- `task_or_group_args` is now initialized to `{}` before the node loop, so it is always defined when passed to the helper (previously it was only assigned inside the loop).

### Verification

- `ruff check cosmos/airflow/graph.py` passes with the suppression removed — complexity is now within the configured limit.
- `black --check` reports no changes.
- The existing `tests/airflow/test_graph.py` suite covers `build_airflow_graph` across the `AFTER_EACH`, `AFTER_ALL` and `BUILD` behaviors. I could not run the full Airflow test matrix locally (a Windows toolchain limitation building `openlineage-sql`), so I am relying on CI for the suite — happy to iterate on anything it surfaces.

## Related Issue(s)

Closes #1943

## Breaking Change?

No — this is an internal refactor with no functional change.

## Checklist

- [x] I have made corresponding changes to the documentation (not required for this internal refactor)

No new tests were added: this is a behavior-preserving refactor, and the existing `build_airflow_graph` tests already exercise the moved code.


00yhj22-debug:
Thanks for the review. Addressing the three points:

1. **`datached_node_name` typo** — fixed in ef7969a, renamed to `detached_node_name`.

2. **`node` carried over from the last loop iteration for detached test tasks** — this behavior is pre-existing and unchanged by this PR. The original `build_airflow_graph` already built the detached-test `test_task_args` from the same `task_or_group_args` dict (the last main-loop iteration value), overriding only `task_meta` and `resource_type`. This PR is a pure behavior-preserving extraction for #1943 (reduce cyclomatic complexity), so I kept that behavior identical. If it is a genuine bug it would be better addressed in a separate focused PR rather than mixed into a complexity refactor.

3. **`task_or_group_args` initialized to `{}`** — same scope note. Before this PR, an empty `nodes` mapping combined with `AFTER_ALL` raised `NameError` (the variable was only bound inside the loop); after it, the helper would raise `TypeError`. Both are unhandled errors on the same degenerate input, so practical behavior is unchanged. Initializing the variable before the loop is required so it is always defined when passed to `_add_test_tasks`. Gracefully handling an empty graph would be a behavior change beyond the scope of this complexity refactor.

00yhj22-debug:
Rebased onto current `main` to clear the merge conflict introduced by #2615 (`Watcher depends on past`) and #2710 (type-ignore cleanup).

The overlap was in the AFTER_ALL test-task block that this PR extracts into `_add_test_tasks`. #2615 added one new line inside that block — `tasks_map[f"{dbt_project_name}_test"] = test_task` — so the leaf test task is visible to `_add_watcher_dependencies` when it propagates `wait_for_downstream`. I pulled that line into the extracted helper so the watcher-depends-on-past behaviour from #2615 is preserved unchanged. The other upstream changes (the type-ignore removals from #2710 and #2615's modifications to `_add_watcher_dependencies` / `_add_watcher_producer_task`) merged cleanly because this PR doesn't touch those areas.

No behaviour change beyond what's already in the original refactor; just bringing the branch up to date.

Let me know if you'd prefer me to also drop the `# noqa: C901` comment from any sibling functions, or if the test-task extraction itself needs a different shape now that it has to keep the helper compatible with the watcher pass.

pankajkoti:
Hi @00yhj22-debug, could you please rebase the PR once? Integration tests are failing, and I guess https://github.com/astronomer/astronomer-cosmos/pull/2752 might help

00yhj22-debug:
Done — rebased onto current `main`, which now includes #2752. This time it was a clean rebase (no conflicts), since #2752 only touches the virtualenv example DAGs. That should let the integration tests pick up the `dbt-core<2.0` pin and go green. Thanks @pankajkoti!

cosmos/airflow/graph.py
```diff
@@ -989,7 +989,78 @@ def _add_teardown_task(
...
+            test_indirect_selection,
+            task_args=task_args,
+            on_warning_callback=on_warning_callback,
+            render_config=render_config,
+            enable_owner_inheritance=render_config.enable_owner_inheritance,
+        )
+        test_task_args = {
+            **task_or_group_args,
+            "task_meta": test_meta,
+            "resource_type": DbtResourceType.TEST,  # type: ignore
```
```suggestion
            "resource_type": DbtResourceType.TEST,
```

cosmos/airflow/graph.py
```diff
@@ -989,7 +989,78 @@ def _add_teardown_task(
...
+                task_args=task_args,
+                on_warning_callback=on_warning_callback,
+                render_config=render_config,
+                node=node,
+                enable_owner_inheritance=render_config.enable_owner_inheritance,
+            )
+            test_task_args = {
+                **task_or_group_args,
+                "task_meta": test_meta,
+                "resource_type": node.resource_type,  # type: ignore
```
```suggestion
                "resource_type": node.resource_type, 
```

cosmos/airflow/graph.py
```diff
@@ -989,7 +989,78 @@ def _add_teardown_task(
...
+            enable_owner_inheritance=render_config.enable_owner_inheritance,
+        )
+        test_task_args = {
+            **task_or_group_args,
+            "task_meta": test_meta,
+            "resource_type": DbtResourceType.TEST,  # type: ignore
+        }
+        # AFTER_ALL test is a single DAG-level task: place it at root so task_id stays e.g. "astro_shop_test"
+        test_task_args["task_group"] = parent_task_group
+        test_task = generate_or_convert_task(**test_task_args)  # type: ignore[arg-type]
```
```suggestion
        test_task = generate_or_convert_task(**test_task_args)
```

cosmos/airflow/graph.py
```diff
@@ -989,7 +989,78 @@ def _add_teardown_task(
...
+                render_config=render_config,
+                node=node,
+                enable_owner_inheritance=render_config.enable_owner_inheritance,
+            )
+            test_task_args = {
+                **task_or_group_args,
+                "task_meta": test_meta,
+                "resource_type": node.resource_type,  # type: ignore
+            }
+            test_task = generate_or_convert_task(**test_task_args)  # type: ignore[arg-type]
```
```suggestion
            test_task = generate_or_convert_task(**test_task_args)
```

cosmos/airflow/graph.py
```suggestion
        task_or_group = generate_task_or_group(**task_or_group_args, filtered_nodes=nodes)
```

pankajkoti (approved):
LGTM. The suggestions from Copilot look correct but fall outside the scope of the refactor. The validation of the existing bugs can be reproduced & tested, and applied in follow-up PRs. Just @00yhj22-debug could you please check the failing mypy type error warnings and fix those(I guess we mostly have to remove those type ignore annotations which I have suggested, need to be tried)? With that, we're good here.

cosmos/airflow/graph.py
```diff
@@ -948,6 +1015,7 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
     tasks_map: dict[str, TaskGroup | BaseOperator] = {}
     task_groups: dict[str, TaskGroup] = {}
     task_or_group: TaskGroup | BaseOperator | None
+    task_or_group_args: dict[str, Any] = {}
```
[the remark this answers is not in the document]
pankajkoti: Hi @00yhj22-debug looks like this is a real issue that was just reported in https://github.com/astronomer/astronomer-cosmos/issues/2813 by @m1racoli. Any chance we can address this issue here? Or if you could please address the failing type check in the CI and we can follow-up immediately with the fix for issue #2813 in a separate PR.

00yhj22-debug:
Thanks @pankajkoti — addressed both in 611b869:

**#2813**: I went ahead and fixed it here. The root cause is exactly as @m1racoli described — with an empty `nodes` mapping the per-node loop never runs, so `task_or_group_args` is never populated. On `main` that surfaces as `UnboundLocalError`; in this PR I'd initialized it to `{}`, which just moved the failure to a `TypeError` when `generate_or_convert_task(**{})` is called without its required args. Since an empty manifest means there are no models to test (and no leaf tasks to attach an `AFTER_ALL` test to), `_add_test_tasks` now returns early when `nodes` is empty. Added a regression test (`test_build_airflow_graph_with_after_all_and_empty_nodes`) that builds the graph with `nodes={}` + `AFTER_ALL` and asserts no crash / no test task.

**Type ignores**: applied all the suggestions — dropped the now-unused `# type: ignore` comments on the test-task and `generate_task_or_group` calls.

Let me know if you'd prefer the empty-nodes guard somewhere else (e.g. earlier in `build_airflow_graph`) — I kept it inside `_add_test_tasks` since that's where the `AFTER_ALL` branch lives.

pankajkoti:
Thanks a lot @00yhj22-debug for addressing issue #2813 here.  However meantime, @m1racoli also contributed and opened  PR #2814 to tackle the issue. I see the approaches are subtly different here and in the other PR, and I have invited a discussion in https://github.com/astronomer/astronomer-cosmos/pull/2814#pullrequestreview-4532825490 to see how we want to tackle this issue.

The PR looks good to me now, and I also prefer the approach we're taking here to resolve #2813 by not creating a test task at all when there are no nodes, just holding off on the merge until the discussion is concluded with more people in the loop.

cc: @tatiana @m1racoli

00yhj22-debug:
Thanks @pankajkoti, and thanks @m1racoli for the quick fix + regression test on #2814 — both PRs stop the crash, so either is a safe fix for 1.15.0.

The one reason I leaned towards skipping the task entirely: with zero rendered models, an `AFTER_ALL` task runs `dbt test` against an empty selection, so it either no-ops or errors at runtime, and the DAG ends up as a single lonely test task with nothing upstream of it. That seemed more surprising to a user than just getting an empty graph — hence not creating it at all. But it's genuinely a judgement call and I don't feel strongly.

Happy to go whichever way you and @tatiana land. If you'd rather take #2814's approach for #2813, I'm glad to drop the empty-nodes guard + test from here so this PR stays focused on just the C901 refactor, and we avoid double-fixing the same issue. Just let me know.

tatiana:
Thanks for your solution and for the great and healthy discussion, @00yhj22-debug - we'll go ahead with your PR and release it as part of 1.15.0.

PR #2692, opened 2026-05-18
Skip AF 3.0 integration test stuck on `example_watcher_with_freshness`

pankajastro:
## Summary

- The `Run-Integration-Tests` matrix has been hanging silently on the **Airflow 3.0 + split 1** cell (all three Python variants) for the past week, getting killed at GitHub Actions' 6h job timeout. Six main-branch runs in a row exhausted CI capacity that way.
- Pinpointed the offender with two non-speculative diagnostic levers: a 30-min wall-clock cap on the job (`.github/workflows/test.yml`) and `pytest-timeout` with `--timeout=180 --timeout-method=thread`. The next CI run failed in ~12 min with the test name + thread stack instead of a silent 6h cancel.
- Hanging test: ``tests/test_example_dags.py::test_example_dag[example_watcher_with_freshness]``. The thread stack pointed at ``cosmos/operators/_watcher/triggerer.py:82``:

  ```
  File ".../cosmos/operators/_watcher/triggerer.py", line 82, in get_xcom_val_af3
      return await sync_to_async(XCom.get_one)(...)
  File ".../airflow/sdk/bases/xcom.py", line 242, in get_one
      from airflow.sdk.execution_time.task_runner import SUPERVISOR_COMMS
  ImportError: cannot import name 'SUPERVISOR_COMMS' from
      'airflow.sdk.execution_time.task_runner'
  ```

## Why this hangs only on AF 3.0 and only in `dag.test()`

- ``dag.test()`` runs the ``WatcherTrigger`` *inline* (no separate triggerer process). The trigger calls ``airflow.sdk.execution_time.xcom.XCom.get_one``, which on AF 3.0 unconditionally imports ``SUPERVISOR_COMMS`` from ``airflow.sdk.execution_time.task_runner``. That symbol does not exist until Airflow 3.1.
- The ``ImportError`` surfaces as a ``NameError`` warning (``state.py:177``), the deferred task is re-queued, and ``dag.test()``'s inline-trigger loop redrives the same task forever (~30ms cycle). On AF 3.1+ ``SUPERVISOR_COMMS`` exists, so the trigger resolves and the loop terminates.
- The watcher path still works in a real triggerer process on AF 3.0, so this only affects the ``dag.test()`` exerciser used by the integration matrix. Real users running the watcher in a deployed scheduler are not blocked.

## Fix

- Add ``watcher_with_freshness_check.py`` to ``.airflowignore`` when ``3.0.0 <= AIRFLOW_VERSION < 3.1.0`` (mirrors the existing ``example_cosmos_cleanup_dag.py`` pattern). The DAG is still parametrized and run on AF 2.9–2.11, 3.1, 3.2.
- Keep the 30-min job timeout and ``pytest-timeout`` diagnostic levers — both are cheap insurance against future silent hangs and they're what surfaced this one. Happy to drop them in review if preferred.

## Test plan

- [x] Reproduced the hang on this branch's first CI run (run ``26031384067``) — all three AF 3.0 split-1 jobs failed in ~12 min with the ``SUPERVISOR_COMMS`` ImportError instead of running to the 6h cap.
- [ ] Verify on this PR's CI run that the AF 3.0 + split 1 cells now finish green, while AF 2.9/2.10/2.11/3.1/3.2 still parametrize and run the DAG.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main, diagnose-af3-split1-hang]
+    branches: [main]
   # Also run on pull requests originating from forks. Jobs that require sensitive secrets (integration tests,
   # kubernetes, code coverage) depend on the Authorize job, which requires manually approving the workflow run for
   # external PRs. Jobs that do not use sensitive secrets (type-check, unit tests, performance, telemetry) run
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -191,8 +191,8 @@ jobs:
     needs: [Authorize, Check-changed-files]
     if: needs.Check-changed-files.outputs.code_changed == 'true'
     runs-on: ubuntu-latest
-    # Diagnostic: cap the job at 30 min so a hanging test surfaces a pytest failure with the test
-    # name and traceback in the log instead of being killed silently by GitHub's 6h default.
+    # Diagnostic: cap the job at 30 min so a hang does not run until GitHub's 6h default
+    # timeout. This is a job-level backstop and does not itself produce a pytest traceback.
     timeout-minutes: 30
     permissions:
       contents: read
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -193,7 +193,7 @@ jobs:
     runs-on: ubuntu-latest
     # Diagnostic: cap the job at 30 min so a hang does not run until GitHub's 6h default
     # timeout. This is a job-level backstop and does not itself produce a pytest traceback.
-    timeout-minutes: 30
+    timeout-minutes: 60
     permissions:
       contents: read
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, diagnose-af3-split1-hang]
   # Also run on pull requests originating from forks. Jobs that require sensitive secrets (integration tests,
   # kubernetes, code coverage) depend on the Authorize job, which requires manually approving the workflow run for
   # external PRs. Jobs that do not use sensitive secrets (type-check, unit tests, performance, telemetry) run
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, diagnose-af3-split1-hang]
```
```suggestion
    branches: [main]
```

scripts/test/integration.sh
```diff
@@ -25,6 +25,8 @@ pytest -vv \
     --cov-report=term-missing \
     --cov-report=xml \
     --durations=0 \
+    --timeout=180 \
```
pankajastro:
```suggestion
    --timeout=300 \
```

tatiana (commented):
@pankajastro I'm supportive of this change once we have a follow-up ticket to investigate and address the actual issue, and to assign it to the next sprint with high priority, starting on Monday, please. Share the link here once it is logged.

pankajastro:
> @pankajastro I'm supportive of this change once we have a follow-up ticket to investigate and address the actual issue, and to assign it to the next sprint with high priority, starting on Monday, please. Share the link here once it is logged.

Created issue https://github.com/astronomer/astronomer-cosmos/pull/2692 and added it to next sprint with priority high

.github/workflows/test.yml
```diff
@@ -191,6 +191,9 @@ jobs:
     needs: [Authorize, Check-changed-files]
     if: needs.Check-changed-files.outputs.code_changed == 'true'
     runs-on: ubuntu-latest
+    # Diagnostic: cap the job at 30 min so a hang does not run until GitHub's 6h default
+    # timeout. This is a job-level backstop and does not itself produce a pytest traceback.
+    timeout-minutes: 30
```
```suggestion
    timeout-minutes: 60
```
Do you think 30 mins is too aggressive? Maybe 45 or 60 to account for occasional latencies for warehouse-related tests is safe IMO.

pankajkoti (approved):

tatiana (approved):
Thanks for addressing the feedback, @pankajastro !

.github/workflows/test.yml
```diff
@@ -191,6 +191,9 @@ jobs:
     needs: [Authorize, Check-changed-files]
     if: needs.Check-changed-files.outputs.code_changed == 'true'
     runs-on: ubuntu-latest
+    # Diagnostic: cap the job at 30 min so a hang does not run until GitHub's 6h default
```
```suggestion
    # Diagnostic: cap the job at 60 min so a hang does not run until GitHub's 6h default
```

.github/workflows/test.yml
pankajastro: Increased to 60 min

tatiana:

PR #2708, opened 2026-05-20
Release 1.14.2

tatiana:
## CHANGELOG entry

1.14.2 (2026-05-21)
-------------------

Behaviour Changes

These changes adjust observable behaviour of the ``ExecutionMode.WATCHER`` execution mode.
None of them breaks the public Cosmos API, but users relying on undocumented internals
(graph wiring assertions, XCom backup Variable names, retry-on-recovery semantics, or
retry log format) should review before upgrading.

* ``ExecutionMode.WATCHER`` + ``depends_on_past=True``: when the producer task has
  ``depends_on_past=True`` (typically set via ``default_args``), the producer-done gateway
  task inside ``DbtTaskGroup`` is now wired downstream of every consumer task, in addition
  to the producer. This is required so that ``wait_for_downstream`` gating behaves
  correctly across DAG runs and the task group acts as a single unit that must fully
  succeed before the next run starts. Users with ``depends_on_past=False`` (the default)
  see no topology change. See #2615.
* ``ExecutionMode.WATCHER`` downstream retry on upstream recovery: dbt models that were
  skipped after an upstream-failure event are now retried in the same DAG run when the
  upstream task succeeds on retry. Previously these models remained skipped for the run.
  See #2684.
* ``ExecutionMode.WATCHER`` consumer-retry log format: the consumer's fallback ``dbt``
  invocation no longer inherits the producer's internal ``--log-format json`` flag, so
  retry task logs now default to dbt's normal text format. Users who relied on JSON output
  in retry logs can opt in via ``operator_args={"dbt_cmd_flags": ["--log-format", "json"]}``.
  See #2713.
* ``ExecutionMode.WATCHER`` XCom-backup Variable key scheme: the per-model XCom backup
  Variable key now includes the full task-group path and sanitises disallowed characters
  (``+`` / ``:``) from ``run_id``. External monitoring or cleanup scripts that match the
  old key pattern will need updating. See #2629 and #2683.

Bug Fixes

* Sanitize disallowed characters from XCom backup variable key by @MichaelRBlack in #2629
* Prevent watcher producers from colliding on one XCom-backup key by @tatiana in #2683
* Retry watcher downstream models on upstream-failure recovery by @tatiana in #2684
* Fix ``ExecutionMode.WATCHER`` interaction with ``depends_on_past`` by @johnhoran in #2615
* Strip ``--log-format`` from producer flags on watcher consumer retry by @tatiana in #2713
[...]

PR #2713, opened 2026-05-20
Fix: watcher consumer retry logging dbt with `--log-format json`

tatiana:
When a watcher consumer task retries via `_fallback_to_non_watcher_run`, it copies the producer's `add_cmd_flags()` output into its own dbt invocation. The producer hardcodes `log_format="json"` so it can parse dbt's structured event stream — but that internal choice was leaking into the consumer's user-facing retry dbt command, producing raw JSON (one structured event per line) instead of dbt's default formatted text in task logs.

Example of what users currently see on a retry task like `watcher_downstream_not_skipped.model_downstream_run`:

```
[2026-05-20 21:45:11] INFO - {"data": {"log_version": 3, "version": "=1.11.8"}, "info": {"category": "", "code": "A001", ...}}
[2026-05-20 21:45:11] INFO - {"data": {"adapter_name": "***", ...}, "info": {"category": "", "code": "E034", ...}}
```

## Fix

Add `--log-format` to `_filter_flags`'s skip list. The producer's hardcoded `--log-format json` is no longer propagated into the consumer's fallback dbt command, so retries default to dbt's normal text output.

## How users can still get JSON

Users who want JSON-formatted dbt output on retry can opt in via:

```python
DbtDag(
    ...,
    operator_args={"dbt_cmd_flags": ["--log-format", "json"]},
)
```

`dbt_cmd_flags` is appended to the dbt command by `build_cmd` outside of `_filter_flags`'s pipeline, so this path keeps working after the fix.

### Automated

- [x] `tests/operators/test_watcher.py` updated:
  - `test_filter_flags` — extended to assert `--log-format <value>` is filtered alongside `--select` / `--exclude`
  - `test_filter_flags_strips_log_format_when_first` (new) — covers the edge case where `--log-format` is the first flag in the input list
  - `test_fallback_strips_producer_log_format_by_default` (new) — end-to-end verifies the consumer's retry command has no `--log-format`
  - `test_fallback_preserves_user_dbt_cmd_flags_opt_in_to_json` (new) — documents the opt-in path via `operator_args["dbt_cmd_flags"]`
- [x] All 12 affected tests pass on `tests.py3.12-2.11-1.11`.

### Manual validation DAGs

Two example DAGs in `dev/failed_dags/` exercise both sides of the change. Both intentionally fail once on `model_retry` and retry via the watcher fallback path — inspect the retry task's logs to verify behaviour. Do **not** run them concurrently (they share `public._cosmos_fail_once_seq`).

- [x] **`dev/failed_dags/example_watcher_downstream_not_skipped.py`** — default config (no `dbt_cmd_flags`). On the `watcher_downstream_not_skipped.model_retry_run` retry, log lines should be dbt's normal **text format** (e.g. `INFO  [MainThread]: Running with dbt=1.11.8`). Before this fix, those lines came through as raw JSON; after the fix, they should read normally.
- [x] **`dev/failed_dags/example_watcher_downstream_not_skipped_json_logs.py`** (new in this PR) — adds `operator_args={"dbt_cmd_flags": ["--log-format", "json"]}`. On the `watcher_downstream_not_skipped_json_logs.model_retry_run` retry, log lines should be structured **JSON** objects (`{"data": ..., "info": ...}`). Confirms the user opt-in still works after the fix.

## How this was discovered

Reported while validating `example_watcher_downstream_not_skipped`'s `model_downstream_run` retry on the [1.14.2a3 release candidate](https://github.com/astronomer/astronomer-cosmos/pull/2708).

## Backport

Targeting `main`; needs a backport to `release-1.14` so it ships in **1.14.2**. Add `Cosmos 1.14.2` milestone after merge.

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -446,7 +446,15 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
 
     @staticmethod
     def _filter_flags(flags: list[str]) -> list[str]:
-        """Filters out dbt flags that are incompatible with retry (e.g., --select, --exclude)."""
+        """Filters out dbt flags that should not propagate from the producer to the consumer's retry:
+
+        - ``--select`` / ``--exclude``: the consumer targets a single model and re-applies its own selector.
+        - ``--log-format``: the producer always sets ``--log-format json`` so it can parse dbt's structured
+          event stream; the consumer's retry is a user-facing dbt run and should default to text. Users who
+          want JSON on the retry can opt in via
+          ``operator_args={"dbt_cmd_flags": ["--log-format", "json"]}``, which is appended by ``build_cmd``
+          outside of this flag pipeline.
+        """
         filtered = []
         skip_next = False
         for token in flags:
@@ -455,7 +463,7 @@ class BaseConsumerSensor(BaseSensorOperator):  # type: ignore[misc]
                     skip_next = False
                 else:
                     continue  # skip value of previous flag
-            if token in ("--select", "--exclude"):
+            if token in ("--select", "--exclude", "--log-format"):
                 skip_next = True
                 continue
             filtered.append(token)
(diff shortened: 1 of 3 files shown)


cosmos/operators/_watcher/base.py
```diff
@@ -455,7 +463,7 @@ def _filter_flags(flags: list[str]) -> list[str]:
                     skip_next = False
                 else:
                     continue  # skip value of previous flag
-            if token in ("--select", "--exclude"):
+            if token in ("--select", "--exclude", "--log-format"):
```
pankajkoti, line 466: Might be a good idea to extract these into a constant _PRODUCER_ONLY_FLAGS in a follow-up PR.

pankajkoti (approved):
LGTM. I guess we need to associate the milestone 1.14.2 for this PR?

cosmos/operators/_watcher/base.py
tatiana: Addressed in https://github.com/astronomer/astronomer-cosmos/pulls

tatiana:

PR #2714, opened 2026-05-21
Disclose `--no-static-parser` flag in DBT_RUNNER mode

### LINKED ISSUE #1925: Docs: Add disclaimer for static-parser when using execution mode local
### Description

I was trying to optimize overhead for small models and saw that "--no-static-parser" was appended to all runs.

After some digging I found a comment in the [source code](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/operators/local.py) stating that static-parser is not available in this execution mode. This should be disclosed in the documentation. I suggest adding a disclaimer [here](https://astronomer.github.io/astronomer-cosmos/getting_started/execution-modes.html).

### Use case/motivation

Transparency of this limitation would have saved me time spent investigating why local+DbtRunner always append --no-static-parser

### Related issues

_No response_

pankajkoti:
## Summary

Cosmos appends `--no-static-parser` to every dbt command invoked through `InvocationMode.DBT_RUNNER` on dbt 1.5.6+, but the behavior was only documented in a source-code comment (`cosmos/operators/local.py:529-540`). Users who tried to opt into dbt's static parser for performance had no way to discover this from the docs.

Adds a note to the invocation-mode page explaining that the flag is injected automatically, why it is needed (temp-path interaction between DAG parsing and task execution can hang tasks), and that `InvocationMode.SUBPROCESS` is the alternative for users who rely on the static parser.

Closes #1925.

### DIFF
--- a/docs/optimize_performance/invocation_mode.rst
+++ b/docs/optimize_performance/invocation_mode.rst
@@ -12,6 +12,10 @@ For ``ExecutionMode.LOCAL`` execution mode, Cosmos supports two invocation modes
    In order to use this mode, dbt must be installed in the same local environment. This mode does not have the overhead of spawning new subprocesses or parsing the output of dbt commands and is faster than ``InvocationMode.SUBPROCESS``. \
    This mode requires dbt version 1.5.0 or higher. It is up to the user to resolve :ref:`execution-modes-local-conflicts` when using this mode.
 
+.. note::
+
+   When ``InvocationMode.DBT_RUNNER`` is used with dbt 1.5.6 or later, Cosmos automatically appends ``--no-static-parser`` to every dbt command it invokes. Cosmos copies the dbt project into a temporary directory at both DAG parse time and task execution time, and the differing temp paths can interfere with dbt's static parser and cause task hangs. Disabling the static parser is a workaround for this interaction. If you rely on dbt's static parser, use ``InvocationMode.SUBPROCESS`` instead.
+
 The invocation mode can be set in the ``ExecutionConfig`` as shown below:
 
 .. code-block:: python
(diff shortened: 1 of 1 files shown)


docs/optimize_performance/invocation_mode.rst
```diff
@@ -12,6 +12,16 @@ For ``ExecutionMode.LOCAL`` execution mode, Cosmos supports two invocation modes
    In order to use this mode, dbt must be installed in the same local environment. This mode does not have the overhead of spawning new subprocesses or parsing the output of dbt commands and is faster than ``InvocationMode.SUBPROCESS``. \
    This mode requires dbt version 1.5.0 or higher. It is up to the user to resolve :ref:`execution-modes-local-conflicts` when using this mode.
 
+.. note::
+
+   When ``InvocationMode.DBT_RUNNER`` is used with dbt-core 1.5.6 or later,
+   Cosmos automatically appends ``--no-static-parser`` to every dbt command
+   it invokes. Cosmos copies the dbt project into a temporary directory at
+   both DAG parse time and task execution time, and the differing temp paths
+   can interfere with dbt's static parser and cause task hangs. Disabling the
```
tatiana, line 21: Would it be worth adding a reference to when/how we reached this conclusion?

tatiana (approved):
Hi @pankajkoti, thanks for documenting this, as it may have unexpected side effects.
I left a minor comment, and I'm happy for the PR to be merged after it is addressed.

docs/optimize_performance/invocation_mode.rst
pankajkoti: yes, added the issue and PR link for traceability.

Issue #2666, opened 2026-05-12
[bug] Cosmos ignores `freshness` set by users (template field is output-only)


tatiana:
### Summary

`freshness` is registered in `template_fields` on local-mode operators
(`cosmos/operators/local.py:188`), which makes it look like a
user-settable, Jinja-templatable operator arg. In practice it is an
internal output field used to expose dbt source freshness JSON in the
Airflow UI — any value a user passes via `operator_args={"freshness": ...}`
is silently discarded.

### Evidence

- `cosmos/operators/local.py:216` — `__init__` unconditionally sets
  `self.freshness = ""`, overwriting any kwarg value.
- `cosmos/operators/local.py:411` — `store_freshness_json` overwrites
  `self.freshness` with `formatted_sources_json` after `dbt source
  freshness` runs.
- `template_fields_renderers` maps `freshness` to the `"json"` renderer
  (line 191), confirming the intent is UI display, not user input.

### Impact

Misleading documentation. `docs/guides/run_dbt/operators/operator-args.rst`
lists `freshness` indirectly as a template field. Users who pass
`freshness` via `operator_args` see no effect and no warning.

### Proposed fix

1. Change the behaviour and allow users to set freshness as a templated field or
2. Document `freshness` (and `compiled_sql`) explicitly as **output-only**
   template fields in `docs/guides/run_dbt/operators/operator-args.rst`. In this case,
   log a warning if a user passes `freshness` or `compiled_sql`
   as a kwarg to a local operator, since the value will be discarded.

00yhj22-debug:
Following up from #2719 — happy to take this as a runtime change: raise an error in `__init__` (and surface it cleanly when the value arrives via `operator_args`) when `compiled_sql` or `freshness` is passed, so the misuse fails fast at task instantiation instead of being silently overwritten.

Could you assign this to me?

pankajkoti:
assigned to you @00yhj22-debug , thanks for taking this up

00yhj22-debug:
Thanks @pankajkoti, on it.

PR #2719, opened 2026-05-21
Document `compiled_sql` and `freshness` as output-only template fields

00yhj22-debug:
## Description

Relates to: #2666.

`compiled_sql` and `freshness` are both registered in the local execution-mode operators' `template_fields` (`cosmos/operators/local.py`), which makes them look like user-settable Jinja-templatable args. In practice both are *output-only*:

- `__init__` unconditionally resets `self.compiled_sql = ""` and `self.freshness = ""` on every operator instantiation.
- `store_freshness_json` later overwrites `self.freshness` with the captured `sources.json` after `dbt source freshness` runs.

So any value a user passes via `operator_args={"freshness": ...}` (or `compiled_sql`) is silently discarded — no effect, no warning.

This PR is the documentation-only option proposed in #2666: expand the existing "Template fields" section in `docs/guides/run_dbt/operators/operator-args.rst` with a small "Output-only template fields" subsection that calls out both fields and explains why values supplied through `operator_args` have no effect. The runtime behavior is unchanged.

If you'd rather pursue option 1 (actually let users pass values), I'm happy to send a follow-up — but the docs clarification is useful regardless.

## Related Issue(s)

Relates to #2666

## Breaking Change?

No — docs-only change.

## Checklist

- [x] I have made corresponding changes to the documentation (this *is* the docs change)

No new tests — the change is documentation only.

### DIFF
--- a/docs/guides/run_dbt/operators/operator-args.rst
+++ b/docs/guides/run_dbt/operators/operator-args.rst
@@ -161,6 +161,19 @@ The following template fields are only selectable when using the operators in a 
 
 Since Airflow resolves template fields during Airflow DAG execution and not DAG parsing, the args above cannot be templated via ``DbtDag`` and ``DbtTaskGroup`` because both need to select dbt nodes during DAG parsing.
 
-Additionally, the SQL for compiled dbt models is stored in the template fields, which is viewable in the Airflow UI for each task run.
-This is provided for telemetry on task execution, and is not an operator arg.
-For more information about this, see the :doc:`Compiled SQL </guides/cosmos_devex/compiled-sql>` docs.
+Output-only template fields
++++++++++++++++++++++++++++
+
+A small number of template fields on the local execution mode operators are
+*output-only*: Cosmos populates them as the task runs so the values appear in the
+Airflow UI, but any value passed in via ``operator_args`` is silently
+overwritten and has no effect.
+
+- ``compiled_sql`` — the SQL Cosmos compiled for a model. See the
+  :doc:`Compiled SQL </guides/cosmos_devex/compiled-sql>` docs for how it
+  is populated and how to disable it via ``should_store_compiled_sql``.
+- ``freshness`` — the JSON Cosmos captures from ``dbt source freshness`` when
+  source nodes run, reset on every task instance.
+
+If you need to set one of these values through ``operator_args`` for an
+upstream task to consume, use a separate, regular template field instead.
--- a/docs/guides/run_dbt/operators/operator-args.rst
+++ b/docs/guides/run_dbt/operators/operator-args.rst
@@ -174,6 +174,3 @@ overwritten and has no effect.
   is populated and how to disable it via ``should_store_compiled_sql``.
 - ``freshness`` — the JSON Cosmos captures from ``dbt source freshness`` when
   source nodes run, reset on every task instance.
-
-If you need to set one of these values through ``operator_args`` for an
-upstream task to consume, use a separate, regular template field instead.


docs/guides/run_dbt/operators/operator-args.rst
```diff
@@ -161,6 +161,19 @@ The following template fields are only selectable when using the operators in a
...
+overwritten and has no effect.
+
+- ``compiled_sql`` — the SQL Cosmos compiled for a model. See the
+  :doc:`Compiled SQL </guides/cosmos_devex/compiled-sql>` docs for how it
+  is populated and how to disable it via ``should_store_compiled_sql``.
+- ``freshness`` — the JSON Cosmos captures from ``dbt source freshness`` when
+  source nodes run, reset on every task instance.
+
+If you need to set one of these values through ``operator_args`` for an
+upstream task to consume, use a separate, regular template field instead.
```
pankajkoti: LGTM, however, I did not understand this part. Could you please elaborate on what we mean by upstream task with an example, and how to use a separate template field to set these values?

00yhj22-debug:
Thanks for the review @pankajkoti! On reflection you are right — that closing line was hand-wavy and did not actually describe a concrete supported workflow. Cosmos already populates these fields itself, so there is no real "use a separate template field instead" pattern to point users at. I dropped the sentence in 4098bbd; the subsection still keeps the core point that the two fields are output-only and that values supplied via  are ignored.

pankajkoti (approved):

docs/guides/run_dbt/operators/operator-args.rst
```diff
@@ -161,6 +161,16 @@ The following template fields are only selectable when using the operators in a
...
+Output-only template fields
++++++++++++++++++++++++++++
+
+A small number of template fields on the local execution mode operators are
+*output-only*: Cosmos populates them as the task runs so the values appear in the
+Airflow UI, but any value passed in via ``operator_args`` is silently
+overwritten and has no effect.
+
+- ``compiled_sql`` — the SQL Cosmos compiled for a model. See the
+  :doc:`Compiled SQL </guides/cosmos_devex/compiled-sql>` docs for how it
```
tatiana, line 173: @00yhj22-debug, please, could you confirm this works with Sphinx when we build & serve the docs via` hatch run docs:serve ` (https://github.com/astronomer/astronomer-cosmos/blob/main/docs/policy/contributing.rst#writing-docs)

tatiana (commented):
@00yhj22-debug thanks a lot for your contribution here! I changed the PR description from "Closes" to "Relates to", since I'd still like to see a runtime change land alongside the docs.

Thinking about it a bit more, both `compiled_sql` and `freshness` are genuinely output-only:

- `compiled_sql` is populated by Cosmos from the actual compiled artefact.
- `freshness` is only ever written by `DbtSourceLocalOperator` via `store_freshness_json`, and only for source nodes. On every other operator, it stays empty - there's nothing meaningful for a user to set.

Given that, I'd propose the follow-up raises an error when a user passes either field via `operator_args` or as an arg to the operators, if they are instantiating them directly, so the misuse is caught at task instantiation rather than silently discarded at runtime.

If you'd like to pick this up, please drop a comment on #2666, and we'll assign it to you - we'd really appreciate the follow-up. No pressure if you'd rather not; happy to find another owner otherwise.

Thanks again for the clean docs change!

00yhj22-debug:
Thanks @tatiana! I'll take the follow-up.

On the Sphinx check — RST parses cleanly via `docutils` and the new subsection follows the `=` → `~` → `+` heading hierarchy from the docs style guide (`+++` "Output-only template fields" under the `~~~` "Template fields" section). I couldn't run `hatch run docs:serve` locally because the docs env install fails on `openlineage-sql` (Rust toolchain needed on Windows). If anything renders wrong in your build, ping me and I'll fix.

Dropping a note on #2666 now to pick up the runtime change.

tatiana (approved):

PR #2722, opened 2026-05-22
Document tradeoffs of remote cache and disabling dbt ls cache

pankajkoti:
## Summary

- Add a "Choosing between Variable cache and remote cache" subsection to both the dbt ls and YAML selectors cache sections in `docs/optimize_performance/caching.rst`, calling out the roughly 2 to 4 seconds of per-parse latency observed with remote cache on busy deployments, and the parse-time penalty of disabling either cache entirely.
- Add one-line tradeoff notes to the `enable_cache_dbt_ls`, `enable_cache_dbt_yaml_selectors`, and `remote_cache_dir` entries in `docs/reference/configs/cosmos-conf.rst`.

Both of these tradeoffs were previously folklore, easy to miss when choosing between cache backends or deciding whether to disable caching. Documenting them inline makes it easier for anyone reading the docs to pick the right setting for their deployment on the first attempt.

### DIFF
--- a/docs/optimize_performance/caching.rst
+++ b/docs/optimize_performance/caching.rst
@@ -42,6 +42,18 @@ Variables. To do so, you need to configure a remote cache directory. See :ref:`r
 user feedback. The ``remote_cache_dir`` will eventually be merged into the :ref:`cache_dir` setting in upcoming
 releases.
 
+**Choosing between Variable cache and remote cache**
+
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Looks great, thank you very much, @pankajkoti. This will help both Cosmos users, CRE, and bots understand the implications and how to configure the cache in future, avoiding the issue we observed a customer faced recently

PR #2726, opened 2026-05-26
Reject `compiled_sql` and freshness in `operator_args` at init

### LINKED ISSUE #2666: [bug] Cosmos ignores `freshness` set by users (template field is output-only)

00yhj22-debug:
## Description

Follow-up to #2719, picking up the runtime change @tatiana proposed on [that PR](https://github.com/astronomer/astronomer-cosmos/pull/2719#pullrequestreview-3460200000).

`compiled_sql` and `freshness` are listed in `template_fields` on the local-execution-mode operators, which makes them look like user-settable Jinja-templatable args, but they are populated by Cosmos at runtime:

- `AbstractDbtLocalBase.__init__` unconditionally resets `self.compiled_sql = ""` and `self.freshness = ""`.
- The dbt run / `store_freshness_json` handlers fill them in afterwards.

So any value a user passed via `operator_args={"compiled_sql": ...}` (or directly to the operator) was silently overwritten and had no effect.

This PR fails fast in `AbstractDbtLocalBase.__init__`: if either field is present in `**kwargs`, raise `CosmosValueError` naming the field(s) so the misuse is caught at task instantiation rather than discarded at runtime.

```text
CosmosValueError: 'compiled_sql' is output-only template field(s) populated by
Cosmos at runtime; remove them from operator_args.
```

## Related Issue(s)

Closes #2666

## Breaking Change?

Technically yes — code that *passed* these fields was always a silent no-op, but it will now raise on task instantiation. The behaviour was never functional, so the surface area is narrow.

## Checklist

- [x] I have made corresponding changes to the documentation (the docs subsection landing in #2719 already calls these fields output-only)
- [x] I have added tests that prove my fix is effective (`tests/operators/test_local.py` — parametrised for both fields, plus a combined case)

Caveat on local verification: I could not run `hatch run tests …` locally because the docs/test env install fails on `openlineage-sql` (Rust toolchain needed on Windows). The change is small and surgical — a 7-line guard in `__init__` plus two test cases — and `ruff check` / `black --check` are clean. Happy to iterate on anything CI surfaces.

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -120,8 +120,6 @@ from cosmos.operators.base import (
 
 logger = get_logger(__name__)
 
-_OUTPUT_ONLY_TEMPLATE_FIELDS: tuple[str, ...] = ("compiled_sql", "freshness")
-
 
 def _read_target_sources_json(project_root: Path) -> dict[str, Any] | None:
     """Parse ``target/sources.json`` under ``project_root`` if the file exists."""
@@ -193,17 +191,6 @@ class AbstractDbtLocalBase(AbstractDbtBase):
     _OUTPUT_ONLY_TEMPLATE_FIELDS: tuple[str, ...] = ("compiled_sql", "freshness")
     _process_log_line_callable: Callable[[str, Any], None] | None = None
 
-    def _warn_on_output_only_fields(self, kwargs: dict[str, Any]) -> None:
-        """Warn when output-only template fields are passed directly to local operators."""
-        for field in _OUTPUT_ONLY_TEMPLATE_FIELDS:
-            if field in kwargs:
-                self.log.warning(
-                    "The '%s' argument passed to %s will be overwritten at runtime; "
-                    "it is an output-only template field.",
-                    field,
-                    self.__class__.__name__,
-                )
-
     def __init__(
         self,
         task_id: str,
@@ -220,14 +207,6 @@ class AbstractDbtLocalBase(AbstractDbtBase):
         dbt_runner_callbacks: list[Callable] | None = None,  # type: ignore[type-arg]
         **kwargs: Any,
     ) -> None:
-        forbidden = [f for f in self._OUTPUT_ONLY_TEMPLATE_FIELDS if f in kwargs]
-        if forbidden:
-            names = ", ".join(repr(f) for f in forbidden)
-            raise CosmosValueError(
-                f"{names} {'is' if len(forbidden) == 1 else 'are'} output-only "
-                "template field(s) populated by Cosmos at runtime; "
-                "remove them from operator_args."
-            )
         self.task_id = task_id
         self.profile_config = profile_config
         self.callback = callback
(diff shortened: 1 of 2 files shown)


cosmos/operators/local.py
```diff
@@ -227,6 +231,14 @@ def __init__(
         dbt_runner_callbacks: list[Callable] | None = None,  # type: ignore[type-arg]
         **kwargs: Any,
     ) -> None:
+        forbidden = [f for f in self._OUTPUT_ONLY_TEMPLATE_FIELDS if f in kwargs]
```
pankajastro: These filed may have dropped here https://github.com/astronomer/astronomer-cosmos/blob/v1.11.0/cosmos/operators/local.py#L949-L961

pankajastro (commented):
I accepted PR #2737, which addresses the same issue. I think emitting a warning is a more relaxed and user-friendly approach than raising an error. Unless we consider this a bug, changing the behavior to raise an error could introduce a regression for some users and would effectively be a breaking change.

Given that, I'd be happy to merge this PR if the other maintainers agree that this should be treated as a bug fix. If we go in that direction, I also think it would be a good opportunity to clean up the code related to the warning behavior.

00yhj22-debug:
Good catch, thank you — you're exactly right. `DbtLocalBaseOperator.__init__` segregates kwargs by the `__init__` argspec before delegating to `AbstractDbtLocalBase.__init__`, and since neither `compiled_sql` nor `freshness` is a constructor argument they were dropped at that split (the lines you linked) and never reached the guard, so it never fired for the `operator_args` path and the tests failed.

I've moved the guard to the top of `DbtLocalBaseOperator.__init__`, before the segregation, so it inspects the raw kwargs. While there I dropped the redundant module-level constant and the warn-only helper, leaving a single class-level `_OUTPUT_ONLY_TEMPLATE_FIELDS`. The existing tests now exercise the real path. Pushed in c5832ee.

00yhj22-debug:
While fixing this I found the branch had drifted into an inconsistent state: it carried *both* a warn-only path (a `_warn_on_output_only_fields` helper + a test asserting only a log warning) *and* the reject-at-init guard (plus two tests asserting a raise). The two contradicted each other, which is what was actually failing CI.

I consolidated on **reject** to match this PR's title and the issue intent — fail fast so the misuse can't be silently ignored. Concretely: removed the warn helper and the duplicate module-level constant, kept a single class-level `_OUTPUT_ONLY_TEMPLATE_FIELDS`, moved the guard ahead of the kwargs segregation, and rewrote the warn test to assert rejection through `DbtRunLocalOperator` (the real `operator_args` path).

One thing worth your call: raising is stricter than warning, so a DAG that currently passes `compiled_sql`/`freshness` and has them silently ignored would now error on parse. I think that's the right trade for output-only fields, but if you'd rather keep it a non-breaking warning for a release cycle first, that's an easy switch — just say the word and I'll flip the guard to a `logger.warning`.

00yhj22-debug:
Rebased onto the latest `main` - the branch had gone stale and was showing a merge conflict after recent changes to `local.py` landed.

The only conflict was mechanical: `main` had reshuffled the surrounding code, and this branch deletes the now-dead `_warn_on_output_only_fields` helper. Resolved by keeping the deletion. The actual behaviour is unchanged from the previously reviewed state - `compiled_sql`/`freshness` are still rejected up front in `DbtLocalBaseOperator.__init__`, before the argspec-based kwargs segregation that would otherwise silently drop them.

No other code changes. CI should re-run on the rebased head. Happy to switch the guard to a warning instead of a hard error if you'd prefer the non-breaking path - just let me know.

pankajastro (commented):
LGTM. 

One thing to check: the guard is in `DbtLocalBaseOperator.__init__`, but the async operator `DbtRunAirflowAsyncBigqueryOperator` inherits `AbstractDbtLocalBase` directly and skips it. So passing  `compiled_sql` via `dbt_kwargs` there is still silently dropped — the same issue this PR fixes.

pankajastro (approved):
Created an issue https://github.com/astronomer/astronomer-cosmos/issues/2851 to track the async path

PR #2501, opened 2026-03-25
fix(config): unify postgres profiles.yml across dbt example projects

yeoreums:
## Description

This PR unifies the postgres_profile configuration between:

- dev/dags/dbt/jaffle_shop
- dev/dags/dbt/altered_jaffle_shop

Changes:
- Replaced incorrect 'pass' field with 'password' (dbt standard)
- Standardized port configuration using env_var
- Aligned both profiles.yml files for consistency

This ensures consistent configuration across examples and prevents potential dbt connection issues.

## Related Issue(s)

Closes #1384 

## Breaking Change?

No

## Checklist

- [x] I have made corresponding changes to the documentation (if required)
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #2735, opened 2026-05-27
fix: unify default and postgres_profile in example profiles.yml files

### LINKED ISSUE #1384: Unify `default` and `postgres_profile` in the example `profiles.yml`
Currently, both dbt projects:
* dev/dags/dbt/jaffle_shop
* dev/dags/dbt/altered_jaffle_shop

Have two profiles that are almost identical, as pointed out in: https://github.com/astronomer/astronomer-cosmos/pull/1374#discussion_r1881521297

The goal of this ticket is to unify them in both files.

Aaditya-git:
## Summary

Closes #1384

The `postgres_profile` in both example dbt projects was inconsistent with the `default` profile despite connecting to the same Postgres instance.

**Changes in both `dev/dags/dbt/jaffle_shop/profiles.yml` and `dev/dags/dbt/altered_jaffle_shop/profiles.yml`:**

| Field | Before | After |
|---|---|---|
| Password key | `pass:` | `password:` |
| Port | `5432` (hardcoded) | `env_var('POSTGRES_PORT') \| int` |
| Threads | missing | `threads: 4` |
| Field order | inconsistent | matches `default` |

Both profiles now use identical connection configuration, eliminating the divergence that caused unexpected behaviour when switching between them.

### DIFF
--- a/dev/dags/dbt/altered_jaffle_shop/profiles.yml
+++ b/dev/dags/dbt/altered_jaffle_shop/profiles.yml
@@ -16,9 +16,10 @@ postgres_profile:
   outputs:
     dev:
       type: postgres
-      dbname: "{{ env_var('POSTGRES_DB') }}"
       host: "{{ env_var('POSTGRES_HOST') }}"
-      pass: "{{ env_var('POSTGRES_PASSWORD') }}"
-      port: 5432 # "{{ env_var('POSTGRES_PORT') | as_number }}"
-      schema: "{{ env_var('POSTGRES_SCHEMA') }}"
       user: "{{ env_var('POSTGRES_USER') }}"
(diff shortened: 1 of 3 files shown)


tatiana:
Hi @Aaditya-git, thanks so much for your contribution here!

Another community member had already started work in this direction in #2501, and that PR has progressed quite a bit: https://github.com/astronomer/astronomer-cosmos/pull/2501

Given the overlap, I think it makes sense for us to continue the work there rather than duplicate efforts in this PR.

Would you be interested in picking up another ticket? One that could be particularly interesting, and that no one has started yet, is #2739: https://github.com/astronomer/astronomer-cosmos/issues/2739

If you’d be up for it, please comment on that issue so I can assign it to you. That should help us avoid a similar overlap in the future.

PR #2744, opened 2026-05-28
Forward Snowflake `insecure_mode` from connection Extra into dbt profile

### LINKED ISSUE #2740: Snowflake profile mappings drop `insecure_mode` from connection extras
## Description

The Cosmos Snowflake profile mappings convert an Airflow Snowflake connection into a dbt `profiles.yml`. The conversion is driven by each mapping's `airflow_param_mapping` allowlist — only fields named in that dict are forwarded into the rendered profile.

`insecure_mode` is not in any of the Snowflake mappings' allowlists. As a result, when a user sets `Insecure Mode: True` on their Airflow Snowflake connection (a first-class field of the `apache-airflow-providers-snowflake` connection form), Cosmos silently drops it from the dbt profile. The user gets no warning, and the resulting profile does not match the connection they configured.

The corresponding `apache-airflow-providers-snowflake` `SnowflakeHook` does pass `insecure_mode` through to the connector, so non-Cosmos Snowflake operators behave correctly with the same connection.

## Example mapping affected

For instance, [`SnowflakeEncryptedPrivateKeyPemProfileMapping.airflow_param_mapping`](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/profiles/snowflake/user_encrypted_privatekey_env_variable.py) forwards: `account`, `user`, `database`, `warehouse`, `schema`, `role`, `private_key`, `private_key_passphrase`, `query_tag`. The same gap exists in all Snowflake mappings under `cosmos/profiles/snowflake/`.

## Impact

`insecure_mode = True` is the documented way to bypass OCSP certificate-revocation checks against Snowflake. Users behind PrivateLink (or in any environment where the Snowflake OCSP responder is unreachable or slow) rely on it to connect at all. When Cosmos drops the field, those users see `250001: Could not connect to Snowflake backend after 11 attempt(s).Aborting` after a multi-minute hang — a failure mode that's hard to diagnose without already knowing about the allowlist.

Workaround today: set the value via `profile_args`, e.g.:

```python
profile_mapping=SnowflakeEncryptedPrivateKeyPemProfileMapping(
    conn_id="snowflake_conn_id",
    profile_args={"insecure_mode": True},
)
```

This works (covered by [#2738](https://github.com/astronomer/astronomer-cosmos/pull/2738)), but it requires users to know it.

## Proposed fix

Add `insecure_mode` to the `airflow_param_mapping` of each Snowflake profile mapping. The mapping value should be `extra.insecure_mode` so it reads from the connection's Extra JSON the same way other extras are read.

A separate, related concern: in `snowflake-connector-python` v3.14.1 (April 2025), `insecure_mode` was deprecated in favour of `disable_ocsp_checks` (same behaviour, just a renamed alias). The parameter has *not* been removed (still present through 4.5.0 on PyPI / 4.6.0 in the latest changelog), so forwarding `insecure_mode` continues to work. It's worth deciding whether Cosmos should:

- (a) forward `insecure_mode` as-is and let the connector emit its own deprecation warning, or
- (b) translate `insecure_mode: True` into `connect_args: { disable_ocsp_checks: true }` when generating the profile.

Option (a) is the minimum fix and matches how `SnowflakeHook` behaves. Option (b) is more user-friendly long-term but couples Cosmos to a moving target. See #_(next issue)_ for the broader allowlist audit.

## Acceptance criteria

- Every Snowflake profile mapping in `cosmos/profiles/snowflake/` forwards `insecure_mode` from the connection Extra into the rendered profile.
- A unit test covers the round-trip for at least the encrypted-PEM mapping (similar to `test_profile_args_insecure_mode` from #2738, but driving the value from the connection Extra rather than `profile_args`).
- No behaviour change for users who don't set `insecure_mode` on the connection.

toor11:
## Summary

The Snowflake profile mappings forward only the fields listed in each mapping's `airflow_param_mapping` allowlist. `insecure_mode` was missing from all of them, so when a user sets **Insecure Mode** on an Airflow Snowflake connection, Cosmos silently dropped it from the rendered dbt profile.

## Root Cause

`BaseProfileMapping.mapped_params` builds the profile only from keys present in `airflow_param_mapping` (`cosmos/profiles/base.py`). None of the five mappings under `cosmos/profiles/snowflake/` included `insecure_mode`, so the value never reached the profile — even though the upstream `SnowflakeHook` forwards it to the connector.

## Changes

- Add `"insecure_mode": "extra.insecure_mode"` to the `airflow_param_mapping` of all five Snowflake mappings, so the value is read from the connection Extra.
- Add two tests on the encrypted-PEM mapping: a connection-Extra round-trip, and an absence check confirming the key is omitted when unset.

## Why This Is Safe

All five Snowflake mappings already null-filter their profile (`filter_null`), and `get_dbt_value` skips falsy values. So `insecure_mode` that is unset or `False` is omitted from the profile — which matches `dbt-snowflake`'s default — meaning **no behaviour change** for connections that don't set it. Only an explicit `True` is forwarded.

## Testing

Full Snowflake profile suite (`tests/profiles/snowflake/`): **55 passed**. Black, Ruff, and mypy pass on the changed files.

### Reproduce bug (before fix)
```text
>   assert profile_mapping.profile["insecure_mode"] is True
E   KeyError: 'insecure_mode'
```

### Validate fix (after fix)
```text
55 passed in 1.21s
```

## Related

- Closes #2740
- Related: #2738 (covers the `profile_args` workaround; this PR adds the connection-Extra path requested in #2740's acceptance criteria)

### DIFF
--- a/tests/profiles/snowflake/test_snowflake_user_encrypted_privatekey_env_variable.py
+++ b/tests/profiles/snowflake/test_snowflake_user_encrypted_privatekey_env_variable.py
@@ -317,3 +317,43 @@ def test_old_snowflake_format() -> None:
             "warehouse": conn.extra_dejson.get("warehouse"),
             "threads": 4,
         }
+
+
+def test_insecure_mode_from_connection_extra() -> None:
+    """
+    Tests that `insecure_mode` set on the Airflow connection's Extra is forwarded to the profile.
+
+    `insecure_mode` is a first-class field of the Airflow Snowflake connection form. It must reach
+    the rendered dbt profile via `airflow_param_mapping` (not only via `profile_args`), otherwise
+    users behind PrivateLink who rely on it to bypass OCSP checks silently lose the setting.
+    """
+    conn = Connection(
+        conn_id="my_snowflake_pk_connection_insecure",
+        conn_type="snowflake",
+        login="my_user",
+        schema="my_schema",
+        extra=json.dumps(
+            {
+                "account": "my_account",
+                "database": "my_database",
+                "warehouse": "my_warehouse",
+                "private_key_content": "my_private_key",
+                "private_key_passphrase": "my_passphrase",
+                "insecure_mode": True,
+            }
+        ),
+    )
+
+    with patch("cosmos.profiles.base.BaseHook.get_connection", return_value=conn):
+        profile_mapping = SnowflakeEncryptedPrivateKeyPemProfileMapping(conn)
+        assert profile_mapping.profile["insecure_mode"] is True
+
+
+def test_insecure_mode_absent_when_not_set(
+    mock_snowflake_conn: Connection,
+) -> None:
+    """Tests that `insecure_mode` is omitted from the profile when it is not set on the connection."""
+    profile_mapping = get_automatic_profile_mapping(
+        mock_snowflake_conn.conn_id,
+    )
+    assert "insecure_mode" not in profile_mapping.profile
--- a/tests/profiles/snowflake/test_snowflake_user_encrypted_privatekey_env_variable.py
+++ b/tests/profiles/snowflake/test_snowflake_user_encrypted_privatekey_env_variable.py
@@ -9,7 +9,11 @@ from airflow.models.connection import Connection
 
 from cosmos.profiles import get_automatic_profile_mapping
 from cosmos.profiles.snowflake import (
+    SnowflakeEncryptedPrivateKeyFilePemProfileMapping,
     SnowflakeEncryptedPrivateKeyPemProfileMapping,
+    SnowflakePrivateKeyFilePemProfileMapping,
+    SnowflakePrivateKeyPemProfileMapping,
+    SnowflakeUserPasswordProfileMapping,
 )
 
 PLAIN_TEXT_KEY = "mocked-private-key-content"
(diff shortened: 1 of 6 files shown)


tests/profiles/snowflake/test_snowflake_user_encrypted_privatekey_env_variable.py
```diff
@@ -317,3 +317,43 @@ def test_old_snowflake_format() -> None:
...
+            }
+        ),
+    )
+
+    with patch("cosmos.profiles.base.BaseHook.get_connection", return_value=conn):
+        profile_mapping = SnowflakeEncryptedPrivateKeyPemProfileMapping(conn)
+        assert profile_mapping.profile["insecure_mode"] is True
+
+
+def test_insecure_mode_absent_when_not_set(
```
tatiana: Would it make sense to use Pytest parametrize (https://docs.pytest.org/en/stable/example/parametrize.html) to have this test run against all of the profile mappings, to make sure we don't have regressions in any of them in the future?

tatiana (commented):
@toor11 thank you very much for your contribution! I left a minor comment inline; if you could review it, it would be great. The implementation does give me the impression that we could maybe have a base snowflake profile mapping and avoid the duplicated logic in multiple places - what do you think? This could be a great work for a follow-up PR.

tests/profiles/snowflake/test_snowflake_user_encrypted_privatekey_env_variable.py
toor11: Great suggestion — done in 810e011. Both `test_insecure_mode_from_connection_extra` and `test_insecure_mode_absent_when_not_set` are now parametrized over all five Snowflake profile mappings (user/password, private-key PEM, private-key file, encrypted private-key file, encrypted private-key PEM), so a regression in any single mapping's `airflow_param_mapping` is caught. All 5×2 cases pass locally.

toor11:
Thanks for the review, @tatiana! 🙏

- **Parametrized tests** (inline comment): done in 810e011 — the `insecure_mode` forwarding/absence tests now run against all five Snowflake profile mappings.
- **Base Snowflake profile mapping**: I agree — a shared base would let us define `insecure_mode` (and the other common `airflow_param_mapping` entries like `role`/`query_tag`) in one place instead of repeating it across the five mappings. To keep this PR minimal and focused on the fix, I have not refactored it here, but I am happy to take it on as a follow-up. Want me to open a tracking issue for it?

tatiana:
> Thanks for the review, @tatiana! 🙏
> 
> * **Parametrized tests** (inline comment): done in [810e011](https://github.com/astronomer/astronomer-cosmos/commit/810e0114db682307d537578f6579ddca9aebd2b9) — the `insecure_mode` forwarding/absence tests now run against all five Snowflake profile mappings.

Thanks a lot, @toor11 , for addressing the feedback!

> * **Base Snowflake profile mapping**: I agree — a shared base would let us define `insecure_mode` (and the other common `airflow_param_mapping` entries like `role`/`query_tag`) in one place instead of repeating it across the five mappings. To keep this PR minimal and focused on the fix, I have not refactored it here, but I am happy to take it on as a follow-up. Want me to open a tracking issue for it?

Yes, please log a follow-up ticket and label it as a "good first issue"; that would be great. We can assign you to it, if you'd like to continue this work stream.

tatiana (approved):

PR #2006, opened 2025-09-26
Allowing exclusions to be passed through to the test tasks

anyapriya:
## Description

Previously, exclusions were not being passed through to test tasks when using test behavior of after each or after all.  This meant that if you were trying to exclude unit tests or a specific test, you were unable to do so easily without a workaround of also passing the exclusion into the operator args.  

While this would be great to also do something similar with the select or selector (ex. select model A and test 2), I think it would be too complex to get the unions / intersections correct.  I added documentation about this limitation so it's at least more clear.  

## Related Issue(s)

Closes #1763
Closes #1865

## Breaking Change?

No 

## Checklist

- [x] I have made corresponding changes to the documentation (if required)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #2746, opened 2026-05-29
Stop stale bot from auto-closing PRs

pankajkoti:
## Description

The stale workflow currently marks PRs stale after 30 days and then auto-closes them 10 days later (`days-before-pr-close: 10`). This closes contributor PRs that are otherwise healthy but simply haven't received a maintainer review, e.g. #2006.

This sets `days-before-pr-close: -1` so PRs are never auto-closed, mirroring the existing behavior for issues (`days-before-issue-close: -1`). PRs are still labeled `stale` after 30 days of inactivity to surface them, but they remain open. The now-unused `close-pr-message` is removed.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -16,10 +16,9 @@ jobs:
           stale-issue-label: "stale"
           stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
           days-before-pr-stale: 30
-          days-before-pr-close: 10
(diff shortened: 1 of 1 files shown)


tatiana (approved):
We had introduced this in https://github.com/astronomer/astronomer-cosmos/pull/1288/changes. I think this makes sense, Pankaj, for now, since we still want to do something with the open PRs. We can review this policy in the future.

PR #2747, opened 2026-05-30
fix: log redacted profile contents on cache miss (#2739)

### LINKED ISSUE #2739: Log rendered profile contents when profile caching is enabled
## Description

Cosmos converts an Airflow connection into a dbt `profiles.yml` via the configured profile mapping. The contents of the rendered profile are only logged at INFO when profile caching is **disabled** — when caching is enabled (the default, since `enable_cache_profile = True`), only the cache path is logged, not the contents.

This is a meaningful diagnostic gap: a substantial class of "Cosmos generates the wrong profile" or "Cosmos drops a connection field" issues can only be confirmed by inspecting the rendered file, which requires either:

1. Toggling `AIRFLOW__COSMOS__ENABLE_CACHE_PROFILE=False` and re-running, or
2. Shelling into the worker and `cat`-ing the file under `/tmp/cosmos/profile/<hash>/profiles.yml`.

Both are workable but slow, and on managed/serverless deployments (1) requires a redeploy and (2) is often impossible.

## Where this happens

[`RenderConfig.ensure_profile`](https://github.com/astronomer/astronomer-cosmos/blob/main/cosmos/config.py) (around lines 373-411 at the time of writing):

- **Non-cached branch** logs the full contents via `logger.info("... with the following contents:\n%s", profile_contents)`.
- **Cached branch** logs only `"Profile not found in cache storing and using profile: %s."` — the contents are silently elided.

## Proposed fix

Log the rendered profile contents on the cached path too, with secret fields redacted (e.g., replace values for fields listed in the mapping's `secret_fields` with `***`). The log line should appear once per cache miss, at the same level as the non-cached branch (INFO).

## Why this matters

Without this, every Cosmos profile-mapping troubleshooting cycle requires a back-and-forth with the user to either flip a config flag or pull the file off the worker. Adding the redacted contents to the standard log surfaces them immediately and saves a round-trip in support / debugging contexts.

## Acceptance criteria

- When profile caching is enabled and a cache miss occurs, the rendered `profiles.yml` contents are logged at INFO.
- Values for fields in the mapping's `secret_fields` are redacted before logging.
- The existing non-cached branch behaviour is unchanged.
- Cache *hits* may also log the path (already happens) — but logging contents on a hit is optional.

Aaditya-git:
## Summary

Fixes #2739.

When profile caching is enabled and a cache miss occurs, the rendered `profiles.yml` contents are now logged at INFO with secret fields redacted (`***`), matching the behaviour of the non-cached branch.

**Before:** only the cache path was logged on a miss, making it hard to diagnose profile mapping issues without toggling `ENABLE_CACHE_PROFILE=False` or shelling into the worker.

**After:** the full profile contents are logged once per cache miss, with any fields listed in `profile_mapping.secret_fields` replaced by `***`.

## Changes

- `cosmos/config.py`: on cache miss, load the profile YAML, redact secret fields, and include the redacted contents in the existing INFO log line
- `tests/test_config.py`: new test `test_profile_config_logs_redacted_contents_on_cache_miss` verifies secrets are redacted and not leaked in the log output

## Test plan

- [x] `test_profile_config_logs_redacted_contents_on_cache_miss` passes
- [x] Full `tests/test_config.py` suite passes (47/47)
- [x] Non-cached branch behaviour unchanged


cosmos/config.py
```diff
@@ -366,7 +366,18 @@ def _get_profile_path(self, use_mock_values: bool = False) -> Path:
                 profile_name=self.profile_name, target_name=self.target_name, use_mock_values=use_mock_values
             )
             profile_path = create_cache_profile(current_profile_version, profile_contents)
-            logger.info("Profile not found in cache storing and using profile: %s.", profile_path)
+            if isinstance(profile_contents, str):
+                profile_data = yaml.safe_load(profile_contents)
+                for field in self.profile_mapping.secret_fields:
+                    profile_data[self.profile_name]["outputs"][self.target_name][field] = "***"
```
pankajkoti, line 389: This redaction reaches into the `profiles.yml` dict shape (`profile_data[self.profile_name]["outputs"][self.target_name]`) that is actually constructed in `BaseProfileMapping.get_profile_file_contents` (`base.py:264-272`). If that structure changes, this breaks silently here. Consider moving redaction into a helper on the profile mapping, e.g. `get_redacted_profile_file_contents(...)`, so the structural knowledge stays in one place and can be reused.

Also: `secret_fields` are rendered as `{{ env_var(...) }}` references, not literal values (`postgres/user_pass.py:47` -> `get_env_var_format`), and `filter_null` drops null entries. So for a secret field that is not present in the rendered output, `profile_data[...][field] = "***"` *adds* a phantom `password: '***'` line to the log that was not actually in the profile. Suggest guarding the assignment, e.g. `if field in outputs: outputs[field] = "***"`.

cosmos/config.py
```diff
@@ -366,7 +366,18 @@ def _get_profile_path(self, use_mock_values: bool = False) -> Path:
...
+                redacted_contents = yaml.dump(profile_data, indent=4)
+                logger.info(
+                    "Profile not found in cache, storing and using profile: %s with the following contents:\n%s",
+                    profile_path,
+                    redacted_contents,
+                )
+            else:
+                logger.info("Profile not found in cache storing and using profile: %s.", profile_path)
```
pankajkoti, line 397: `get_profile_file_contents` is annotated `-> str` and always returns `str(yaml.dump(...))` (`base.py:274`), so this `else` is effectively unreachable. If kept as a defensive guard, note it still carries the old message `"...cache storing..."` (missing comma) while the new branch uses `"...cache, storing..."`, so the two diverge. Either align the wording or drop the dead branch.

cosmos/config.py
```diff
@@ -366,7 +366,18 @@ def _get_profile_path(self, use_mock_values: bool = False) -> Path:
                 profile_name=self.profile_name, target_name=self.target_name, use_mock_values=use_mock_values
             )
             profile_path = create_cache_profile(current_profile_version, profile_contents)
-            logger.info("Profile not found in cache storing and using profile: %s.", profile_path)
+            if isinstance(profile_contents, str):
```
pankajkoti, line 386: The PR description says this "matches the behaviour of the non-cached branch," but the non-cached branch logs `profile_contents` **unredacted** (`config.py:410-413` and `421-426`). After this change the cached path redacts and the non-cached path does not, which is the opposite of parity. If redaction is the goal, consider factoring a single helper and applying it at all three log sites so behaviour is consistent.

tests/test_config.py
```diff
@@ -215,6 +215,41 @@ def test_profile_config_ensure_profile_with_caching_calls_get_profile_file_conte
...
+            profile_name="test", target_name="cosmos_target", profile_mapping=profile_mapping
+        )
+
+        mock_create_cache_profile.return_value = Path("/tmp/cosmos/profile/abc/profiles.yml")
+
+        profile_config._get_profile_path(use_mock_values=False)
+
+    log_call_args = " ".join(str(a) for a in mock_logger.info.call_args_list)
+    assert "***" in log_call_args
+    assert "supersecret" not in log_call_args
```
pankajkoti, line 250: `assert "supersecret" not in log_call_args` is vacuously true: the raw password lives in `env_vars`, never in `get_profile_file_contents` output (it is rendered as `{{ env_var('COSMOS_CONN_TEST_PASSWORD') }}`), so it could never appear in this log line, and the assertion would pass even if the redaction loop were deleted. To actually guard redaction, assert the env-var reference is gone and `***` is present, e.g.:

```python
assert "env_var(" not in log_call_args
assert "***" in log_call_args
```

Worth adding a case for a `secret_fields` entry that is not in the rendered output (the phantom-key scenario) too.

pankajkoti (commented):
A few notes on the redaction approach, placement, parity with the non-cached branch, and the test assertions. Details inline.

tatiana:
@Aaditya-git when you get a chance, it would be great if you could address the feedback. I feel we're really close to merging this change, and if we're able to get it until tomorrow, we may be able to ship it as part of Cosmos 1.15.0, to be released on 1st July

PR #2748, opened 2026-05-31
feat: extend Snowflake profile mappings with authenticator, client_session_keep_alive, host, and port

### LINKED ISSUE #2741: Audit Snowflake profile mapping allowlists for missing connection parameters
## Description

Cosmos's Snowflake profile mappings convert an Airflow Snowflake connection into a dbt `profiles.yml`, but each mapping's `airflow_param_mapping` is a narrow, fixed allowlist of fields. Anything in the Airflow connection's Extra that isn't named in the allowlist is silently dropped on the way to the dbt profile.

Compared with what the Airflow `SnowflakeHook` accepts and forwards to `snowflake-connector-python`, the Cosmos mappings are missing several first-class connection extras that real users rely on. The result is a recurring pattern of "works with the non-Cosmos Snowflake operator but breaks in Cosmos" tickets, where the underlying cause is "field X never made it into the rendered profile."

#2740 covers the most acute case (`insecure_mode`). This issue tracks a broader audit and extension.

## Mappings in scope

All files under [`cosmos/profiles/snowflake/`](https://github.com/astronomer/astronomer-cosmos/tree/main/cosmos/profiles/snowflake):

- `user_pass.py` (`SnowflakeUserPasswordProfileMapping`)
- `user_privatekey.py` (`SnowflakePrivateKeyPemProfileMapping`)
- `user_privatekey_file.py` (`SnowflakePrivateKeyPemFileProfileMapping`)
- `user_encrypted_privatekey.py` / `user_encrypted_privatekey_env_variable.py` (`SnowflakeEncryptedPrivateKey*Mapping`)
- `user_encrypted_privatekey_file.py` (`SnowflakeEncryptedPrivateKeyFilePemProfileMapping`)

## Candidate parameters to forward

Per the Snowflake provider connection form and the underlying connector's documented parameters, the following are reasonable candidates. Each should be verified against the dbt-snowflake adapter's accepted profile fields (some are passed as top-level keys, others belong inside a `connect_args` block) before adding:

- `insecure_mode` (tracked separately in #2740)
- `authenticator` — required for SSO / OAuth setups
- `application` — labels query history in Snowflake; commonly set by integrators
- `host` / `port` — explicit overrides for testing or non-standard endpoints
- `ocsp_fail_open` — softer OCSP behaviour (warn instead of fail)
- `client_session_keep_alive` — long-running jobs
- `login_timeout` — environments with slow auth paths
- `network_timeout` — same rationale
- `client_request_mfa_token` / `client_store_temporary_credential` — federated auth flows

Some of these need to land in `connect_args` rather than as top-level profile keys (e.g., `disable_ocsp_checks`, which is the modern replacement for `insecure_mode` per snowflake-connector-python v3.14.1+).

## Suggested approach

Two possible directions, not mutually exclusive:

1. **Allowlist extension** — add each verified field to the appropriate mapping's `airflow_param_mapping`. Pro: explicit and reviewable. Con: requires per-field maintenance and may lag the upstream connector.
2. **Pass-through of unknown extras** — opt-in mechanism (e.g., a class attribute `forward_unknown_extras: bool` or a per-instance constructor argument) that copies any unrecognized keys in the connection Extra straight into the profile (or into a `connect_args` block). Pro: future-proof. Con: risks rendering invalid profiles if dbt-snowflake doesn't accept a given key.

Worth discussing on this issue before implementation.

## Acceptance criteria

- A documented decision on the approach (allowlist extension vs. pass-through, or a hybrid).
- The agreed-upon fields are forwarded from the Snowflake connection into the rendered profile for every Snowflake mapping.
- Unit tests cover the new fields with the connection-Extra → profile path.
- Documentation under [`docs/profiles/`](https://github.com/astronomer/astronomer-cosmos/tree/main/docs/profiles) notes which Snowflake connection fields are supported.

## Related

- #2738 (unit tests pinning `insecure_mode` / `connect_args` propagation via `profile_args`)
- #2740 (forward `insecure_mode` from the connection — narrowest immediate fix)
- #2739 (log rendered profile when caching is enabled — diagnostic gap, separate concern)

Aaditya-git:
## Summary

Closes #2741

Extends all five Snowflake profile mappings with missing connection parameters that are silently dropped today, causing "works outside Cosmos, breaks inside Cosmos" failures.

### Fields added

| dbt profile field | Airflow connection extra | Applies to |
|---|---|---|
| `authenticator` | `extra.authenticator` | All 5 mappings |
| `client_session_keep_alive` | `extra.client_session_keep_alive` | All 5 mappings |
| `host` | `extra.host` | 4 mappings (already present in `user_pass`) |
| `port` | `extra.port` | 4 mappings (already present in `user_pass`) |

### Decision: allowlist extension (approach a)

Added each verified field explicitly to each mapping's `airflow_param_mapping`. All four fields are accepted as top-level keys by dbt-snowflake and forwarded by the Airflow `SnowflakeHook`, so the mapping is safe and straightforward.

Fields that require `connect_args` nesting (e.g., `disable_ocsp_checks`, `login_timeout`, `network_timeout`) are deferred — they need additional handling in the profile render logic and are out of scope for this PR.

`insecure_mode` is excluded because PR #2744 already covers it.

## Test plan

- 69 existing Snowflake profile tests still pass (`pytest tests/profiles/snowflake/ -x -q`)
- New tests added to all 5 test files covering:
  - `authenticator` present in profile when set
  - `authenticator` absent from profile when not set (user_pass only, mirrors existing pattern)
  - `client_session_keep_alive` present in profile when set
  - `client_session_keep_alive` absent from profile when not set (user_pass only)
  - `host` and `port` forwarded (for the 4 mappings that didn't have them)


tatiana (approved):
Thanks for the contribution, @Aaditya-git !

tatiana (commented):
@Aaditya-git I feel like we have a lot of duplicate code in this PR, particularly the tests. Could we leverage `pytest.parametrize` and follow a similar strategy to https://github.com/astronomer/astronomer-cosmos/pull/2744/changes?

Aaditya-git:
> @Aaditya-git I feel like we have a lot of duplicate code in this PR, particularly the tests. Could we leverage `pytest.parametrize` and follow a similar strategy to https://github.com/astronomer/astronomer-cosmos/pull/2744/changes?

Thanks for the feedback @tatiana! Good call, I'll refactor the new tests to use pytest.parametrize following the same pattern as #2744. Will push the update shortly.

Aaditya-git:
> > @Aaditya-git I feel like we have a lot of duplicate code in this PR, particularly the tests. Could we leverage `pytest.parametrize` and follow a similar strategy to https://github.com/astronomer/astronomer-cosmos/pull/2744/changes?
> 
> Thanks for the feedback @tatiana! Good call, I'll refactor the new tests to use pytest.parametrize following the same pattern as #2744. Will push the update shortly.

Since I was adding fields to all 5 mappings simultaneously, I wrote the tests as I went, one per mapping file, and ended up around 15 near-identical test functions before realizing the duplication.

tatiana (approved):
@Aaditya-git thanks a lot for addressing the feedback!

Please, could you rebase? The `main` branch changed a bit and there are a few conflicts.

One point that could be further improved would be to create an abstraction and have a base class, so we avoid all the code duplication, as discussed in: https://github.com/astronomer/astronomer-cosmos/pull/2744#pullrequestreview-4389750073

That said, I believe @toor11 will be taking this over (https://github.com/astronomer/astronomer-cosmos/pull/2744#issuecomment-4587900083), so don't worry about it.

tatiana (changes requested):
Hey @Aaditya-git! Thanks so much for this PR - really appreciate the work you've put into it! We're excited to get it merged. Could you do a quick rebase to resolve the merge conflicts when you get a chance? Let us know if you need any help with that!

Aaditya-git:
> Hey @Aaditya-git! Thanks so much for this PR - really appreciate the work you've put into it! We're excited to get it merged. Could you do a quick rebase to resolve the merge conflicts when you get a chance? Let us know if you need any help with that!

Rebased onto main and resolved all merge conflicts. Here is what changed during the rebase:
  
The upstream main branch had added insecure_mode support across all five Snowflake profile mappings and introduced parametrized tests for it in test_snowflake_user_encrypted_privatekey_env_variable.py. Our branch added client_session_keep_alive, authenticator, host, and port across the same five mappings.

All conflicts were purely additive, both sides added new fields to airflow_param_mapping. The resolution keeps both insecure_mode (from upstream) and client_session_keep_alive (from this PR) in all five mapping files.

For the test file, the upstream insecure_mode parametrized tests and our refactored cross-mapping parametrized tests are both preserved. The _build_snowflake_conn helper was updated to use **kwargs instead of a fixed insecure_mode parameter, which makes it flexible enough to support any field including insecure_mode=True, authenticator, client_session_keep_alive, host, and port without changing the call sites.

All Snowflake profile tests pass locally after the rebase.

tatiana (approved):
Thanks for improving this, @Aaditya-git !

PR #2246, opened 2026-01-05
Add seed rendering behavior support in Cosmos

tuantran0910:
## Description

This PR introduces a new `SeedRenderingBehavior` configuration option that allows users to control how dbt seed nodes are rendered in Airflow DAGs. Previously, Cosmos would always render seeds and attempt to run `dbt seed`. This new feature provides flexibility similar to how `SourceRenderingBehavior` works for source nodes.

### Motivation

In most production scenarios, seeds do not need to be continuously run on every DAG execution. This feature allows users to:
- Skip seed rendering entirely when seeds are managed outside of Cosmos
- Only run seeds when the underlying CSV file has changed, optimizing DAG execution time
- Maintain backward compatibility for users who want the original behavior

### New Configuration Options

The `seed_rendering_behavior` parameter in `RenderConfig` accepts:

| Value | Description |
|-------|-------------|
| `ALWAYS` | Always render and run seeds (default behavior) |
| `NONE` | Don't render any seeds in DAG/TaskGroup |
| `WHEN_SEED_CHANGES` | Only execute seeds if the CSV file has changed since last successful run |

### How `WHEN_SEED_CHANGES` Works

1. At task execution time, computes a SHA256 hash of the seed CSV file
2. Compares against the stored hash (persisted as an Airflow Variable)
3. Skips execution if hashes match; runs the seed if they differ
4. Stores the new hash after successful execution

### Test Behavior

Test behavior for seeds is now controlled **exclusively via `TestBehavior`**, not through `SeedRenderingBehavior`. This separation of concerns simplifies configuration:

```python
# Run seeds with tests
RenderConfig(
    seed_rendering_behavior=SeedRenderingBehavior.ALWAYS,
    test_behavior=TestBehavior.AFTER_EACH,
)

# Run seeds without tests
RenderConfig(
    seed_rendering_behavior=SeedRenderingBehavior.ALWAYS,
    test_behavior=TestBehavior.NONE,
)
```

### Example Usage

```python
import datetime
from pathlib import Path

from airflow.providers.standard.operators.empty import EmptyOperator
from airflow.sdk import DAG, Asset, Param
from cosmos import DbtTaskGroup, ProjectConfig, ProfileConfig, ExecutionConfig, RenderConfig
from cosmos.constants import TestBehavior, ExecutionMode, SeedRenderingBehavior


# Paths configuration
DAGS_DIR = Path("/opt/airflow/dags")
[...]

PR #2755, opened 2026-06-02
Add configurable seed rendering behavior

### LINKED ISSUE #1576: [Feature] Improve how Cosmos renders & runs seeds
### Description

Currently, if the dbt project contains seeds, Cosmos will render them and attempt to run `dbt seed` unless the end-user customises how Seeds should be rendered:
https://astronomer.github.io/astronomer-cosmos/configuration/render-config.html#customizing-how-nodes-are-rendered-experimental

However, in most cases, seeds do not need to be continuously run.

It would be great if Cosmos could allow users to opt for different ways of rendering/running seeds, similar to [source nodes](https://astronomer.github.io/astronomer-cosmos/configuration/source-nodes-rendering.html):

- `SeedRenderingBehavior.ALWAYS`: Current behaviour of Cosmos, always add the seed tasks in the DAG/TaskGroup and run seeds
- `SeedRenderingBehavior.NONE`: New behavior, don't render any seeds in DAG/TaskGroup 
- `SeedRenderingBehavior.WHEN_SEED_CHANGES`: Only render and run a seed if the `csv` file has changed since the last execution

There are a few open questions:
- Should we still render seeds so users understand the topology, even if we don't run them in `SeedRenderingBehavior.NONE` and `SeedRenderingBehavior.WHEN_SEED_CHANGES`, and use an `EmptyOperator` in those cases?
- Where could we store the state if a `.csv` was run and if it changed? As an Airflow variable? In a remote object store? Somewherelse..?

I'd love your thoughts on these ideas.

### Use case/motivation

Allow users to opt out of running seeds and only run seeds when needed.

### Related issues

_No response_

pankajkoti:
## Description

Adds a `seed_rendering_behavior` option to `RenderConfig`, giving control over how Cosmos renders and runs dbt seeds (analogous to `source_rendering_behavior`):

- `ALWAYS` (default): render the seed and run `dbt seed` on every execution (the original Cosmos behaviour).
- `WHEN_SEED_CHANGES`: render the seed, but only run `dbt seed` when its CSV content has changed since the last successful run. The checksum is read from dbt's manifest (falling back to hashing the CSV) and the last-seen value is persisted as an Airflow Variable scoped per `DbtDag`/`DbtTaskGroup` and seed. Supported for `ExecutionMode.LOCAL`, `VIRTUALENV` and `AIRFLOW_ASYNC`, and incompatible with `TestBehavior.BUILD`; both raise `CosmosValueError`.
- `RENDER_ONLY`: render the seed as a no-op `EmptyOperator` placeholder.
- `NONE`: do not render the seed at all.

The gate runs before the `TestBehavior.BUILD` branch so `NONE`/`RENDER_ONLY` are honoured under every test behaviour. On an unchanged run the seed task succeeds without running `dbt seed`, without emitting its dataset, and without skip-propagating downstream. Change detection delegates to `AbstractDbtBase.execute()`, preserving `extra_context`, debug-mode tracking and `**kwargs`.

Under `ExecutionMode.WATCHER` a single `dbt build` runs all seeds regardless of this setting; only `ALWAYS` is meaningful there.

Fresh implementation building on the approach explored in #2246 by @tuantran0910.

## Related Issue(s)

Closes #1576


pankajkoti:
The PR is ready for review. Please do not merge this PR. It relies on https://github.com/astronomer/astronomer-cosmos/pull/2758 to be merged first.

tuantran0910:
Hi @pankajkoti, thank you very much for helping me continue with the PR adding seed rendering behavior. I am so sorry that I don't have time to continue putting effort in my original PR because I was so busy with my work. I am happy to see this PR and looking forward to seeing it merged in the future.

cosmos/airflow/graph.py
```diff
@@ -365,6 +367,32 @@ def create_task_metadata(  # noqa: C901
...
+                    resource_suffix=node.resource_type.value,
+                    execution_mode=execution_mode,
+                )
+                # EmptyOperator does not accept custom dbt parameters (e.g. profile_args); recreate the args.
+                args = {"task_display_name": args["task_display_name"]} if "task_display_name" in args else {}
+                return TaskMetadata(id=task_id, operator_class=EMPTY_OPERATOR_CLASS, arguments=args)
+            if render_config.seed_rendering_behavior == SeedRenderingBehavior.WHEN_SEED_CHANGES:
+                # Render the seed as usual, but flag the task so the operator skips running
+                # `dbt seed` when the CSV content is unchanged since the last successful run.
+                extra_context["uses_seed_change_detection"] = True
```
```suggestion
                extra_context["should_run_if_seed_changed"] = True
```

cosmos/dbt/seed.py
```diff
@@ -0,0 +1,99 @@
...
+_VARIABLE_KEY_PREFIX = "cosmos_seed_checksum"
+
+
+def _seed_variable_key(dag_task_group_identifier: str, unique_id: str) -> str:
+    """Return a bounded, stable Airflow Variable key scoped to a DbtDag/DbtTaskGroup and seed."""
+    digest = hashlib.sha1(f"{dag_task_group_identifier}:{unique_id}".encode()).hexdigest()
+    return f"{_VARIABLE_KEY_PREFIX}__{digest}"
+
+
+def _compute_csv_checksum(seed_file_path: str | Path) -> str | None:
```
tatiana: I strongly suggest we add a checksum property to `DbtNode` and define it there. Even though `manifest.json` includes a checksum, I believe we should always calculate it from the seed node. This way, we'd have a consistent behaviour if users set `SeedRenderingBehavior.WHEN_SEED_CHANGES` and switch between `LoadMode.MANIFEST` and `LoadMode.DBT_LS`,

cosmos/dbt/graph.py
```diff
@@ -106,6 +106,9 @@ class DbtNode:
     has_non_detached_test: bool = False
     downstream: list[str] = field(default_factory=lambda: [])
     fqn: list[str] | None = None
+    # dbt records a content checksum per node in the manifest (e.g. sha256 of a seed's CSV). Used by
+    # SeedRenderingBehavior.WHEN_SEED_CHANGES to detect seed changes. None when loading via dbt ls.
+    checksum: str | None = None
```
tatiana: I believe we should make this a property of `DbtNode`, bringing the logic currently defined in https://github.com/astronomer/astronomer-cosmos/pull/2755/changes#diff-e79f48cd2b08a9aab68ee4ce5346072726e24910cd374c3ef847c8b9825179f1R36 into the node itself.

cosmos/dbt/seed.py
```diff
@@ -0,0 +1,99 @@
+"""Utilities supporting :attr:`~cosmos.constants.SeedRenderingBehavior.WHEN_SEED_CHANGES`.
```
tatiana: I vote that we keep cache-related logic in `cache.py` and the checksum in the `DbtNode`. I don't think we should have a new module called `seed.py` in Cosmos. An improvement I believe we could make in the future is to make `cache.py` a package with dedicated files per resource we're caching, but I believe this is outside the scope of the current PR.

cosmos/dbt/graph.py
```diff
@@ -200,6 +203,7 @@ def context_dict(self) -> dict[str, Any]:
             "has_non_detached_test": self.has_non_detached_test,
             "resource_name": self.resource_name,
             "name": self.name,
+            "checksum": self.checksum,
```
tatiana, line 273: If we follow https://github.com/astronomer/astronomer-cosmos/pull/2755/changes#r3356041356, this is not needed

cosmos/dbt/graph.py
```diff
@@ -363,6 +367,7 @@ def parse_dbt_ls_output(project_path: Path | None, ls_stdout: str) -> dict[str,
                         else False
                     ),
                     fqn=node_dict.get("fqn"),
+                    checksum=(node_dict.get("checksum") or {}).get("checksum"),
```
tatiana: If we follow https://github.com/astronomer/astronomer-cosmos/pull/2755/changes#r3356041356, this is not needed

cosmos/dbt/graph.py
```diff
@@ -412,6 +417,7 @@ def _build_dbt_node_from_manifest_resource(
             is_freshness_effective(node_dict.get("freshness")) if resource_type == DbtResourceType.SOURCE else False
         ),
         fqn=node_dict.get("fqn"),
+        checksum=(node_dict.get("checksum") or {}).get("checksum"),
```
tatiana: If we follow https://github.com/astronomer/astronomer-cosmos/pull/2755/changes#r3356041356, this is not needed

cosmos/dbt/seed.py
```diff
@@ -0,0 +1,99 @@
...
+    if current is None or not dag_task_group_identifier or not unique_id:
+        logger.info(
+            "Seed change detection could not resolve a checksum for seed `%s` (checksum resolved: %s); "
+            "running `dbt seed` without change detection.",
+            unique_id,
+            current is not None,
+        )
+        return True, None, None
+
+    variable_key = _seed_variable_key(dag_task_group_identifier, unique_id)
```
tatiana: What will happen if there are two concurrent DAG Runs for the same DAG?

pankajkoti:
When `seed_rendering_behavior=NONE`, no seeds rendered
Screenshot 2026-06-02 at 4 35 24 PM

When `seed_rendering_behavior=RENDER_ONLY`, rendered as empty operators
Screenshot 2026-06-02 at 4 36 46 PM

When `seed_rendering_behavior=WHEN_SEED_CHANGES`, and seed has not changed by checking the checksum in Variables, does not re-run seed
Screenshot 2026-06-02 at 4 42 37 PM
Screenshot 2026-06-02 at 4 45 30 PM

When `seed_rendering_behavior=WHEN_SEED_CHANGES`, and `TEST_BEHAVIOR=BUILD`, incompatible
Screenshot 2026-06-02 at 4 48 43 PM

`TEST_BEHAVIOR=BUILD` continues to work with `seed_rendering_behavior=ALWAYS` (the default)
Screenshot 2026-06-02 at 4 49 22 PM

cosmos/operators/local.py
```diff
@@ -1036,6 +1037,40 @@ class DbtSeedLocalOperator(DbtSeedMixin, DbtLocalBaseOperator):
...
+        # request always runs the seed, bypassing change detection.
+        extra_context = self.extra_context or {}
+        if extra_context.get("uses_seed_change_detection") and not self._is_full_refresh():
+            node_config = extra_context.get("dbt_node_config") or {}
+            should_run, checksum, variable_key = _evaluate_seed_change(
+                extra_context.get("dbt_dag_task_group_identifier"),
+                node_config.get("unique_id"),
+                node_config.get("checksum"),
+                node_config.get("file_path"),
+            )
```
tatiana: I believe this interface would make the code cleaner and easier to maintain:
```
should_run = False
last_executed_seed_checksum = ...
current_seed_checksum = node_config.get("checksum")
if last_run_seed_checksum != current_seed_checksum:
    should_run = True
```

tatiana (commented):
@pankajkoti, thanks a lot for this feature; it will benefit many users.

I left some inline feedback.

Regarding this PR comment:
> Under `ExecutionMode.WATCHER`, a single dbt build runs all seeds regardless of this setting; only ALWAYS is meaningful there.

We could adopt a similar strategy to the one we used to add support for `SourceRenderingBehavior` in `ExecutionMode.WATCHER`. The work was implemented/improved throughout a few PRs:
- https://github.com/astronomer/astronomer-cosmos/pull/2467
- https://github.com/astronomer/astronomer-cosmos/pull/2563
- https://github.com/astronomer/astronomer-cosmos/pull/2570
- https://github.com/astronomer/astronomer-cosmos/pull/2572

WDYT of logging a follow-up ticket to add support for `SeedRenderingBehavior` in `ExecutionMode.WATCHER`?

cosmos/dbt/graph.py
pankajkoti: Done. `DbtNode.checksum` is now a property that always recomputes the SHA256 from the seed CSV (streamed in chunks), so the value stays consistent across `LoadMode.MANIFEST` and `LoadMode.DBT_LS`. Removed the dataclass field and both constructor sites that populated it from the manifest.

cosmos/dbt/seed.py
pankajkoti: Done. The checksum is now a property on `DbtNode` and is always calculated from the seed file rather than read from the manifest, so it stays consistent when switching between `LoadMode.MANIFEST` and `LoadMode.DBT_LS`.

cosmos/dbt/seed.py
pankajkoti: `cosmos/dbt/seed.py` is removed. The checksum now lives on `DbtNode`, and the Airflow Variable persistence (`get_seed_checksum`/`store_seed_checksum` plus key generation) now lives in `cache.py`.

cosmos/dbt/graph.py
pankajkoti: I kept `checksum` in `context_dict`: the operator still needs the render time checksum to compare against the stored one at run time (as in your `local.py` suggestion, `current_seed_checksum = node_config.get("checksum")`). It is now property derived (always from the file) rather than manifest derived.

One tradeoff to flag: this hashes seeds at parse time. That is cheap for typical seeds, but I am happy to move the computation to runtime (during task execution) if you would rather avoid parse-time I/O on the default `ALWAYS` path.

cosmos/dbt/graph.py
pankajkoti: Removed. The constructor no longer populates `checksum`; it is computed by the property.

cosmos/dbt/graph.py
pankajkoti: Removed. The constructor no longer populates `checksum`; it is computed by the property.

cosmos/operators/local.py
pankajkoti: Adopted. `execute()` now uses the `should_run` flow with `last_run_seed_checksum != current_seed_checksum`.

cosmos/dbt/seed.py
pankajkoti: Change detection is best effort and the checksum is persisted only after `dbt seed` succeeds. Two concurrent runs both read the same stored value, and on a changed CSV both re-run `dbt seed` (idempotent, it reloads the table) and write the same new checksum. The variable is never advanced before the data is actually loaded, so a run cannot be wrongly skipped by an in flight sibling. Worst case is a redundant `dbt seed`, never a missed load.

cosmos/airflow/graph.py
pankajkoti: Applied, renamed to `should_run_if_seed_changed`.

cosmos/dbt/seed.py
```diff
@@ -0,0 +1,99 @@
...
+    stored = Variable.get(variable_key, default_var=None)
+    return stored != current, current, variable_key
```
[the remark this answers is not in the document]
pankajkoti: `get_seed_checksum()` in `cache.py` now catches `AirflowException` and `SQLAlchemyError`, logs a warning, and returns `None` so the seed runs. This is symmetric with the write path.

cosmos/dbt/seed.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def _compute_csv_checksum(seed_file_path: str | Path) -> str | None:
+    """Return the SHA256 checksum of a seed CSV file, or ``None`` if it cannot be read."""
+    try:
+        content = Path(seed_file_path).read_bytes()
+    except OSError as exc:
+        logger.warning("Unable to read seed file `%s` for change detection: %s", seed_file_path, exc)
+        return None
+    return hashlib.sha256(content).hexdigest()
```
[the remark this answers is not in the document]
pankajkoti: Done. The checksum now streams the file in 1 MiB chunks via `_calculate_file_checksum` in `graph.py`.

cosmos/operators/local.py
```diff
@@ -1036,6 +1037,40 @@ class DbtSeedLocalOperator(DbtSeedMixin, DbtLocalBaseOperator):
...
+            super().execute(context, **kwargs)
+            _persist_seed_checksum(variable_key, checksum)
+            return
+        super().execute(context, **kwargs)
+
+    def _is_full_refresh(self) -> bool:
+        """Return whether --full-refresh is requested, mirroring DbtSeedMixin.add_cmd_flags handling."""
+        if isinstance(self.full_refresh, str):
+            return to_boolean(self.full_refresh)
+        return bool(self.full_refresh)
```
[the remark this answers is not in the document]
pankajkoti: Fixed. `_is_full_refresh()` now calls `to_boolean(self.full_refresh.strip())`, mirroring `_should_install_deps`.

cosmos/airflow/graph.py
```diff
@@ -366,6 +367,32 @@ def create_task_metadata(  # noqa: C901
...
+                    args=args,
+                    use_task_group=use_task_group,
+                    normalize_task_id=render_config.normalize_task_id,
+                    normalize_task_display_name=render_config.normalize_task_display_name,
+                    resource_suffix=node.resource_type.value,
+                    execution_mode=execution_mode,
+                )
+                # EmptyOperator does not accept custom dbt parameters (e.g. profile_args); recreate the args.
+                args = {"task_display_name": args["task_display_name"]} if "task_display_name" in args else {}
+                return TaskMetadata(id=task_id, operator_class=EMPTY_OPERATOR_CLASS_PATH, arguments=args)
```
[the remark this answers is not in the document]
pankajkoti: This matches the existing source to EmptyOperator path (the `SourceRenderingBehavior.ALL` branch in `graph.py`), which recreates the args the same way. Preserving BaseOperator kwargs on placeholder tasks would be a separate change spanning both the source and seed paths, so I would prefer to handle it as a follow-up if we want that behavior.

cosmos/converter.py
```diff
@@ -130,6 +131,19 @@ def validate_arguments(
...
+    if render_config.seed_rendering_behavior == SeedRenderingBehavior.WHEN_SEED_CHANGES and (
+        execution_config.execution_mode
+        not in (ExecutionMode.LOCAL, ExecutionMode.VIRTUALENV, ExecutionMode.AIRFLOW_ASYNC)
+    ):
+        raise CosmosValueError(
+            "SeedRenderingBehavior.WHEN_SEED_CHANGES is only supported with ExecutionMode.LOCAL, "
+            "ExecutionMode.VIRTUALENV or ExecutionMode.AIRFLOW_ASYNC, which run dbt directly on the "
+            f"Airflow worker. The configured execution_mode is {execution_config.execution_mode}. Use "
+            "SeedRenderingBehavior.ALWAYS, or switch to a supported execution mode."
+        )
```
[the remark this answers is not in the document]
pankajkoti: the current messaging is correct.

cosmos/airflow/graph.py
```diff
@@ -365,6 +367,32 @@ def create_task_metadata(  # noqa: C901
...
+                    args=args,
+                    use_task_group=use_task_group,
+                    normalize_task_id=render_config.normalize_task_id,
+                    normalize_task_display_name=render_config.normalize_task_display_name,
+                    resource_suffix=node.resource_type.value,
+                    execution_mode=execution_mode,
+                )
+                # EmptyOperator does not accept custom dbt parameters (e.g. profile_args); recreate the args.
+                args = {"task_display_name": args["task_display_name"]} if "task_display_name" in args else {}
+                return TaskMetadata(id=task_id, operator_class=EMPTY_OPERATOR_CLASS, arguments=args)
```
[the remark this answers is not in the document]
pankajkoti: This matches the existing source to EmptyOperator path (the SourceRenderingBehavior.ALL branch in graph.py), which recreates the args the same way. Preserving BaseOperator kwargs on placeholder tasks would be a separate change spanning both the source and seed paths, so I would prefer to handle it as a follow-up if we want that behavior.

cosmos/dbt/seed.py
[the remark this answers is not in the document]
pankajkoti: get_seed_checksum() in cache.py now catches AirflowException and SQLAlchemyError, logs a warning, and returns None so the seed runs. This is symmetric with the write path.

pankajkoti:
Thanks @tatiana for the thorough review. I have pushed changes addressing the inline feedback (replies are in each thread):

- `checksum` is now a property on `DbtNode`, always computed from the seed CSV (streamed in chunks), so it is consistent across `LoadMode.MANIFEST` and `LoadMode.DBT_LS`.
- Removed `cosmos/dbt/seed.py`; the Airflow Variable persistence now lives in `cache.py`, and reads are best effort (they log and fall back to running the seed on backend errors).
- Adopted the cleaner `should_run` interface in the operator and renamed the render flag to `should_run_if_seed_changed`.
- Also handled the Copilot notes: streaming checksum, and stripping whitespace before `to_boolean` in `_is_full_refresh`.

On the watcher: agreed on logging a follow-up. I opened #2772 to add `SeedRenderingBehavior` support under `ExecutionMode.WATCHER`, following the `SourceRenderingBehavior` approach you linked.

When you have a moment, could you please take another look? I have re-requested your review.

cosmos/dbt/graph.py
```diff
@@ -69,6 +70,23 @@
...
+    digest = hashlib.sha256()
+    try:
+        with open(file_path, "rb") as file:
+            for chunk in iter(lambda: file.read(_CHECKSUM_READ_CHUNK_SIZE), b""):
+                digest.update(chunk)
+    except OSError as exc:
+        logger.warning("Unable to read file `%s` to compute its checksum: %s", file_path, exc)
+        return None
+    return digest.hexdigest()
+
```
tatiana, line 89: This would be better placed in the `cosmos/cache.py`

cosmos/dbt/graph.py
```diff
@@ -113,6 +131,18 @@ def file_path(self) -> Path:
         """Combined path to the node's file (path_base / original_file_path)."""
         return self.path_base / self.original_file_path
 
+    @property
```
tatiana, line 134: Would there be any benefit in marking this as a cached property?

cosmos/operators/local.py
```diff
@@ -1035,6 +1037,46 @@ class DbtSeedLocalOperator(DbtSeedMixin, DbtLocalBaseOperator):
...
+        # request always runs the seed, bypassing change detection.
+        extra_context = self.extra_context or {}
+        if not extra_context.get("should_run_if_seed_changed") or self._is_full_refresh():
+            super().execute(context, **kwargs)
+            return
+
+        node_config = extra_context.get("dbt_node_config") or {}
+        dag_task_group_identifier = extra_context.get("dbt_dag_task_group_identifier")
+        unique_id = node_config.get("unique_id")
+        current_seed_checksum = node_config.get("checksum")
```
tatiana, line 1069: BTW: I logged a ticket for us to improve how we represent the dbt node within the operator, this is not new to this PR, but it is a room for improvement:
https://github.com/astronomer/astronomer-cosmos/issues/2768

cosmos/cache.py
```diff
@@ -140,6 +146,45 @@ def create_cache_key(cache_identifier: str) -> str:
...
 
+def _create_seed_checksum_key(dag_task_group_identifier: str, unique_id: str) -> str:
+    """Return a bounded, stable Airflow Variable key scoped to a DbtDag/DbtTaskGroup and seed.
+
+    Airflow Variable keys are length-bounded (the ``variable.key`` column is commonly 250 chars), so we
+    store the checksum under a fixed-length key: a readable prefix plus a stable hash of the full
+    identifier, ensuring long DAG ids / nested task groups / package+resource names never overflow it.
+    """
+    digest = hashlib.sha1(f"{dag_task_group_identifier}:{unique_id}".encode()).hexdigest()
+    return f"{VAR_KEY_SEED_CHECKSUM_PREFIX}{digest}"
```
tatiana, line 157: Should we override with https://github.com/astronomer/astronomer-cosmos/pull/2755/changes#r3363120159?

It would be great to be consistent on how we're creating other hashes if we're not doing this already (use the same library/function).

cosmos/cache.py
```diff
@@ -140,6 +146,45 @@ def create_cache_key(cache_identifier: str) -> str:
...
+
+
+def get_seed_checksum(dag_task_group_identifier: str, unique_id: str) -> str | None:
```
tatiana, line 160:
```suggestion
def get_cache_seed_checksum(dag_task_group_identifier: str, unique_id: str) -> str | None:
```

It would be great to include in the function name that this captures the checksum of the last cached seed.

tatiana (approved):
@pankajkoti thanks a lot for addressing the feedback! Minor feedback inline - non-blocking to merge this PR (you could implement the improvements in a follow-up PR).

Please, rebase before merging

cosmos/dbt/graph.py
pankajkoti: will address this in a followup PR, logged an isssue for next cycle to address this https://linear.app/astronomer/issue/BOSS-443/address-feedback-on-seed-rendering-pr

cosmos/operators/local.py
pankajkoti: will address this in a followup PR, logged an isssue for next cycle to address this https://linear.app/astronomer/issue/BOSS-443/address-feedback-on-seed-rendering-pr

cosmos/cache.py
pankajkoti: will address this in a followup PR, logged an isssue for next cycle to address this https://linear.app/astronomer/issue/BOSS-443/address-feedback-on-seed-rendering-pr

cosmos/cache.py
pankajkoti: will address this in a followup PR, logged an isssue for next cycle to address this https://linear.app/astronomer/issue/BOSS-443/address-feedback-on-seed-rendering-pr

cosmos/dbt/graph.py
pankajkoti: will check and address this in a followup PR, logged an isssue for next cycle to address this https://linear.app/astronomer/issue/BOSS-443/address-feedback-on-seed-rendering-pr

cosmos/operators/local.py
```diff
@@ -1035,6 +1037,46 @@ class DbtSeedLocalOperator(DbtSeedMixin, DbtLocalBaseOperator):
...
+
```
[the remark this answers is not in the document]
pankajkoti: Not planning to address this for now; can revisit in future if the DAG parsing stops to be happening on the worker.

pankajkoti:
@tatiana Thanks for the review. I have logged a follow-up [ticket](https://linear.app/astronomer/issue/BOSS-443/address-feedback-on-seed-rendering-pr) and scheduled it for the upcoming cycle to address the remaining feedback here.

PR #2758, opened 2026-06-02
Centralize version-aware EmptyOperator imports in a compatibility module

pankajkoti:
Introduces `cosmos/airflow/compatibility.py`, a small module that centralizes the version-specific import of Airflow objects whose path changed between Airflow 2 and 3. The first such object is `EmptyOperator`, which moved to the standard provider in Airflow 3. The legacy `airflow.operators.empty` path still resolves on Airflow 3 but emits a `DeprecatedImportWarning`, so the module imports from the standard provider on Airflow 3 and from the legacy path on Airflow 2. The selection compares on the Airflow major version, so pre-releases such as `3.0.0rc1` are treated as Airflow 3.

The module also exposes `EMPTY_OPERATOR_CLASS_PATH`, the dotted import path for the version-appropriate `EmptyOperator`, derived from the imported class so it always stays in sync.

Call sites reference the operator by name or via the constant instead of duplicating the version check:
- `watcher_kubernetes.py` and `_watcher/base.py` import `EmptyOperator` from the module, replacing their duplicated try/except import blocks.
- `graph.py` serializes the `operator_class` string via `EMPTY_OPERATOR_CLASS_PATH`.

### DIFF
--- a/cosmos/airflow/compatibility.py
+++ b/cosmos/airflow/compatibility.py
@@ -13,7 +13,7 @@ from cosmos.constants import _AIRFLOW3_MAJOR_VERSION, AIRFLOW_VERSION
 if AIRFLOW_VERSION.major >= _AIRFLOW3_MAJOR_VERSION:
     try:
         from airflow.providers.standard.operators.empty import EmptyOperator as EmptyOperator
-    except ImportError as exc:
+    except ImportError as exc:  # pragma: no cover
         raise ImportError(
             "Cosmos on Airflow 3 requires `apache-airflow-providers-standard` to import `EmptyOperator`."
         ) from exc
--- a/cosmos/airflow/compatibility.py
+++ b/cosmos/airflow/compatibility.py
@@ -1,64 +1,28 @@
 """Version-aware imports for Airflow objects whose import path differs across Airflow 2 and 3.
 
-Names exported here are resolved lazily on first attribute access (PEP 562), so importing
-this module is free: the underlying Airflow object is only imported when the name is actually
-used. Use sites keep referencing the object by its real name (e.g. ``EmptyOperator``) rather
-than a version-specific dotted path.
+``EmptyOperator`` moved to the standard provider in Airflow 3; the legacy
+``airflow.operators.empty`` path still resolves there but emits a ``DeprecatedImportWarning``,
+so on Airflow 3 we import it from the standard provider. Compare on the major version so that
+Airflow 3 pre-releases (e.g. 3.0.0rc1) are treated as Airflow 3.
 """
 
 from __future__ import annotations
 
-import importlib
-from typing import TYPE_CHECKING
-
 from cosmos.constants import _AIRFLOW3_MAJOR_VERSION, AIRFLOW_VERSION
 
-if TYPE_CHECKING:
-    # Resolved for type checkers / IDEs only; no import happens at runtime here. The standard
-    # provider path is tried first so the type resolves to the real class on Airflow 3 (the
-    # legacy module is a deprecated shim typed as ``Any`` there).
+if AIRFLOW_VERSION.major >= _AIRFLOW3_MAJOR_VERSION:
     try:
         from airflow.providers.standard.operators.empty import EmptyOperator as EmptyOperator
-    except ImportError:
-        from airflow.operators.empty import EmptyOperator as EmptyOperator  # type: ignore[no-redef]
+    except ImportError as exc:
+        raise ImportError(
+            "Cosmos on Airflow 3 requires `apache-airflow-providers-standard` to import `EmptyOperator`."
+        ) from exc
+else:
+    # The redundant ``as EmptyOperator`` alias marks the name as an explicit re-export for type
+    # checkers; the ``no-redef`` ignore silences the duplicate binding mypy sees across branches.
+    from airflow.operators.empty import EmptyOperator as EmptyOperator  # type: ignore[no-redef]
 
-# Single source of truth for where ``EmptyOperator`` lives. The operator moved to the standard
-# provider in Airflow 3; the legacy ``airflow.operators.empty`` path still resolves there but
-# emits a ``DeprecatedImportWarning``, so on Airflow 3 we select the standard provider path.
-# Compare on the major version so Airflow 3 pre-releases (e.g. 3.0.0rc1) are treated as Airflow 3.
-_EMPTY_OPERATOR_MODULE = (
-    "airflow.operators.empty"
-    if AIRFLOW_VERSION.major < _AIRFLOW3_MAJOR_VERSION
-    else "airflow.providers.standard.operators.empty"
-)
-
-# Maps the exported name to the module it should be imported from for this Airflow version.
-_LAZY_IMPORTS = {
-    "EmptyOperator": _EMPTY_OPERATOR_MODULE,
-}
-
-
-def __getattr__(name: str) -> object:
-    module_path = _LAZY_IMPORTS.get(name)
-    if module_path is None:
-        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
-    # Cache the resolved symbol in the module namespace so subsequent attribute access skips
-    # __getattr__ entirely (PEP 562 only invokes it for names missing from globals()).
-    resolved = getattr(importlib.import_module(module_path), name)
-    globals()[name] = resolved
-    return resolved
-
-
-def get_version_aware_operator_class_path(operator: type) -> str:
-    """Return the fully qualified import path for the given operator class.
-
-    The path is read from the class itself (``__module__`` + ``__name__``), so it reflects
-    the actual module the class lives in for the running Airflow version. Pair it with the
-    version-aware classes exported from this module, e.g.::
-
-        get_version_aware_operator_class_path(EmptyOperator)
-
-    Used where Cosmos stores the operator as a dotted string for later dynamic import
-    (e.g. ``Task.operator_class``) instead of referencing the class directly.
-    """
-    return f"{operator.__module__}.{operator.__name__}"
+# Dotted import path for the version-appropriate EmptyOperator, for the places that store the
+# operator as a string for later dynamic import (e.g. ``Task.operator_class``) rather than
+# referencing the class. Derived from the class itself, so it always matches the imported one.
+EMPTY_OPERATOR_CLASS_PATH = f"{EmptyOperator.__module__}.{EmptyOperator.__name__}"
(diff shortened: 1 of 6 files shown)


tatiana (commented):
Hi @pankajkoti ,

I feel uneasy about using a variable to represent a well-known class. This tends to make the human maintenance of the code harder.

That said, I agree we should unify how we handle different import paths (AF2/AF3) across our codebase.

Please, could you take a look at this library - it may be a better solution:
https://airflow.apache.org/docs/apache-airflow-providers-common-compat/stable/index.html

If it is not, I suggest we create a module in `cosmos/airflow/compatibility.py` to handle the import, keeping the class name as `EmptyOperator`.

pankajkoti:
Thanks @tatiana for the quick review. 


> Please, could you take a look at this library - it may be a better solution:
https://airflow.apache.org/docs/apache-airflow-providers-common-compat/stable/index.html
If it is not, I suggest we create a module in cosmos/airflow/compatibility.py to handle the import, keeping the class name as EmptyOperator.

I checked `apache-airflow-providers-common-compat`, and it does not ship an `EmptyOperator`. Its [standard/operators.py](https://github.com/apache/airflow/blob/main/providers/common/compat/src/airflow/providers/common/compat/standard/operators.py) only re-exports `PythonOperator`, `ShortCircuitOperator`, and a few SDK base classes, so common-compat is not an option here.

I went with your second suggestion: a new `cosmos/airflow/compatibility.py` that handles the version-specific import and keeps the name as `EmptyOperator`. It resolves names lazily (PEP 562 module `__getattr__`), mirroring the pattern the standard provider itself uses, so importing the module stays cheap and call sites reference the real class rather than a string variable.

Changes in the latest push:
- `watcher_kubernetes.py` and `_watcher/base.py` now import `EmptyOperator` from the new module instead of duplicating try/except blocks.
- `graph.py` builds the serialized `operator_class` path via `get_version_aware_operator_class_path(EmptyOperator)`.
- The `EMPTY_OPERATOR_CLASS` constant is removed.

The module is intentionally generic so we can fold other Airflow 2/3 import splits into it over time.

cosmos/airflow/compatibility.py
```diff
@@ -0,0 +1,59 @@
...
+        from airflow.providers.standard.operators.empty import EmptyOperator as EmptyOperator
+    except ImportError:
+        from airflow.operators.empty import EmptyOperator as EmptyOperator  # type: ignore[no-redef]
+
+# Single source of truth for where ``EmptyOperator`` lives. The operator moved to the standard
+# provider in Airflow 3; the legacy ``airflow.operators.empty`` path still resolves there but
+# emits a ``DeprecatedImportWarning``, so on Airflow 3 we select the standard provider path.
+_EMPTY_OPERATOR_MODULE = (
+    "airflow.operators.empty" if AIRFLOW_VERSION < Version("3.0") else "airflow.providers.standard.operators.empty"
+)
```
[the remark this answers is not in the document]
pankajkoti: Good catch. Switched to `AIRFLOW_VERSION.major < _AIRFLOW3_MAJOR_VERSION`, matching the convention used elsewhere in the codebase, so Airflow 3 pre-releases resolve to the standard-provider path.

cosmos/airflow/compatibility.py
```diff
@@ -0,0 +1,59 @@
...
+_LAZY_IMPORTS = {
+    "EmptyOperator": _EMPTY_OPERATOR_MODULE,
+}
+
+
+def __getattr__(name: str) -> object:
+    module_path = _LAZY_IMPORTS.get(name)
+    if module_path is None:
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+    return getattr(importlib.import_module(module_path), name)
```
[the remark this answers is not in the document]
pankajkoti: Done. `__getattr__` now caches the resolved symbol into `globals()`, so subsequent attribute access skips the lookup.

cosmos/airflow/compatibility.py
```diff
@@ -0,0 +1,64 @@
...
+    The path is read from the class itself (``__module__`` + ``__name__``), so it reflects
+    the actual module the class lives in for the running Airflow version. Pair it with the
+    version-aware classes exported from this module, e.g.::
+
+        get_version_aware_operator_class_path(EmptyOperator)
+
+    Used where Cosmos stores the operator as a dotted string for later dynamic import
+    (e.g. ``Task.operator_class``) instead of referencing the class directly.
+    """
+    return f"{operator.__module__}.{operator.__name__}"
```
tatiana: I’d prefer we solve the deprecation/import-path issue directly, without introducing a general compatibility layer for a single symbol.

It feels like a lot of abstraction for a small Airflow 2/3 import path difference - something that, even though it was isolated from the core `apache-airflow` library since Airflow 3, is still a part of its standard operators.

If we need to make this more complex and add lazy evaluation in the future, we can review it.

```suggestion
from __future__ import annotations

from cosmos.constants import _AIRFLOW3_MAJOR_VERSION, AIRFLOW_VERSION

if AIRFLOW_VERSION.major >= _AIRFLOW3_MAJOR_VERSION:
    try:
        from airflow.providers.standard.operators.empty import EmptyOperator
    except ImportError as exc:
        raise ImportError(
            "Cosmos on Airflow 3 requires `apache-airflow-providers-standard` "
            "to import `EmptyOperator`."
        ) from exc
else:
    from airflow.operators.empty import EmptyOperator
```

cosmos/operators/_watcher/base.py
```diff
@@ -51,7 +51,7 @@
         from airflow.sdk import DAG
     except ImportError:
         from airflow.models.dag import DAG  # type: ignore[assignment]
-    from airflow.operators.empty import EmptyOperator
+    from cosmos.airflow.compatibility import EmptyOperator
```
tatiana, line 54: If we have concerns about memory utilisation, we could move this import close to the only part of the code that actually needs it, inside the function.

cosmos/airflow/compatibility.py
tatiana: @pankajkoti, we could also add a line like the following, to deal with the `cosmos/airflow/graph.py` case:
```
EMPTY_OPERATOR_CLASS_PATH = f"{EmptyOperator.__module__}.{EmptyOperator.__name__}"
```

pankajkoti: Isn't that contradictory to the original comment https://github.com/astronomer/astronomer-cosmos/pull/2758#pullrequestreview-4416277854? It has to be constant for repeated usage across the code base for dotted empty operator class path.

tatiana (approved):
@pankajkoti approving, so you can move forward with the other tasks. Even though we have different views on the approach, this should not block your progress.

pankajkoti:
> @pankajkoti approving, so you can move forward with the other tasks. Even though we have different views on the approach, this should not block your progress.

Thanks @tatiana. I've pushed the simpler version: a direct version-gated import in cosmos/airflow/compatibility.py plus an `EMPTY_OPERATOR_CLASS_PATH` constant derived from the imported class, and removed the lazy `__getattr__` layer and the helper.  `graph.py` now serialises the path via that constant.

I checked the full apache-airflow-providers-common-compat provider. The only operators it ships are `PythonOperator` and `ShortCircuitOperator` (alongside `TimeDeltaTrigger`, `BaseNotifier`, a few security permission constants, and OpenLineage helpers). It does not cover `EmptyOperator`, nor the `BaseOperator` and `TaskGroup` Airflow 2/3 splits, nor any of the heavy provider operators we guard (Docker, AWS ECS, GCP Cloud Run, Azure Container Instances, BigQuery, KubernetesPod). So common-compat would not absorb the imports we keep duplicating, and we would still need our own shim.

For a single lightweight symbol like `EmptyOperator`, the plain import is clearly the right call, and that is what is in the PR now after the last commit. I did want to record the reasoning behind the lazier version, since I think it earns its keep once this module grows past one symbol:
  - We already repeat the same Airflow 2/3 import dance in many places. The BaseOperator SDK-vs-`airflow.models` split shows up in 9 modules, and the `TaskGroup` SDK-vs-`airflow.utils.task_group` split in 6. Those are exactly the imports this shared compatibility module that you suggested to add here in the PR would own, so we stop copy-pasting the `try/except` into every file.
  - Several symbols behind these try/except guards (not always Airflow2/3, but provider-specific version-aware imports that could be integrated here) are heavy provider imports (Docker, AWS ECS, GCP Cloud Run, Azure Container Instances, BigQuery, KubernetesPod). For those, resolving on first access rather than at module import is a genuine win, since importing the compatibility module would not eagerly pull in every provider.

So the intent was to make this the one place that owns version-aware imports, and during the housekeeping activities fold the scattered `try/except` blocks into it, with lazy resolution for the heavy ones. The symbols and the justification already exist today, namely the duplications above plus the heavy provider operators, so this is a scope and timing decision rather than waiting for the case to build up. I would like to bring back the lazy form as part of that cleanup.

 The eager, class-derived `EMPTY_OPERATOR_CLASS_PATH` constant is in slight tension with lazy resolution, since computing the dotted path at import forces the class to resolve. If we move heavy operators to lazy access, their path constants would need to be derived lazily too rather than as module-level constants. Easy to handle, just worth keeping in mind so we do not reintroduce eager imports through the path strings.

cosmos/airflow/compatibility.py
```diff
@@ -0,0 +1,28 @@
...
+        ) from exc
+else:
+    # The redundant ``as EmptyOperator`` alias marks the name as an explicit re-export for type
+    # checkers; the ``no-redef`` ignore silences the duplicate binding mypy sees across branches.
+    from airflow.operators.empty import EmptyOperator as EmptyOperator  # type: ignore[no-redef]
+
+# Dotted import path for the version-appropriate EmptyOperator, for the places that store the
+# operator as a string for later dynamic import (e.g. ``Task.operator_class``) rather than
+# referencing the class. Derived from the class itself, so it always matches the imported one.
+EMPTY_OPERATOR_CLASS_PATH = f"{EmptyOperator.__module__}.{EmptyOperator.__name__}"
```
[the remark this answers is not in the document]
pankajkoti: updated the PR description.

cosmos/operators/_watcher/base.py
pankajkoti: responded in https://github.com/astronomer/astronomer-cosmos/pull/2758#issuecomment-4621789162

cosmos/airflow/compatibility.py
pankajkoti: I took the suggestion.

PR #2754, opened 2026-06-02
Add dbt-core 1.12 (beta) to test matrix

pankajastro:
## Summary

Adds `dbt-core` `1.12` to the test matrix in `astronomer-cosmos` in preparation for the upcoming 1.12.0 release line.

> **Note:** `dbt-core 1.12.0b2` is **not yet published on PyPI** as of 2026-06-02 — latest is 1.11.11. This PR is **preparatory**. The Hatch matrix slot `"1.12"` resolves via `~=$DBT_VERSION` in `scripts/test/pre-install-airflow.sh`, so CI will pick up the latest available 1.12.x as soon as it ships — no further matrix edit is needed at release time. The CI workflow matrices (`Run-Unit-Tests`, `Run-Integration-Tests`) are deliberately **not** modified — they still target stable versions (`1.10` / `1.11`) and should only be bumped to `1.12` after a GA release of 1.12.x and after the key adapters catch up (see below).

## Touchpoints applied

| File | Change |
|---|---|
| `pyproject.toml` | Inserted `"1.12"` between `"1.11"` and `"2.0"` in `[tool.hatch.envs.tests.matrix].dbt`. |
| `docs/policy/compatibility-policy.rst` | Inserted `1.12` in the dbt Core supported-versions line. |

## Touchpoints intentionally NOT modified

| File | Why |
|---|---|
| `.github/workflows/test.yml` (unit + integration matrices) | Beta version; adding to CI would make CI jobs pull a non-existent version. Defer until 1.12.x GA + adapters catch up. |
| `scripts/test/integration-setup.sh` | Takes `DBT_VERSION` as an arg; no per-version hardcoding to update. |
| `scripts/test/pre-install-airflow.sh` | Same — no hardcoded dbt version branches needed for 1.12 at this time. |
| `dev/dags/example_virtualenv*.py`, `dev/Dockerfile.*` | All pin `dbt-core<2.0`; 1.12 satisfies this. No change needed. |

## Pre-flight scan results

- **Upstream changelog:** N/A — dbt-core 1.12.0b2 not yet released. Re-scan when the version ships.
- **Adapter lag (PyPI snapshot, 2026-06-02):**

  | Adapter | Latest on PyPI | Status for dbt-core 1.12 |
  |---|---|---|
  | `dbt-databricks` | `1.12.0` | ✅ Ahead — already publishes a 1.12 line |
  | `dbt-bigquery` | `1.11.1` | ⚠️ Behind — no 1.12 release yet |
[...]

PR #2761, opened 2026-06-02
CI: carve dbt-loom tests into a dedicated job

pankajastro:
## Summary

Carves the one dbt-loom-dependent test path out of the main integration suite into a dedicated CI job, so the main matrix can move to dbt versions that dbt-loom doesn't yet support.

## Motivation

`dbt-loom 0.9.4` (latest on PyPI as of 2026-06-02) declares `dbt-core>=1.6.0,<1.12.0`. When dbt-loom is installed alongside dbt-core in the common pre-install path, `uv` silently downgrades `dbt-core` to satisfy the cap — so a future bump of the integration matrix to dbt-1.12+ would resolve back to dbt-1.11 without erroring. Pulling dbt-loom out of the common path removes that downgrade pressure and lets the main matrix bump independently.

This is a prerequisite for [#2754 (Add dbt-core 1.12 to test matrix)](https://github.com/astronomer/astronomer-cosmos/pull/2754) actually delivering dbt-1.12 CI coverage.

## What changes

| File | Change |
|---|---|
| `scripts/test/integration-dbt-loom.sh` *(new)* | Pins dbt-core to a loom-supported version, installs `dbt-loom` in both the main env and `venv-subprocess/`, then runs only `tests/test_example_dags.py::test_example_dag[cross_project_manifest_dag]`. Modeled on `integration-dbt-async.sh`. |
| `.github/workflows/test.yml` | Adds `Run-Integration-Tests-DBT-Loom` job — single matrix cell (py3.11, airflow-3.2, dbt-1.11) with the postgres service that `cross_project_manifest_dag`'s profile expects. |
| `pyproject.toml` | Adds `test-integration-dbt-loom = 'sh scripts/test/integration-dbt-loom.sh {matrix:dbt}'` hatch script entry. |
| `scripts/test/pre-install-airflow.sh` | Removes `uv pip install dbt-loom` from the else-branch and from the `venv-subprocess` install. Both call-sites get a comment pointing at the dedicated job. |
| `tests/test_example_dags.py` | If `dbt_loom` isn't importable, appends `cross_project_manifest_dag.py` to `IGNORED_DAG_FILES`. Lets the main example-DAG test pass on matrix cells where dbt-loom isn't installed; the dedicated job covers the DAG on supported versions. |

## Footprint check

`dbt-loom` is referenced by name in `cosmos/dbt/graph.py`, `cosmos/dbt/selector.py`, `tests/dbt/test_selector.py`, `tests/dbt/test_graph.py`, and two example DAGs — but **no Python code in cosmos imports the `dbt_loom` package**. The references are in docstrings, defensive handling of external-node data shapes, and example-DAG strings/tags. The only runtime dependency on the installed `dbt-loom` package is `cross_project_manifest_dag.py`, which spawns dbt subprocesses (via `venv-subprocess/bin/dbt`) against a project that uses dbt-loom's cross-project references. That single DAG is what the new job covers.

`cross_project_dbt_ls_dag.py` is already in `IGNORED_DAG_FILES` (pre-existing exclusion, unrelated to dbt-loom).

## Re-merge condition

Once `dbt-loom` publishes a release that supports the main matrix's dbt version, this job can either:
- Update its matrix to follow main's (and continue running as a dedicated cell), or
- Be merged back into `Run-Integration-Tests` and the file deleted.

## Follow-up

`Run-Integration-Tests-DBT-Loom` shares ~25 lines of boilerplate (checkout + cache + setup-python + install) with `Run-Integration-Tests-DBT-Async` and `Run-Integration-Tests`. A separate follow-up PR will extract a `setup-cosmos-test-env` composite action and refactor the loom + async jobs to use it. Deliberately kept out of this PR so reviewers can focus on the carve-out without bikeshedding the action's interface.

## Test plan

- [x] CI runs the new `Run-Integration-Tests-DBT-Loom` job to completion
- [ ] Main integration tests still pass — `cross_project_manifest_dag` is conditionally ignored on cells where dbt-loom isn't installed; still runs on cells where it is (current dbt-1.11 cells)
- [ ] Async + unit jobs unaffected

CI: https://github.com/astronomer/astronomer-cosmos/actions/runs/26838876022/job/79145312925

## Reviewer checklist

- [ ] Carve-out scope matches the footprint check (one DAG, no `dbt_loom` imports in source/tests)
- [ ] The `dbt_loom` install removals from `pre-install-airflow.sh` don't break any non-carved test path
- [ ] Conditional `IGNORED_DAG_FILES` logic in `test_example_dags.py` reads cleanly
- [ ] New hatch entry + workflow job follow repo conventions
- [ ] Re-merge plan acceptable
- [ ] Ready to mark non-draft

---

Drafted with the [`lib-upgrader`](https://github.com/astronomer/oss-integrations-private/pull/438) Claude Code skill (BOSS-432).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,11 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    # TEMP: ci/separate-dbt-loom-tests is included so the new
-    # Run-Integration-Tests-DBT-Loom job (added in this branch) actually runs in
-    # CI for review. pull_request_target uses the base branch's workflow file,
-    # so the new job would not otherwise appear. REVERT this line before merge.
-    branches: [main, ci/separate-dbt-loom-tests]
+    branches: [main]
   # Also run on pull requests originating from forks. Jobs that require sensitive secrets (integration tests,
   # kubernetes, code coverage) depend on the Authorize job, which requires manually approving the workflow run for
   # external PRs. Jobs that do not use sensitive secrets (type-check, unit tests, performance, telemetry) run
@@ -910,6 +906,7 @@ jobs:
       - Authorize
       - Run-Unit-Tests
       - Run-Integration-Tests
+      - Run-Integration-Tests-DBT-Loom
       - Run-Integration-Tests-Expensive
       - Run-Kubernetes-Tests
       - Run-Telemetry-Tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -539,7 +539,7 @@ jobs:
       POSTGRES_SCHEMA: public
       POSTGRES_PORT: 5432
 
-  Run-Integration-Tests-DBT-Loom:
+  Run-Integration-Tests-dbt-Loom:
     # Dedicated job for cross_project_manifest_dag.py, which exercises dbt-loom.
     # Carved out of Run-Integration-Tests so the loom job's dbt + loom versions
     # can move independently of the main matrix.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,11 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    # TEMP: ci/separate-dbt-loom-tests is included so the new
+    # Run-Integration-Tests-DBT-Loom job (added in this branch) actually runs in
+    # CI for review. pull_request_target uses the base branch's workflow file,
+    # so the new job would not otherwise appear. REVERT this line before merge.
+    branches: [main, ci/separate-dbt-loom-tests]
   # Also run on pull requests originating from forks. Jobs that require sensitive secrets (integration tests,
   # kubernetes, code coverage) depend on the Authorize job, which requires manually approving the workflow run for
   # external PRs. Jobs that do not use sensitive secrets (type-check, unit tests, performance, telemetry) run
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -539,6 +539,85 @@ jobs:
       POSTGRES_SCHEMA: public
       POSTGRES_PORT: 5432
 
+  Run-Integration-Tests-DBT-Loom:
+    # Dedicated job for cross_project_manifest_dag.py, which exercises dbt-loom.
+    # Carved out of Run-Integration-Tests so the loom job's dbt + loom versions
+    # can move independently of the main matrix.
+    needs: [Authorize, Check-changed-files]
+    if: needs.Check-changed-files.outputs.code_changed == 'true'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.11"]
+        airflow-version: ["3.2"]
+        dbt-version: ["1.11"]
+    services:
+      postgres:
+        image: postgres@sha256:4cd697181d4bd3ddc41a09012f339fa8cb5a8cd3d8b30130ea8378c176b6c494  # 14.18
+        env:
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,11 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    # TEMP: ci/separate-dbt-loom-tests is included so the new
+    # Run-Integration-Tests-DBT-Loom job (added in this branch) actually runs in
+    # CI for review. pull_request_target uses the base branch's workflow file,
+    # so the new job would not otherwise appear. REVERT this line before merge.
+    branches: [main, ci/separate-dbt-loom-tests]
```
```suggestion
    branches: [main]
```

.github/workflows/test.yml
```diff
@@ -539,6 +539,85 @@ jobs:
       POSTGRES_SCHEMA: public
       POSTGRES_PORT: 5432
 
+  Run-Integration-Tests-DBT-Loom:
```
```suggestion
  Run-Integration-Tests-dbt-Loom:
```
In the past we used the incorrect capitalization - if we can use the project's capitalization, it would be great.

.github/workflows/test.yml
```diff
@@ -539,6 +539,85 @@ jobs:
...
+    if: needs.Check-changed-files.outputs.code_changed == 'true'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.11"]
+        airflow-version: ["3.2"]
+        dbt-version: ["1.11"]
```
tatiana, line 556: Would it be worth running dbt Loom against a larger matrix? @pankajkoti you worked closer with dbt Loom, any thoughts on this one?

tatiana (approved):
@pankajastro minor feedback inline, happy for us to merge.

.github/workflows/test.yml
pankajastro: Good point — I'd rather keep this PR small and focused on the carve-out, but I've opened #2769 to track expanding the loom matrix as a follow-up. Happy to size it together with you once we agree on which dimensions matter most.

.github/workflows/test.yml
```diff
@@ -827,6 +906,7 @@ jobs:
       - Authorize
       - Run-Unit-Tests
       - Run-Integration-Tests
+      - Run-Integration-Tests-DBT-Loom
```
```suggestion
      - Run-Integration-Tests-dbt-Loom
```

tests/test_example_dags.py
```diff
@@ -29,6 +30,12 @@
     "cross_project_dbt_ls_dag.py",
 ]
 
+# cross_project_manifest_dag.py runs dbt subprocesses that need dbt-loom.
+# When dbt-loom isn't installed, the dedicated Run-Integration-Tests-DBT-Loom
+# CI job covers this DAG instead.
```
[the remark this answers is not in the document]
pankajastro: Good catch — fixed in 629f7ab7. Updated the comment to reference the actual job id `Run-Integration-Tests-dbt-Loom`.

scripts/test/integration-dbt-loom.sh
```diff
@@ -0,0 +1,59 @@
...
+    venv-subprocess/bin/pip install -U "dbt-core>=$DBT_VERSION,<$NEXT_MINOR_VERSION" dbt-loom dbt-postgres
+fi
+
+actual_dbt_version=$(dbt --version | awk '/installed:/ { split($3, v, "."); print v[1]"."v[2] }')
+if [ "$actual_dbt_version" = "$DBT_VERSION" ]; then
+    echo "Version is as expected: $DBT_VERSION"
+else
+    echo "Version does not match. Expected: $DBT_VERSION, but got: $actual_dbt_version"
+    exit 1
+fi
```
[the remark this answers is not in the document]
pankajastro: Thanks, but I verified this is a false positive and left the parsing as-is. `dbt --version` for the dbt-core this job pins (`>=1.11,<1.12`) prints:

```
Core:
  - installed: 1.11.11
  - latest:    1.11.11 - Up to date!
```

On the `installed:` line the version is field **3** (`-`=$1, `installed:`=$2, `1.11.11`=$3), so the current `split($3, v, ".")` correctly yields `1.11`. The `Core:` line has no version token, so the `Core:`/`$2` approach you suggest (from `pre-install-airflow.sh`) actually returns an empty string here and would fail the check.

The missing-`installed:`-line case only happens with dbt-fusion 2.0, which this job never installs (it pins dbt-core <1.12). This also matches the existing parsing in `scripts/test/integration-dbt-async.sh`, so I am keeping them consistent.

PR #2773, opened 2026-06-04
Add dbt-core 1.12 to test matrix

pankajastro:
## Summary

Adds **dbt-core 1.12 (stable, `1.12.0`)** to the Cosmos test matrix **and wires the CI install path** so the new version is actually exercised in CI rather than left as a no-op matrix label.

Supersedes the closed preparatory PR #2754 (which only added the Hatch matrix slot while `1.12` was pre-release / blocked on `dbt-loom`).

## Why this is more than a one-line matrix bump

The unit and integration jobs install dbt from **pinned `requirements/requirements-airflow-X.Y-dbt-Z.txt` lockfiles**, not from the matrix value directly. Before this PR every Airflow version installed the hardcoded `…-dbt-1.11.txt` file regardless of the matrix `dbt-version`, so adding `"1.12"` to the matrix alone would install dbt 1.11 under a "1.12" label. To make CI genuinely run 1.12 the install path now selects the lockfile by `(airflow, dbt)`.

## Touchpoints applied

| File | Change |
|---|---|
| `pyproject.toml` | `[tool.hatch.envs.tests.matrix].dbt` — inserted `"1.12"` between `"1.11"` and `"2.0"`. `pre-install-commands` now passes `{matrix:dbt}` as a third arg. |
| `scripts/test/pre-install-airflow.sh` | `$3=DBT_VERSION` is now **required** (no default). Selects `requirements-airflow-${AIRFLOW}-dbt-${DBT}.txt`; for a missing **1.11 or 1.12** lockfile this now **hard-fails** (`::error::`) rather than silently substituting — every supported Airflow version is expected to have both. Other dbt minors (1.5-1.10, 2.0) still fall back to the dbt-1.11 baseline with a `::warning::`, since jobs that need them (dbt-1-5-4, dbt-async, dbtf, performance, telemetry, integration-expensive) either re-pin dbt in their own setup step afterward or rely on this baseline directly. The dbt version-check now uses the working `installed:` parser. |
| `requirements/requirements-airflow-{2.9,2.10,2.11,3.0,3.1,3.2,3.3}-dbt-1.12.txt` | **New** — 7 pinned lockfiles for `dbt-core==1.12.0` (stable), generated in `python:3.10` linux/amd64 Docker. |
| `.github/workflows/test-unprivileged.yml` | `Run-Unit-Tests` matrix `dbt-version: ["1.10"] → ["1.11", "1.12"]` (additive, not a replace — see below). (This job lived in `test.yml` when this PR started; it has since moved to the new unprivileged workflow as part of an unrelated CI split on `main` — the dbt-version bump now applies there instead.) |
| `.github/workflows/test.yml` | `Run-Integration-Tests` matrix `dbt-version: ["1.11"] → ["1.12"]` (integration coverage moved off 1.11 entirely, see below). Airflow `3.3` (added on `main` after this PR was opened) is preserved in the merged matrix. |
| `tests/dbt/test_runner.py` | `test_handle_exception_if_needed_after_exception` now picks the expected missing-package error text by the **installed** dbt major/minor (`<= 1.11` keeps the old wording, else dbt 1.12's reworded one), so the test passes whichever dbt lockfile a job installs — including jobs that fall back to the dbt-1.11 baseline. |
| `docs/policy/compatibility-policy.rst` | Added `1.12` to the dbt Core supported-versions line. |

## dbt-core: beta → stable

The branch originally pinned `dbt-core==1.12.0b2` (the only build available at the time). **dbt-core 1.12.0 has since shipped stable on PyPI**, so all 6 lockfiles were regenerated against it:

- Per file: installed the existing (beta) lockfile in a fresh `python:3.10` linux/amd64 container, then `uv pip install -P dbt-core "dbt-core==1.12.0"` (upgrades only `dbt-core` and whatever it strictly requires — every other pin is untouched), then `uv pip freeze`.
- Diff per file is exactly 3 lines: `dbt-core` `1.12.0b2 → 1.12.0`, `dbt-core-experimental-parser` `2.0.0a1 → 2.0.0a4`, `dbt-adapters` `1.24.2 → 1.24.5`.
- `uv pip check` clean on all 6 files.
- `dbt-postgres`/`dbt-bigquery` are no longer pre-release builds either (they resolved to stable versions compatible with dbt-core 1.12.0 during the regenerate) — the "Residual TODOs" item about pre-release adapter pins from the original PR body no longer applies.

## Integration coverage: dropped 1.11, moved fully to 1.12 — unit tests keep both

A later commit on this branch (`Move integration tests to dbt 1.12 and assert dbt 1.12 error wording`) changed `Run-Integration-Tests`'s `dbt-version` from `["1.11", "1.12"]` to `["1.12"]` only — integration coverage moves to 1.12 rather than running both.

`Run-Unit-Tests` initially followed the same pattern (`1.10` → `1.12`, replace not additive). Tatiana's review pointed out that testing units only against the newest dbt drops coverage of the more broadly-used 1.11 line, so it's now `dbt-version: ["1.11", "1.12"]` (additive) — unit tests run both, integration runs only 1.12. This roughly doubles the `Run-Unit-Tests` matrix (up to 56 cells before exclusions instead of 28), which is the direct cost of that broader coverage.

## Base-image pip/wheel pin bug (fixed on this branch)

An early pass at generating the lockfiles picked up the `python:3.10-slim` base image's bootstrap `pip==23.0.1`/`wheel==0.45.1` in the freeze. That `pip` downgrade broke py3.12/3.13 (`AttributeError: module 'pkgutil' has no attribute 'ImpImporter'`, removed in 3.12). Fixed by stripping those two pins from all 6 files, matching the dbt-1.11 lockfiles which don't pin them either. (This fix carried through the beta→stable regeneration — the regenerated files were built from the already-fixed lockfiles, and `uv pip freeze` doesn't reintroduce base-image pins since it only freezes the venv.)

## Rebase onto `main`

This branch was rebased onto current `main` (which independently landed Airflow 3.3 support and split `test.yml` into `test.yml` (secret-dependent jobs) + `test-unprivileged.yml` (`Type-Check`/`Run-Unit-Tests`/telemetry/performance, gated on plain `pull_request` instead of `pull_request_target`)). Conflicts were resolved by union-ing both sides' matrix changes (Airflow 3.3 + dbt 1.12) rather than dropping either, and by re-homing the unit-test dbt-version bump into `test-unprivileged.yml` since that's where `Run-Unit-Tests` now lives.

## Airflow 3.3 × dbt 1.12: added the missing lockfile (Copilot catch)

Because Airflow 3.3 landed on `main` after this branch's original 6 dbt-1.12 lockfiles were generated, there was no `requirements-airflow-3.3-dbt-1.12.txt` — Copilot's review correctly flagged that the `airflow-version: 3.3` cells in both matrices were labeled `dbt-version: 1.12` but `pre-install-airflow.sh`'s fallback would silently install dbt 1.11 instead, so those cells would have passed while mislabeled. Added `requirements-airflow-3.3-dbt-1.12.txt` the same way as the other 6 (installed the existing 3.3-dbt-1.11 lockfile, then moved `dbt-core` to `1.12.0` and let `dbt-databricks` move with it via `uv pip install -P dbt-core -P dbt-databricks`, since the 1.11 file's `dbt-databricks==1.12.2` caps `dbt-core<1.11.13`; stripped the same base-image `wheel` pin the other files had removed). `uv pip check` clean; installed dbt-core resolves to `1.12.0`. All 7 airflow versions now have a real dbt-1.12 lockfile — the fallback path is no longer hit in the normal matrix, but `pre-install-airflow.sh` now logs a `::warning::` if a future (airflow, dbt) gap like this one ever recurs.

## Hard-fail on a missing 1.11/1.12 lockfile (Tatiana)

Tatiana flagged that `pre-install-airflow.sh`'s silent fallback-to-dbt-1.11 masked real gaps: once the dbt-version assertion actually worked (fixed earlier on this branch), the fallback just reassigned "expected" to match whatever got installed, so a missing lockfile silently passed instead of failing. Scoped the fix narrowly to this PR's actual concern rather than reverting the fallback wholesale: a missing **1.11 or 1.12** lockfile now hard-fails with `::error::` (every one of the 7 Airflow versions is expected to have both), while other dbt minors (1.5-1.10, 2.0) keep the silent fallback + warning, since their jobs either re-pin dbt themselves afterward or rely on this baseline directly and a blanket hard-fail would break hatch env creation for all of them before any re-pin step could run.

**Still open:** Tatiana followed up suggesting we go further — fail (and generate a real lockfile) for *every* missing `(airflow, dbt)` pair, not just 1.11/1.12, rather than falling back at all. Investigated this: `uv`'s strict resolver genuinely can't produce a coherent env for some old-dbt + new-Airflow combos (e.g. dbt 1.5 + Airflow 3.0 conflict on `jinja2`), which is presumably why the job-specific `*-setup.sh` re-pin scripts use permissive plain `pip install` rather than `uv`. Generating real lockfiles for the full `1.5-2.0 × 7-Airflow-version` matrix (or switching that install path to plain `pip`) is a bigger undertaking than this PR's scope. Filed astronomer/astronomer-cosmos#2902 to track raising the minimum supported dbt-core version instead, which would shrink/eliminate the problem by dropping the oldest, most conflict-prone dbt minors — left open for discussion there rather than attempted in this PR.

## ⚠️ Reviewer attention: 2.9 / 3.1 lockfile footprint

The hand-curated `dbt-1.11` lockfiles for **Airflow 2.9 and 3.1** omitted the `google-cloud-*` / full `dbt-bigquery` dependency tree (a conflict-avoidance workaround). With dbt-1.12's adapter set that tree now resolves cleanly (`uv pip check` passes), so the `2.9`/`3.1` dbt-1.12 files **include it** (+~95 lines each). This is **new coverage** — the curated dbt-1.11 files are untouched and still serve the fallback path — but the `2.9×1.12` / `3.1×1.12` cells should be watched on the first CI run for any runtime aiobotocore/google issues that `uv pip check` can't see. The other four files (2.10/2.11/3.0/3.2) are minimal dbt-stack-only diffs.

## Test plan

- [ ] CI unit tests pass on dbt-1.12 across the Python × Airflow matrix (via `test-unprivileged.yml`).
- [ ] CI integration tests pass on dbt-1.12 (`test.yml`, integration fully moved off 1.11).
- [ ] 2.9×1.12 and 3.1×1.12 cells pass (the expanded-footprint files).
- [ ] Pre-existing jobs (async / dbtf / loom / kubernetes / expensive / telemetry / perf) unaffected — they fall back to the dbt-1.11 baseline as before.
- [ ] Airflow 3.3 × dbt 1.12 cells now install the real 1.12 lockfile (no fallback warning in the logs).

## Reviewer checklist

- [ ] Touchpoint table matches the diff.
- [ ] Agree with unit tests keeping both 1.11 and 1.12 while integration moves fully to 1.12 (see coverage section above).
- [ ] Weigh in on the still-open "fail instead of dynamically installing older dbt" discussion (see hard-fail section above / astronomer/astronomer-cosmos#2902).
- [ ] 2.9/3.1 footprint expansion accepted (see callout above).
- [ ] CI green → ready to mark non-draft.

---

Drafted with the `lib-upgrader` Claude Code skill.


.github/workflows/test.yml
```diff
@@ -2,7 +2,11 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    # TEMP(revert before merge): also run on the dbt-1.12 bump branch so the new
+    # 1.12 matrix cells actually execute pre-merge. `pull_request_target` reads
+    # the workflow + matrix from the base branch, so it never runs new matrix
+    # cells defined only on the PR head; a `push` run uses the branch's workflow.
+    branches: [main, bump/dbt-core-1.12.0b2]
```
```suggestion
    # TEMP(revert before merge): also run on the dbt-1.12 bump branch so the new
    # 1.12 matrix cells actually execute pre-merge. `pull_request_target` reads
    # the workflow + matrix from the base branch, so it never runs new matrix
    # cells defined only on the PR head; a `push` run uses the branch's workflow.
    branches: [main]
```

.github/workflows/test.yml
```diff
@@ -142,7 +146,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         airflow-version: ["2.9", "2.10", "2.11", "3.0", "3.1", "3.2"]
-        dbt-version: ["1.10"]
+        dbt-version: ["1.12"]
```
pankajkoti: Do we want to run our tests only on the beta version? I am worried that, since it's still in beta and may bring other changes that may break us during development PRs, if we get this PR in, it may hamper other PRs mid-sprint/mid-week, and we may have to spend effort getting the CI up. WDYT?

docs/policy/compatibility-policy.rst
```diff
@@ -47,7 +47,7 @@ In some cases, Cosmos may continue to support older Airflow versions, depending
 dbt Core
 ++++++++
 
-- **Supported versions**: 1.5, 1.6, 1.7, 1.8, 1.9, 1.10, 1.11, 2.0 (dbt Fusion)
+- **Supported versions**: 1.5, 1.6, 1.7, 1.8, 1.9, 1.10, 1.11, 1.12, 2.0 (dbt Fusion)
```
pankajkoti, line 50: 1.12 is not released yet right? It's still in beta. so would be good to reflect that. Or we wait until 1.12 is released as stable.

pankajkoti (approved):
The PR description says the resolved stack was `dbt-postgres==1.11.0b1`, `dbt-bigquery==1.12.b1`, but the lock files reflect pinned stable versions `dbt-postgres==1.10.0` and `dbt-biguqery==1.11.0`. Can we resolve these discrepancies?

Also, I feel we should wait to get this in before stable 1.12 is released based on the inline comment that it may affect our development work midflight during the sprint.

requirements/requirements-airflow-3.2-dbt-1.12.txt
```diff
@@ -0,0 +1,403 @@
...
+db-dtypes==1.5.1
+dbt-adapters==1.24.2
+dbt-bigquery==1.11.1
+dbt-common==1.38.0
+dbt-core==1.12.0b2
+dbt-core-experimental-parser==2.0.0a1
+dbt-databricks==1.10.9
+dbt-duckdb==1.10.1
+dbt-extractor==0.6.0
+dbt-postgres==1.10.0
```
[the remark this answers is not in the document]
pankajastro: Addressed in f3ace621 — dbt-core 1.12.0 shipped stable since this review; lockfiles were regenerated against it and the PR description now reflects the actual (non-beta) adapter pins.

.github/workflows/test.yml
pankajastro: Addressed in f3ace621 — dbt-core 1.12.0 has since shipped stable on PyPI, so all lockfiles were regenerated against the stable release rather than the beta.

docs/policy/compatibility-policy.rst
pankajastro: Addressed in f3ace621 — dbt-core 1.12.0 shipped stable and this PR's lockfiles now pin the stable release, so this line's `1.12` entry is accurate.

.github/workflows/test.yml
```diff
@@ -133,7 +137,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         airflow-version: ["2.9", "2.10", "2.11", "3.0", "3.1", "3.2", "3.3"]
-        dbt-version: [ "1.11" ]
+        dbt-version: [ "1.12" ]
         split-group: [1, 2, 3]
```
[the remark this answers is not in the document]
pankajastro: Addressed in 5e8efe65 — added requirements-airflow-3.3-dbt-1.12.txt, so this matrix cell now installs real dbt 1.12 instead of silently falling back to 1.11.

.github/workflows/test-unprivileged.yml
```diff
@@ -135,7 +135,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         airflow-version: ["2.9", "2.10", "2.11", "3.0", "3.1", "3.2", "3.3"]
-        dbt-version: ["1.10"]
+        dbt-version: ["1.12"]
         exclude:
```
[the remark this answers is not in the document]
pankajastro: Addressed in 5e8efe65 — added requirements-airflow-3.3-dbt-1.12.txt, so this matrix cell now installs real dbt 1.12 instead of silently falling back to 1.11.

scripts/test/pre-install-airflow.sh
```diff
@@ -24,20 +27,23 @@ rm -f $AIRFLOW_HOME/airflow.db
...
+  # Install from the pinned lockfile matching this (airflow, dbt) pair. When no
+  # lockfile exists for the requested dbt minor, fall back to the dbt-1.11
+  # baseline; jobs that need a different dbt minor re-pin it in their own setup.
+  REQUIREMENTS_FILE="requirements/requirements-airflow-${AIRFLOW_VERSION}-dbt-${DBT_VERSION}.txt"
+  if [ -f "$REQUIREMENTS_FILE" ]; then
+    EFFECTIVE_DBT_VERSION="$DBT_VERSION"
+  else
+    REQUIREMENTS_FILE="requirements/requirements-airflow-${AIRFLOW_VERSION}-dbt-1.11.txt"
+    EFFECTIVE_DBT_VERSION="1.11"
+  fi
```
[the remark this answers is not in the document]
pankajastro: Addressed in 5e8efe65 — added a `::warning::` log line when the script falls back to the dbt-1.11 baseline lockfile.

requirements/requirements-airflow-3.2-dbt-1.12.txt
```diff
@@ -0,0 +1,403 @@
...
+daff==1.4.2
+databricks-sdk==0.47.0
+databricks-sql-connector==4.2.5
+db-dtypes==1.5.1
+dbt-adapters==1.24.2
+dbt-bigquery==1.11.1
+dbt-common==1.38.0
+dbt-core==1.12.0b2
+dbt-core-experimental-parser==2.0.0a1
+dbt-databricks==1.10.9
```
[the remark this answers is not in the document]
pankajastro: Addressed in f3ace621 — dbt-core 1.12.0 shipped stable, lockfiles were regenerated against it, and the PR description now matches the actual (stable) pins.

scripts/test/pre-install-airflow.sh
```diff
@@ -24,20 +27,23 @@ rm -f $AIRFLOW_HOME/airflow.db
...
-  uv pip install -r requirements/requirements-airflow-3.1-dbt-1.11.txt
-elif [ "$AIRFLOW_VERSION" = "3.2" ] ; then
-  uv pip install -r requirements/requirements-airflow-3.2-dbt-1.11.txt
-elif [ "$AIRFLOW_VERSION" = "3.3" ] ; then
-  uv pip install -r requirements/requirements-airflow-3.3-dbt-1.11.txt
+# Tracks the dbt minor we can assert on after install. Set when a pinned
+# lockfile actually pins the requested dbt minor, or by the dynamic path below.
+EFFECTIVE_DBT_VERSION=""
+
+if [ "$AIRFLOW_VERSION" = "2.9" ] || [ "$AIRFLOW_VERSION" = "2.10" ] || [ "$AIRFLOW_VERSION" = "2.11" ] || [ "$AIRFLOW_VERSION" = "3.0" ] || [ "$AIRFLOW_VERSION" = "3.1" ] || [ "$AIRFLOW_VERSION" = "3.2" ] || [ "$AIRFLOW_VERSION" = "3.3" ] ; then
```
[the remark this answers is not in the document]
pankajastro: Addressed in 5e8efe65 — added requirements-airflow-3.3-dbt-1.12.txt so Airflow 3.3 × dbt 1.12 no longer falls back to 1.11.

requirements/requirements-airflow-3.2-dbt-1.12.txt
```diff
@@ -0,0 +1,403 @@
...
+daff==1.4.2
+databricks-sdk==0.47.0
+databricks-sql-connector==4.2.5
+db-dtypes==1.5.1
+dbt-adapters==1.24.5
+dbt-bigquery==1.11.1
+dbt-common==1.38.0
+dbt-core==1.12.0
+dbt-core-experimental-parser==2.0.0a4
+dbt-databricks==1.10.9
```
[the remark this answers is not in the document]
pankajastro: Addressed in f3ace621 — dbt-core 1.12.0 GA shipped since this review; the lockfiles and PR description now both reflect the stable release, not the beta.

.github/workflows/test-unprivileged.yml
```diff
@@ -135,7 +135,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         airflow-version: ["2.9", "2.10", "2.11", "3.0", "3.1", "3.2", "3.3"]
-        dbt-version: ["1.10"]
+        dbt-version: ["1.12"]
```
tatiana: Maybe we should keep testing against a version of dbt that is more broadly used, instead of only running tests against the latest.
```suggestion
        dbt-version: ["1.11", "1.12"]
```

pankajastro: Addressed in 644a502e — restored `dbt-version: ["1.11", "1.12"]` for Run-Unit-Tests so 1.11 stays covered alongside 1.12.

scripts/test/pre-install-airflow.sh
```diff
@@ -6,6 +6,9 @@ set -e
 
 AIRFLOW_VERSION="$1"
 PYTHON_VERSION="$2"
+# dbt minor passed by Hatch via {matrix:dbt}. Defaults to the pinned baseline
+# when the script is invoked with only two args (e.g. local/manual runs).
+DBT_VERSION="${3:-1.11}"
```
tatiana: This line can lead to silently downgrade for lockfile-less matrix cells. A hatch env like tests.py3.10-2.9-1.5 now falls back to the dbt-1.11 lockfile and the assert passes (against 1.11). Let's revert to the old behaviour (hard assert failure, without a default value.

scripts/test/pre-install-airflow.sh
```diff
@@ -24,20 +27,24 @@ rm -f $AIRFLOW_HOME/airflow.db
...
+EFFECTIVE_DBT_VERSION=""
+
+if [ "$AIRFLOW_VERSION" = "2.9" ] || [ "$AIRFLOW_VERSION" = "2.10" ] || [ "$AIRFLOW_VERSION" = "2.11" ] || [ "$AIRFLOW_VERSION" = "3.0" ] || [ "$AIRFLOW_VERSION" = "3.1" ] || [ "$AIRFLOW_VERSION" = "3.2" ] || [ "$AIRFLOW_VERSION" = "3.3" ] ; then
+  # Install from the pinned lockfile matching this (airflow, dbt) pair. When no
+  # lockfile exists for the requested dbt minor, fall back to the dbt-1.11
+  # baseline; jobs that need a different dbt minor re-pin it in their own setup.
+  REQUIREMENTS_FILE="requirements/requirements-airflow-${AIRFLOW_VERSION}-dbt-${DBT_VERSION}.txt"
+  if [ -f "$REQUIREMENTS_FILE" ]; then
+    EFFECTIVE_DBT_VERSION="$DBT_VERSION"
+  else
```
tatiana: I believe we should fail instead of pinning to older dependencies - and force ourselves to generate the file.

We already had lots of issues with installing dependencies dynamically in the past, and believe the fallback plan can lead to apparent success, but not testing what it should.

tatiana (approved):
@pankajastro I gave minor feedback inline, happy for this PR to be merged once tests are passing.

Please, update the PR description to mention we're using the GA dbt-core 1.12

scripts/test/pre-install-airflow.sh
pankajastro: Addressed in 6a9be10e — scoped narrowly rather than a blanket revert: missing 1.11/1.12 lockfiles now hard-fail with `::error::` (every one of the 7 Airflow versions is expected to have both). Other dbt minors (1.5-1.10, 2.0) keep the silent fallback since their jobs (dbt-1-5-4, dbt-async, dbtf, performance, telemetry, integration-expensive) either re-pin dbt in their own setup step after this script runs, or rely on this baseline install without a per-Airflow lockfile — hard-failing those would break env creation before their setup step ever ran.

scripts/test/pre-install-airflow.sh
pankajastro: Tried exactly this — routing airflow versions through the dynamic install path whenever no lockfile exists (instead of substituting the dbt-1.11 lockfile), so it installs the actually-requested dbt minor and fails if it can't. Ran it for real combos in a `python:3.11` container (e.g. `airflow=3.0, dbt=1.5`, used by `Run-Integration-Tests-DBT-1-5-4`), and `uv` correctly refuses to resolve it:

```
And because apache-airflow-core==3.0.0 depends on jinja2>=3.1.5 and
apache-airflow==3.0.0 depends on apache-airflow-core==3.0.0, we can
conclude that apache-airflow==3.0.0 and dbt-duckdb>=1.5.0,<=1.5.2 are
incompatible.
And because you require dbt-duckdb>=1.5,<1.6 and apache-airflow==3.0, we
can conclude that your requirements are unsatisfiable.

hint: `dbt-core` was requested with a pre-release marker (e.g., all of:
    dbt-core>=1.5.1rc1,<=1.5.1rc2
    dbt-core>=1.5.2rc1,<=1.5.2rc2
    dbt-core==1.5.4rc1
), but pre-releases weren't enabled (try: `--prerelease=allow`)
```

dbt 1.5 (2023) and Airflow 3.0 genuinely don't have a mutually-compatible `jinja2` under a strict resolver — this isn't a pinning mistake on our end, it's a real conflict in the package universe. That's presumably exactly why `integration-dbt-1-5-4.sh` (the job's own re-pin step) uses plain `pip install`, not `uv pip install` — pip's legacy resolver doesn't hard-block on full-graph consistency the way `uv` does, so it can land in a state that works even though `uv` would call it unsatisfiable.

So making `pre-install-airflow.sh` fail-and-force-a-real-lockfile for every `(airflow, dbt)` pair would mean either: (a) generating real lockfiles for the full `1.5–2.0 × 7-Airflow-version` matrix — many combos of which may not have any mutually-satisfiable dependency set under `uv` at all — or (b) switching that install path to plain `pip` to match the permissive style the `*-setup.sh` scripts already rely on. Both are bigger than this PR's scope.

Filed #2902 to track raising the minimum supported dbt-core version, which would shrink/eliminate this problem by dropping the oldest, most conflict-prone dbt minors. Happy to discuss (b) as a nearer-term option too — thoughts?

PR #2775, opened 2026-06-05
Re-enable cross-project dbt ls example DAG in the dbt-loom CI job

pankajkoti:
`cross_project_dbt_ls_dag` (the `LoadMode.DBT_LS` cross-project example that exercises dbt-loom) has been disabled in the example-DAG tests since #2343. At that time dbt-loom had to be installed in a separate virtual environment due to dependency conflicts, so it was not resolvable in the DAG-parse process where `dbt ls` runs, and the DAG was added to the unconditional `IGNORED_DAG_FILES` list.

That constraint no longer applies in the dedicated dbt-loom CI job added in #2761:

- `scripts/test/integration-dbt-loom.sh` installs dbt-loom in both the main env and `venv-subprocess`, so loom is resolvable where `dbt ls` runs.
- The upstream `manifest.json` that dbt-loom reads (via `downstream/dbt_loom.config.yml`) is committed under `dev/dags/dbt/cross_project/upstream/target`, so it already exists at parse time with no generation step required.

The manifest DAG was re-enabled conditionally in #2761, but the dbt ls DAG was left unconditionally ignored, only as leftover from #2343.

This change:

- Removes `cross_project_dbt_ls_dag.py` from the unconditional `IGNORED_DAG_FILES` and guards it behind `find_spec("dbt_loom")` alongside `cross_project_manifest_dag.py`, so non-loom runs still skip it.
- Runs the dbt ls DAG in the dbt-loom CI job via a second `TEST_SINGLE_DAG` invocation.
- Updates the related job comment and step name to cover both cross-project DAGs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -540,9 +540,10 @@ jobs:
       POSTGRES_PORT: 5432
 
   Run-Integration-Tests-dbt-Loom:
-    # Dedicated job for cross_project_manifest_dag.py, which exercises dbt-loom.
-    # Carved out of Run-Integration-Tests so the loom job's dbt + loom versions
-    # can move independently of the main matrix.
+    # Dedicated job for the cross-project example DAGs (cross_project_manifest_dag.py
+    # and cross_project_dbt_ls_dag.py), which exercise dbt-loom. Carved out of
+    # Run-Integration-Tests so the loom job's dbt + loom versions can move
+    # independently of the main matrix.
     needs: [Authorize, Check-changed-files]
     if: needs.Check-changed-files.outputs.code_changed == 'true'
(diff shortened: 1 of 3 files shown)


pankajkoti:
Successful dag run: https://github.com/astronomer/astronomer-cosmos/actions/runs/27004015749/job/79691043153?pr=2775#step:7:636

.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, enable-dbt-ls-cross-project-loom-ci]
```
```suggestion
    branches: [main]
```

tatiana (approved):

PR #2784, opened 2026-06-08
Refactor seed checksum hashing and caching helpers

pankajkoti:
## Description

Follow-up to #2755 addressing review feedback on the seed rendering work.

- Move the streaming file-checksum helper out of `cosmos/dbt/graph.py` into `cosmos/fs.py` as `_calculate_file_checksum`, alongside `safe_copy`, since it is a generic, dependency-free file utility. The seed-cache persistence stays in `cosmos/cache.py`.
- Make `DbtNode.checksum` a `cached_property`: the value is derived from the seed file at parse time and read again when building the task `extra_context`.
- Standardize seed checksum hashing on `md5`, matching the other non-cryptographic content and identifier hashes across Cosmos (`cosmos/cache.py`, `cosmos/versioning.py`, `cosmos/dbt/selector.py`). Both the seed content checksum and the Airflow Variable key derivation now use `md5`.
- Rename `get_seed_checksum`/`store_seed_checksum` to `get_cache_seed_checksum`/`store_cache_seed_checksum` to convey that they read and write the last cached seed checksum.

The seed-content checksum algorithm changed from `sha256` to `md5` and the Variable key derivation from `sha1` to `md5`. This feature has not been released, so no persisted checksum Variables depend on the previous derivation.

## Related Issue(s)

Closes astronomer/oss-integrations-private#444

### DIFF
--- a/cosmos/fs.py
+++ b/cosmos/fs.py
@@ -1,9 +1,35 @@
 from __future__ import annotations
 
+import hashlib
 import os
 import shutil
 import tempfile
 from pathlib import Path
+
+from cosmos.log import get_logger
+
+logger = get_logger(__name__)
+
+# Read the file in fixed-size chunks when checksumming so a large file (e.g. a seed CSV) does not have
+# to be loaded into memory all at once.
+_CHECKSUM_READ_CHUNK_SIZE = 1024 * 1024
+
+
+def calculate_file_checksum(file_path: Path) -> str | None:
+    """Return the MD5 checksum of a file, streaming it in chunks, or ``None`` if it cannot be read.
+
+    MD5 is used for consistency with the other non-cryptographic content/identifier hashes across Cosmos
+    (e.g. ``cosmos/cache.py``, ``cosmos/versioning.py``); this is change detection, not a security boundary.
+    """
+    digest = hashlib.md5()
+    try:
+        with open(file_path, "rb") as file:
+            for chunk in iter(lambda: file.read(_CHECKSUM_READ_CHUNK_SIZE), b""):
+                digest.update(chunk)
+    except OSError as exc:
+        logger.warning("Unable to read file `%s` to compute its checksum: %s", file_path, exc)
+        return None
+    return digest.hexdigest()
 
 
 def safe_copy(src: Path, dst: Path) -> None:
--- a/cosmos/fs.py
+++ b/cosmos/fs.py
@@ -15,7 +15,7 @@ logger = get_logger(__name__)
 _CHECKSUM_READ_CHUNK_SIZE = 1024 * 1024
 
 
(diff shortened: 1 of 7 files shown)


pankajkoti:
Failing CodeCov job in the CI will be resolved after getting in PR #2786

cosmos/fs.py
```diff
@@ -1,10 +1,36 @@
...
+from cosmos.log import get_logger
+
+logger = get_logger(__name__)
+
+# Read the file in fixed-size chunks when checksumming so a large file (e.g. a seed CSV) does not have
+# to be loaded into memory all at once.
+_CHECKSUM_READ_CHUNK_SIZE = 1024 * 1024
+
+
+def calculate_file_checksum(file_path: Path) -> str | None:
```
pankajastro: Shall we keep this private?

cosmos/fs.py
```diff
@@ -1,10 +1,36 @@
...
+    """Return the MD5 checksum of a file, streaming it in chunks, or ``None`` if it cannot be read.
+
+    MD5 is used for consistency with the other non-cryptographic content/identifier hashes across Cosmos
+    (e.g. ``cosmos/cache.py``, ``cosmos/versioning.py``); this is change detection, not a security boundary.
+    """
+    digest = hashlib.md5()
```
pankajastro, line 24: Maybe ok to change hash algorithm from sha265 —> md5 here but worth checking with @tatiana too

pankajastro (approved):
LGTM

cosmos/fs.py
pankajkoti: yes, okay made in internal by prefixing with `_`

cosmos/fs.py
pankajkoti: This is an effort to be consistent with our existing hashing algorithms. We have 8 existing call sites where we use md5, and there is only one occurrence of using sha in [profiles/base.py:version](https://github.com/astronomer/astronomer-cosmos/blob/d90dfc13fb7b62e1afde2b43725f8855d246be22/cosmos/profiles/base.py#L112), which we can standardise to md5 as well if we agree, but on a majority we have md5 everywhere else.

cosmos/fs.py
```diff
@@ -1,10 +1,36 @@
...
+    try:
+        with open(file_path, "rb") as file:
+            for chunk in iter(lambda: file.read(_CHECKSUM_READ_CHUNK_SIZE), b""):
+                digest.update(chunk)
+    except OSError as exc:
+        logger.warning("Unable to read file `%s` to compute its checksum: %s", file_path, exc)
+        return None
+    return digest.hexdigest()
+
```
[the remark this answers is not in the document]
pankajkoti: We haven't released this yet so we're safe.

PR #2788, opened 2026-06-08
Reduce Airflow 3 Asset URI migration warning noise in WATCHER producer runs

### LINKED ISSUE #2778: [Improvement] Airflow 3 WATCHER logs Asset URI migration warning on every model even with USE_DATASET_AIRFLOW3_URI_STANDARD
### Astronomer Cosmos Version

1.14.2

1.9.10

### Versions of dbt adapters

dbt-trino 1.9.3

### LoadMode

DBT_LS

### ExecutionMode

WATCHER

### InvocationMode

DBT_RUNNER

### airflow version

3.1.8

### Operating System

Linux (Docker)

Chrome

### Deployment

Official Apache Airflow Helm Chart

### Deployment details

CeleryExecutor, 580 models, Trino adapter. emit_datasets=False, test_behavior=NONE.

### What happened?

After upgrading to Cosmos 1.14.2 on Airflow 3, the WATCHER producer task fills the logs with repeated warnings about Asset URI validation rules changing. We set `AIRFLOW__COSMOS__USE_DATASET_AIRFLOW3_URI_STANDARD=True` as documented [here](<https://astronomer.github.io/astronomer-cosmos/reference/configs/cosmos-conf.html#use-dataset-airflow3-uri-standard>). The URIs in the warning already use the Airflow 3 format, but the warning is still logged once per model on every run.

In a single producer run we saw roughly 1168 warnings of this type, plus hundreds of `Variable.set is deprecated` messages (likely related to astronomer/astronomer-cosmos#2736).

We expected the config to reduce this noise, or at least not log a migration warning on every node when we are already on Airflow 3 and not using dataset scheduling.

### Relevant log output

```shell
WARNING - Airflow 3.0.0 Asset (Dataset) URIs validation rules changed and OpenLineage URIs (standard used by Cosmos) are no longer accepted. Therefore, if using Cosmos with Airflow 3, the Airflow Asset (Dataset) URI is now <trino://.../trusted_ci/dbt_ci_run_local_test/gupy_ats/applications>. Before, with Airflow 2.x, the URI used to be <trino://.../trusted_ci.dbt_ci_run_local_test.gupy_ats.applications>. Please, change any DAGs that were scheduled using the old standard to the new one.
```

### How to reproduce

How to reproduce

1. Airflow 3.1.x + Cosmos 1.14.2
2. `DbtTaskGroup` with `ExecutionMode.WATCHER`, `InvocationMode.DBT_RUNNER`, `LoadMode.DBT_LS`
3. `RenderConfig(emit_datasets=False)`
4. Set `AIRFLOW__COSMOS__USE_DATASET_AIRFLOW3_URI_STANDARD=True`
5. Run the producer on a project with a few hundred models
6. Check task logs: the Asset URI warning appears once per model

### Anything else :)?

*No response*

### Contact Details

*No response*

faridnsh:
## Description

This PR makes two related changes in their own commits:

- WATCHER now preserves the original `RenderConfig.emit_datasets` value before forcing the producer task's own `emit_datasets=False`. When consumers have `emit_datasets=False`, the producer skips dataset namespace and outlet URI preparation entirely.
- When WATCHER does need to compute outlet URIs from `manifest.json`, Cosmos logs the Asset URI migration warning once per manifest computation instead of once per dbt node. This could be done in several other ways, wasn't sure if this is the most elegant way possible.

This fixes the reported problem where WATCHER producer logs could contain hundreds or thousands of repeated Asset URI warnings even when dataset emission was disabled.

## Related Issue(s)

Fixes #2778

## Breaking Change?

No.

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -767,6 +767,10 @@ def _add_watcher_producer_task(
     The producer task is the task that will be used to produce the events for the watcher execution mode.
     """
     producer_task_args = task_args.copy()
+    # The producer never emits datasets itself (consumer tasks do), but it must still compute
+    # per-model outlet URIs when consumers will emit. Carry that intent as an explicit flag so the
+    # producer can decide whether to generate model URIs without overloading other signals.
+    producer_task_args["_should_generate_model_uris"] = task_args.get("emit_datasets", True)
     # Producer should not emit datasets — consumer tasks handle their own emission
     producer_task_args["emit_datasets"] = False
     if tests_per_model is not None:
--- a/cosmos/operators/watcher.py
+++ b/cosmos/operators/watcher.py
@@ -214,6 +214,9 @@ class DbtProducerWatcherOperator(DbtBuildMixin, DbtLocalBaseOperator):
         self.tests_per_model: dict[str, list[str]] = kwargs.pop("tests_per_model", {})
         self.test_results_per_model: dict[str, dict[str, str]] = {}
         self._check_source_freshness: bool = kwargs.pop("_check_source_freshness", False)
+        self._should_generate_model_uris: bool = kwargs.pop(
+            "_should_generate_model_uris", kwargs.get("emit_datasets", True)
+        )
         self._freshness_callback: Callable[
             [Context, Any, TaskGroup | None, dict[str, DbtNode] | None, dict[str, Any] | None],
             list[tuple[str, str]],
@@ -252,6 +255,7 @@ class DbtProducerWatcherOperator(DbtBuildMixin, DbtLocalBaseOperator):
             test_results_per_model=self.test_results_per_model,
             model_outlet_uris=self._model_outlet_uris,
             dataset_namespace=self._dataset_namespace,
+            should_generate_model_uris=self._should_generate_model_uris,
             upstream_failure_skipped_ids=self._upstream_failure_skipped_ids,
         )
 
@@ -449,8 +453,14 @@ class DbtProducerWatcherOperator(DbtBuildMixin, DbtLocalBaseOperator):
         self._apply_node_state_tokens(context, freshness_results)
 
     def execute(self, context: Context, **kwargs: Any) -> Any:
-        # Pre-compute the dataset namespace for per-model outlet URI generation.
-        self._dataset_namespace = get_dataset_namespace(self.profile_config)
+        # Resolve the dataset namespace only when the producer will generate per-model outlet
+        # URIs. With emit_datasets=False the producer does no dataset work, so we skip the
+        # namespace's profile resolution and leave it None. URI generation is gated explicitly by
+        # should_generate_model_uris in the log parser -- this None is just the disabled-state
+        # invariant, not a control signal.
+        self._dataset_namespace = (
+            get_dataset_namespace(self.profile_config) if self._should_generate_model_uris else None
+        )
         self._model_outlet_uris.clear()
         self._upstream_failure_skipped_ids.clear()
 
--- a/cosmos/dataset.py
+++ b/cosmos/dataset.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import urllib.parse
+from collections.abc import Callable
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
@@ -193,7 +194,7 @@ def get_dataset_namespace(profile_config: ProfileConfig) -> str | None:
     return None
 
 
-def construct_dataset_uri(namespace: str, name: str) -> str:
+def construct_dataset_uri(namespace: str, name: str, warn_uri_migration: bool = True) -> str:
     """
     Construct an Airflow Asset/Dataset URI from an OL-compatible namespace and
     dataset name.
@@ -204,6 +205,9 @@ def construct_dataset_uri(namespace: str, name: str) -> str:
 
     :param namespace: The OL-compatible dataset namespace (e.g. ``postgres://host:5432``).
     :param name: The dot-delimited dataset name (e.g. ``database.schema.table``).
+    :param warn_uri_migration: Whether to emit the Airflow 3 URI migration warning for this URI.
+        Callers that create multiple URIs in one task run should pass ``False`` after
+        the first generated URI.
     """
     airflow_2_uri = namespace + "/" + urllib.parse.quote(name)
     airflow_3_uri = namespace + "/" + urllib.parse.quote(name).replace(".", "/")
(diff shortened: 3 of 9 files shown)


faridnsh:
Copilots comments aren't exactly related to my changes, so I'm not sure whether they are worthwhile to change and whether they should be fixed in this PR since I didn't change the warning messages themselves.

pankajastro (approved):
Nice!

We currently call `construct_dataset_uri` from `local.py`, and it is now always being invoked with `warn_uri_migration=True`. We may want to audit those call sites to ensure this behavior is intentional. That said, I think this can be addressed in a separate PR.

cosmos/airflow/graph.py
```diff
@@ -767,6 +767,7 @@ def _add_watcher_producer_task(
     The producer task is the task that will be used to produce the events for the watcher execution mode.
     """
     producer_task_args = task_args.copy()
+    producer_task_args["_emit_datasets_for_consumers"] = task_args.get("emit_datasets", True)
```
```suggestion
    producer_task_args["_should_generate_model_uris"] = task_args.get("emit_datasets", True)
```
I think this name is better suited because the producer by itself does not generate the datasets, but is generating the model URIs that the consumer uses to emit datasets.

cosmos/operators/watcher.py
```diff
@@ -214,6 +214,9 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.tests_per_model: dict[str, list[str]] = kwargs.pop("tests_per_model", {})
         self.test_results_per_model: dict[str, list[str]] = {}
         self._check_source_freshness: bool = kwargs.pop("_check_source_freshness", False)
+        self._emit_datasets_for_consumers: bool = kwargs.pop(
+            "_emit_datasets_for_consumers", kwargs.get("emit_datasets", True)
+        )
```
pankajkoti: Need to update the variable name if the above comment is considered.

cosmos/operators/watcher.py
```diff
@@ -448,7 +451,9 @@ def _apply_source_freshness(self, context: Context) -> None:
 
     def execute(self, context: Context, **kwargs: Any) -> Any:
         # Pre-compute the dataset namespace for per-model outlet URI generation.
-        self._dataset_namespace = get_dataset_namespace(self.profile_config)
+        self._dataset_namespace = (
+            get_dataset_namespace(self.profile_config) if self._emit_datasets_for_consumers else None
+        )
```
pankajkoti, line 463: I think we're indirectly deciding not to let generating model URI computation proceed by setting _dataset_namespace to None here. It might be difficult to read and understand why we're setting the dataset namespace to `None` here and how that decides whether consumers would emit datasets or not.

Is there a chance we can make this explicit?  e.g. maybe by adding a new argument `should_generate_model_uris` to the `store_dbt_resource_status_from_log` method and then using that flag to decide whether to call `_ensure_subprocess_model_outlet_uris` where we compute the model URIs

Something like 
```
def store_dbt_resource_status_from_log(
      line, extra_kwargs, *,
      ...,
      dataset_namespace: str | None = None,
      should_generate_model_uris: bool = True,          # NEW
      upstream_failure_skipped_ids=None,
  ):
      ...
      outlet_uris: list[str] = []
      if generate_model_uris and dbt_node_resource_type in _DATASET_EMITTING_RESOURCE_TYPES:
          project_dir = extra_kwargs.get("project_dir")
          _ensure_subprocess_model_outlet_uris(model_outlet_uris, dataset_namespace, project_dir)
          outlet_uris = model_outlet_uris.get(unique_id, []) if model_outlet_uris else []
```

With that, we would not need a dataset namespace as a control signal.

pankajkoti (commented):
It looks like a solid work here @faridnsh.

I have some suggestions inline on naming the properties to better reflect their purpose and tie to the role they contribute to, and also making the conditional explicit on when to populate model URIs rather than relying on setting dataset_namespace=None to make it more readable. Looking forward to your thoughts here.

Thanks a lot for the effort here

pankajkoti:
Hi @faridnsh, following up to see if you would like to discuss this here or share any thoughts on my earlier review comments. We're planning to release Cosmos 1.15.0 on July 1st, and it would be nice to get this fix in.

faridnsh:
@pankajkoti I'm a bit busier than usual this week and next week. I agree with the comments, but I need to sit down and understand the code better to make sure it's meaningful. Latest, I can push the fixes around the 27-28th June weekend, hopefully it's not too late for you guys.

faridnsh:
Hi @pankajkoti

Turns out I managed to find some time! I added some comments because I was also kind of confused about the why initially. I still would appreciate another review, I'm still a bit blur about the whole end-to-end process. Once reviewed, then I can rebase to the latest and squash it to 2 commits to keep git history clean.

cosmos/operators/watcher.py
```diff
@@ -447,8 +451,12 @@ def _apply_source_freshness(self, context: Context) -> None:
...
 
     def execute(self, context: Context, **kwargs: Any) -> Any:
-        # Pre-compute the dataset namespace for per-model outlet URI generation.
-        self._dataset_namespace = get_dataset_namespace(self.profile_config)
+        # Whether per-model outlet URIs are generated is decided explicitly by
+        # _should_generate_model_uris, which is forwarded to the log parser. The namespace is only
+        # an input to that generation, so resolve it only when we will actually generate URIs.
+        self._dataset_namespace = (
+            get_dataset_namespace(self.profile_config) if self._should_generate_model_uris else None
+        )
```
pankajkoti:
```suggestion
        # Pre-compute the dataset namespace for per-model outlet URI generation.
        self._dataset_namespace = get_dataset_namespace(self.profile_config)
```
Do we need to do that anymore? I guess the previous code can be kept as is, no?

pankajkoti (commented):
The approach is looking good to me now from a quick overview. Do you have a local setup where we can confirm the fix works cleanly and the noise is reduced? It would be nice if you get a chance to test and confirm. Thanks for finding time earlier for this iteration 🙌🏽

pankajkoti:
> Hi @pankajkoti
> 
> Turns out I managed to find some time! I added some comments because I was also kind of confused about the why initially. I still would appreciate another review, I'm still a bit blur about the whole end-to-end process. Once reviewed, then I can rebase to the latest and squash it to 2 commits to keep git history clean.

Thanks a lot @faridnsh . Yes, don't worry about squashing. We will take care of it while merging the PR. The merged OR will reflect as a single commit into main's history, because we squash while merging using the GitHub UI's "Squash and Merge" feature.

cosmos/operators/watcher.py
faridnsh: Ok, I applied the change, and then remembered why I did what I did. Basically the idea was when `emit_datasets` is false then we shouldn't do anything related to datasets, so in that case `_dataset_namespace` should be set to false. Basically fixing what was being complained in #2778 that they set `emit_datasets` to false yet it's still complaining about datasets.

faridnsh: I have reverted it back now.

faridnsh:
@pankajkoti  I tested it locally with dev/dags/example_watcher.py shows the warning once when emit_datasets=True is enabled and no warning emit_datasets=False.

pankajkoti (approved):
LGTM @faridnsh. Thanks a lot for considering the comments and addressing those! We will release this improvement in Cosmos 1.15.0, planned for July 1st, 2026

PR #2796, opened 2026-06-09
Avoid deprecated `airflow.exceptions.AirflowSkipException` import on Airflow 3.1+

pankajastro:
## Problem

On Airflow 3.2+, importing `AirflowSkipException` from `airflow.exceptions` is deprecated and emits a `UserWarning` at import time:

```
WARNING - airflow.exceptions.AirflowSkipException is deprecated and will be removed in a future
version. Use airflow.sdk.exceptions.AirflowSkipException instead. category=UserWarning
```

Cosmos imports it from `airflow.exceptions` in `local.py`, `watcher.py`, `_watcher/base.py`, and `_k8s_common.py`, so the warning shows up at DAG parse / task execution.

## Fix

`airflow.sdk.exceptions.AirflowSkipException` only exists on Airflow 3.2+ — the Task SDK does not expose it on Airflow 2 / 3.0 / 3.1, where `airflow.exceptions.AirflowSkipException` is **not** deprecated. So a plain switch would break older versions.

Add a version-tolerant re-export in the existing `cosmos/airflow/compatibility.py` (which already centralises Airflow 2/3 import differences) that prefers the Task SDK path and falls back to `airflow.exceptions`, and import `AirflowSkipException` from there in the operator modules. `AirflowException` is unaffected (not deprecated) and is left as-is.

## Verification

- **Airflow 3.2.0**: importing the Cosmos operators emits **no** `AirflowSkipException` deprecation warning; it resolves to `airflow.sdk.exceptions.AirflowSkipException`.
- **Airflow 3.0 / 3.1**: the Task SDK does not expose `AirflowSkipException`, so it falls back to `airflow.exceptions.AirflowSkipException` (which is not deprecated there, so no warning).
- New regression test (`tests/airflow/test_compatibility.py`) asserts the shim resolves to the Task SDK on Airflow ≥ 3.2 and to `airflow.exceptions` below it.

### DIFF
--- a/cosmos/airflow/compatibility.py
+++ b/cosmos/airflow/compatibility.py
@@ -67,8 +67,8 @@ def delete_variable(key: str) -> None:
         Variable.delete(key)
 
 
-# ``airflow.exceptions.AirflowSkipException`` is deprecated on Airflow 3.1+; the Task SDK path does
-# not exist on Airflow 2 / 3.0 (hence the fallback; ignore is for the mypy env, pinned apache-airflow<3.1).
+# ``airflow.exceptions.AirflowSkipException`` is deprecated on Airflow 3.2+; the Task SDK path only
+# exposes it from 3.2 onwards, hence the fallback for Airflow 2 / 3.0 / 3.1.
 try:
     from airflow.sdk.exceptions import AirflowSkipException as AirflowSkipException  # type: ignore[attr-defined]
 except ImportError:
--- a/cosmos/airflow/compatibility.py
+++ b/cosmos/airflow/compatibility.py
@@ -65,3 +65,13 @@ def delete_variable(key: str) -> None:
         from airflow.models import Variable
 
         Variable.delete(key)
+
+
+# ``airflow.exceptions.AirflowSkipException`` is deprecated on Airflow 3.1+ in favour of the Task SDK
+# path, which does not exist on Airflow 2 / 3.0.
+try:
+    # attr-defined is ignored because the mypy env pins apache-airflow<3.1, where this path does not
+    # yet exist; at runtime the except branch handles Airflow 2 / 3.0.
+    from airflow.sdk.exceptions import AirflowSkipException as AirflowSkipException  # type: ignore[attr-defined]
+except ImportError:
+    from airflow.exceptions import AirflowSkipException as AirflowSkipException
--- a/cosmos/airflow/compatibility.py
+++ b/cosmos/airflow/compatibility.py
@@ -67,11 +67,9 @@ def delete_variable(key: str) -> None:
         Variable.delete(key)
 
 
-# ``airflow.exceptions.AirflowSkipException`` is deprecated on Airflow 3.1+ in favour of the Task SDK
-# path, which does not exist on Airflow 2 / 3.0.
+# ``airflow.exceptions.AirflowSkipException`` is deprecated on Airflow 3.1+; the Task SDK path does
+# not exist on Airflow 2 / 3.0 (hence the fallback; ignore is for the mypy env, pinned apache-airflow<3.1).
 try:
-    # attr-defined is ignored because the mypy env pins apache-airflow<3.1, where this path does not
-    # yet exist; at runtime the except branch handles Airflow 2 / 3.0.
     from airflow.sdk.exceptions import AirflowSkipException as AirflowSkipException  # type: ignore[attr-defined]
 except ImportError:
     from airflow.exceptions import AirflowSkipException as AirflowSkipException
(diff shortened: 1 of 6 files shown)


cosmos/airflow/compatibility.py
```diff
@@ -26,3 +26,13 @@
...
+
+
+# ``airflow.exceptions.AirflowSkipException`` is deprecated on Airflow 3.1+ in favour of the Task SDK
+# path, which does not exist on Airflow 2 / 3.0.
+try:
+    # attr-defined is ignored because the mypy env pins apache-airflow<3.1, where this path does not
+    # yet exist; at runtime the except branch handles Airflow 2 / 3.0.
+    from airflow.sdk.exceptions import AirflowSkipException as AirflowSkipException  # type: ignore[attr-defined]
+except ImportError:
+    from airflow.exceptions import AirflowSkipException as AirflowSkipException
```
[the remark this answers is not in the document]
pankajastro: Thanks, but I verified this is actually `attr-defined`, not `import-not-found`, and the ignore is required. `airflow.sdk.exceptions` *does* exist on the mypy-pinned `apache-airflow<3.1` — it just lacks the `AirflowSkipException` attribute there. So `ignore_missing_imports = true` does not help (it suppresses missing *modules*, not missing *attributes* of an existing module). Dropping the ignore fails mypy:

```
cosmos/airflow/compatibility.py:36: error: Module "airflow.sdk.exceptions" has no attribute "AirflowSkipException"  [attr-defined]
```

Kept the `# type: ignore[attr-defined]` and trimmed the surrounding comments in 83483caa.

cosmos/airflow/compatibility.py
```diff
@@ -65,3 +65,11 @@ def delete_variable(key: str) -> None:
         from airflow.models import Variable
 
         Variable.delete(key)
+
+
+# ``airflow.exceptions.AirflowSkipException`` is deprecated on Airflow 3.1+; the Task SDK path does
+# not exist on Airflow 2 / 3.0 (hence the fallback; ignore is for the mypy env, pinned apache-airflow<3.1).
+try:
+    from airflow.sdk.exceptions import AirflowSkipException as AirflowSkipException  # type: ignore[attr-defined]
```
tatiana, line 73: Can we verify the Airflow version boundary here? In the repo’s Airflow 3.1 matrix (`tests.py3.12-3.1-1.9`, Airflow 3.1.7), `airflow.sdk.exceptions` exists but does not expose `AirflowSkipException`, so this import falls back to `airflow.exceptions`.

That means the helper still resolves to the legacy class for Airflow 3.1.7, despite the comment/PR saying Airflow 3.1+. Please either narrow the version claim to the Airflow version where the SDK export actually exists, or add a regression test in the relevant Airflow matrix proving this resolves away from `airflow.exceptions`.

pankajastro: Good catch — you're right, and I verified it against the Airflow source:

- **Airflow 3.1.7**: `airflow.sdk.exceptions` exists but does **not** define `AirflowSkipException`, and `airflow.exceptions.AirflowSkipException` is **not** deprecated there (no warning). So the shim falls back to `airflow.exceptions`, which is harmless on 3.1.
- **Airflow 3.2.0**: `airflow.sdk.exceptions.AirflowSkipException` exists, and `airflow.exceptions.AirflowSkipException` is listed in `_DEPRECATED_EXCEPTIONS` and warns via `__getattr__`. The shim resolves to the SDK path and the warning is avoided.

So the real boundary is **3.2+**, not 3.1+. Addressed in 8cca2eb9:

1. Corrected the comment (3.1+ → 3.2+) and updated the PR description.
2. Added a regression test (`tests/airflow/test_compatibility.py`) asserting the shim resolves to `airflow.sdk.exceptions` on Airflow ≥ 3.2 and falls back to `airflow.exceptions` below it — so the 3.2 matrix proves it resolves away from the deprecated path.
3. While here: `cosmos/operators/_k8s_common.py` (added after this branch was cut, in #2488) still imported the deprecated path — routed it through the shim too. Also rebased on main, which dropped the now-obsolete `watcher_kubernetes.py` change (that skip logic moved into `_k8s_common.py`).

tatiana: thanks a lot, @pankajastro !

tatiana (approved):
Hi @pankajastro This needs a rebase before merge. Current `main` has refactored the Kubernetes watcher retry path into `cosmos/operators/_k8s_common.py`, and this branch now conflicts in `cosmos/operators/watcher_kubernetes.py`.

After rebase, please also update the moved Kubernetes helper: on current `main`, `cosmos/operators/_k8s_common.py` still imports `AirflowSkipException` from `airflow.exceptions`, so the Kubernetes watcher path would keep the deprecated import even though this PR updates the old `watcher_kubernetes.py` location.

PR #2764, opened 2026-06-03
docs: document `self.log` vs `get_logger` logging convention

pankajastro:
## Summary

The "Python Coding Standards → Logging" section in `AGENTS.md` told
contributors to "get loggers via `cosmos.log.get_logger`"
unconditionally. Read literally, that guides someone working in an
operator to add a module-level `get_logger` — the exact inconsistency
[#1157](https://github.com/astronomer/astronomer-cosmos/issues/1157) set
out to fix.

This scopes that guidance to library / module-level code and states the
operator rule explicitly: inside operator instance methods, log via
`self.log`. Cosmos's `AbstractDbtBase` routes `self.log` through
`get_logger`, so messages keep the `(astronomer-cosmos)` prefix and
scoped log levels. Module-level `get_logger(__name__)` stays correct
where `self` isn't available (module-level helpers, `@staticmethod`s).
The Yes/No examples show both cases.

## Why

The logging cleanup PRs (#2670, #2680, #2681) established the
conventions in code but they weren't all written down:

- `cosmos.log.get_logger` for library code — already documented ✅
- lazy logging — already documented ✅
- **`self.log` inside operator instance methods — not documented** ❌ (this PR)

Capturing it keeps human contributors and AI agents (Cursor, Copilot,
Claude) from re-introducing the workaround #1157 removed.

Docs-only; one-paragraph + example change to `AGENTS.md`.

## Related

- closes the documentation gap from #1157
- documents the convention landed in #2798 (operator logging consistency)
- follows #2670, #2680, #2681 (logging cleanup) and #2679 (logging docs)

PR #2798, opened 2026-06-09
Log via self.log inside dbt operators for consistency

pankajastro:
## Description

Operator logging was split between module-level `get_logger(__name__)` calls and `self.log` — the inconsistency tracked in #1157. This standardizes on **`self.log` inside instance methods of `AbstractDbtBase`-derived operators**.

`AbstractDbtBase` overrides `self.log` to return `get_logger(__name__)` (see `cosmos/operators/base.py`) and sits first in the MRO of every dbt operator, so `self.log` *is* the Cosmos logger: messages keep the `(astronomer-cosmos)` prefix, honour scoped log levels, and stay on the `cosmos.*` logger namespace that pytest's `caplog` can capture. The `{filename:lineno}` shown in the log still points at the real call site (Python derives it from the call stack, not the logger name).

### What changed
- `cosmos/operators/local.py` — 38 `logger.* → self.log.*` conversions
- `cosmos/operators/watcher.py` — 8 conversions (producer/source/test watcher operators, plus the `_event_callback` closure that captures `self`)
- `cosmos/operators/_asynchronous/bigquery.py` — 1 conversion; removed the now-unused `logger`/`get_logger` import

### What was intentionally kept on module-level `get_logger(__name__)`
Where no operator instance is in scope or `self.log` is *not* Cosmos-routed:
- module-level functions and `@staticmethod`s
- the sensors/triggers under `cosmos/operators/_watcher/` (`BaseConsumerSensor`, `WatcherTrigger`), which inherit the plain Airflow `LoggingMixin` logger rather than the overridden one

### Verification
- `ruff check` passes
- mypy on changed files adds no new errors (the only flagged item, `bigquery.py:26` unused `type: ignore`, pre-exists on `main` in this local env)
- `tests/operators/test_local.py`, `test_watcher.py`, `_asynchronous/test_bigquery.py`: identical results to baseline (the failures present are pre-existing `sqlite3.OperationalError: no such table` env/DB issues, unrelated to this change); the name-scoped `caplog` tests pass despite the logger name now resolving to `cosmos.operators.base`

## Related

- Follow-up to #1157 (operator logging consistency)
- Complements the docs PR #2764, which documents this `self.log` vs `get_logger` convention. The AGENTS.md wording there will be finalized to match this PR once the code direction is confirmed.

### DIFF
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -480,7 +480,7 @@ class AbstractDbtLocalBase(AbstractDbtBase):
                 ).delete()
                 session.add(rtif)
             else:
-                logger.info("Warning: ti is of type TaskInstancePydantic. Cannot update template_fields.")
+                self.log.info("Warning: ti is of type TaskInstancePydantic. Cannot update template_fields.")
 
         _override_rtif_airflow_2_x()
 
@@ -1031,7 +1031,7 @@ class DbtLocalBaseOperator(AbstractDbtLocalBase, BaseOperator):
                         outlets.append(DatasetAlias(name=dataset_alias_name))
 
                     else:
-                        logger.warning(f"Unknown outlet type {outlet}")  # Otherwise, pass
+                        self.log.warning(f"Unknown outlet type {outlet}")  # Otherwise, pass
 
                 operator_kwargs["outlets"] = outlets + [
                     DatasetAlias(name=get_dataset_alias_name(dag_id, task_group_id, self.task_id))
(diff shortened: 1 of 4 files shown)


cosmos/operators/local.py
pankajkoti, line 359: how about these that are still in the class inside a static method?

pankajkoti (approved):

cosmos/operators/local.py
pankajastro: Good catch — but for now leaving this one on the module logger since it's a `@staticmethod` with no self.

PR #2804, opened 2026-06-11
Use SDK Variable in WATCHER XCom backup to reduce log noise

jhorlima:
## Environment

- astronomer-cosmos 1.14.2
- Apache Airflow 3.1.8
- ExecutionMode.WATCHER

## What happened

The WATCHER XCom backup imports `airflow.models.Variable` directly. On Airflow 3 that path emits a `Variable.set from airflow.models is deprecated` warning on every `safe_xcom_push`, which adds up to hundreds of warnings per producer run.

## Fix

- Add small Variable compatibility helpers to `cosmos/airflow/compatibility.py` (the existing home for Airflow 2/3 import differences, alongside `EmptyOperator`). They prefer `airflow.sdk.Variable` and fall back to `airflow.models.Variable` on Airflow 2.
- Route the WATCHER XCom backup get/set/delete through these helpers instead of importing `airflow.models.Variable` directly.

The helpers are thin wrapper functions rather than a bare class re-export like `EmptyOperator`, because the `get` signature differs across versions (`airflow.sdk.Variable.get(key, default=...)` vs `airflow.models.Variable.get(key, default_var=...)`), so a direct re-export would not give callers a stable interface.

## Scope

This PR focuses only on the deprecated `Variable` log noise. The related Airflow 3 Asset URI migration warning is addressed separately in #2788, which takes a broader approach (warn once per manifest computation and skip URI generation when consumers have `emit_datasets=False`).

## Tests

- Added `tests/operators/_watcher/test_xcom.py` for the Variable helpers.
- Updated `tests/operators/_watcher/test_state.py` and `tests/operators/test_watcher.py` to patch the helpers instead of `airflow.models.Variable`, so the mocks intercept the call on Airflow 3 as well.

## Validation

Tested on Airflow 3.1.8 with a WATCHER DAG (around 906 nodes, 225 dbt models per run): the deprecated `Variable` warnings dropped from hundreds per run to 0, with all models completing normally.

### DIFF
--- a/cosmos/airflow/compatibility.py
+++ b/cosmos/airflow/compatibility.py
@@ -8,6 +8,8 @@ Airflow 3 pre-releases (e.g. 3.0.0rc1) are treated as Airflow 3.
 
 from __future__ import annotations
 
+from typing import Any
+
 from cosmos.constants import _AIRFLOW3_MAJOR_VERSION, AIRFLOW_VERSION
 
 if AIRFLOW_VERSION.major >= _AIRFLOW3_MAJOR_VERSION:
@@ -26,3 +28,40 @@ else:
 # operator as a string for later dynamic import (e.g. ``Task.operator_class``) rather than
 # referencing the class. Derived from the class itself, so it always matches the imported one.
 EMPTY_OPERATOR_CLASS_PATH = f"{EmptyOperator.__module__}.{EmptyOperator.__name__}"
+
+
+# ``Variable`` moved to the Task SDK in Airflow 3. ``airflow.sdk.Variable`` avoids the
+# deprecation warning that ``airflow.models.Variable`` emits per access on Airflow 3, but it only
+# works inside the task supervisor; outside it (e.g. ``dag.test()``) it raises ImportError. These
+# helpers prefer the SDK and fall back to ``airflow.models.Variable`` when it is not usable.
+def get_variable(key: str, default: Any = None) -> Any:
+    try:
+        from airflow.sdk import Variable
+
+        return Variable.get(key, default=default)
+    except ImportError:
+        from airflow.models import Variable as LegacyVariable
+
+        return LegacyVariable.get(key, default_var=default)
+
+
+def set_variable(key: str, value: str) -> None:
+    try:
+        from airflow.sdk import Variable
+
+        Variable.set(key, value)
+    except ImportError:
+        from airflow.models import Variable
+
+        Variable.set(key, value)
+
+
+def delete_variable(key: str) -> None:
+    try:
+        from airflow.sdk import Variable
+
+        Variable.delete(key)
+    except ImportError:
+        from airflow.models import Variable
+
+        Variable.delete(key)
--- a/cosmos/operators/_watcher/xcom.py
+++ b/cosmos/operators/_watcher/xcom.py
@@ -14,6 +14,7 @@ import re
 import zlib
 from typing import Any
 
+from cosmos.airflow.compatibility import delete_variable, get_variable, set_variable
 from cosmos.log import get_logger
 from cosmos.operators._watcher.state import safe_xcom_push
 
@@ -94,10 +95,8 @@ def _persist_backup(var_key: str, backup_buffer: dict[str, Any]) -> None:
     if not backup_buffer:
         return
 
-    from airflow.models import Variable
-
     compressed = base64.b64encode(zlib.compress(json.dumps(backup_buffer, default=str).encode("utf-8"))).decode("utf-8")
-    Variable.set(var_key, compressed)
+    set_variable(var_key, compressed)
     logger.debug("Persisted %d XCom entries to Variable '%s'", len(backup_buffer), var_key)
 
 
@@ -125,9 +124,7 @@ def _delete_xcom_backup_variable(context: Any) -> None:
     if not isinstance(var_key, str):
         return
     try:
-        from airflow.models import Variable
-
-        Variable.delete(var_key)
+        delete_variable(var_key)
         logger.debug("Deleted XCom backup Variable '%s'", var_key)
     except KeyError:
         pass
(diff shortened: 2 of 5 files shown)


cosmos/_utils/variable.py
```diff
@@ -0,0 +1,42 @@
+"""Airflow Variable compatibility helpers.
```
pankajkoti: I think c`osmos/airflow/compatibility.py`, which we introduced recently in #2758, module maybe a better home for these airflow compatibility mappings instead of creating a new module under `_utils`. Can we consider moving it there and using the style of exposure that we already have for `EmptyOperator` and its usages, using an import from that module into the existing modules where `Variable` gets used?

cosmos/_utils/variable.py
```diff
@@ -0,0 +1,42 @@
...
+
+def _variable_cls() -> Any:
+    try:
+        from airflow.sdk import Variable
+
+        return Variable
+    except ImportError:
+        from airflow.models import Variable
+
+        return Variable
```
pankajkoti: we probably don't need to define these then if we consider the above proposal

cosmos/dataset.py
```diff
@@ -224,15 +224,16 @@ def construct_dataset_uri(namespace: str, name: str) -> str:
...
+    if not settings.use_dataset_airflow3_uri_standard:
+        logger.warning(
+            "Airflow 3.0.0 Asset (Dataset) URIs validation rules changed and OpenLineage URIs "
+            "(standard used by Cosmos) are no longer accepted. "
+            "Therefore, if using Cosmos with Airflow 3, the Airflow Asset (Dataset) URI is now <%s>. "
+            "Before, with Airflow 2.x, the URI used to be <%s>. "
+            "Please, change any DAGs that were scheduled using the old standard to the new one.",
+            airflow_3_uri,
+            airflow_2_uri,
+        )
```
pankajkoti: we also have PR #2788 addressing the dataset warnings issue. Suggestion is to also have a look at it and see if that solves the problem you're facing with the noise of dataset warnings.

pankajkoti (commented):
Thanks for contributing this PR @jhorlima . I think this will definitely help reduce the Variable log noise you're observing and for the dataset warnings I think #2788 might be preferrable. 

I have some inline comments if you could consider. Also looks like some tests are failing the CI. Could you please take a look at it?

cosmos/_utils/variable.py
pankajkoti: But please check, and if you feel we need to take this route because the interfaces/args are different and the bare export may not work directly, please mention that and we can take another look if we want to keep this style of exposing the class or their methods.

jhorlima:
> Thanks for contributing this PR @jhorlima . I think this will definitely help reduce the Variable log noise you're observing and for the dataset warnings I think #2788 might be preferrable.
> 
> I have some inline comments if you could consider. Also looks like some tests are failing the CI. Could you please take a look at it?

Thanks for the review!

I've reworked the PR to focus only on the `Variable` deprecation noise, which is the part #2788 doesn't touch. Changes:

 - Dropped the `dataset.py` change and left the Asset URI warning to #2788.
 - Moved the Variable helpers into `cosmos/airflow/compatibility.py` as you suggested. I kept them as thin wrapper functions rather than a bare class re-export like `EmptyOperator`, because the `get` signature differs across versions (`airflow.sdk.Variable.get(key, default=...)` vs `airflow.models.Variable.get(key, default_var=...)`), so a direct re-export wouldn't give callers a stable interface.
 - Fixed the failing tests — they were patching `airflow.models.Variable`, which the helper no longer hits on Airflow 3. They now patch the helper.

cosmos/airflow/compatibility.py
```diff
@@ -26,3 +28,36 @@
...
+
+        return LegacyVariable.get(key, default_var=default)
+
+
+def variable_set(key: str, value: str, *, serialize_json: bool = False) -> None:
+    _variable_cls().set(key, value, serialize_json=serialize_json)
+
+
+def variable_delete(key: str) -> None:
+    _variable_cls().delete(key)
```
pankajkoti:
```suggestion
try:
        from airflow.sdk import Variable
except ImportError:
        from airflow.models import Variable

# ``Variable`` moved to the Task SDK in Airflow 3. It cannot be re-exported as-is like
# ``EmptyOperator`` because ``get`` differs across versions (``default`` vs ``default_var``),
# so this thin wrapper exposes a stable interface.
def get_variable(key: str, default: Any = None) -> Any:
    if AIRFLOW_VERSION.major >= _AIRFLOW3_MAJOR_VERSION:
        return Variable.get(key, default=default)
    else:
        return Variable.get(key, default_var=default)

```

WDYT we have this minimum code here and then use 
`from cosmos.airflow.compatibility import Variable, get_variable` ?

pankajkoti (approved):
Thanks a lot for addressing the previous feedback.

In this round, I felt that we could maybe reduce the number of utilities that we're writing here.  I agree on having `variable_get` -> suggestion to rename to `get_variable` or `get_airflow_variable` and then we might not need the other utilities if we directly use `Variable` from this module. Let me know what you think. Thanks again!

cosmos/airflow/compatibility.py
jhorlima: Done, I applied the suggestion. Thanks for the reviews!

pankajkoti:
Hi @jhorlima, thanks for considering and addressing the feedback so promptly. Looks like some tests are failing in the CI with this change. Would you have a chance to take a look at what could be causing them to fail?

jhorlima:
> Hi @jhorlima, thanks for considering and addressing the feedback so promptly. Looks like some tests are failing in the CI with this change. Would you have a chance to take a look at what could be causing them to fail?

Pushed a fix for the CI failures. Turns out `airflow.sdk.Variable` only works inside the task supervisor, so it blew up under `dag.test()` in the integration tests. The helpers now prefer the SDK and fall back to `airflow.models.Variable` when it's not available, so production stays warning-free and the test paths work. Also rebased on latest main.

pankajkoti (approved):
Thanks a lot for the quick iterations on the PR here @jhorlima. Appreciate those. I guess let's get this in and see if we can optimise further in a follow-up PR. Thanks again!

pankajkoti:
Another follow-up idea here is to introduce a pre-commit hook so that we don't introduce this issue again for future variable references directly from Airflow models, but always reference them from the `cosmos.airflow.compatibility` module if you'd like to contribute @jhorlima :)

jhorlima:
> Another follow-up idea here is to introduce a pre-commit hook so that we don't introduce this issue again for future variable references directly from Airflow models, but always reference them from the `cosmos.airflow.compatibility` module if you'd like to contribute @jhorlima :)

Sounds good, happy to do the pre-commit guardrail as a follow-up. I'll look into the impact across the project and open a PR once it's set up.

Issue #2456, opened 2026-03-11
Include non-error logs in consumer task output


pankajastro:
This PR builds on [#2431](https://github.com/astronomer/astronomer-cosmos/pull/2431)
, which added a mechanism to ship error logs from the producer to the consumer task.

Currently, other (non-error) logs are stored in XCom, but they get overwritten by newer logs.

This change:
- Includes non-error logs in the shipping mechanism.
- Avoids duplicate logs, ensuring all relevant logs are captured in the consumer task.

michal-mrazek:
Hi @pankajastro do you plan to tackle this or should I attempt? :)

pankajastro:
Hi @michal-mrazek, there has been some improvement in 1.14.0, but please feel free to improve it further.

tatiana:
Next steps:
- Log ticket for the duplicated log, prioritise it for today (timebox)
- Reasure customer that we are not running command twice - it is just a logging issue

tatiana:
@michal-mrazek just checkling, do you still have bandwith to work on this one, or should we unassign so someone else can work on it?

michal-mrazek:
Hi @tatiana , I currently have in flight change for the test sensor to include something like "3 out of 5 tests failed (test1, test2, test3)". So I will push this soon. Feel free to assign this to someone else in the meantime :)

PR #2806, opened 2026-06-11
Capture dbt-runner error message in WATCHER consumer logs

pankajastro:
## Problem

In `ExecutionMode.WATCHER` the producer ships each node's structured dbt event to the consumer via XCom. The error text lives in different places depending on invocation mode:

- **dbt-runner mode:** `NodeFinished.data.run_result.message`
- **subprocess mode:** `RunResultError.data.msg`

`_process_dbt_log_event` read only `data.msg` / `info.msg`, which are **empty on `NodeFinished`**. So a failed **dbt-runner** node reached the consumer with no error text — the consumer logged `[ERROR] Start.. Finish..` with no detail and pointed users at the producer log.

## Fix

Read `run_result.message` first when building the per-node `_dbt_event`, so the dbt-runner error reaches the consumer.

## Scope — read before reviewing

This is split PR 2 of 3 carved out of #2792 (per review feedback):
1. duplicate consumer logs → #2805
2. **this PR** — dbt-runner error capture, on the existing **per-event** write path
3. per-node event buffering (architectural change)

**Important:** this fixes the **dbt-runner** case cleanly and statelessly. The **subprocess** case is *not* fully fixed here: there the error arrives in a *later* `RunResultError` (after the terminal event) and must not be overwritten by a trailing empty message — which requires retained per-node state. That robustness lands with the **buffering** PR (3), where the `RunResultError` re-flush and "don't overwrite the error"/"ignore `\"None\"` status" guards live.

## Testing

- Unit test `test_process_dbt_log_event_captures_error_from_run_result_message`: a `NodeFinished` with `run_result.message` set and empty `data.msg`/`info.msg` pushes the error to `_dbt_event`.
- Integration test `test_process_dbt_log_event_surfaces_dbt_runner_error_from_real_event` (added per review feedback — aligned to dbt's real event interface, not a mock): runs real dbt-runner against the failing `watcher_failing_tests` project and asserts `_process_dbt_log_event` surfaces the error from the actual `NodeFinished.run_result.message`. It first asserts the premise (`data.msg` empty) and **fails without the fix** (the consumer would otherwise get the generic `Finished running node ...`).
- Full `tests/operators/_watcher/` suite passes on the AF 3.0 hatch env; `pre-commit` (ruff/black/mypy) passes.

Before/after consumer-log screenshots below.

AFTER

Screenshot 2026-06-27 at 7 15 04 PM

BEFORE

Screenshot 2026-06-27 at 7 17 04 PM

Reproduce: add a failing model and run a WATCHER DbtDag (dbt-runner mode); the error now appears in the failing node's **consumer** task log.
```sql
select this_column_does_not_exist_at_all
```
Relates to #2456.

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -113,7 +113,12 @@ def _process_dbt_log_event(task_instance: Any, dbt_log: dict[str, Any]) -> None:
     unique_id = node_info.get("unique_id")
     start_time = node_info.get("node_started_at")
     finish_time = node_info.get("node_finished_at")
-    msg = data.get("msg") or info.get("msg")
+    # The error text lives in different places depending on the invocation mode: ``run_result.message`` on
+    # ``NodeFinished`` (dbt-runner mode) and ``data.msg`` on ``RunResultError`` (subprocess mode). The old code
+    # read only ``data.msg``/``info.msg``, which are empty on ``NodeFinished`` -- so a failed dbt-runner node
+    # reached the consumer with no error message. Read ``run_result.message`` first to capture it.
+    run_result = data.get("run_result") or {}
+    msg = run_result.get("message") or data.get("msg") or info.get("msg")
 
     if unique_id:
         dbt_event = {
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -114,9 +114,7 @@ def _process_dbt_log_event(task_instance: Any, dbt_log: dict[str, Any]) -> None:
     start_time = node_info.get("node_started_at")
     finish_time = node_info.get("node_finished_at")
     # The error text lives in different places depending on the invocation mode: ``run_result.message`` on
-    # ``NodeFinished`` (dbt-runner mode) and ``data.msg`` on ``RunResultError`` (subprocess mode). The old code
-    # read only ``data.msg``/``info.msg``, which are empty on ``NodeFinished`` -- so a failed dbt-runner node
-    # reached the consumer with no error message. Read ``run_result.message`` first to capture it.
+    # ``NodeFinished`` (dbt-runner mode) and ``data.msg`` on ``RunResultError`` (subprocess mode).
     run_result = data.get("run_result") or {}
     msg = run_result.get("message") or data.get("msg") or info.get("msg")
 
--- a/tests/operators/_watcher/test_watcher_base.py
+++ b/tests/operators/_watcher/test_watcher_base.py
@@ -91,6 +91,31 @@ class TestBaseConsumerSensor:
             _process_dbt_log_event(task_instance, dbt_log)
             mock_push.assert_not_called()
 
+    def test_process_dbt_log_event_captures_error_from_run_result_message(self):
+        """dbt-runner mode: the error text lives in NodeFinished.data.run_result.message (data.msg/info.msg
+        are empty there). It must reach the consumer's _dbt_event."""
+        task_instance = Mock()
+        dbt_log = {
+            "data": {
+                "node_info": {
+                    "unique_id": "model.test.bad_model",
+                    "node_status": "error",
+                    "node_started_at": "2024-01-01T00:00:00",
+                    "node_finished_at": "2024-01-01T00:01:00",
+                },
+                "run_result": {"status": "error", "message": "Runtime Error ... Catalog Error: Table does not exist!"},
+                "msg": "",
+            },
+            "info": {"name": "NodeFinished", "msg": ""},
+        }
+
+        with patch("cosmos.operators._watcher.base.safe_xcom_push") as mock_push:
+            _process_dbt_log_event(task_instance, dbt_log)
+
+        value = mock_push.call_args.kwargs["value"]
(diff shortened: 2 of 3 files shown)


cosmos/operators/_watcher/base.py
```diff
@@ -113,7 +113,12 @@ def _process_dbt_log_event(task_instance: Any, dbt_log: dict[str, Any]) -> None:
     unique_id = node_info.get("unique_id")
     start_time = node_info.get("node_started_at")
     finish_time = node_info.get("node_finished_at")
-    msg = data.get("msg") or info.get("msg")
+    # The error text lives in different places depending on the invocation mode: ``run_result.message`` on
+    # ``NodeFinished`` (dbt-runner mode) and ``data.msg`` on ``RunResultError`` (subprocess mode). The old code
+    # read only ``data.msg``/``info.msg``, which are empty on ``NodeFinished`` -- so a failed dbt-runner node
+    # reached the consumer with no error message. Read ``run_result.message`` first to capture it.
```
tatiana: AI agents can be very verbose in the comments they add, describing things that are no longer relevant after the fix was done. In the living code, most times, we do not care about the previous code state - we care about the comment that explains why it is the way it is.
```suggestion
    # ``NodeFinished`` (dbt-runner mode) and ``data.msg`` on ``RunResultError`` (subprocess mode).```

File: tests/operators/_watcher/test_watcher_base.py [hunk 2]
```diff
@@ -91,6 +91,31 @@ def test_process_dbt_log_event_skips_when_no_unique_id(self):
             _process_dbt_log_event(task_instance, dbt_log)
             mock_push.assert_not_called()
 
+    def test_process_dbt_log_event_captures_error_from_run_result_message(self):
```
tatiana, line 94: Could we have a functional/integration test to cover this? Mocks can easily get out dated, and since this fix is aligned with a dbt interface, I believe another type of test would be more meaningful.

tatiana (approved):
Thanks for the fix, @pankajastro !

Please, before merging this PR:
1. Address the inline comment
2. Add an example of a DAG that leads to the problem (if possible, an existing DAG in our repo - we have a few examples in the failed_dags folder)
3 Share a screenshot of the error not being displayed in the consumer task before fix
4. Share a screenshot of the error being captured/displayed in the consumer task after the fix

cosmos/operators/_watcher/base.py
pankajastro: Done in b77990af — trimmed the comment to just why the lookup order exists, dropping the previous-code narrative.

tests/operators/_watcher/test_watcher_base.py
pankajastro: Added an integration test (0b47bf19) in `tests/operators/test_watcher.py::test_process_dbt_log_event_surfaces_dbt_runner_error_from_real_event`: it runs real dbt-runner against the failing `watcher_failing_tests` project and asserts `_process_dbt_log_event` surfaces the error from the real `NodeFinished.run_result.message` (`data.msg`/`info.msg` are empty there). Verified it fails without the fix (the consumer would get the generic `Finished running node ...` instead of the dbt error), so it's aligned with dbt's event interface rather than a mock.

pankajastro:
@tatiana thanks for the review — addressed your feedback:

1. **Comment trimmed** (`da483e85`) — kept only why the lookup order exists, dropped the previous-code narrative.
2. **Functional/integration test added** (`f7f5ec6c`) — `test_process_dbt_log_event_surfaces_dbt_runner_error_from_real_event` runs **real dbt-runner** against the failing `watcher_failing_tests` project and asserts the error is surfaced from the actual `NodeFinished.run_result.message`. It's aligned to dbt's event interface (not a mock) and **fails without the fix** (the consumer would otherwise get the generic `Finished running node ...`).
3. **Before/after screenshots** added to the PR description, reproduced on a jaffle_shop project with one failing model.

tests/operators/test_watcher.py
```diff
@@ -2435,6 +2436,54 @@ def test_dbt_dag_with_watcher_and_failing_model(caplog):
...
+
+    project_dir = tmp_path / "watcher_failing_tests"
+    shutil.copytree(DBT_WATCHER_FAILING_TESTS_PATH, project_dir)
+    profiles_dir = tmp_path / "profiles"
+    profiles_dir.mkdir()
+    (profiles_dir / "profiles.yml").write_text(
+        "default:\n  target: dev\n  outputs:\n    dev:\n      type: postgres\n      host: 0.0.0.0\n"
+        "      port: 5432\n      user: postgres\n      password: postgres\n      dbname: postgres\n"
+        "      schema: public\n      threads: 4\n"
+    )
```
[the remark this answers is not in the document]
pankajastro: Good call — fixed in d3ea8a23. The integration test now generates `profiles.yml` via the shared `example_conn`-based `profile_config.ensure_profile()` (the same ProfileConfig the other watcher integration tests use), so connection details are no longer hardcoded or duplicated and the `host: 0.0.0.0` bind-all address is gone. Verified it still passes with the fix and fails without it.

PR #2814, opened 2026-06-17
Fix UnboundLocalError in build_airflow_graph with empty nodes

### LINKED ISSUE #2813: [Bug] UnboundLocalError: 'task_or_group_args' in build_airflow_graph with empty nodes + TestBehavior.AFTER_ALL
## Bug

`build_airflow_graph()` raises `UnboundLocalError: cannot access local variable 'task_or_group_args' where it is not associated with a value` when it is called with an **empty `nodes`** mapping **and** `RenderConfig(test_behavior=TestBehavior.AFTER_ALL)`.

```
UnboundLocalError: cannot access local variable 'task_or_group_args' where it is not associated with a value
  File "cosmos/airflow/graph.py", line 1028, in build_airflow_graph
    **task_or_group_args,
```

## Root cause

`task_or_group_args` is assigned **inside** the per-node loop in `cosmos/airflow/graph.py`:

```python
    ...
    task_or_group_args = { ... }      # only bound if the loop body runs
```

It is then read **after** the loop in the `AFTER_ALL` branch:

```python
if render_config.test_behavior == TestBehavior.AFTER_ALL:
    ...
    test_task_args = {
        **task_or_group_args,         # UnboundLocalError when nodes was empty
        ...
    }
```

If `nodes` is empty, the loop body never executes, so `task_or_group_args` is never bound, and the `AFTER_ALL` block fails.

The `BUILD` / `AFTER_EACH` branch references the same variable, but it only does so while iterating over `detached_nodes` (also empty when `nodes` is empty), so it never trips there. **`AFTER_ALL` is the only trigger.**

## How we hit it

DAG-parse / integrity tests that mock the dbt manifest to `{}` (e.g. `LoadMode.DBT_MANIFEST` with no warehouse/S3 access in CI), where the DAG uses `test_behavior=TestBehavior.AFTER_ALL`. The empty manifest yields empty `filtered_nodes`, and graph construction blows up at parse time instead of producing an empty graph.

## Expected behavior

An empty `nodes` mapping should not raise. It should produce an empty `tasks_map` (no leaves → no meaningful `AFTER_ALL` test task to wire up).

## Suggested fix

Hoist the static parts of `task_or_group_args` out of the loop (the only per-node key is `"node"`), and/or guard the `AFTER_ALL` block when there are no nodes/leaves.

## Versions

- Cosmos: 1.14.2 (also present on `main` as of this writing — `cosmos/airflow/graph.py`)
- Reproduces with `ExecutionMode.LOCAL`, `LoadMode.DBT_MANIFEST`, `TestBehavior.AFTER_ALL`

## Related

PR #2689 (refactor of `build_airflow_graph`) touches this code and may address it.

m1racoli:
## Description

`build_airflow_graph` raised `UnboundLocalError: cannot access local variable 'task_or_group_args'` when called with an empty `nodes` mapping (e.g. an empty or mocked dbt manifest) together with `RenderConfig(test_behavior=TestBehavior.AFTER_ALL)`, because `task_or_group_args` was only bound inside the per-node loop yet read afterwards in the `AFTER_ALL` branch. This moves the shared `task_or_group_args` to before the loop so it is always bound, merging in only the per-node `task_group`/`node` inside the loop. Behaviour for non-empty manifests is unchanged; an empty manifest now yields just the standalone project test task instead of crashing. A regression test covering the empty-nodes + `AFTER_ALL` scenario is included.

## Related Issue(s)

closes #2813

## Breaking Change?

No.

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections import OrderedDict, defaultdict
 from collections.abc import Callable
 from copy import deepcopy
+from pathlib import Path
 from typing import Any
 
 try:  # Airflow 3
@@ -1060,13 +1061,12 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
             tests_per_model=tests_per_model,
         )
 
-    # Arguments shared by every node's task/group as well as the AFTER_ALL/detached test tasks below.
-    # Defined before the loop so it is always bound, even when ``nodes`` is empty
-    task_or_group_args: dict[str, Any] = {
+    # Immutable base args shared by every node's task/group as well as the AFTER_ALL/detached test tasks below.
+    # ``node`` and ``task_group`` are intentionally omitted so per-node dicts are built cleanly each iteration.
+    base_task_or_group_args: dict[str, Any] = {
         # Arguments to this method:
         "dag": dag,
         "task_group": parent_task_group,
-        "node": None,
         "task_args": task_args,
         "dbt_project_name": dbt_project_name,
         "render_config": render_config,
@@ -1080,6 +1080,7 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
         "node_converters": render_config.node_converters or {},
     }
 
+    last_node: DbtNode | None = None
     for node_id, node in nodes.items():
         task_group = (
             create_task_groups_based_on_folder(dag, node, parent_task_group, task_groups)
@@ -1087,7 +1088,7 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
             else task_group
         )
         task_or_group_args = {
-            **task_or_group_args,
+            **base_task_or_group_args,
             "task_group": task_group,
             "node": node,
         }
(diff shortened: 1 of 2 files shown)


cosmos/airflow/graph.py
```diff
@@ -1060,31 +1060,40 @@ def build_airflow_graph(  # noqa: C901 TODO: https://github.com/astronomer/astro
             tests_per_model=tests_per_model,
         )
 
+    # Arguments shared by every node's task/group as well as the AFTER_ALL/detached test tasks below.
+    # Defined before the loop so it is always bound, even when ``nodes`` is empty (e.g. an empty or
+    # mocked manifest), which would otherwise raise UnboundLocalError in the AFTER_ALL branch.
```
```suggestion
    # Defined before the loop so it is always bound, even when ``nodes`` is empty
```

tatiana (approved):
@m1racoli thanks a lot for the quick fix & regression test!
Please, could you address Copilot's feedback before we merge it?
We'll be shipping this as part of Cosmos 1.15.0, expected to be released on 1st July.

pankajkoti (commented):
Thanks for contributing this fix @m1racoli. I am wondering whether we should create a test task (with AFTER_ALL mode) at all when there are no nodes or models rendered. What does the Graph look like with this change? Does it just show the test task and no other tasks? Do we want to have that, or do users want to have that?

After issue #2813, which was marked as related to #2689, I asked @00yhj22-debug whether they could also resolve it in their PR, since Copilot had flagged the same issue there. And I see they have attempted to address the issue by not creating a test task if there are no nodes, and I am leaning toward preferring that option. Would be nice to discuss further and finalise the approach before we merge one of these two PRs.

cc: @tatiana

tatiana:
@pankajkoti I agree with your suggestion. I'm closing this PR in favour of #2689

Regardless, @m1racoli, thanks a lot for working on this to make Cosmos better!

PR #2818, opened 2026-06-18
fix: validate manifest file and improve error handling in _load_manifest_from_file

April-Sonnet:
Improve error handling when loading a dbt manifest JSON file. Previously, attempting to load an empty or invalid manifest file would raise an opaque JSONDecodeError with no indication of which file failed or why.

Changes:
1. **Empty file check**: Before parsing, verify the file is non-empty and raise CosmosLoadDbtException with the path
2. **JSON decode error handling**: Wrap json.load/orjson.loads in try/except for JSONDecodeError/ValueError
3. **File stat error handling**: Wrap stat() call to catch OSError

Closes #2817

PR #2819, opened 2026-06-18
fix: Raise exception on empty or invalid manifest and tests

### LINKED ISSUE #2817: Raise a clear error when the dbt manifest is empty or invalid (LoadMode.DBT_MANIFEST)
### Description

When using `LoadMode.DBT_MANIFEST`, Cosmos loads the manifest in `DbtGraph.load_from_dbt_manifest` via `json.load(fp)` on the configured `manifest_path`. If the file is empty, truncated, or not valid JSON, `json.load` raises a raw `json.JSONDecodeError`. Because this runs at DAG-parse time (inside `DbtTaskGroup`/`DbtDag` construction), the unhandled exception **breaks DAG import** and surfaces only as an opaque `JSONDecodeError` deep in Cosmos internals — with no indication of which manifest path failed or that the file is malformed.

### Current behavior

A malformed/empty `manifest.json` fails with a bare traceback through `cosmos/dbt/graph.py` → `load_from_dbt_manifest` → `json.load`:

```
File ".../cosmos/airflow/task_group.py", line ..., in __init__
File ".../cosmos/converter.py", line ..., in __init__
    self.dbt_graph.load(method=..., execution_mode=...)
File ".../cosmos/dbt/graph.py", line ..., in load_from_dbt_manifest
    manifest = json.load(fp)
...
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

The error gives no hint about the cause or which path was read.

### Expected behavior

Validate the manifest before parsing and raise an actionable `CosmosLoadDbtException` that includes the path and the underlying cause, e.g.:

```
Failed to load dbt manifest at <path>: file is empty or not valid JSON (<original error>)
```

### Proposed implementation

* In `load_from_dbt_manifest`, check the file exists / is non-empty and wrap the `json.load` call to re-raise as `CosmosLoadDbtException` with the manifest path and original error.
* Add a test with an empty/truncated manifest fixture.

### Notes

Related parsing work: astronomer/astronomer-cosmos#2294, astronomer/astronomer-cosmos#2486. This is a robustness / error-message improvement on the `LoadMode.DBT_MANIFEST` path (not a performance change).

karanw330:
## Description

Currently, when using LoadMode.DBT_MANIFEST, astronomer-cosmos parses the configured manifest.json file. If the file is empty, truncated, or malformed, standard Python libraries (json / orjson) raise raw, generic exceptions (JSONDecodeError / ValueError). This breaks Airflow DAG imports at parse time with an opaque error, without indicating which manifest path failed or why.

## Related Issue(s)

Closes #2817

## Breaking Change?

No. This only handles unhandled exceptions to provide better error messages.

## Checklist


pankajastro (commented):
We already have a PR: https://github.com/astronomer/astronomer-cosmos/pull/2818 which aim to improve same code path.

karanw330:
Got it. I noticed that the referenced pr doesn't include the same exception handling for orjson and tests for empty/invalid json. Happy to keep this one opened if you want, the rest I'll leave upto you!

cosmos/dbt/graph.py
```diff
@@ -1358,14 +1358,31 @@ def _load_manifest_from_file(self, manifest_path: Path | ObjectStoragePath) -> d
...
+            try:
+                with manifest_path.open("r") as fp:
+                    content = fp.read()
+                    if not content or content.isspace():
+                        raise CosmosLoadDbtException(f"Failed to load dbt manifest at `{manifest_path}`: file is empty")
+                    manifest = json.loads(content)
+            except (json.JSONDecodeError, UnicodeDecodeError) as e:
+                raise CosmosLoadDbtException(
+                    f"Failed to load dbt manifest at `{manifest_path}`: file is not valid JSON ({e})"
+                ) from e
```
tatiana: These lines look almost identical to 1363-1372. Could we further improve the code to avoid duplication?

tatiana (commented):
Hi @karanw330, thanks a lot for your contribution to this. It is an accurate implementation, and unit tests checks passed, so we are sticking with your change. 

Please, could you address the feedback inline?

We're aiming to release Cosmos 1.15.0 by 1st July. If you could address this feedback soon, we can ship it as part of this release.

karanw330:
@tatiana updated graph.py with the requested changes, refactored to avoid code duplication. Now both of the parsers use a shared try catch block.

tatiana (approved):
Thanks a lot for addressing the feedback, @karanw330 !

PR #2824, opened 2026-06-19
Fix `group_nodes_by_folder` collapsing folders that share a leaf name

anor4k:
## Description

When `RenderConfig.group_nodes_by_folder=True`, `create_task_groups_based_on_folder` cached the generated `TaskGroup`s by the **bare folder name**. As a result, folders that share a leaf name under different parents — e.g. `models/staging/aurora` and `models/marts/aurora` — collapsed into a single `TaskGroup`: the first `aurora` group created was reused for every later folder named `aurora`, so its models were placed under the wrong parent.

This re-keys the group cache by the **cumulative folder path** instead of the leaf name, so same-named folders under different parents render as distinct task groups (`models.staging.aurora` vs `models.marts.aurora`) while shared ancestors (e.g. `models/`) are still reused. The `group_id` shown in the UI is unchanged (still the leaf folder name, scoped under its parent).

### Before
```
models/
  aurora/          # contains BOTH staging/aurora and marts/aurora models
```

### After
```
models/
  staging/
    aurora/        # staging models only
  marts/
    aurora/        # marts models only
```

## Tests

- **Added** `test_create_task_groups_based_on_folder_distinguishes_same_leaf_name` — a regression test that builds two nodes under `staging/aurora` and `marts/aurora` and asserts they map to distinct task groups (`models.staging.aurora` vs `models.marts.aurora`) with the cache keyed by cumulative path.
- **Updated existing folder-grouping tests** in `tests/airflow/test_graph.py` that had encoded the old (buggy) behaviour. In the test fixtures, `child`/`child2_v2` live under `gen3/models/` while `parent` lives under `gen2/models/`; previously the `gen3/models` folder collapsed into `gen2.models`, so the tests asserted task ids like `gen2.models.child_run`. With the fix they are correctly `gen3.models.child_run`, and the `len(task_groups)` assertions increase by one because `gen3.models` is now its own group:
  - `test_build_airflow_graph_with_after_each`: task ids `gen2.models.child*` → `gen3.models.child*`; group count `4` → `5`.
  - `test_build_airflow_graph_with_after_all`: same id update; group count `3` → `4`.
  - `test_build_airflow_graph_with_build`: same id update; group count `3` → `4`.
  - `test_custom_meta`: the custom `pool`/`queue` assertions now key off `gen3.models.child*` ids.

  (`test_build_airflow_graph_with_override_profile_config` only references `parent`, which stays under `gen2.models`, so it is unchanged.)

## Breaking Change?

Yes, task groups will be rendered with a different name than PR - since their name will contain a hierarchy.

## Checklist

- [ ] I have made corresponding changes to the documentation (not required)
- [x] I have added tests that prove my fix is effective (`test_create_task_groups_based_on_folder_distinguishes_same_leaf_name`), and updated existing folder-grouping tests that encoded the previous behaviour

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -927,12 +927,18 @@ def create_task_groups_based_on_folder(
     """
     task_group = None
     resource_file_path_parts = str(node.original_file_path).split("/")[:-1]
+    # Cache groups by their cumulative path rather than the bare folder name, so
+    # that folders sharing a leaf name under different parents (e.g.
+    # ``staging/aurora`` and ``marts/aurora``) become distinct task groups instead
+    # of collapsing into the first one created.
+    cumulative_path = ""
     for resource_file_path_part in resource_file_path_parts:
-        if resource_file_path_part in task_groups:
-            task_group = task_groups[resource_file_path_part]
+        cumulative_path = f"{cumulative_path}/{resource_file_path_part}" if cumulative_path else resource_file_path_part
+        if cumulative_path in task_groups:
+            task_group = task_groups[cumulative_path]
         else:
             task_group = TaskGroup(dag=dag, group_id=resource_file_path_part, parent_group=parent_task_group)
-            task_groups[resource_file_path_part] = task_group
+            task_groups[cumulative_path] = task_group
         parent_task_group = task_group
     return task_group
 
--- a/tests/airflow/test_graph.py
+++ b/tests/airflow/test_graph.py
@@ -28,6 +28,7 @@ from cosmos.airflow.graph import (
     calculate_detached_node_name,
     calculate_leaves,
     calculate_operator_class,
+    create_task_groups_based_on_folder,
     create_task_metadata,
     create_test_task_metadata,
     exclude_detached_tests_if_needed,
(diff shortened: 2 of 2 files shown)


tatiana (commented):
@anor4k thanks a lot for reporting/fixing this issue. This looks great!

My only concern is that this changes the names of the generated Cosmos task groups. That would be a breaking change for users who reference those task groups from other tasks. I’m aware of at least one user with that pattern, e.g.:

```
non_cosmos_taks = ...
cosmos_dbt_task_group = task_group.children["tg_a.models"].children[f"tg_a.models.marts"]

non_cosmos_taks >> cosmos_dbt_task_group
```

Could we make this fix configurable instead? To avoid a breaking change, I’d suggest defaulting it to `False` and following the same config pattern as the existing Cosmos settings: https://astronomer.github.io/astronomer-cosmos/reference/configs/cosmos-conf.html

A possible config name is `enable_hierarchical_naming_for_group_nodes_by_folder`, though I’m open to other naming suggestions.

We may be able to add this fix to the 1.15.0 release, planned for 1st July, depending on your availability to address the feedback.

tatiana (approved):
We really want to ship this change as part of Cosmos 1.15.0, so I'm merging and creating a follow-up PR to add the configuration and disable it by default, just so we don't introduce a breaking change. In Cosmos 2.0 we'll stick to this original implementation and always name the folder based on the hierarchy.

anor4k:
sorry, I missed the follow-up! great to hear it's merged, gating behind a config makes sense to avoid breaking changes. thanks for following up!

PR #2835, opened 2026-06-24
docs: Document retrying models but not tests

### LINKED ISSUE #2130: Identify and document/create an example with the best way to retry model runs but not tests
During a [user-session](https://us-35700.app.gong.io/call?id=3308719594734248781&email_type=call-ready-notification&xtid=3pi87k5s8durzd4in7z), it was raised that they would like to retry models, but not tests, because some of the tests would take 30 minutes to run, and while they were running, following DAG runs were blocked.

I suggested one option (untested) could be potentially to use:
https://astronomer.github.io/astronomer-cosmos/configuration/operator-args.html#overriding-operator-arguments-per-dbt-node-or-group-of-nodes

Defining the retry number similar to this example:
https://astronomer.github.io/astronomer-cosmos/profiles/index.html#customising-the-profile-config-per-dbt-node

We should test, validate and document what is the reocmmended way to accomplish this.

pankajastro:
## Problem

Issue #2130: users want to retry **model runs** but **not tests** — slow tests (e.g. 30 min) would otherwise be retried and block subsequent DAG runs.

## Solution

Documents how to retry models without retrying tests in the testing-behavior guide:

- **`TestBehavior.AFTER_EACH` (default)**: since #2936, a test's own `operator_kwargs` (e.g. `retries: 0`) overrides what it would otherwise inherit from the model it tests. This is the primary recipe — see `overriding-operator-args.rst`.
- **`TestBehavior.AFTER_ALL`**: documented here as an alternative. The single end-of-DAG test task has no associated node, so it keeps the DAG-level `retries` default while per-node `retries` apply only to model tasks.

This PR originally proposed `AFTER_ALL` as the only way to solve this, since `AFTER_EACH` unconditionally inherited the model's `retries` at the time and ignored any `operator_kwargs` set on the test itself. #2936 has since merged and fixed that directly, so this PR was rebased on top: the stale "`AFTER_EACH` ignores test kwargs" note and the regression test asserting that now-fixed behavior were removed, and the section now points to #2936's docs first.

## Verification

- `test_retries_after_all_separates_model_and_test` and `test_retries_after_each_test_inherits_model_operator_kwargs` pass, alongside the full `tests/airflow/test_graph.py` suite (166 passed).
- `sphinx-build` is clean; the `:ref:operator-args-test-nodes` cross-reference resolves.

Closes #2130

### DIFF
--- a/docs/guides/translate_dbt_to_airflow/testing-behavior.rst
+++ b/docs/guides/translate_dbt_to_airflow/testing-behavior.rst
@@ -49,6 +49,41 @@ Finally, an example DAG and how it is rendered in the `Apache Airflow® <https:/
 
 .. image:: ../../_static/test_behavior_build.png
 
+Retrying models but not tests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A common requirement is to retry failing model runs while never retrying tests, for example when tests are slow and
+retrying them would delay or block subsequent DAG runs.
+
+With the default ``TestBehavior.AFTER_EACH``, this is done by setting ``retries`` on the models and a different
+``retries`` on their dbt tests: see :ref:`operator-args-test-nodes`.
+
+If you use ``TestBehavior.AFTER_ALL`` instead, tests run as a single task at the end of the DAG that is independent
+of any model node, so it keeps the DAG-level ``retries`` default while per-node ``retries`` apply only to the model
+(run) tasks. Set ``retries=0`` as the DAG-wide default and override ``retries`` for the models in your
+``dbt_project.yml``:
+
+.. code-block:: python
+
+    from cosmos import DbtDag, RenderConfig
+    from cosmos.constants import TestBehavior
+
+    DbtDag(
+        # ...
+        operator_args={"retries": 0},
+        render_config=RenderConfig(test_behavior=TestBehavior.AFTER_ALL),
+    )
+
+.. code-block:: yaml
+
+    # dbt_project.yml — retry every model run twice, leaving the final test task at retries=0
+    models:
+      my_dbt_project:
+        +meta:
+          cosmos:
+            operator_kwargs:
+              retries: 2
+
 Warning behavior
 ~~~~~~~~~~~~~~~~
 
--- a/tests/airflow/test_graph.py
+++ b/tests/airflow/test_graph.py
@@ -2933,6 +2933,54 @@ def test_create_task_metadata_seed_rendering_always_no_flag():
     assert "should_run_if_seed_changed" not in metadata.extra_context
 
 
+def _model_with_node_retries(retries):
+    return DbtNode(
+        unique_id=f"{DbtResourceType.MODEL.value}.proj.my_model",
+        resource_type=DbtResourceType.MODEL,
+        depends_on=[],
+        path_base=Path("."),
+        original_file_path=Path("models/my_model.sql"),
+        tags=[],
+        config={"materialized": "table", "meta": {"cosmos": {"operator_kwargs": {"retries": retries}}}},
+        has_test=True,
+    )
+
+
+def test_retries_after_all_separates_model_and_test():
+    """Per-node ``retries`` reaches the model task, while the AFTER_ALL test task (node=None) keeps the DAG default.
+
+    This is the supported way to retry model runs but not tests (issue #2130).
+    """
+    node = _model_with_node_retries(2)
+    model_meta = create_task_metadata(
+        node, execution_mode=ExecutionMode.LOCAL, args={"retries": 0}, dbt_dag_task_group_identifier="dag"
+    )
+    test_meta = create_test_task_metadata(
+        "proj_test",
+        ExecutionMode.LOCAL,
+        TestIndirectSelection.EAGER,
+        task_args={"retries": 0},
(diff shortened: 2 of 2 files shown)


pankajkoti (commented):
@pankajastro were you able to test this out with a dbt project and observe the DAG render/task runs accordingly?

pankajastro:
> @pankajastro were you able to test this out with a dbt project and observe the DAG render/task runs accordingly?

Yes, I have tested it. Here's a small demo.

https://github.com/user-attachments/assets/d8facba5-b93a-4415-aeae-d00b8a17c25d

pankajkoti:
> > @pankajastro were you able to test this out with a dbt project and observe the DAG render/task runs accordingly?
> 
> Yes, I have tested it. Here's a small demo.
> 
>  Screen.Recording.2026-06-30.at.4.27.02.PM.mov

@pankajastro I saw the 3 DAG runs in the demo, but none seem to show the behaviour this PR claims, i.e. retry models but not tests. Only one DAG run(the last one) showed both the model and tests failed, but both seem to have had 3 tries. Am I missing something here?

pankajastro:
> @pankajastro I saw the 3 DAG runs in the demo, but none seem to show the behaviour this PR claims, i.e. retry models but not tests. Only one DAG run(the last one) showed both the model and tests failed, but both seem to have had 3 tries. Am I missing something here?

Ok, Sir, here we go.


https://github.com/user-attachments/assets/55127c3f-63e9-427a-a451-42e009b01b0e

Currently
For TestBehavior.AFTER_EACH we can't set test retry 0 and model retry >0 
For TestBehavior.AFTER_ALL we can set test retry 0 and model retry >0

docs/guides/translate_dbt_to_airflow/testing-behavior.rst
```diff
@@ -49,6 +49,52 @@ Finally, an example DAG and how it is rendered in the `Apache Airflow® <https:/
 
 .. image:: ../../_static/test_behavior_build.png
 
+Retrying models but not tests
```
pankajkoti: Can we explicitly mention here in the title that this applies only to AFTER_ALL?
```suggestion
Retrying models but not tests when using TestBehavior.AFTER_ALL
```

pankajkoti (approved):
Can we please alter the PR title that clarifies it only applies to TestBehavior.AFTER_ALL.

At the moment, looking at the PR title which would go into commit history it would appear as a generic feature applicable which is not the case. 

Would be nice to have another pair of eyes here to have a wider consensus on the public facing docs.

cc: @tatiana if you could also review please

docs/guides/translate_dbt_to_airflow/testing-behavior.rst
pankajastro: Applied suggestion

pankajastro:
Thanks for the review @pankajkoti! Addressed both points in `7d0906ba`:

- **Section heading** — accepted your suggestion, now reads *"Retrying models but not tests when using TestBehavior.AFTER_ALL"* (bumped the RST underline to match so the docs build stays warning-free).
- **PR title** — updated to *"Document how to retry models but not tests with TestBehavior.AFTER_ALL"* so it doesn't read as a generic feature in the commit history.

@tatiana, tagging you per @pankajkoti request — would appreciate a second pair of eyes on the public-facing docs change in `docs/guides/translate_dbt_to_airflow/testing-behavior.rst` before we merge.

docs/guides/translate_dbt_to_airflow/testing-behavior.rst
```diff
@@ -49,6 +49,52 @@ Finally, an example DAG and how it is rendered in the `Apache Airflow® <https:/
...
+              retries: 2
+
+.. note::
+
+    This does not work with the default ``TestBehavior.AFTER_EACH``. With ``after_each``, the test task is derived from
+    its model node and inherits the model's task arguments. As a result:
+
+    - Setting ``retries`` on a model (via ``meta.cosmos.operator_kwargs``) also retries its after-each test.
+    - Setting ``operator_kwargs`` on the test itself (via the ``tests``/``data_tests`` config in ``dbt_project.yml`` or
+      the test's ``meta`` in ``schema.yml``) is ignored — the after-each test always uses the model's arguments.
```
tatiana: Even though Cosmos doesn't have a declarative way of doing this right now, users could accomplish the desired behaviour by doing something like this:

```
dag = DbtDag(..., operator_args={"retries": 2})
for task in dag.tasks:
    if type(task).__name__.startswith("DbtTest"):
        task.retries = 0
```

Another possible way to accomplish this would be to use `node_converters`, but this would be overkill.

tatiana (commented):
@pankajastro please, take a look in the inline comment - but, before of that, I think I have a solution for the `AFTER_EACH`: https://github.com/astronomer/astronomer-cosmos/pull/2936

tatiana:
We may want to update this PR if we go ahead and merge https://github.com/astronomer/astronomer-cosmos/pull/2936

docs/guides/translate_dbt_to_airflow/testing-behavior.rst
pankajastro: This works, but it feels a bit hacky to document. I'd prefer to avoid adding it to the docs.

PR #2841, opened 2026-06-26
refactor: avoid hard-coded status strings in watcher

### LINKED ISSUE #2585: Improve watcher implementation to avoid hard-coded constants
We have a mixed definition/utilisation approach:
-  "skipped"
-  `WatcherEventReason.PRODUCER_SKIPPED`

The same applies to other statuses.

We should normalise and avoid hard-coded strings in the code.

Feedback from:
https://github.com/astronomer/astronomer-cosmos/pull/2559#discussion_r3117026015

pankajastro:
## Description

Addresses #2585. The watcher used a mixed approach for the same status values — sometimes a typed constant/enum, sometimes a bare string literal (e.g. the literal `"skipped"`/`"failed"` vs `WatcherEventReason.PRODUCER_SKIPPED`). This normalises the status vocabularies and removes the remaining hard-coded literals.

### Changes

**`_watcher/state.py`** (single source of truth):
- Add `ProducerTaskState(str, Enum)` for Airflow producer task states, which previously had **no** enum at all — only membership frozensets. As a `str` subclass its members compare equal to the raw state strings, so `==` and set membership are unchanged.
- Rebuild `PRODUCER_TERMINAL_STATES` / `PRODUCER_FINAL_STATES` from the enum instead of duplicating literals.
- Add `DbtNodeStatus(str, Enum)` holding every raw dbt node status value (`success`/`pass`/`warn`/`failed`/`fail`/`error`/`runtime error`/`skipped`), and rebuild the `DBT_SUCCESS_STATUSES` / `DBT_FAILED_STATUSES` / `DBT_SKIPPED_STATUSES` frozensets from its members. A `__str__` override returns the raw value so logs/XCom render `skipped`/`failed` rather than `DbtNodeStatus.SKIPPED`.

**`_watcher/triggerer.py`**: replace `producer_task_state == "failed"/"skipped"/"success"` with `ProducerTaskState.*`.

**`_watcher/base.py`**: replace `dbt_node_status == "skipped"` → `is_dbt_node_status_skipped(...)`; the upstream-failure rewrite `return "failed"` → `DbtNodeStatus.FAILED`; `status == "success"` / `status != "failed"` → `EventStatus.SUCCESS`/`EventStatus.FAILED`; two `producer_task_state` literals → `ProducerTaskState.*`.

**`operators/watcher.py`**: the freshness callback's `(uid, "skipped")` → `(uid, DbtNodeStatus.SKIPPED)`.

This is **behaviour-neutral** — enum members are `str` subclasses carrying the identical string values, and the `__str__` override keeps their rendered/serialised form unchanged.

## Testing

- `ruff`, `black`, `mypy` pass (full pre-commit suite green on the commit).
- Watcher unit tests pass: `tests/operators/_watcher/` (163 passed, 1 skipped).

Closes #2585

### DIFF
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -23,6 +23,7 @@ from cosmos.listeners.dag_run_listener import EventStatus
 from cosmos.log import get_logger
 from cosmos.operators._watcher.aggregation import get_tests_status_xcom_key, push_test_result_or_aggregate
 from cosmos.operators._watcher.state import (
+    DbtNodeStatus,
     ProducerTaskState,
     _iso_to_string,
     _log_dbt_event,
@@ -282,7 +283,7 @@ def _rewrite_upstream_failure_skip_status(
     if is_dbt_upstream_failure_skip_event(log_line.get("info", {}).get("name")):
         upstream_failure_skipped_ids.add(unique_id)
     if is_dbt_node_status_skipped(dbt_node_status) and unique_id in upstream_failure_skipped_ids:
-        return "failed"
+        return DbtNodeStatus.FAILED
     return dbt_node_status
 
 
--- a/cosmos/operators/_watcher/base.py
+++ b/cosmos/operators/_watcher/base.py
@@ -23,6 +23,7 @@ from cosmos.listeners.dag_run_listener import EventStatus
 from cosmos.log import get_logger
 from cosmos.operators._watcher.aggregation import get_tests_status_xcom_key, push_test_result_or_aggregate
 from cosmos.operators._watcher.state import (
+    ProducerTaskState,
     _iso_to_string,
     _log_dbt_event,
     build_producer_state_fetcher,
@@ -280,7 +281,7 @@ def _rewrite_upstream_failure_skip_status(
         return dbt_node_status
     if is_dbt_upstream_failure_skip_event(log_line.get("info", {}).get("name")):
         upstream_failure_skipped_ids.add(unique_id)
-    if dbt_node_status == "skipped" and unique_id in upstream_failure_skipped_ids:
+    if is_dbt_node_status_skipped(dbt_node_status) and unique_id in upstream_failure_skipped_ids:
         return "failed"
     return dbt_node_status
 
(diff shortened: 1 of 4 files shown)


cosmos/operators/_watcher/base.py
```diff
@@ -277,7 +278,7 @@ def _rewrite_upstream_failure_skip_status(
         return dbt_node_status
     if is_dbt_upstream_failure_skip_event(log_line.get("info", {}).get("name")):
         upstream_failure_skipped_ids.add(unique_id)
-    if dbt_node_status == "skipped" and unique_id in upstream_failure_skipped_ids:
+    if is_dbt_node_status_skipped(dbt_node_status) and unique_id in upstream_failure_skipped_ids:
         return "failed"
```
tatiana: Should we also replace this hard-coded "failed" string? If `DBT_NODE_STATUS_FAILED` doesn't exist, maybe we should introduce and use it.

tatiana (approved):
Thanks for refactoring and improving this, @pankajastro !
Please, could you rebase and resolve any conflicts?

Please, could you also address the inline feedback: https://github.com/astronomer/astronomer-cosmos/pull/2841/changes#r3492796945

cosmos/operators/_watcher/base.py
pankajastro: Good call. Rather than add a lone `DBT_NODE_STATUS_FAILED` constant, I introduced a single `DbtNodeStatus(str, Enum)` holding every raw dbt node status value (`success`/`pass`/`warn`/`failed`/`fail`/`error`/`runtime error`/`skipped`), mirroring the existing `ProducerTaskState`. The `DBT_SUCCESS_STATUSES` / `DBT_FAILED_STATUSES` / `DBT_SKIPPED_STATUSES` sets are now built from its members, and the `"failed"` literal here (plus the previous `DBT_NODE_STATUS_SKIPPED`) is replaced with `DbtNodeStatus.FAILED` / `DbtNodeStatus.SKIPPED`.

A `__str__` override returns the raw value, so logs and XCom still render `skipped`/`failed` (not `DbtNodeStatus.SKIPPED`) — no behavioural change. Done in 9753284e.

cosmos/operators/watcher.py
```diff
@@ -123,7 +127,7 @@ def _default_freshness_callback(
     # and test hash-suffixed unique_ids are not valid dbt --exclude selectors.
     excludable = [uid for uid in visited if nodes.get(uid) and nodes[uid].resource_type in _excludable_resource_types]
     logger.info("Nodes to skip due to stale sources: %s", excludable)
-    return [(uid, "skipped") for uid in excludable]
+    return [(uid, DbtNodeStatus.SKIPPED) for uid in excludable]
```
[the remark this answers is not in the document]
pankajastro: Thanks — updated the PR description to match the implementation. It now documents the `DbtNodeStatus(str, Enum)` approach (with the `DBT_*_STATUSES` frozensets built from its members and a `__str__` override) rather than the earlier `DBT_NODE_STATUS_SKIPPED` constant, so the description and code are consistent.

PR #2844, opened 2026-06-26
Add dedicated producer, consumer, and retry queues for `ExecutionMode.WATCHER`

### LINKED ISSUE #2729: [Feature] Support Additional Queue Configuration for Retries when using `ExecutionMode.WATCHER`
### Description

Per the Cosmos docs, both the initial producer task and all retries leverage the `watcher_dbt_execution_queue` when executing. See the snippet from the docs below.

> - For watcher producer tasks (``DbtProducerWatcherOperator``), the configured queue is used during their first execution
> - For watcher consumer tasks (``DbtConsumerWatcherSensor``), from their first retry onwards, if ``watcher_dbt_execution_queue`` is configured, the task is automatically assigned to the specified queue
> - This behavior is enforced by Cosmos via an `Airflow cluster policy <https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/cluster-policies.html>`_ (``task_instance_mutation_hook``) that mutates ``task_instance.queue`` at runtime for retry attempts

Long-term, I think it would make sense to have three environment variables, which control the queue that Tasks are run on.

- `AIRFLOW__COSMOS__WATCHER_DBT_PRODUCER_QUEUE`
- `AIRFLOW__COSMOS__WATCHER_DBT_WATCHER_QUEUE`
- `AIRFLOW__COSMOS__WATCHER_DBT_RETRY_QUEUE`

### Use case/motivation

Running a large number of producer tasks heavily uses the "execution" queue. When there is a failure, the retries go to this same queue, which causes additional congestion. Ideally, I'd like to have these two processes running in separate queues to isolate my initial producer tasks from retries.

### Related issues

_No response_

jroachgolf84:
## Description

This PR adds support for three queues for users leveraging `ExecutionMode.WATCHER`:

- `AIRFLOW__COSMOS__WATCHER_DBT_PRODUCER_QUEUE`
- `AIRFLOW__COSMOS__WATCHER_DBT_WATCHER_QUEUE`
- `AIRFLOW__COSMOS__WATCHER_DBT_RETRY_QUEUE`

Previously, a user could only specify a `AIRFLOW__COSMOS__WATCHER_DBT_EXECUTION_QUEUE`, which was used for both Producer and Retry tasks. Now, users can specify an explicit queue for Producer, Watcher, and Retry tasks.

## Related Issue(s)

closes: https://github.com/astronomer/astronomer-cosmos/issues/2729

## Breaking Change?

No, backward compatibility is ensured.

## Checklist


jroachgolf84:
Still need to review docs, will do that shortly.

cosmos/operators/watcher.py
```diff
@@ -48,7 +48,7 @@
     DbtRunLocalOperator,
     DbtSourceLocalOperator,
 )
-from cosmos.settings import watcher_dbt_execution_queue
+from cosmos.settings import watcher_dbt_producer_queue, watcher_dbt_watcher_queue
```
pankajkoti: Can we think of and call something better, please, for the second queue, because there's 2 watcher in the second queue's name.

cosmos/settings.py
```diff
@@ -62,12 +62,13 @@
 enable_setup_async_task = conf.getboolean("cosmos", "enable_setup_async_task", fallback=True)
 enable_teardown_async_task = conf.getboolean("cosmos", "enable_teardown_async_task", fallback=True)
 
-# DBT Watcher Execution Mode Watcher Task Retry Queue
-# in watcher mode, if the producer watcher fails, the consumer tasks run the individual models on retry.
-# since these tasks are sensors that require low memory/cpu on their first try,
-# this setting allows retries to run on a queue with larger resources, which is often necessary for larger dbt projects
-# this would also be used to run the producer task
-watcher_dbt_execution_queue = conf.get("cosmos", "watcher_dbt_execution_queue", fallback=None)
```
pankajkoti, line 77: Can we make this backwards compatible, please, instead of removing it altogether? Users who already understand what `watcher_dbt_execution_queue` is meant for should be able to continue using it for the producer and retries.

cosmos/operators/watcher.py
```diff
@@ -227,7 +227,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         kwargs["should_store_compiled_sql"] = False
         kwargs.setdefault("priority_weight", PRODUCER_WATCHER_DEFAULT_PRIORITY_WEIGHT)
         kwargs.setdefault("weight_rule", WATCHER_TASK_WEIGHT_RULE)
-        kwargs["queue"] = watcher_dbt_execution_queue or kwargs.get("queue") or DEFAULT_QUEUE
+        kwargs["queue"] = watcher_dbt_producer_queue or kwargs.get("queue") or DEFAULT_QUEUE
```
pankajkoti, line 234: same suggestion as https://github.com/astronomer/astronomer-cosmos/pull/2844/changes#r3489430942 to ensure we have backwards compatability.

cosmos/plugin/cluster_policy.py
```diff
@@ -42,11 +42,11 @@ def task_instance_mutation_hook(task_instance: TaskInstance) -> None:
     else:
         retry_number = 2
 
-    if watcher_dbt_execution_queue and task_instance.try_number and _is_watcher_sensor(task_instance):
+    if watcher_dbt_retry_queue and task_instance.try_number and _is_watcher_sensor(task_instance):
```
pankajkoti, line 45: same suggestion as https://github.com/astronomer/astronomer-cosmos/pull/2844/changes#r3489430942 to ensure we have backwards compatibility here.

docs/guides/run_dbt/airflow-worker/watcher-execution-mode.rst
pankajkoti, line 1: let's also ensure here that once the backwards compatibility is addressed we mention the same here.

docs/reference/configs/cosmos-conf.rst
```diff
@@ -388,19 +388,29 @@ This page lists all available `Apache Airflow® <https://airflow.apache.org/>`_
     - Default: ``False``
     - Environment Variable: ``AIRFLOW__COSMOS__ENABLE_ORJSON_PARSER``
 
-.. _watcher_dbt_execution_queue:
+.. _watcher_dbt_producer_queue:
 
-`watcher_dbt_execution_queue`_:
```
pankajkoti, line 409: Same suggestion about backwards compatibility instead of removing this. We can mark this as deprecated and mention that it will be removed in 2.0.0

pankajkoti (commented):
Thanks for working on this @jroachgolf84 . I have some suggestions in-line.

Also, could we please standardise the PR title to match the phrasing with the other project PRs, notably, we don't need the `feature/issue-2729` prefix and the title could begin with an imperative verb "Add ...." instead of "Adding ...". The recommendation for writing a good commit message title can be applied to PR titles as well https://cbea.ms/git-commit/

Also CI seems to be failing on some tests. Could you please take a look at that too? thanks.

cosmos/operators/watcher.py
tatiana: This may be a good alternative: `watcher_dbt_consumer_queue`

jroachgolf84:
@pankajkoti - I will definitely check those out. Seems like tests are only failing on 3.3 due to an issue installing (not my actual tests).

pankajkoti:
> @pankajkoti - I will definitely check those out. Seems like tests are only failing on 3.3 due to an issue installing (not my actual tests).

Thanks @jroachgolf84 . Rebasing the PR might help I believe

cosmos/operators/watcher.py
jroachgolf84: That sounds good to me - I can make that change.

jroachgolf84:
Apologies for the delay, starting to look at this again.

cosmos/operators/watcher.py
jroachgolf84: Implemented in next commit.

cosmos/settings.py
jroachgolf84: Yes, will add this in the next commit.

cosmos/operators/watcher.py
jroachgolf84: Yes, addressed in next commit.

cosmos/plugin/cluster_policy.py
jroachgolf84: Handled in `cosmos/settings.py`, which is in the next commit.

docs/reference/configs/cosmos-conf.rst
jroachgolf84: Addressed in next commit.

jroachgolf84:
Feedback implemented! @pankajkoti, can you take another look?

jroachgolf84:
@pankajkoti - can you take another look at this? Thanks!

pankajkoti (approved):
LGTM. Thanks for addressing the feedback, @jroachgolf84, and for ensuring we have backward compatibility here!

PR #2826, opened 2026-06-21
Surface dbt stdout/stderr in CosmosLoadDbtException

DaPillah:
## Description

Include exit code, stderr, and stdout in all CosmosLoadDbtException messages raised by run_command_with_subprocess. Also tighten the error-detection condition from truthy-returncode to returncode != 0, so that a None exit code (undefined result) is correctly treated as a failure rather than silently swallowed.

## Related Issue(s)

Closes #2822

## Breaking Change?

No.

## Checklist

- [ ] I have made corresponding changes to the documentation (if required)
- [x] I have added tests that prove my fix is effective or that my feature works

🤖 Generated with Claude Code (https://claude.com/claude-code)

PR #2848, opened 2026-06-29
ci: Restore being able to run integration tests on external committers PRs

tatiana:
## Description

The recently adopted [`actions/checkout` v7.0.0](https://github.com/astronomer/astronomer-cosmos/pull/2833) refuses to check out fork PR code in a `pull_request_target` workflow unless `allow-unsafe-pr-checkout: true` is set on the step (it guards against "pwn request" attacks). This broke `test.yml` for fork PRs: the `Check-changed-files` job failed outright, and the secret-dependent jobs (integration, kubernetes, coverage) were skipped behind it. Surfaced by the CI failure on #2826.

The fix follows the repo's existing safe-`pull_request_target` model — the `Authorize` owner-review environment gate:
Validated locally: full pre-commit suite passes (incl. `check-yaml`); `zizmor` reports no findings.

> Note: `pull_request_target` loads the workflow from the base branch, so this only takes effect once merged to `main` — re-running an open fork PR's CI beforehand still uses the old workflow.

## Related Issue(s)

Surfaced by the CI failure on #2826 (does not close it).

## Breaking Change?

No — CI configuration only.

## Checklist

- [x] I have made corresponding changes to the documentation (if required) — N/A (CI config only)
- [x] I have added tests that prove my fix is effective or that my feature works — N/A; validated via `zizmor` + pre-commit, and verifiable once a fork PR's CI runs against the merged workflow

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,10 +38,15 @@ jobs:
     outputs:
       code_changed: ${{ steps.check.outputs.code_changed }}
     steps:
+      # Only the `push` branch below uses the working tree (`git diff`). For pull_request /
+      # pull_request_target the changed-file list comes from the GitHub API, so we must NOT
+      # check out untrusted fork code in this ungated job — that is the pull_request_target
+      # "pwn request" risk actions/checkout v7 refuses by default.
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        if: github.event_name == 'push'
         with:
           fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          ref: ${{ github.sha }}
           persist-credentials: false
 
       - id: check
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -38,10 +38,15 @@ jobs:
     outputs:
       code_changed: ${{ steps.check.outputs.code_changed }}
     steps:
+      # Only the `push` branch below uses the working tree (`git diff`). For pull_request /
+      # pull_request_target the changed-file list comes from the GitHub API, so we must NOT
+      # check out untrusted fork code in this ungated job — that is the pull_request_target
+      # "pwn request" risk actions/checkout v7 refuses by default.
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        if: github.event_name == 'push'
```
pankajkoti, line 46: The only point that I was wondering about in this PR was whether we should simply add `authorize` gate for this job too. 

But after chatting with Claude, it's helping me understand that this approach is safer too and that we don't rely on local checkout code in forked PRs anyway with this job. So, I guess we're good.

pankajkoti (approved):
Thanks a lot @tatiana for observing the issue that affects our setup with what as added as a feature in https://github.com/actions/checkout/pull/2454 and fixing it here for our setup.

LGTM.

PR #2860, opened 2026-06-30
Add dbt-core 2.0 (alpha) to the integration test matrix

tatiana:
This started as a simple "add dbt-core 2.0 to the test matrix" and turned into a useful probe of what dbt Core 2.0 actually *is* — so I'm keeping it as a draft to capture the finding and agree on next steps before we commit to anything.

## What

**dbt Core 2.0** is dbt Labs' ground-up **Rust rewrite** of the dbt engine — Apache-2.0 licensed, currently in alpha (`2.0.0a3`). It's built on the **same foundations as the dbt Fusion engine**: dbt Labs open-sourced much of the Fusion code and shipped it as dbt Core v2 (see their [announcement](https://docs.getdbt.com/blog/dbt-core-v2-is-here)). The two are close cousins:

- **dbt Core 2.0** — the open-source (Apache-2.0) Rust core. CLI-first.
- **dbt Fusion** — the ELv2-licensed **superset binary** on the same Rust foundation, with extra capabilities ("can do more out of the box"). So Fusion ⊋ dbt Core 2.0.

The catch we ran into: the 2.0 alpha is **CLI-only** — it ships **no importable Python `dbt` package** at all (`import dbt` → `ModuleNotFoundError`; only `importlib.metadata.version("dbt-core")` resolves). That breaks the way we talk to dbt from Python today — `from dbt.version import __version__` and `InvocationMode.DBT_RUNNER` (which needs `dbt.cli.main.dbtRunner`) can't work against a binary that has no Python surface. dbt Labs' [v2 roadmap](https://github.com/dbt-labs/dbt-core/blob/main/docs/roadmap/2026-06-announcing-v2.md) lists "a programmatic (Python) interface for invoking dbt Core v2.0" as a parity gap they intend to close before GA, so this should ease over time.

This PR wires dbt Core 2.0 into the integration matrix to see how far we get, while leaving the existing dbt-Fusion job in place — that one drives the Fusion *binary* via subprocess, touches no Python API, and stays green. The contrast is the whole story: the Python-integrated cells fail 72/72, the subprocess Fusion cells pass 4/4 (full detail in the [findings comment](https://github.com/astronomer/astronomer-cosmos/pull/2860#issuecomment-4842205861), CI run [28435105169](https://github.com/astronomer/astronomer-cosmos/actions/runs/28435105169)).

## What we're doing in this PR

- Switch dbt-version detection to `importlib.metadata.version("dbt-core")` (`operators/local.py`) so it no longer imports the removed `dbt.version` module — works on dbt 1.x and 2.x.

## Next steps

- [ ] Revert the temporary CI-trigger commit (`7652a4a8`) before this is marked ready / merged — it only exists so the new 2.0 cells run on the branch (`test.yml` is `pull_request_target`).
- [ ] Drop the Postgres `2.0` cells and instead add a **duckdb-based** 2.0 cell: dbt 2.0's Fusion engine doesn't support the Postgres adapter our suite uses ([`dbt1005`](https://github.com/astronomer/astronomer-cosmos/pull/2860#issuecomment-4843703981)), but duckdb is in Fusion's supported set, runs locally, and we already ship `dbt-duckdb`. That's the real way to exercise dbt-core 2.0.

## Pending — on us (cosmos)

- [ ] Raise a clear, actionable error when `InvocationMode.DBT_RUNNER` is explicitly requested but the Python dbt API is unavailable (dbt-core 2.0 / Fusion), pointing users at `InvocationMode.SUBPROCESS` — instead of today's opaque `ImportError` from `dbt.cli.main`. (Auto-detect already falls back to subprocess; this covers the *explicit* case.)
- [ ] Gate dbt-core 2.0 to `ExecutionMode.LOCAL` + `InvocationMode.SUBPROCESS` in `config.py` validation, with a clear message for unsupported combos.
- [ ] Audit the remaining `dbt.*` import sites for a no-Python-dbt environment — `operators/virtualenv.py`, `dbt/parser/output.py`, and `operators/_asynchronous/bigquery.py` (the async path) — and guard any that aren't already (`dbt/runner.py` and `local.py:80` are already guarded via `is_available()` / `TYPE_CHECKING`).
- [ ] Decide whether the `dbt-core<2.0` → `<3.0` cap raises and the `require-dbt-version` fixture bumps land here or in a separate PR.

## Depends on dbt Labs (not us)

- A programmatic Python interface for dbt Core 2.0 is on their GA roadmap. If/when it lands, `DBT_RUNNER` mode could work again and we'd revisit the gating above.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,10 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    # TEMP — REVERT BEFORE MERGE. Also run on this bump branch so the new dbt-core 2.0
+    # matrix cells are exercised pre-merge: test.yml is pull_request_target, so PR runs read
+    # the matrix from `main`; a push event runs the branch's own workflow (and 2.0 matrix).
+    branches: [main, bump/dbt-core-2.0.0a3]
   # Also run on pull requests originating from forks. Every test job here is secret-dependent
   # (integration tests, kubernetes, code coverage) and depends on the Authorize job, which requires
   # manually approving the workflow run for external PRs. Jobs that need no secrets (type-check,
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -133,7 +133,10 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13"]
         airflow-version: ["2.9", "2.10", "2.11", "3.0", "3.1", "3.2", "3.3"]
-        dbt-version: [ "1.11" ]
+        # "2.0" = dbt Core 2.0, the open-source build of the dbt Fusion engine
+        # (https://docs.getdbt.com/blog/dbt-core-v2-is-here). Installed with pre-releases
+        # allowed via integration-setup.sh (alpha-only today).
+        dbt-version: [ "1.11", "2.0" ]
         split-group: [1, 2, 3]
         exclude:
           # Apache Airflow versions prior to 3.1.0 have not been tested with Python 3.13.
--- a/cosmos/dbt/executable.py
+++ b/cosmos/dbt/executable.py
@@ -1,5 +1,11 @@
+from __future__ import annotations
+
 import shutil
+from functools import cache
+from importlib.metadata import PackageNotFoundError, version
 from importlib.util import find_spec
+
+from packaging.version import Version
 
 
 def get_system_dbt() -> str:
@@ -7,6 +13,22 @@ def get_system_dbt() -> str:
     Tries to identify which is the path to the dbt executable, return "dbt" otherwise.
     """
     return shutil.which("dbt") or "dbt"
+
+
+@cache
+def get_dbt_version() -> Version | None:
+    """
+    Return the installed dbt-core version, or None if dbt-core is not installed.
+
+    Reads the version from the installed package metadata rather than the ``dbt.version``
+    module: dbt-core 2.0 is CLI-only and no longer ships an importable ``dbt`` package, so
+    ``from dbt.version import __version__`` raises there. Package metadata works for both
+    dbt 1.x and 2.x. Cached for the process lifetime since the installed version is fixed.
+    """
+    try:
+        return Version(version("dbt-core"))
+    except PackageNotFoundError:
+        return None
 
 
 def is_dbt_installed_in_same_environment() -> bool:
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -551,7 +551,11 @@ class AbstractDbtLocalBase(AbstractDbtBase):
             self.profile_config.target_name,
         ]
         if self.invocation_mode == InvocationMode.DBT_RUNNER:
-            from dbt.version import __version__ as dbt_version
+            from importlib.metadata import version as _installed_version
+
+            # dbt-core 2.0 removed the importable `dbt.version` module (it is CLI-only); read the
+            # installed version from package metadata instead, which works on both dbt 1.x and 2.x.
+            dbt_version = _installed_version("dbt-core")
 
             if Version(dbt_version) >= Version("1.5.6"):
                 # PR #1484 introduced the use of dbtRunner during DAG parsing. As a result, invoking dbtRunner again
--- a/cosmos/operators/local.py
+++ b/cosmos/operators/local.py
@@ -551,13 +551,10 @@ class AbstractDbtLocalBase(AbstractDbtBase):
             self.profile_config.target_name,
         ]
         if self.invocation_mode == InvocationMode.DBT_RUNNER:
-            from importlib.metadata import version as _installed_version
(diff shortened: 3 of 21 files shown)


tatiana:
## Findings — dbt-core 2.0 has no Python API (CI run [28435105169](https://github.com/astronomer/astronomer-cosmos/actions/runs/28435105169))

Exercised the new matrix cells pre-merge via a temporary `push:` trigger. Result:

| Cells | Result |
|---|---|
| **New `Run-Integration-Tests (…, 2.0, …)`** (regular, Python-integrated path) | **72 / 72 FAILED** |
| **`Run-Integration-dbt-fusion-Tests (…, 2.0)`** (pre-existing, subprocess path) | **4 / 4 passed** |

### Root cause

dbt-core 2.0 ships **no importable Python `dbt` package** — it's a CLI-only Rust binary. Verified against `dbt-core==2.0.0a3`:

```text
import dbt            -> ModuleNotFoundError: No module named 'dbt'
dbt.version          -> ModuleNotFoundError          # cosmos/operators/local.py:552
dbt.cli.main         -> ModuleNotFoundError          # InvocationMode.DBT_RUNNER needs dbtRunner
importlib.metadata.version("dbt-core") = 2.0.0a3     # only package metadata is available
```

Consequences for cosmos:
- `from dbt.version import __version__` (`local.py:552`) raises — the first visible failure.
- **`InvocationMode.DBT_RUNNER` is impossible** on dbt-core 2.0 (no Python `dbtRunner`).
- Any code importing `dbt.*` internals breaks.

The **Fusion job passes** precisely because it drives the engine **binary via subprocess** and touches no Python dbt API — the same constraint, handled the right way.

### Was this knowable from dbt Labs' announcement?

Partly. The [v2 blog](https://docs.getdbt.com/blog/dbt-core-v2-is-here) signaled the direction (Rust rewrite, "removes the need to fight with Python's virtual environments", ADBC/Arrow adapters) but did **not** spell out the missing Python API. The [v2 roadmap](https://github.com/dbt-labs/dbt-core/blob/main/docs/roadmap/2026-06-announcing-v2.md) is explicit, though, listing as a remaining parity gap:

> "closing the last few gaps in parity (such as a programmatic (Python) interface for invoking dbt Core v2.0)"

So the **no-Python-API state is an explicitly-acknowledged alpha gap, expected to be closed before GA** — not a permanent removal. dbt-core 2.0 is the open-source (Apache-2.0) Rust core that shares Fusion's foundations; Fusion is the ELv2 superset binary. Both are CLI-first.

### Implication

The **regular, Python-integrated matrix is the wrong vehicle** for dbt-core 2.0 today. Coverage belongs on a **subprocess-only** path (which the Fusion job already provides). Genuine OSS-core 2.0 support would be a dedicated effort — gate to `ExecutionMode.LOCAL` + `InvocationMode.SUBPROCESS`, guard every `dbt.*` import, and replace `dbt.version` with `importlib.metadata` — and may be partly mooted once the GA Python interface lands.

**Disposition:** this draft did its job (proving the constraint). Proposing to drop the regular-matrix `2.0` cells and revert the temp CI trigger; track real OSS dbt-core 2.0 support separately.

tatiana:
### Update: the 2.0 cells now fail on the adapter, not on us

After wiring the new dbt-core 2.0 cells to actually run (via the temporary push trigger), two import fixes landed and did their job — `operators/local.py` and the test suite no longer import the removed `dbt.version` (we route through `cosmos.dbt.executable.get_dbt_version()` now), so we're past the collection errors.

The cells still fail 72/72, but now they reach the real dbt parse and hit this (CI run [28443831346](https://github.com/astronomer/astronomer-cosmos/actions/runs/28443831346)):

```
[error] [InvalidConfig (dbt1005)]: Configuration Error: The 'postgres' adapter is not yet
supported by dbt Fusion. Supported adapters: snowflake, bigquery, databricks, redshift,
duckdb, salesforce. To use an experimental adapter, set DBT_ALLOW_EXPERIMENTAL_ADAPTERS=true.
```

dbt-core 2.0's Rust/Fusion engine ships its own adapters, and Postgres isn't one of them. Our whole integration suite runs against Postgres, so every example DAG fails to parse (`CosmosLoadDbtException`) and the cells go red. The failure chain was: `dbt.version` import in the operator → fixed; `dbt.version` import in the tests → fixed; dbt 2.0 has no Postgres adapter → not something we can fix.

**Takeaway:** the regular Postgres-based matrix can't validate dbt-core 2.0, full stop. To exercise 2.0 for real we'd need a Fusion-supported adapter, and **duckdb** is the obvious one — it's in the supported list, runs locally with no warehouse, and we already ship `dbt-duckdb` + duckdb examples. That's the right shape for the follow-up, rather than forcing Postgres through with `DBT_ALLOW_EXPERIMENTAL_ADAPTERS=true` (unstable, not a real signal).

Adjusting the plan: the "add a subprocess-only 2.0 cell" item should specifically be a **duckdb-based** cell, and the Postgres `2.0` cells added here should come back out.

Issue #971, opened 2024-05-15
Cosmos 1.4 issue: Unpack failed: incomplete input


tatiana:
Reported in [#airflow-astronomer](https://apache-airflow.slack.com/archives/C03T0AVNA6A) on 15 May 2024 by Jason T at:
https://apache-airflow.slack.com/archives/C03T0AVNA6A/p1715795666218829

Details:

After upgrading to Cosmos v1.4, I've been getting the follow error intermittently on some tasks. Rerunning the task succeeds, but the error continues on subsequent DAG runs.
```
  File "/home/airflow/.local/lib/python3.8/site-packages/cosmos/cache.py", line 134, in _copy_partial_parse_to_project
    data = msgpack.unpack(f)
  File "/home/airflow/.local/lib/python3.8/site-packages/msgpack/__init__.py", line 47, in unpack
    return unpackb(data, **kwargs)
  File "msgpack/_unpacker.pyx", line 205, in msgpack._cmsgpack.unpackb
ValueError: Unpack failed: incomplete input
```

tatiana:
Fix in review: https://github.com/astronomer/astronomer-cosmos/pull/972

PR #2867, opened 2026-07-01
Make partial-parse cache reads resilient to NFS ESTALE races

mungiyo:
## Description

Follow-up to #2815 (atomic `partial_parse.msgpack` cache writes).

After #2815 switched the partial-parse cache writes to `safe_copy` (write to a temp file, then `os.replace`), we started seeing tasks **crash** on a shared network filesystem (NFS/EFS) when the cache is written concurrently:

```
OSError: [Errno 116] Stale file handle:
  '.../cosmos/<dag>/target/partial_parse.msgpack' -> '/tmp/.../target/tmpXXXX'
  File ".../cosmos/cache.py", line 323, in _copy_partial_parse_to_project
  File ".../cosmos/fs.py", line 53, in safe_copy
  File ".../shutil.py", line 273, in copyfile   # _fastcopy_sendfile
```

### Root cause

`safe_copy` publishes via `os.replace` (rename), which unlinks the previous inode. When a **reader** task is mid-`shutil.copyfile` on the shared source (`partial_parse.msgpack` / `manifest.json`) and a **writer** task atomically replaces it, the reader's open handle is invalidated. The POSIX guarantee that "an open reader keeps the old, now-anonymous inode alive" holds on a local filesystem but **not across NFS clients**, so the reader gets `ESTALE` instead of a consistent read — and the task fails.

In other words, #2815 correctly removed the *torn read* (previously surfaced as a caught `ValueError` from `msgpack.unpack` and silently degraded to a full parse), but on NFS it traded that for an **uncaught `ESTALE` that crashes the task**.

### Change

Partial parse is a pure optimisation, so the fix restores the pre-#2815 behaviour: absorb the copy race **at the cache layer and fall back to a full `dbt` parse**, in the same place the torn read was already tolerated (`_copy_partial_parse_to_project`, right next to `patch_partial_parse_content`'s `ValueError` catch).

- **`cosmos/cache.py`** — wrap the partial-parse copy in `try/except OSError` and fall back to a full parse instead of failing the task. The log level is split by cause: transient `ESTALE` (a concurrent task atomically replaced the source mid-copy on a shared FS) logs at `info`; any other `OSError` (e.g. `EACCES` on a misconfigured `cache_dir`, `ENOSPC`) logs at `warning` so a persistent problem that silently disables partial parse stays visible.
- **`cosmos/operators/local.py` and `cosmos/fs.py` are unchanged** — the critical operator execute path is untouched, and `safe_copy` stays as-is.

Tests move the fallback coverage to `tests/test_cache.py` (parametrised: `ESTALE` → `info`, `EACCES` → `warning`), verifying the copy race is swallowed and never raises.

### Notes for reviewers (anticipated questions)

- **Why catch `OSError` broadly?** Partial parse must never fail the task. Narrowing the catch to only `ESTALE` would turn an unrelated, recoverable-by-full-parse error (e.g. a transient cache-dir permission blip) into a hard task failure. Instead we keep the broad catch but raise the log level for non-transient causes so real misconfigurations are not swallowed silently.
- **Why not retry `safe_copy` on `ESTALE`?** An earlier revision of this PR added a bounded retry/backoff to `safe_copy`. It was dropped: the full-parse fallback is the real safety net (it already makes the race non-fatal), while a retry only salvages the *performance* of partial parse in the rare `ESTALE` window — at the cost of adding heuristic retry/backoff/sleep logic to a generic, widely-used utility that other callers share. If the full-parse fallback ever proves too costly under heavy NFS contention, a retry can be re-added later with constants tuned from real traces.

## Related Issue(s)

Follow-up to #2815. Related to the partial-parse race in #971 / #972.

## Breaking Change?

No.

## Checklist

### DIFF
--- a/cosmos/cache.py
+++ b/cosmos/cache.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 import errno
 import functools
 import hashlib
@@ -338,8 +339,12 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
         # and a failed rewrite in ``patch_partial_parse_content`` (which truncates via ``open("wb")`` first)
         # can leave a corrupt msgpack. Remove both copied artifacts so the advertised full-parse fallback is
         # actually what happens instead of handing dbt a half-written or manifest-less cache.
-        target_partial_parse_file.unlink(missing_ok=True)
-        target_manifest_filepath.unlink(missing_ok=True)
+        # The cleanup itself must stay best-effort: ``unlink(missing_ok=True)`` only swallows a missing file,
+        # so suppress any other OSError (e.g. a permission issue or a concurrent process touching the target)
+        # too — otherwise the cleanup error would mask the original one and still fail the task.
+        for artifact in (target_partial_parse_file, target_manifest_filepath):
+            with contextlib.suppress(OSError):
+                artifact.unlink(missing_ok=True)
         # ``errno.ESTALE`` is not defined on every platform; guard the lookup so a missing constant
         # degrades to the warning branch instead of raising ``AttributeError`` while handling the error.
         estale = getattr(errno, "ESTALE", None)
--- a/cosmos/cache.py
+++ b/cosmos/cache.py
@@ -326,27 +326,14 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
         if source_manifest_filepath.exists():
             safe_copy(source_manifest_filepath, target_manifest_filepath)
     except OSError as err:
-        # Partial parse is a best-effort optimisation: if the cache cannot be copied we simply skip it
-        # and let dbt do a full parse, rather than failing the task. This mirrors how the torn-read race
-        # was absorbed here (as a ValueError in ``patch_partial_parse_content``) before the cache writes
-        # became atomic in #2815. The log level is split by cause: ESTALE is an expected, transient race
-        # on a shared network filesystem (a concurrent task atomically replaced the source mid-copy), so
-        # it stays at info; any other OSError (e.g. EACCES on a misconfigured cache_dir, ENOSPC) is likely
-        # a persistent problem that would otherwise silently disable partial parse, so it is surfaced at
-        # warning.
-        # A mid-sequence failure can leave the target inconsistent: a failed manifest copy leaves a lone
-        # partial_parse.msgpack behind (so dbt would still partial-parse despite the "full parse" message),
-        # and a failed rewrite in ``patch_partial_parse_content`` (which truncates via ``open("wb")`` first)
-        # can leave a corrupt msgpack. Remove both copied artifacts so the advertised full-parse fallback is
-        # actually what happens instead of handing dbt a half-written or manifest-less cache.
-        # The cleanup itself must stay best-effort: ``unlink(missing_ok=True)`` only swallows a missing file,
-        # so suppress any other OSError (e.g. a permission issue or a concurrent process touching the target)
-        # too — otherwise the cleanup error would mask the original one and still fail the task.
-        for artifact in (target_partial_parse_file, target_manifest_filepath):
-            with contextlib.suppress(OSError):
-                artifact.unlink(missing_ok=True)
-        # ``errno.ESTALE`` is not defined on every platform; guard the lookup so a missing constant
-        # degrades to the warning branch instead of raising ``AttributeError`` while handling the error.
+        # Best-effort: never fail the task over the partial parse cache — fall back to a full dbt parse.
+        # ESTALE means a concurrent task replaced the cache mid-copy on a shared filesystem (NFS/EFS,
+        # see #2867): transient, so info; other OSErrors likely persist, so warning.
+        # Remove the possibly half-written msgpack, but leave the manifest — it may be the user's,
+        # copied earlier by ``_clone_project``. Suppress cleanup errors so they cannot mask ``err``.
+        with contextlib.suppress(OSError):
+            target_partial_parse_file.unlink(missing_ok=True)
+        # ``errno.ESTALE`` is not defined on every platform.
         estale = getattr(errno, "ESTALE", None)
         if estale is not None and getattr(err, "errno", None) == estale:
             logger.info("Partial parse cache was replaced concurrently (%r); falling back to a full dbt parse", err)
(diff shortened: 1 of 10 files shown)


tests/test_fs.py
```diff
@@ -58,6 +59,65 @@ def test_safe_copy_cleans_up_temp_file_on_failure(tmp_path: Path) -> None:
...
 
 
+def test_safe_copy_retries_on_stale_file_handle(tmp_path: Path) -> None:
+    # A concurrent writer replacing ``src`` on a network filesystem surfaces as ESTALE; safe_copy
+    # must retry (re-opening the new inode) rather than propagating the error.
+    src = tmp_path / "src.txt"
+    src.write_text("hello")
+    dst = tmp_path / "dst.txt"
+
+    real_copyfile = __import__("shutil").copyfile
```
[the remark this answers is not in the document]
mungiyo: Done in c4b657eb — dropped the `__import__("shutil")` in favour of a normal `import shutil`.

One nuance worth flagging: since `cosmos.fs` does `import shutil`, `cosmos.fs.shutil` is the *same* module object as `shutil` here, so `patch("cosmos.fs.shutil.copyfile")` also replaces `shutil.copyfile`. Calling `shutil.copyfile(...)` inside the fake would therefore recurse into the mock. I kept a capture of the real implementation (`real_copyfile = shutil.copyfile`) taken *before* the patch context so the retry can actually perform the copy, and added a comment explaining why.

cosmos/cache.py
```diff
@@ -315,12 +316,37 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
...
+        # it stays at info; any other OSError (e.g. EACCES on a misconfigured cache_dir, ENOSPC) is likely
+        # a persistent problem that would otherwise silently disable partial parse, so it is surfaced at
+        # warning.
+        # A mid-sequence failure can leave the target inconsistent: a failed manifest copy leaves a lone
+        # partial_parse.msgpack behind (so dbt would still partial-parse despite the "full parse" message),
+        # and a failed rewrite in ``patch_partial_parse_content`` (which truncates via ``open("wb")`` first)
+        # can leave a corrupt msgpack. Remove both copied artifacts so the advertised full-parse fallback is
+        # actually what happens instead of handing dbt a half-written or manifest-less cache.
+        target_partial_parse_file.unlink(missing_ok=True)
+        target_manifest_filepath.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
mungiyo: Good catch — fixed in e5e48803. `unlink(missing_ok=True)` only swallows a missing file, so a permission error (or an ESTALE on the unlink itself on the same shared FS) would have propagated out of the `except` block, masked the original `OSError`, and still failed the task — especially since `_handle_partial_parse` no longer wraps this call. The two cleanup unlinks are now wrapped in `contextlib.suppress(OSError)` so the fallback can never crash the task. Added `test__copy_partial_parse_to_project_cleanup_is_best_effort` (patches `Path.unlink` to raise) to lock this in.

cosmos/cache.py
```diff
@@ -315,12 +317,41 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
...
+        # warning.
+        # A mid-sequence failure can leave the target inconsistent: a failed manifest copy leaves a lone
+        # partial_parse.msgpack behind (so dbt would still partial-parse despite the "full parse" message),
+        # and a failed rewrite in ``patch_partial_parse_content`` (which truncates via ``open("wb")`` first)
+        # can leave a corrupt msgpack. Remove both copied artifacts so the advertised full-parse fallback is
+        # actually what happens instead of handing dbt a half-written or manifest-less cache.
+        # The cleanup itself must stay best-effort: ``unlink(missing_ok=True)`` only swallows a missing file,
+        # so suppress any other OSError (e.g. a permission issue or a concurrent process touching the target)
+        # too — otherwise the cleanup error would mask the original one and still fail the task.
+        for artifact in (target_partial_parse_file, target_manifest_filepath):
```
pankajastro: This unlinks `target_manifest_filepath` even when the failure happens before the manifest step is reached. If `manifest_filepath` is configured, `_clone_project` already copied the user's manifest here (`local.py:530`, runs before `_handle_partial_parse` at line 699) — we'd be deleting a file this call never touched. Should probably only remove an artifact if this call actually wrote it.

pankajastro (commented):
Thanks for the PR. I have a few questions and comments otherwise looks good

cosmos/cache.py
```diff
@@ -315,12 +317,41 @@ def _copy_partial_parse_to_project(partial_parse_filepath: Path, project_path: P
...
-        safe_copy(source_manifest_filepath, target_manifest_filepath)
+    try:
+        safe_copy(partial_parse_filepath, target_partial_parse_file)
+
+        patch_partial_parse_content(target_partial_parse_file, project_path)
+
+        if source_manifest_filepath.exists():
+            safe_copy(source_manifest_filepath, target_manifest_filepath)
+    except OSError as err:
+        # Partial parse is a best-effort optimisation: if the cache cannot be copied we simply skip it
```
pankajastro: This comment block is fairly long relative to the code it explains (~16 lines of comments for ~8 lines of logic). Might be worth trimming to the genuinely non-obvious parts — why both artifacts get removed together, and why cleanup suppresses `OSError` — and dropping the changelog-style history ("mirrors pre-#2815 behavior..."), which reads more naturally in the PR description than as a code comment.

cosmos/cache.py
mungiyo: Good catch, thank you — fixed in 1e6edf6e. You're right that this could delete the user's manifest placed there by `_clone_project` (`copy_manifest_file_if_exists` writes to exactly this path) before this call ever ran.

Looking closer, the manifest unlink turned out to be unnecessary altogether: `safe_copy` is atomic (temp file + `os.replace`), so a failed manifest copy leaves the destination untouched — this function can never leave a half-written manifest behind. The only artifact it can leave inconsistent is the msgpack (`patch_partial_parse_content` truncates via `open("wb")`), and removing that alone is enough to make the full-parse fallback deterministic. So the cleanup now only unlinks `partial_parse.msgpack` and never touches the manifest.

Updated the test accordingly: it now pre-seeds a user manifest at the target (simulating `_clone_project`) and asserts it survives the error path untouched.

cosmos/cache.py
mungiyo: Fair point, thanks — trimmed in 1e6edf6e. The block is down from ~16 lines to 5, keeping only the non-obvious parts: the NFS/ESTALE context for the log-level split (with a pointer to this PR), why only the msgpack is removed while the manifest is left alone, and why the cleanup suppresses `OSError`. The changelog-style history now lives only in the PR description.

pankajastro (approved):
LGTM

tatiana (approved):
Thanks for fixing this, @mungiyo ! We'll be shipping in Cosmos 1.15.1

PR #2868, opened 2026-07-01
Reject output-only template fields on the async operator_args path

### LINKED ISSUE #2851: Async operators silently drop output-only template fields (compiled_sql/freshness)
## Context

Follow-up to #2726, which makes local operators **fail fast** when output-only template fields (`compiled_sql`, `freshness`) are passed via `operator_args`. Previously these were silently overwritten at runtime; #2726 raises `CosmosValueError` at instantiation instead.

## Problem

The guard added in #2726 lives in `DbtLocalBaseOperator.__init__` and inspects the raw `kwargs` before they are segregated by argspec.

However, `DbtRunAirflowAsyncBigqueryOperator` (`cosmos/operators/_asynchronous/bigquery.py`) inherits `AbstractDbtLocalBase` **directly** and does not route through `DbtLocalBaseOperator.__init__`, so it bypasses the guard. It also:

- lists `compiled_sql` in its own `template_fields`, and
- resets `self.compiled_sql = ""` at the end of `__init__`.

As a result, passing `compiled_sql` via `dbt_kwargs` to the async BigQuery operator is **still silently dropped** — the exact misuse #2726 set out to eliminate, just on the async execution path.

This is not a regression (the previous warning didn't cover the async path either), but since #2726 moved `_OUTPUT_ONLY_TEMPLATE_FIELDS` onto the shared `AbstractDbtLocalBase`, it now looks like both paths are covered when only the sync one is.

## Why a single chokepoint is awkward

The guard must inspect raw kwargs *before* segregation: `DbtLocalBaseOperator.__init__` pre-strips unknown kwargs (so a guard inside `AbstractDbtLocalBase.__init__` would only ever see the already-filtered `base_kwargs`), while the async path forwards `**self.dbt_kwargs` raw. No single existing location cleanly catches both.

## Suggested fix

Enforce the rejection on the async path too — e.g. apply the same `_OUTPUT_ONLY_TEMPLATE_FIELDS` check against `dbt_kwargs` in `DbtRunAirflowAsyncBigqueryOperator.__init__`, or factor the check into a small shared helper invoked by both entry points. Add a test mirroring the sync-path tests added in #2726.

## References

- PR #2726 (sync-path fix): https://github.com/astronomer/astronomer-cosmos/pull/2726
- Review comment: https://github.com/astronomer/astronomer-cosmos/pull/2726#pullrequestreview-4593006566
- Relevant code: `cosmos/operators/_asynchronous/bigquery.py` (`DbtRunAirflowAsyncBigqueryOperator.__init__`)

pankajastro:
## Description

Follow-up to #2852.

#2852 added a guard on `DbtRunAirflowAsyncBigqueryOperator` that rejects output-only template fields (`compiled_sql`, `freshness`) passed via `dbt_kwargs`. However, the user-facing path is still unprotected.

`compiled_sql`/`freshness` are **template fields, not `__init__` parameters**. When a user sets them via `operator_args`, they arrive at `DbtRunAirflowAsyncOperator` as top-level kwargs (`operator_args` flow verbatim as `**task_kwargs` in `cosmos/core/airflow.py`). The kwarg split in `DbtRunAirflowAsyncOperator.__init__` routes anything that isn't a recognized init param into `clean_kwargs`, so these fields never land in `dbt_kwargs` and the #2852 guard never sees them.

Result today: the user gets an opaque `TypeError: Invalid arguments were passed ...` from the underlying async operator instead of the intended `CosmosValueError`.

## Fix

Reject the output-only fields against the raw `kwargs` at the top of `DbtRunAirflowAsyncOperator.__init__`, before the split. Reuses the existing `AbstractDbtLocalBase._reject_output_only_template_fields` helper.

After the fix, `operator_args={"compiled_sql": ...}` (or `freshness`) raises:

```
'compiled_sql' is output-only template field(s) populated by Cosmos at runtime; remove them from operator_args.
```

## Tests

Adds a parametrized unit test on `DbtRunAirflowAsyncOperator` (which had no prior coverage) for both `compiled_sql` and `freshness`. Top-level kwargs are exactly how `operator_args` reach the operator, so this covers the user path without needing a metadata DB.

Closes #2851

### DIFF
--- a/cosmos/operators/airflow_async.py
+++ b/cosmos/operators/airflow_async.py
@@ -54,6 +54,10 @@ class DbtRunAirflowAsyncOperator(DbtRunAirflowAsyncFactoryOperator):
         **kwargs: Any,
     ) -> None:
 
+        # Reject output-only template fields up front: the split below routes them into
+        # clean_kwargs, bypassing the inner operator's dbt_kwargs guard.
+        AbstractDbtLocalBase._reject_output_only_template_fields(kwargs)
+
         # Cosmos attempts to pass many kwargs that async operator simply does not accept.
         # We need to pop them.
         clean_kwargs = {}
--- a/tests/operators/test_airflow_async.py
+++ b/tests/operators/test_airflow_async.py
@@ -130,3 +130,31 @@ def test_dbt_run_airflow_async_operator_rejects_output_only_template_fields(fiel
             profile_config=profile_config,
             **{field: "value-the-user-tried-to-set"},
         )
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("field", ["compiled_sql", "freshness"])
+def test_airflow_async_operator_args_rejects_output_only_template_fields(mock_bigquery_conn, field):
+    """End-to-end: an output-only template field in ``operator_args`` fails DAG parsing."""
+    profile_mapping = get_automatic_profile_mapping(mock_bigquery_conn.conn_id, {})
+
+    profile_config = ProfileConfig(
+        profile_name="airflow_db",
+        target_name="bq",
+        profile_mapping=profile_mapping,
+    )
+
+    with pytest.raises(CosmosValueError, match=field):
+        DbtDag(
+            project_config=ProjectConfig(dbt_project_path=DBT_PROJECTS_ROOT_DIR / DBT_PROJECT_NAME),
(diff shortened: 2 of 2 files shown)


tatiana (approved):
Thanks for making this consistent with the non-async tasks, @pankajastro !

PR #2871, opened 2026-07-01
Add a `retries` param to `gcs_upload` callback

### LINKED ISSUE #2870: [Feature] Add retries to GCS uploads
### Description

We've a case where a GCS upload will fail when using the tarball upload feature added in 1.15.

As it happens, the airflow GCS hook includes a num_max_attempts parameter. We could simply invoke this:

https://airflow.apache.org/docs/apache-airflow-providers-google/stable/_api/airflow/providers/google/cloud/hooks/gcs/index.html#airflow.providers.google.cloud.hooks.gcs.GCSHook.upload

### Use case/motivation

This should minimize transient DAG errors during uploads.

### Related issues

_No response_

TTMichaelA:
## Description

This adds retries into GCS uploads. Transient errors during artifact uploads can result in failing DAG tasks unnecessarily. This uses the built in `num_max_attempts` param in `GCSHook.upload()` (which relies on exponential backoff) to retry uploads.

https://airflow.apache.org/docs/apache-airflow-providers-google/stable/_api/airflow/providers/google/cloud/hooks/gcs/index.html#airflow.providers.google.cloud.hooks.gcs.GCSHook.upload

## Related Issue(s)

Closes #2870.

## Breaking Change?

No. This leaves present invocations untouched. Should function tidily in place.

## Checklist

### DIFF
--- a/cosmos/io.py
+++ b/cosmos/io.py
@@ -81,6 +81,7 @@ def upload_to_gcp_gs(
     gcp_conn_id: str | None = None,
     source_subpath: str = DEFAULT_TARGET_PATH,
     use_tarball: bool = False,
+    num_max_attempts: int = 1,
     **kwargs: Any,
 ) -> None:
     """
@@ -91,6 +92,7 @@ def upload_to_gcp_gs(
     :param gcp_conn_id: GCP connection ID to use when uploading files.
     :param source_subpath: Path of the source directory sub-path to upload files from.
     :param use_tarball: If True, uploads a single tar.gz archive instead of individual files.
+    :param num_max_attempts: Number of upload attempts. Defaults to 1.
     """
     from airflow.providers.google.cloud.hooks.gcs import GCSHook
 
@@ -116,6 +118,7 @@ def upload_to_gcp_gs(
             object_name=f"{run_prefix}/{source_subpath}.tar.gz",
             data=buf.read(),
             mime_type="application/gzip",
+            num_max_attempts=num_max_attempts,
         )
     else:
         # Iterate over the files in the target dir and upload them to GCP GS
@@ -126,6 +129,7 @@ def upload_to_gcp_gs(
                     filename=f"{dirpath}/{filename}",
                     bucket_name=bucket_name,
                     object_name=object_name,
+                    num_max_attempts=num_max_attempts,
                 )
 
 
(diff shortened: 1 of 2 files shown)


tatiana (commented):
This is a good improvement, @TTMichaelA , thank you very much!

Before approving/merging, please, could you confirm how hard it would be to expose this parameter in `upload_to_azure_wasb`, `upload_to_aws_s3` and `upload_to_cloud_storage`? Ideally we'd have the interfaces consistent. I understand hooks may be not consistent - and I'm happy for us to log a follow up ticket.

pankajkoti (approved):
LGTM. Thanks for adding this @TTMichaelA

TTMichaelA:
> This is a good improvement, @TTMichaelA , thank you very much!
> 
> Before approving/merging, please, could you confirm how hard it would be to expose this parameter in `upload_to_azure_wasb`, `upload_to_aws_s3` and `upload_to_cloud_storage`? Ideally we'd have the interfaces consistent. I understand hooks may be not consistent - and I'm happy for us to log a follow up ticket.

Good question - so this is tidily built into the GCS hook (as linked above).

The AWS uses `load_bytes`, which has no analogue: https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/_api/airflow/providers/amazon/aws/hooks/s3/index.html#airflow.providers.amazon.aws.hooks.s3.S3Hook.load_bytes

Same with Azure: https://airflow.apache.org/docs/apache-airflow-providers-microsoft-azure/stable/_api/airflow/providers/microsoft/azure/hooks/wasb/index.html#airflow.providers.microsoft.azure.hooks.wasb.WasbHook.upload

I don't really have a ton of familiarity with AWS and almost none at all with Azure, but I understand from a little scanning the docs (and some light interrogation of Claude) that this setting is connection level for both AWS and Azure. So the GCS hook is a bit unusual in that it addresses it right in the upload method.

I think this can just be emulated with a try-except block in a loop, but I haven't explored exactly what this would look like yet.

Happy to take a crack at it in a followup. In the meantime, am I OK to merge this?

tatiana (approved):
Thanks for analysing, @TTMichaelA , yes - let's merge this, and if/when there is a need to expose other properties in the other callbacks, we can think about ways of generalising this.

PR #2879, opened 2026-07-09
Render `env_var()` Jinja when deriving WATCHER dataset namespace

### LINKED ISSUE #2843: [Bug] WATCHER mode emits invalid asset URIs from profiles.yml with unrendered dbt env_var()
### Astronomer Cosmos Version

1.14.2

1.11.7

### Versions of dbt adapters

dbt-trino==1.9.3

### LoadMode

AUTOMATIC

### ExecutionMode

WATCHER

### InvocationMode

SUBPROCESS

### airflow version

3.2.1

### Operating System

_No response_

### Deployment

Official Apache Airflow Helm Chart

### Deployment details

  Airflow 3.2.1 on Kubernetes (KubernetesExecutor) via Helm. dbt project (dbt-trino,
  Starburst/Trino target) mounted at runtime by a git-sync sidecar. Cosmos installed as
  astronomer-cosmos[openlineage]==1.14.2. dbt runs in a separate venv (dbt-trino==1.9.3).

### What happened?

When using ExecutionMode.WATCHER together with a ProfileConfig built from a static
profiles.yml (profiles_yml_filepath=...), the Airflow Assets/Datasets emitted for dbt
models have invalid URIs whenever the profile uses dbt Jinja for connection fields.

Our profiles.yml has:
host: "{{ env_var('STARBURST_HOST') }}"

The emitted asset URI keeps the unrendered Jinja literal, e.g.:
trino://{{ env_var('STARBURST_HOST') }}:443/lake/trusted/my_model

Expected: in WATCHER mode the namespace should reflect the **resolved** host, as it does
in LOCAL/VIRTUALENV modes.

### Relevant log output

```shell
trino://{{ env_var('STARBURST_HOST') }}:443/lake/trusted/my_model
```

### How to reproduce

1. Airflow 3.x, astronomer-cosmos 1.14.2, dbt-trino.
2. profiles.yml using dbt Jinja for the host:
 my_profile:
   outputs:
     prod:
       type: trino
       method: ldap
       host: "{{ env_var('STARBURST_HOST') }}"
       port: 443
       database: lake
       schema: prod
       ...
3. Build a DbtDag with:
 ExecutionConfig(execution_mode=ExecutionMode.WATCHER)
 ProfileConfig(profile_name=..., target_name="prod",
               profiles_yml_filepath="/path/to/profiles.yml")
4. Trigger the DAG and inspect the Assets emitted for the models.
-> The asset URIs contain the literal "{{ env_var('STARBURST_HOST') }}" instead of
the resolved host.

### Anything else :)?

_No response_

### Contact Details

_No response_

pankajastro:
## Summary
- `ProfileConfig.profiles_yml_filepath` targets are read directly with `yaml.safe_load`, so any field using dbt's `env_var()` Jinja (e.g. `host: "{{ env_var('STARBURST_HOST') }}"`) was passed through literally into the WATCHER dataset-namespace resolvers, producing invalid Asset/Dataset URIs like `trino://{{ env_var('STARBURST_HOST') }}:443/...`.
- `_get_profile_dict` now renders `env_var()` Jinja on string values before returning the target dict, reusing the existing `_resolve_env_var` helper from `cosmos/dbt/project.py` — the same rendering dbt itself performs on `profiles.yml` at run time.
- The `profile_mapping` (Airflow-connection based) path is untouched, since those values already come from a resolved connection rather than a raw `profiles.yml`.

Fixes #2843

## Test plan
- [x] Added a regression test (`test_profiles_yml_filepath_renders_env_var`) reproducing the exact scenario from #2843.
- [x] `TestGetDatasetNamespace` (17/17) and the full `test_dataset.py` (35/35) pass.
- [x] `pre-commit run --all-files` (ruff, black, mypy, etc.) passes.
- [x] Manually reproduced the bug against pre-fix code (unrendered `trino://{{ env_var('STARBURST_HOST') }}:443`) and confirmed the fix resolves it end-to-end through the real `ProfileConfig` + `get_dataset_namespace` path (no mocks), with a real profiles.yml on disk and a real env var.

### DIFF
--- a/cosmos/dataset.py
+++ b/cosmos/dataset.py
@@ -11,6 +11,7 @@ from packaging.version import Version
 
 from cosmos import settings
 from cosmos.constants import _DATASET_EMITTING_RESOURCE_TYPES, AIRFLOW_VERSION
+from cosmos.dbt.project import _resolve_env_var
 from cosmos.log import get_logger
 
 if TYPE_CHECKING:
@@ -145,6 +146,8 @@ def _get_profile_dict(profile_config: ProfileConfig) -> tuple[str, dict[str, Any
         with open(profile_config.profiles_yml_filepath) as f:
             profiles = yaml.safe_load(f)
         target = profiles[profile_config.profile_name]["outputs"][profile_config.target_name]
+        # dbt renders env_var() Jinja in profiles.yml; replicate that since we read the file directly.
+        target = {key: _resolve_env_var(value) if isinstance(value, str) else value for key, value in target.items()}
         adapter_type = target.get("type", "")
         return adapter_type, target
 
--- a/cosmos/dataset.py
+++ b/cosmos/dataset.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import json
+import os
+import re
 import urllib.parse
 from collections.abc import Callable
 from pathlib import Path
@@ -11,7 +13,6 @@ from packaging.version import Version
 
 from cosmos import settings
 from cosmos.constants import _DATASET_EMITTING_RESOURCE_TYPES, AIRFLOW_VERSION
-from cosmos.dbt.project import _resolve_env_var
 from cosmos.log import get_logger
 
 if TYPE_CHECKING:
@@ -129,6 +130,23 @@ def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_i
     return "__".join(identifiers_list)
 
 
+# Matches dbt's `{{ env_var('NAME') }}` / `{{ env_var('NAME', 'default') }}` Jinja calls only.
+# Deliberately narrower than a general Jinja renderer: profiles.yml values are only ever
+# substituted for this specific dbt construct, so no other Jinja syntax is evaluated.
+_ENV_VAR_PATTERN = re.compile(
+    r"""\{\{\s*env_var\(\s*(['"])(?P<name>.*?)\1\s*(?:,\s*(['"])(?P<default>.*?)\3\s*)?\)\s*\}\}"""
+)
+
+
+def _render_env_var(value: str) -> str:
+    """Substitute dbt-style ``env_var('NAME'[, 'default'])`` Jinja calls with the environment value."""
+
+    def _replace(match: re.Match[str]) -> str:
+        return os.getenv(match.group("name"), match.group("default") or "")
+
+    return _ENV_VAR_PATTERN.sub(_replace, value)
+
+
 def _get_profile_dict(profile_config: ProfileConfig) -> tuple[str, dict[str, Any]]:
     """
     Extract the adapter type and profile dict from a ProfileConfig.
@@ -147,7 +165,7 @@ def _get_profile_dict(profile_config: ProfileConfig) -> tuple[str, dict[str, Any
             profiles = yaml.safe_load(f)
         target = profiles[profile_config.profile_name]["outputs"][profile_config.target_name]
         # dbt renders env_var() Jinja in profiles.yml; replicate that since we read the file directly.
-        target = {key: _resolve_env_var(value) if isinstance(value, str) else value for key, value in target.items()}
+        target = {key: _render_env_var(value) if isinstance(value, str) else value for key, value in target.items()}
         adapter_type = target.get("type", "")
         return adapter_type, target
 
--- a/cosmos/dataset.py
+++ b/cosmos/dataset.py
@@ -1,18 +1,18 @@
 from __future__ import annotations
 
 import json
-import os
-import re
 import urllib.parse
 from collections.abc import Callable
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 import yaml
+from jinja2.exceptions import TemplateError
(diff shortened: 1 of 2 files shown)


cosmos/dataset.py
```diff
@@ -140,6 +141,8 @@ def _get_profile_dict(profile_config: ProfileConfig) -> tuple[str, dict[str, Any
         with open(profile_config.profiles_yml_filepath) as f:
             profiles = yaml.safe_load(f)
         target = profiles[profile_config.profile_name]["outputs"][profile_config.target_name]
+        # dbt renders env_var() Jinja in profiles.yml; replicate that since we read the file directly.
+        target = {key: _resolve_env_var(value) if isinstance(value, str) else value for key, value in target.items()}
```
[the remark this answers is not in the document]
pankajastro: Addressed in ba56ae46: replaced the general-purpose `_resolve_env_var` (unsandboxed `jinja2.Template`) with a targeted regex that only substitutes the dbt `env_var('NAME'[, 'default'])` form. Any other Jinja in `profiles.yml` is now left untouched rather than evaluated.

cosmos/dataset.py
```diff
@@ -123,6 +125,23 @@ def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_i
...
+
+def _render_env_var(value: str) -> str:
+    """Substitute dbt-style ``env_var('NAME'[, 'default'])`` Jinja calls with the environment value."""
+
+    def _replace(match: re.Match[str]) -> str:
+        return os.getenv(match.group("name"), match.group("default") or "")
+
+    return _ENV_VAR_PATTERN.sub(_replace, value)
+
+
```
tatiana: Why are we not using `_resolve_env_var from `cosmos.dbt.project`? I feel like it implements a safer approach, it feels using regex to resolve this type of templating in Jinja is a suboptimal approach.
https://github.com/astronomer/astronomer-cosmos/blob/7f941f02c8b643402b5d7de490e923891a14efa3/cosmos/dbt/project.py#L46-L64

tatiana (commented):
Thanks for tracking this down, the #2843 repro and the end-to-end manual verification are great.

When I started reading this PR, I was surprised with the re-implementation of resolving dbt `env_var` - I was sure I had implemented this in the past in Cosmos. I checked and I found it here:
https://github.com/astronomer/astronomer-cosmos/pull/2879/changes#r3659396422

The summary says this renders `env_var()` *"reusing the existing `_resolve_env_var` helper from `cosmos/dbt/project.py`"*, but the implementation actually adds a **new** regex-based renderer (`_ENV_VAR_PATTERN` + `_render_env_var`) in `cosmos/dataset.py`. So we'd end up with two independent env_var renderers in the codebase - which is suboptimal.

The existing `_resolve_env_var` is used by `get_dbt_packages_subpath` and covered by tests in `tests/dbt/test_project.py` (simple, default, complex-conditional, unset-var).

In addition to the duplication, the regex is strictly worse than the helper we already have. It only matches the bare `{{ env_var('X') }}` / `{{ env_var('X','default') }}` form, so these common `profiles.yml` shapes would still pass through unrendered and produce a broken namespace:

| value | `_resolve_env_var` | regex in this PR |
|---|---|---|
| `{{ env_var('HOST') }}` | ✅ | ✅ |
| `{{ env_var('HOST') ~ ':443' }}` | ✅ | ❌ literal |
| `{{ env_var('HOST') \| upper }}` | ✅ | ❌ literal |
| `{{ "_" + env_var("e","") if ... }}` | ✅ | ❌ literal |

Could we just reuse the existing helper? No circular-import issue: `cosmos/dbt/project.py` only imports `cosmos.constants` + `cosmos.log`, and `dataset.py` doesn't import `dbt.*` at runtime.

```python
from cosmos.dbt.project import _resolve_env_var
...
target = {k: _resolve_env_var(v) if isinstance(v, str) else v for k, v in target.items()}
```

That drops ~20 lines, matches dbt's own rendering semantics, and keeps a single source of truth. `get_dataset_namespace` already has the catch-all → `None` path (`test_returns_none_on_error`), so a malformed template won't break a DAG.

cosmos/dataset.py
pankajastro: Reverted to `_resolve_env_var` (full Jinja) per your review — the regex here missed common `profiles.yml` shapes (`~` concatenation, `|` filters), matching the table above.

On the security concern: we'd switched to the regex earlier specifically in response to Copilot's pushback on `_resolve_env_var`'s unsandboxed Jinja rendering. But `profiles.yml` isn't attacker-controlled — it's authored by the same person as the DAG, same trust boundary Cosmos already uses this exact helper for elsewhere (`get_dbt_packages_subpath`). No new privilege boundary crossed.

Added `jinja2.exceptions.TemplateError` to the catch-all so non-`env_var` Jinja that fails to render still returns `None` instead of crashing DAG parsing. All in 8f83abb7, with test coverage for both the concatenation case and the undefined-Jinja fallback.

tatiana (approved):
Thanks for addressing the feedback, @pankajastro ! Looks good

PR #2885, opened 2026-07-10
adapter-native semantic layer (databricks & snowflake) : render semantic layer node types

corsettigyg:
## Description

Databricks and Snowflake have both introduced their platform-native semantic layer, which as of now can directly be plugged into dbt-core (it runs via `dbt run`). This PR just reclassifies the nodes to its own type, allowing better management of resources via airflow. for now, I see that only databricks and snowflake have their own semantic layer, but if others introduce it, we just need to add it to the materialization type below

```python
DBT_SEMANTIC_LAYER_MATERIALIZATIONS = {"metric_view", "semantic_view"}
```

## Internal discussion

this is **not** for MetricFlow. At GetYourGuide we weighted both options and MetricFlow lacks a good platform integration as a whole, meanwhile for companies on databricks like us, metric_view is already pluggable by default into the whole ecosystem (UC lineage, genie queries, etc). 

## Breaking Change?

small risk of breaking something in case users are treating semantic nodes as run nodes

## Checklist


tatiana:
Very happy to see progress in this stream, @corsettigyg ! 🎉 
We'll review once its out of draft

corsettigyg:
hey @tatiana , so the PR is ready for review since it ended up being simpler than I expected 

Screenshot 2026-07-13 at 20 01 51

So giving more context here

## native semantic layer vs Metric Flow
we considered metric flow on our side since it is agnostic to the warehouse and the "go to" semantic layer for dbt, but this comes with the downside that for warehouses like databricks or snowflake that have a native semantic layer, MetricFlow is suboptimal and lives outside of the warehouse. For example, in databricks we can leverage UC lineage and directly interact with the semantic layers using the default metric views.

## behaviour 
since dbt-databricks abstracts the metric view as a dbt run, it works by default with cosmos with the dbt run operator. That all being said, running it as dbt run still flags it as a `metric view` type of node in the dbt CLI. the purpose of this PR is solely to reclassify the node type via UI and also in a way that people can manage semantic layer nodes in their own way.

## testing
it worked fine for databricks. for snowflake, the process was replicated but I did not directly test so cannot be 100% sure.


I am open to make changes or change the directions if you deem it appropriate 😄  currently, this would only work for databricks & snowflake, although I feel that having an ambitious goal of adding MetricFlow into cosmos would be by itself a whole different beast that would achieve similar or worse results for the aforementioned warehouses

pankajkoti:
Hi @corsettigyg, planning to review the PR this week. Looks like there are now some conflicts. If you get a chance, meanwhile, could you please resolve those? Thanks!

dev/dags/dbt/jaffle_shop/models/customer_metrics.sql
```diff
@@ -0,0 +1,3 @@
+{{ config(materialized='metric_view') }}
```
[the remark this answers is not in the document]
corsettigyg: this is for testing purposes so suggestion is not needed

corsettigyg:
@pankajkoti @tatiana any news here?

dev/dags/dbt/jaffle_shop/models/customer_metrics.sql
pankajkoti: This fixture is why CI is red. Adding a metric_view model to the shared jaffle_shop project bumps its node count, so hardcoded assertions go stale across the integration suite: tests/listeners/test_dag_run_listener.py:52 and :65 (total_cosmos_tasks == 13, now 14), and tests/operators/test_watcher.py:1888 and :1890 (filtered_nodes == 23 and task_dict == 14, now 24 and 15, plus the expected_task_names list below). In WATCHER mode that test also runs dbt build, so a metric_view against the postgres profile would fail there too.

Two ways to fix, both fine. Move the fixture into an isolated dbt project (or one exercised only through a static manifest) so it stops perturbing the shared project, or keep it here and update every count above. If it stays, note the model still cannot actually run against postgres. Please confirm the integration suite passes with whichever route you pick.

cosmos/dbt/graph.py
```diff
@@ -301,6 +302,14 @@ def is_freshness_effective(freshness: dict[str, Any] | None) -> bool:
     return False
 
 
+def _classify_resource_type(resource_type: DbtResourceType, config: dict[str, Any]) -> DbtResourceType:
```
pankajkoti, line 305: Reclassification is unconditional, so an existing metric_view or semantic_view model changes task identity on upgrade: my_model / DbtRun* becomes my_model_semantic_layer / DbtSemantic* (and _semantic_layer_build under BUILD). Airflow sees a brand new task, so task history, manual clears, and references to the old task_id break.

Two asks if you think it makes sense. Add a Behaviour Changes entry to CHANGELOG.rst (there is none in this PR), and consider gating the reclassification behind a setting (new behavior on by default), like the existing enable_dags_folder_exclusion_from_dbt flag. A migration note that resource_type:model selectors and node_converters keyed on MODEL stop matching these nodes would help too.

docs/guides/translate_dbt_to_airflow/managing-semantic-layer.rst
```diff
@@ -0,0 +1,61 @@
...
+How it works
+~~~~~~~~~~~~
+
+dbt has no distinct resource type for these objects — they're plain ``model`` nodes with
+``config.materialized: metric_view`` (Databricks) or ``semantic_view`` (Snowflake). Cosmos detects
+that materialization at parse time and reclassifies the node as ``semantic_layer``, which changes:
+
+- **Task identity:** a ``_semantic_layer`` task name suffix and a ``DbtSemantic*`` operator instead of ``DbtRun*``. The underlying dbt command is still ``dbt run``.
+- **TestBehavior.BUILD:** collapses into ``DbtBuild``, same as any other buildable resource.
+- **Selectors:** ``resource_type:model`` no longer matches it — use ``resource_type:semantic_layer`` instead. This
```
pankajkoti: This says the reclassification is only applied when Cosmos performs selection (e.g. DBT_MANIFEST), but that is not accurate. parse_dbt_ls_output also reclassifies under DBT_LS, so the _semantic_layer suffix and DbtSemantic* operator apply there too. What is actually limited to Cosmos-side selection is only the resource_type:semantic_layer selector, because DBT_LS hands select and exclude to the dbt CLI before reclassification. Please reword so the caveat is about the selector, not the reclassification. The mirrored sentence in selecting-excluding.rst has the same issue.

cosmos/operators/base.py
```diff
@@ -471,6 +471,13 @@ def add_cmd_flags(self) -> list[str]:
         return flags
 
 
+class DbtSemanticMixin(DbtRunMixin):
```
pankajkoti, line 474: The new DbtSemantic* operators are not exported in cosmos/operators/__init__.py or the lazy-import map in cosmos/__init__.py, while every peer (DbtSnapshotOperator, DbtSeedOperator, and the per-mode variants) is. The DAG still renders since loading is by dotted path, but users cannot do from cosmos.operators import DbtSemanticOperator. Please add the exports to match the peers.

tests/airflow/test_graph.py
```diff
@@ -904,6 +904,91 @@ def test_create_task_metadata_ephemeral_model_disabled_renders_dbt_build_in_buil
...
+        resource_type=DbtResourceType.SEMANTIC_LAYER,
+        depends_on=[],
+        path_base=Path("."),
+        original_file_path=Path("."),
+        tags=[],
+        config={"materialized": materialization},
+    )
+
+
+@pytest.mark.parametrize("materialization", ["metric_view", "semantic_view"])
```
pankajkoti: This test and the build-mode one below are parametrized over metric_view and semantic_view, but _semantic_layer_node hardcodes resource_type=SEMANTIC_LAYER and create_task_metadata never reads materialized, so both cases run identical code. The distinction is already covered by the reclassification tests in tests/dbt/test_graph.py. You can drop the parametrize on these two.

pankajkoti (approved):
hi @corsettigyg , thanks for contributing this PR. Looks good to me. Sorry this took longer to review, caught up with some internal priorities and now prepping up for the Airflow Summit. It looks like a good one to me and additionally I sought Claude's help to review (the inline comments are Claude-generated) and with that the CI needs fixing, I gave it a run and see tests failing. 

@tatiana mentioned she is interested in reviewing the PR as well. She's currently on time-off and will be back on 7 September; she can review then.

PR #2901, opened 2026-07-17
Fix watcher multi-parent test aggregation: credit all parent models

### LINKED ISSUE #2900: [Bug] ExecutionMode.WATCHER + TestBehavior.AFTER_EACH: a test with multiple parent models is credited to only one parent, so the other parent's test sensor hangs until the producer terminates
### Astronomer Cosmos Version

1.15.0

dbt Core 1.12.0

### Versions of dbt adapters

dbt-duckdb 1.10.1

### LoadMode

DBT_LS

### ExecutionMode

WATCHER

### InvocationMode

SUBPROCESS

### airflow version

3.2.1

### Operating System

Linux (Astro Runtime container image, Debian-based), via the Astro CLI on macOS host.

_No response_

### Deployment

Astronomer

### Deployment details

Local reproduction via the Astro CLI (Docker).

### What happened?

**Setup.** Under `ExecutionMode.WATCHER` with `TestBehavior.AFTER_EACH`, each model gets a
per-model test sensor that waits for that model's aggregated test status (an XCom the
producer pushes once all of the model's tests have reported). A dbt test that depends on
**more than one model** — e.g. any `relationships` test, which depends on both the model
it's defined on and the model it references — is counted in *every* parent's expected-test
set (dbt's default eager indirect-selection behavior, mirrored by Cosmos).

**What happens.** When that shared test finishes, its result is recorded for **only the
first parent** Cosmos finds. The other parent's expected-test set is therefore never
completed, so its aggregated `<model>_tests_status` XCom is never pushed, and its test
sensor **defers forever** — even though the test actually ran and reported. Concretely,
with two models `parent_a` and `child_b` where a `relationships` test on `child_b`
references `parent_a`:

- `child_b`'s test sensor completes normally (it received the shared test's result).
- `parent_a`'s test sensor stays **deferred** for the entire remaining duration of the
  producer's `dbt build`, blocking everything downstream of `parent_a` in the Airflow
  graph. It only resolves when the producer task reaches a terminal state:
  - producer **succeeds** → the sensor resolves *success* via the `NODE_NOT_RUN`
    fallback, i.e. it goes green **without ever having observed a test result**. Correct
    only by luck (the tests happened to pass); nothing was verified.
  - producer **fails** (e.g. because the shared test failed → `dbt build` exits non-zero)
    → the sensor fails with `PRODUCER_FAILED` ("check the producer logs") **instead of the
    actual test verdict** — a delayed, misattributed failure on the model whose references
    the test validates.

On a small project the extra wait is seconds; on a large single-`dbt build` graph it is
the full remaining build time (observed: a hub model's test sensor deferred for ~2h46m
while the rest of the build ran).

### Relevant log output

```shell

```

### How to reproduce

Minimal, self-contained (DuckDB — no external services). Two models where a `relationships`
test creates a shared (multi-parent) test:

**`models/parent_a.sql`**
```sql
select * from (values (1), (2), (3)) as t(id)
```

**`models/child_b.sql`**
```sql
select a.id as a_id from {{ ref('parent_a') }} a
```

**`models/schema.yml`**
```yaml
version: 2
models:
  - name: parent_a
    columns:
      - name: id
        tests: [not_null, unique]           # parent_a's own tests
  - name: child_b
    columns:
      - name: a_id
        tests:
          - not_null
          - relationships:                  # SHARED test: depends on child_b AND parent_a
              arguments:
                to: ref('parent_a')
                field: id
```

`parent_a` now has an expected-test set of 3 (its own `not_null` + `unique`, plus the
shared `relationships` test); `child_b` has 2 (its own `not_null` + the shared test).

**DAG** (WATCHER + AFTER_EACH):
```python
import cosmos

repro = cosmos.DbtDag(
    dag_id="watcher_multi_parent_repro",
    schedule=None,
    start_date=__import__("datetime").datetime(2024, 1, 1),
    project_config=cosmos.ProjectConfig(dbt_project_path="/path/to/project"),
    profile_config=cosmos.ProfileConfig(
        profile_name="repro", target_name="dev",
        profiles_yml_filepath="/path/to/project/profiles.yml",
    ),
    execution_config=cosmos.ExecutionConfig(execution_mode=cosmos.ExecutionMode.WATCHER),
    render_config=cosmos.RenderConfig(test_behavior=cosmos.TestBehavior.AFTER_EACH),
)
```

**Trigger the DAG and observe:**
1. `child_b`'s test sensor completes; `parent_a`'s test sensor stays **deferred**.
2. `parent_a`'s aggregated XCom key (`model__<project>__parent_a_tests_status`) is never
   written (404 in the XCom API / absent in the producer task's XCom tab), even though the
   producer log shows the shared test ran and passed.
3. `parent_a`'s test sensor only resolves when the producer task terminates
   (success → green via `NODE_NOT_RUN`; failure → `PRODUCER_FAILED`).

To see the failing-case behavior, make the shared test fail (e.g. add a `child_b` row whose
`a_id` has no match in `parent_a.id`): the shared test fails, the producer fails on dbt's
non-zero exit, and `parent_a`'s test sensor ends `PRODUCER_FAILED` rather than reporting the
actual test failure.

### Anything else :)?

_No response_

### Contact Details

ignaskr@gmail.com

hellowithchicken:
## Description

Under `ExecutionMode.WATCHER` + `TestBehavior.AFTER_EACH`, a dbt test with more than
one parent model (e.g. a `relationships` test) is listed under every parent in
`tests_per_model` but its result was credited to only the first parent in
`accumulate_test_result` (early `return`). The other parent's expected-test set never
completed, so its `<model>_tests_status` XCom was never pushed and its test sensor
deferred until the producer terminated (then resolved unverified on producer success,
or `PRODUCER_FAILED` on producer failure).

### Fix
`accumulate_test_result` had an early `return` that stopped at the first parent. Drop it so
the result is recorded under every parent (return type `str | None` -> `list[str]`), and have
`push_test_result_or_aggregate` aggregate/push for each parent returned. Single-parent tests
are unchanged (the loop matches exactly one model); locking and keying are untouched.

Tests: added regression coverage in `tests/operators/_watcher/test_aggregation.py` (pass/fail
for a shared test, plus a no-shared-test control) and `tests/dbt/test_graph.py`.

## Related Issue(s)

closes #2900 

## Breaking Change?

No.

## Checklist

### DIFF
--- a/cosmos/operators/_watcher/aggregation.py
+++ b/cosmos/operators/_watcher/aggregation.py
@@ -23,20 +23,25 @@ def accumulate_test_result(
     status: str,
     tests_per_model: dict[str, list[str]],
     test_results_per_model: dict[str, dict[str, str]],
-) -> str | None:
-    """Record a test's terminal status into test_results_per_model for its parent model.
+) -> list[str]:
+    """Record a test's terminal status into test_results_per_model for every parent model.
+
+    A test can belong to more than one model (e.g. a ``relationships`` test depends on
+    both the model it is defined on and the model it references), so it may appear under
+    several keys in ``tests_per_model``; its status is recorded under each.
 
     Results are keyed by ``test_unique_id`` so that a duplicated or replayed log line for
     the same test (the Kubernetes log reader can replay recent lines after an interruption)
     overwrites its entry instead of being counted as another distinct test.
 
-    Returns the parent model's unique_id if found, else None.
+    Returns the unique_ids of all parent models the test was recorded for (empty if none).
     """
+    matched_model_uids: list[str] = []
     for model_uid, test_uids in tests_per_model.items():
         if test_unique_id in test_uids:
             test_results_per_model.setdefault(model_uid, {})[test_unique_id] = status
-            return model_uid
-    return None
+            matched_model_uids.append(model_uid)
+    return matched_model_uids
 
 
 def get_aggregated_test_status(
@@ -86,8 +91,8 @@ def push_test_result_or_aggregate(
     :param task_instance: The Airflow task instance used for XCom push.
     """
     with _test_results_lock:
-        model_uid = accumulate_test_result(test_unique_id, status, tests_per_model, test_results_per_model)
-        if model_uid is not None:
+        model_uids = accumulate_test_result(test_unique_id, status, tests_per_model, test_results_per_model)
+        for model_uid in model_uids:
             aggregated = get_aggregated_test_status(model_uid, tests_per_model, test_results_per_model)
             if aggregated is not None:
                 safe_xcom_push(
--- a/tests/dbt/test_graph.py
+++ b/tests/dbt/test_graph.py
@@ -1625,6 +1625,39 @@ def test_tests_per_model_populated():
     assert any("unique_customers_customer_id" in t for t in customers_tests)
 
 
+def test_tests_per_model_lists_multi_parent_test_under_every_parent():
+    """A test with multiple parents (e.g. a relationships test) is listed under each parent.
+
+    This is the graph-level fact the watcher test-aggregation logic relies on: a shared
+    test must be recorded/aggregated for every parent model, not just one. Asserting it
+    against a real loaded graph keeps the aggregation unit tests honest — they hand-build
+    ``tests_per_model``, and this guards that the builder actually produces that shape.
+    """
+    project_config = ProjectConfig(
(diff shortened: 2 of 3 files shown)


tatiana (approved):
Thank you very much, @hellowithchicken, for this fix. We'll be shipping this in Cosmos 1.15.1.

hellowithchicken:
@tatiana Awesome! Very happy to contribute. Thanks for your work on Cosmos!

PR #2914, opened 2026-07-23
Document verified Cosmos/Runtime/dbt Core compatibility

pankajastro:
## Description

Adds three empirically-verified compatibility tables to `docs/policy/compatibility-policy.rst`, replacing/supplementing the previously declared-only version ranges:

- **Verified Python × dbt Core support** — each dbt Core release from 1.5 to 2.0 pip-installed into plain Python 3.10–3.14 images and import-checked. Surfaces two real breakages: dbt Core 1.5–1.7 fail on Python 3.12+ (`distutils` removal), and 1.8–1.11 fail on Python 3.14 (`mashumaro` incompatibility). dbt Fusion (2.0) isn't pip-installable as `dbt-core` at all.
- **Astronomer Runtime** table simplified to Runtime series → Airflow → Python, cross-linked to the dbt Core table, with a note pointing readers to each package's PyPI page for the live, authoritative Python-support metadata.
- **Cosmos and Astronomer Runtime compatibility** — every Cosmos release from 1.5.1 to 1.15.0 pip-installed into the official Runtime Docker image for each Runtime series (11, 12, 13, 3.0, 3.1, 3.2, 3.3), then checked via `airflow plugins` (the real plugin-loading path, not a bare import). Documents exactly which Cosmos releases work on which Runtime series, with root-cause tracebacks for each failure mode (including a severe one: Cosmos 1.6.0–1.13.1 crash the entire `airflow` process, not just plugin loading, on Airflow 3.2/3.3).

### Follow-up fixes from review

- Replaced a dangling "existing Runtime 3.0 compatibility note" reference with a real `:doc:` link to `docs/policy/airflow3-compatibility.rst`, which documents the actual constraint (Cosmos's AF3 plugin requires Airflow >= 3.1).
- Joined an inline literal that spanned a line break (the `distutils` exception text) onto a single line.
- Corrected a `list-table` cell's indentation, after Copilot Autofix's own fix for this introduced a different misalignment.
- Runtime 3.0's table row was rewritten twice in response to pankajkoti's feedback: it originally read as "0 Cosmos versions work on Airflow 3.0," which is wrong — Cosmos itself has been supported on Airflow 3.0 since Cosmos 1.10.0, only the dbt docs UI plugin needs Airflow >= 3.1. The row now states that directly instead of "N/A".
- Added a "Python (bundled)" column to the Runtime compatibility table and split Runtime 11 out from 12/13, since its `python-3.9` variant has a different (and narrower) Cosmos version support story.
- Simplified the "Cosmos and Astronomer Runtime compatibility" section: dropped "(verified)" from the heading and removed a disclaimer paragraph ("this only tells you if the plugin loads, not if Cosmos works") that was both redundant with the Runtime 3.0 row's own explanation and, per pankajkoti's feedback, read as needlessly unsure of what the section actually establishes. Kept a link to the Airflow 3 compatibility notes for readers who want the full picture.
- Rebased onto main to pick up the dbt 1.12 / Python 3.14 test matrix additions.

## Related Issue(s)

Linear: [BOSS-571](https://linear.app/astronomer/issue/BOSS-571/version-compatibility-matrix-published-and-linked-from-the-cosmos-docs)

closes: BOSS-571

## Breaking Change?

No — documentation only.

## Checklist

### DIFF
--- a/docs/policy/compatibility-policy.rst
+++ b/docs/policy/compatibility-policy.rst
@@ -201,8 +201,7 @@ The 13 Cosmos releases tested, oldest to newest: 1.5.1, 1.6.0, 1.7.1, 1.8.2,
    * - 3.0
      - 3.0
      - None
-     - All 13 tested releases (1.5.1 – 1.15.0). See the existing Runtime 3.0
-       compatibility note.
+      - All 13 tested releases (1.5.1 – 1.15.0).
    * - 3.1
      - 3.1
      - 1.11.3, 1.12.1, 1.13.1, 1.14.0, 1.14.1, 1.14.2, 1.15.0
--- a/docs/policy/compatibility-policy.rst
+++ b/docs/policy/compatibility-policy.rst
@@ -96,9 +96,8 @@ into a plain Python image, then ``import dbt.cli.main`` — checked
      - n/a
      - n/a
 
-- dbt Core 1.5–1.7 fail on Python 3.12+ with ``ModuleNotFoundError: No
-  module named 'distutils'`` — Python 3.12 removed the stdlib ``distutils``
-  module those releases still import.
+- dbt Core 1.5–1.7 fail on Python 3.12+ with ``ModuleNotFoundError: No module named 'distutils'`` —
+  Python 3.12 removed the stdlib ``distutils`` module those releases still import.
 - dbt Core 1.8–1.11 fail on Python 3.14 inside ``dbt_common``'s
   ``mashumaro`` dependency (e.g. ``mashumaro.exceptions.UnserializableField``),
   which hasn't caught up to Python 3.14's typing changes.
@@ -201,7 +200,7 @@ The 13 Cosmos releases tested, oldest to newest: 1.5.1, 1.6.0, 1.7.1, 1.8.2,
    * - 3.0
      - 3.0
      - None
-      - All 13 tested releases (1.5.1 – 1.15.0).
+     - All 13 tested releases (1.5.1 – 1.15.0). The Cosmos AF3 plugin requires Airflow >= 3.1 — see :doc:`Airflow 3 compatibility notes </policy/airflow3-compatibility>`.
    * - 3.1
      - 3.1
      - 1.11.3, 1.12.1, 1.13.1, 1.14.0, 1.14.1, 1.14.2, 1.15.0
@@ -209,8 +208,7 @@ The 13 Cosmos releases tested, oldest to newest: 1.5.1, 1.6.0, 1.7.1, 1.8.2,
    * - 3.2, 3.3
      - 3.2, 3.3
      - 1.14.0, 1.14.1, 1.14.2, 1.15.0
-     - 1.5.1, 1.6.0, 1.7.1, 1.8.2, 1.9.2, 1.10.3, 1.11.3, 1.12.1, 1.13.1 — see
-       below.
+     - 1.5.1, 1.6.0, 1.7.1, 1.8.2, 1.9.2, 1.10.3, 1.11.3, 1.12.1, 1.13.1 — see below.
 
 - **Runtime 3.1, Cosmos 1.5.1–1.9.2** fail with an explicit plugin-import
   error:
--- a/docs/policy/compatibility-policy.rst
+++ b/docs/policy/compatibility-policy.rst
@@ -172,22 +172,20 @@ policy pages directly for the current state.
    for the exact combinations). If you rely on an older dbt Core minor with an
    uncommon Airflow/Python pair, validate it in your own CI.
 
-Cosmos and Astronomer Runtime compatibility (verified)
-++++++++++++++++++++++++++++++++++++++++++++++++++++++
+Cosmos and Astronomer Runtime compatibility
++++++++++++++++++++++++++++++++++++++++++++
 
-The ranges above describe what Cosmos declares and tests in CI. The table
-below reflects targeted, empirical verification instead: each Cosmos release
-from 1.5.1 to 1.15.0 was ``pip install``-ed into the official Astronomer
-Runtime Docker image for each Runtime series, then checked by running
-``airflow plugins`` — the same plugin-loading path Airflow itself uses at
-startup — rather than a bare ``import cosmos``. Verified 2026-07-23 against
-Runtime images 11.20.0 (default and ``python-3.9`` variant), 12.12.0, 13.8.0,
-3.0-16, 3.1-17, 3.2-6, and 3.3-2.
+The table below comes from installing each Cosmos release (1.5.1 to 1.15.0)
+into the official Astronomer Runtime Docker image for every Runtime series,
+then running ``airflow plugins`` to check if the Cosmos plugin loads.
+Verified 2026-07-23 against Runtime images 11.20.0 (default and
+``python-3.9`` variant), 12.12.0, 13.8.0, 3.0-16, 3.1-17, 3.2-6, and 3.3-2.
 
-Because this checks plugin *registration* specifically, it is not the same
-question as "does Cosmos work." The Runtime 3.0 row below is the clearest
-example: no tested Cosmos release registers a plugin there, but that is
-expected — it does not mean Cosmos itself is unsupported on Airflow 3.0.
+This only tells you if the plugin loads, not if Cosmos works. Cosmos itself
+has been supported on Airflow 3.0 since Cosmos 1.10.0, but the dbt docs UI
+plugin needs Airflow >= 3.1, so it never loads on Runtime 3.0. See
+:doc:`Airflow 3 compatibility notes </policy/airflow3-compatibility>` for
+what's supported on Airflow 3.0 versus 3.1+.
 
 The 13 Cosmos releases tested, oldest to newest: 1.5.1, 1.6.0, 1.7.1, 1.8.2,
 1.9.2, 1.10.3, 1.11.3, 1.12.1, 1.13.1, 1.14.0, 1.14.1, 1.14.2, 1.15.0.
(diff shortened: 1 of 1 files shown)


docs/policy/compatibility-policy.rst
```diff
@@ -55,6 +55,205 @@ dbt Core
...
+     - Cosmos versions confirmed failing
+   * - 11, 12, 13
+     - 2.9, 2.10, 2.11
+     - All 13 tested releases (1.5.1 – 1.15.0)
+     - None, except on Runtime 11's ``python-3.9`` variant — see below.
+   * - 3.0
+     - 3.0
+     - None
+     - All 13 tested releases (1.5.1 – 1.15.0). See the existing Runtime 3.0
+       compatibility note.
```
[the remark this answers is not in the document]
pankajastro: updated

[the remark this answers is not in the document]
pankajastro: Fixed in 7b6499d8 — this now links to the actual doc that covers it (`docs/policy/airflow3-compatibility.rst`) instead of referencing a non-existent note, and states the real constraint inline: the Cosmos AF3 plugin requires Airflow >= 3.1.

docs/policy/compatibility-policy.rst
```diff
@@ -55,6 +55,205 @@ dbt Core
...
+   * - 2.0 (Fusion)
+     - n/a
+     - n/a
+     - n/a
+     - n/a
+     - n/a
+
+- dbt Core 1.5–1.7 fail on Python 3.12+ with ``ModuleNotFoundError: No
+  module named 'distutils'`` — Python 3.12 removed the stdlib ``distutils``
+  module those releases still import.
```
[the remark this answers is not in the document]
pankajastro: Fixed in 7b6499d8 — joined the exception text onto a single inline literal so it no longer spans a line break.

docs/policy/compatibility-policy.rst
```diff
@@ -55,6 +55,204 @@ dbt Core
...
+     - Cosmos versions confirmed working
+     - Cosmos versions confirmed failing
+   * - 11, 12, 13
+     - 2.9, 2.10, 2.11
+     - All 13 tested releases (1.5.1 – 1.15.0)
+     - None, except on Runtime 11's ``python-3.9`` variant — see below.
+   * - 3.0
+     - 3.0
+     - None
+      - All 13 tested releases (1.5.1 – 1.15.0).
```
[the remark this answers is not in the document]
pankajastro: Fixed in 7b6499d8. Note: the Copilot Autofix applied for this finding (6fe86a03) actually introduced a *new* misalignment — it re-indented the cell to 6 spaces while every other cell in this list-table uses 5, which would have broken the table's column parsing. 7b6499d8 corrects it to match the table's consistent indentation.

docs/policy/compatibility-policy.rst
```diff
@@ -55,6 +55,202 @@ dbt Core
...
+     - Airflow
+     - Cosmos versions confirmed working
+     - Cosmos versions confirmed failing
+   * - 11, 12, 13
+     - 2.9, 2.10, 2.11
+     - All 13 tested releases (1.5.1 – 1.15.0)
+     - None, except on Runtime 11's ``python-3.9`` variant — see below.
+   * - 3.0
+     - 3.0
+     - None
```
pankajkoti: This looks off. It reads as if we don't support Cosmos on AF3. We did add support for Airflow 3 in Cosmos by disabling only the plugin, not loading it. Can we correct this by mentioning the supported Cosmos versions?

docs/policy/compatibility-policy.rst
```diff
@@ -55,6 +55,202 @@ dbt Core
...
+   :header-rows: 1
+
+   * - Runtime series
+     - Airflow
+     - Cosmos versions confirmed working
+     - Cosmos versions confirmed failing
+   * - 11, 12, 13
+     - 2.9, 2.10, 2.11
+     - All 13 tested releases (1.5.1 – 1.15.0)
+     - None, except on Runtime 11's ``python-3.9`` variant — see below.
```
pankajkoti: the exception here on Python-3.9 raises a question on whether we should have here a column for supported Python versions offered by the Runtime releases against AF*cosmos versions here.

pankajkoti (commented):
couple of suggestions/asks inline, rest of the information is good

docs/policy/compatibility-policy.rst
pankajastro: Good catch — fixed in c6703525. The table now marks Runtime 3.0 as N/A instead of "0 versions working," and I added a note above the table plus a bullet explaining that this only measures plugin *registration*: Cosmos core DAG/task functionality has supported Airflow 3 since 1.10.0, and only the FastAPI plugin (dbt docs UI) requires Airflow >= 3.1 by design. Cross-linked to the Airflow 3 compatibility doc for the full picture.

docs/policy/compatibility-policy.rst
pankajastro: Added in c6703525: a "Python (bundled)" column, with Runtime 11 split out from 12/13 since it has a different range (3.9–3.11 vs 3.10–3.12). That makes the python-3.9 outlier visible in the table itself instead of only in the footnote, and I reworded the footnote to call out that it's a Python-version constraint (Cosmos >= 1.12.0 needs >= 3.10), not an Airflow-version one.

docs/policy/compatibility-policy.rst
```diff
@@ -55,6 +55,231 @@ dbt Core
...
+     - Cosmos >= 1.12.0, but only on the ``python-3.9`` variant — see below.
+   * - 12, 13
+     - 2.10, 2.11
+     - 3.10 – 3.12
+     - All 13 tested releases (1.5.1 – 1.15.0)
+     - None
+   * - 3.0
+     - 3.0
+     - 3.11 – 3.12
+     - N/A — no plugin registration is expected here; see below
```
pankajkoti: I don't think this helps someone trying to find which Cosmos versions are supported on a specific runtime version. The title of the section reads "Cosmos and Astronomer Runtime compatibility (verified)". 

So we should aim to reflect that for each of the runtime versions, which are the supported Cosmos versions.

Checking the plugin installation is a strategy we have followed; maybe there is another strategy to test this, and we can explicitly mention that the plugin is not supported on 3.0, but the intention to reflect Cosmos versions supported on specific runtime versions should not be defeated here, I believe.

pankajastro: Fixed in 840c77d4. This info was already in the paragraph above and the bullet below the table — the row itself just said "N/A". Now the same info (supported since Cosmos 1.10.0, dbt docs UI plugin needs Airflow >= 3.1) is in the table cell too, so it's visible without having to read outside the table. Removed the now-redundant standalone bullet, and pre-1.10.0 releases (1.5.1–1.9.2) go in the failing column since Airflow 3 support did not exist yet.

docs/policy/compatibility-policy.rst
```diff
@@ -55,6 +55,220 @@ dbt Core
...
+Cosmos and Astronomer Runtime compatibility
++++++++++++++++++++++++++++++++++++++++++++
+
+The table below comes from installing each Cosmos release (1.5.1 to 1.15.0)
+into the official Astronomer Runtime Docker image for every Runtime series,
+then running ``airflow plugins`` to check if the Cosmos plugin loads.
+Verified 2026-07-23 against Runtime images 11.20.0 (default and
+``python-3.9`` variant), 12.12.0, 13.8.0, 3.0-16, 3.1-17, 3.2-6, and 3.3-2.
+
+This only tells you if the plugin loads, not if Cosmos works. Cosmos itself
```
pankajkoti: This wording might leave users confused. What can guarantee that Cosmos works? Is there a way to guarantee this, especially because the title says "Cosmos and Astronomer Runtime compatibility" for the section? Would be nice if you could take another read of the section and ensure users are not left confused, and adjust accordingly. I believe there is a high chance of leaving users confused after reading this section. Will leave the decision to you @pankajastro

pankajkoti (approved):
LGTM mostly, have one question inline.

docs/policy/compatibility-policy.rst
pankajkoti: yeah the table is what catches the eyes of the readers so that N/A definitely did not help earlier

docs/policy/compatibility-policy.rst
pankajastro: Good point — fixed in:

```
ca97e8a1
```

Removed that disclaimer entirely rather than rewording it — it was redundant with the Runtime 3.0 row's own cell (which already explains the plugin-vs-core distinction for the one case where it matters), and kept restating it up top just made the section sound less sure of itself. The intro now states directly what the table shows, plus a link to the Airflow 3 compatibility notes for the full picture.

docs/policy/compatibility-policy.rst
```diff
@@ -55,6 +55,231 @@ dbt Core
...
+   * - 12, 13
+     - 2.10, 2.11
+     - 3.10 – 3.12
+     - All 13 tested releases (1.5.1 – 1.15.0)
+     - None
+   * - 3.0
+     - 3.0
+     - 3.11 – 3.12
+     - N/A — no plugin registration is expected here; see below
+     - N/A — no plugin registration is expected here; see below
```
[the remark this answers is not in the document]
pankajastro: Resolved as a side effect of ca97e8a1 — the only remaining N/A-style cells in the doc are the lowercase `n/a` row in the dbt Core table, consistently.

docs/policy/compatibility-policy.rst
```diff
@@ -55,6 +55,231 @@ dbt Core
...
+   uncommon Airflow/Python pair, validate it in your own CI.
+
+Cosmos and Astronomer Runtime compatibility (verified)
+++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+The ranges above describe what Cosmos declares and tests in CI. The table
+below reflects targeted, empirical verification instead: each Cosmos release
+from 1.5.1 to 1.15.0 was ``pip install``-ed into the official Astronomer
+Runtime Docker image for each Runtime series, then checked by running
+``airflow plugins`` — the same plugin-loading path Airflow itself uses at
```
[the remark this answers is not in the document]
pankajastro: Resolved as a side effect of ca97e8a1 — that whole sentence was removed when the intro paragraph was simplified.

PR #2917, opened 2026-07-24
Fix watcher fallback silently dropping `--full-refresh` overrides

### LINKED ISSUE #2908: [Bug] WATCHER_KUBERNETES: consumer fallback drops --full-refresh (reads producer's unrendered add_cmd_flags)
### Astronomer Cosmos Version

1.15.0 (latest release; verified the affected code path is unchanged on `main`)

dbt-core 1.11.12

### Versions of dbt adapters

dbt-athena 1.10.2

### LoadMode

DBT_MANIFEST

### ExecutionMode

WATCHER_KUBERNETES

### InvocationMode

SUBPROCESS (producer pod runs `dbt build --log-format json`, parsed by the watcher)

### airflow version

3.2.1

### Operating System

Amazon Linux 2 (AWS MWAA 3.2.1 workers); dbt runs in EKS pods via KubernetesPodOperator

### Deployment

Amazon (AWS) MWAA

### Deployment details

MWAA 3.2.1, Python 3.12, `mw1.small`. `DbtTaskGroup` with `ExecutionMode.WATCHER_KUBERNETES`; `full_refresh` is driven by an Airflow DAG Param, so the operators receive `full_refresh="{{ params.full_refresh }}"` (templated).

### What happened?

On a `full_refresh=true` run, when the watcher **producer** dies mid-build (in our case its pod was evicted before flushing some nodes' status XComs), the affected **consumer sensors** fall back to running the model in their own pod via `_fallback_to_non_watcher_run`. **That fallback pod silently drops `--full-refresh` and runs the model incrementally**, even though the run is `full_refresh=true`.

The run then reports success, but the fallen-back incremental models were **not** rebuilt — a silent correctness gap on a full-refresh run.

**Root cause.** `_fallback_to_non_watcher_run` (`cosmos/operators/_watcher/base.py`, `main` lines 521–526) builds the fallback command from the **producer** task object, fetched from the DAG structure:

```python
upstream_task = context["ti"].task.dag.get_task(self.producer_task_id)
extra_flags = []
if upstream_task and hasattr(upstream_task, "add_cmd_flags"):
    raw_flags = upstream_task.add_cmd_flags()
    extra_flags = self._filter_flags(raw_flags)
```

`dag.get_task()` returns the producer operator from the DAG definition, whose template fields are **not rendered** for this consumer's task instance. So `upstream_task.full_refresh` is the literal string `"{{ params.full_refresh }}"`. In `add_cmd_flags()` (`cosmos/operators/base.py`) that goes through `to_boolean(self.full_refresh)`, and `to_boolean("{{ params.full_refresh }}")` returns `False`, so `--full-refresh` is never emitted. It is **not** stripped by `_filter_flags` (`_PRODUCER_ONLY_FLAGS` is only `--select`/`--exclude`/`--log-format`) — it is never generated.

The same mechanism drops **any** templated value surfaced through `add_cmd_flags` (e.g. `--vars`), not just `--full-refresh`.

**Expected:** the fallback should apply the run-time-rendered `full_refresh` (and other templated flags) — e.g. read them from the consumer's own rendered task instance, or render the producer's fields against the current context — so a fallback on a `full_refresh=true` run rebuilds the model.

### Relevant log output

```shell
# Consumer sensor fallback on a full_refresh=true DAG run:
Falling back to running model 'model.my_project.my_incremental_model' from project '/dbt' using DbtRunWatcherKubernetesOperator
Running command: ['dbt', 'run', '--select', 'fqn:...my_incremental_model', '--select', 'my_incremental_model', '--profile', '...', '--target', '...', '--project-dir', '/dbt']
# note: NO --full-refresh, despite params.full_refresh == True
1 of 1 START sql incremental model ...my_incremental_model ... [RUN]
1 of 1 OK created sql incremental model ...my_incremental_model ... [OK ... ]
```

### How to reproduce

1. Build a `DbtDag`/`DbtTaskGroup` with `ExecutionMode.WATCHER_KUBERNETES` containing at least one incremental model.
2. Drive `full_refresh` from an Airflow Param so the operators get `full_refresh="{{ params.full_refresh }}"` (templated), and trigger with `full_refresh=true`.
3. Cause the producer to terminate before it flushes that model's status XCom (e.g. kill/evict the producer pod mid-build).
4. The model's consumer sensor falls back to its own pod. Observe the launched command: it is `dbt run --select <model>` **without** `--full-refresh`, and dbt runs the model incrementally instead of rebuilding it.

Expected: the fallback runs `dbt run --select <model> --full-refresh`.

### Anything else :)?

Occurs every time the producer dies mid-build on a templated `full_refresh=true` run, for each model that hasn't yet reported a status XCom (so it must fall back). It does not occur when the producer completes normally (no fallback), nor when `full_refresh` is passed as a static `True` boolean (then the un-rendered producer attribute is already truthy). It is specific to **templated** `add_cmd_flags` values reaching the fallback via the un-rendered producer task.

pankajastro:
## Summary

Fixes #2908.

`_fallback_to_non_watcher_run` (`cosmos/operators/_watcher/base.py`) is shared by all three watcher execution modes (`WATCHER`, `WATCHER_KUBERNETES`, `WATCHER_GCP_GKE`). It had two related bugs affecting the flags used when a consumer sensor falls back to running its model directly:

1. **Original bug.** It built the fallback command's extra flags from the **producer** task object, fetched via `context["ti"].task.dag.get_task(self.producer_task_id)`. That object is the producer as defined in the DAG — Airflow never renders its template fields for this run, since it isn't the executing task instance. So a templated `full_refresh` (e.g. `full_refresh="{{ params.full_refresh }}"`, the common pattern for driving full-refresh from an Airflow DAG Param) stayed a literal Jinja string on that object. dbt's `to_boolean()` treats that string as falsy, so `--full-refresh` was silently never added to the fallback command — even on a `full_refresh=true` run. The model then ran (and reported success) incrementally instead of being rebuilt.
2. **Regression from the first fix.** Reading flags from `self.add_cmd_flags()` (the consumer's own rendered instance) instead of the producer fixed (1), but silently discarded any producer-only override supplied via `ExecutionConfig.setup_operator_args` — e.g. a `full_refresh=True` set only on the producer, while the consumer's own `full_refresh` stays at its default `False`. That reintroduces the same class of bug from the other direction.

## Fix

The producer (local, `WATCHER_KUBERNETES`, and `WATCHER_GCP_GKE` variants) now publishes its own already-rendered `add_cmd_flags()` output to XCom (`producer_cmd_flags` key) as it starts. `_fallback_to_non_watcher_run` reads that first, so `setup_operator_args` overrides survive and the flags always reflect the producer's actual rendered state for this run — falling back to the consumer's own `add_cmd_flags()` only if the producer never published (e.g. it errored before reaching that point). Also kept tolerant of custom consumer sensors that don't define `add_cmd_flags` at all, falling back to `[]` instead of raising `AttributeError`.

## Test plan

- Producer publishes its rendered flags to XCom: `test_dbt_producer_watcher_operator_publishes_own_cmd_flags_to_xcom` (local) and `test_execute_watcher_producer_publishes_own_cmd_flags_to_xcom` (shared `WATCHER_KUBERNETES`/`WATCHER_GCP_GKE` path).
- `test_fallback_prefers_producer_published_flags`: consumer's own `full_refresh=False` but the producer published `--full-refresh` (simulating a `setup_operator_args` override) — asserts the producer's flags win.
- `test_fallback_uses_own_rendered_full_refresh_when_producer_xcom_missing`: producer never published (XCom missing) — asserts fallback to the consumer's own rendered `full_refresh`.
- Updated the existing filter / `dbt_cmd_flags` opt-in / versioned-selector fallback tests to the new producer-XCom-first call pattern.
- Ran the full `test_watcher.py`, `test_k8s_common.py`, `_watcher/`, and `test_watcher_kubernetes_unit.py` suites (399 tests) — all passing.
- `pre-commit run --all-files` (ruff, black, mypy) clean.


cosmos/operators/_watcher/base.py
```diff
@@ -518,12 +518,10 @@ def _fallback_to_non_watcher_run(self, try_number: int, context: Context) -> boo
                 self.__class__.__name__,
             )
 
-        upstream_task = context["ti"].task.dag.get_task(self.producer_task_id)
-
-        extra_flags: list[str] = []
-        if upstream_task and hasattr(upstream_task, "add_cmd_flags"):
-            raw_flags = upstream_task.add_cmd_flags()
-            extra_flags = self._filter_flags(raw_flags)
```
tatiana, line 526: By removing this logic, we're introducing a regression, since producer-only overrides are. lost. I think this flags that we need to have better test coverage for this feature.

tatiana (commented):
@pankajastro `setup_operator_args` can intentionally override producer parameters without changing the consumer’s `operator_args`. For example, with consumer `full_refresh=False` and producer `full_refresh=True`, the producer runs a full refresh, but these lines now discard that effective producer flag and the fallback silently runs incrementally.

Could we preserve producer-specific overrides while resolving their templates against the current context-perhaps by rendering a copy of the producer or propagating its effective fallback flags-and add coverage for differing producer/consumer values?

cosmos/operators/_watcher/base.py
pankajastro: Good catch — fixed in c2215eeb. The producer now publishes its own already-rendered `add_cmd_flags()` to XCom as it starts (`producer_cmd_flags` key), and the consumer's fallback reads that first, falling back to its own `add_cmd_flags()` only if the producer never published (e.g. it errored before that point). This preserves `setup_operator_args` overrides (e.g. a producer-only `full_refresh`) while still fixing the original unrendered-`dag.get_task()` bug, since the producer now always reports its own rendered state instead of a clone being re-fetched later. Added coverage for: producer publishing flags (local + K8s/GKE), fallback preferring the producer's published flags over the consumer's own, and fallback-to-consumer when the producer's XCom is missing.

tatiana (approved):

PR #2918, opened 2026-07-24
Honor `on_warning_callback` passed via `operator_args`

### LINKED ISSUE #2869: `on_warning_callback` silently ignored when passed via `operator_args` (should warn or honor it)
## Problem

## Root cause

```python
task_args = dict(task_args)
```

## Reproduction

```python

```

## Impact

## Proposed improvement

Sanjays2402:
## Description

`create_test_task_metadata` unconditionally overwrote `task_args["on_warning_callback"]` with the top-level argument, which defaults to `None`, so a callback passed through `operator_args` was silently dropped and never fired. It now falls back to the `task_args` value when the top-level argument is unset — mirroring how the sibling `exclude` argument already preserves operator-supplied values. An explicit top-level callback still wins.

## Related Issue(s)

Closes #2869

## Breaking Change?

No.

## Checklist


tatiana (commented):
Hi @Sanjays2402 thanks a lot for fixing this. The change is simple, effective and the test is representative - I confirmed it fails without the fix.

The CI issues seem to be unrelated to this change, I'll fix them, and it would be great if you could rebase once they are merged:
- #2933

I think the fix only covers part of the problem. I believe the same issue will happne in two parts that the PR doesn't touch:
- `TestBehavior.BUILD `
- `DbtResourceType.SOURCE`

So someone on build, or with source tests, who passes `on_warning_callback` through operator_args would still lose it. I don't think #2869 is fully closed yet.

Is there any chance you could isolate the fix into a function, `_honor_on_warning_callback`, and use it in the three places?

I noticed this and PR #2920 attempt to solve the problem with the same implementation. I'll be merging the one that addresses the feedback before, and makes all the CI checks pass.

Sanjays2402:
Done, pushed as f7f1ade (also merged main now that #2933 is in).

Extracted `_honor_on_warning_callback(args, on_warning_callback)` and applied it in the three places: `create_test_task_metadata`, the `TestBehavior.BUILD` branch, and the `DbtResourceType.SOURCE` branch. You were right that both of the latter dropped it — under BUILD the tests run inline with `dbt build` so there's no separate test task to carry the callback, and sources assign it before the freshness/test rendering.

One behaviour change while isolating it: the helper checks `on_warning_callback is not None` rather than truthiness, so an explicitly passed falsy top-level value isn't treated as unset (Copilot's note on the earlier revision).

Added `test_create_task_metadata_build_mode_preserves_task_args_on_warning_callback` and `test_create_task_metadata_source_preserves_task_args_on_warning_callback`; both fail on the previous commit and pass with the helper. `tests/airflow/test_graph.py` is 152 passed locally — the 3 `test_test_behavior_for_watcher_kubernetes_mode` failures are pre-existing on this branch and on main here (no kubernetes provider in my env).

Happy for you to take #2920 instead if it lands cleaner, no objection either way.

tatiana:
Closing in favour of #2940, crediting you as a co-author

PR #2920, opened 2026-07-27
fix: honor `on_warning_callback` passed via `operator_args`

### LINKED ISSUE #2869: `on_warning_callback` silently ignored when passed via `operator_args` (should warn or honor it)

lowbyteguy:
## Description

`create_test_task_metadata` copied `task_args` and then unconditionally overwrote `on_warning_callback` with the top-level argument. That argument defaults to `None`, so a callback supplied through `operator_args` (which arrives inside `task_args`) was always clobbered and never attached to the test task.

The top-level value now takes precedence only when it is actually set, which mirrors how `RenderConfig.exclude` already avoids clobbering an `exclude` coming from `operator_args`. Two tests were added next to the existing `create_test_task_metadata` tests: one fails without the fix (callback dropped), and one guards that an explicit top-level callback still wins.

## Related Issue(s)

closes #2869

## Breaking Change?

No. Behaviour only changes when `on_warning_callback` is present in `operator_args` and the top-level argument is unset — previously a silent no-op.

## Checklist


tatiana (commented):
Hi @lowbyteguy thanks a lot for fixing this. The change is simple, effective and the test is representative - I confirmed it fails without the fix.

The CI issues seem to be unrelated to this change; I'll fix them, and it would be great if you could rebase once they are merged:
- #2933

I think the fix only covers part of the problem. I believe the same issue will happen in two parts that the PR doesn't touch:
- `TestBehavior.BUILD `
- `DbtResourceType.SOURCE`

So someone on build, or with source tests, who passes `on_warning_callback` through operator_args would still lose it. I don't think #2869 is fully closed yet.

Is there any chance you could isolate the fix into a function, `_honor_on_warning_callback`, and use it in the three places?

I noticed this, and PR #2918 attempts to solve the problem with the same implementation. I'll be merging the one that addresses the feedback and makes all the CI checks pass.

tatiana:
Closing in favour of #2940, crediting you as a co-author

PR #2921, opened 2026-07-27
Add DAG-bundle hint when a dbt project path is missing

pankajkoti:
## Description

On Airflow 3, DAGs can run from a **versioned DAG bundle** (e.g. an Astronomer deployment or a git-based bundle) whose filesystem location differs between DAG parsing and task execution. A hardcoded **absolute** `dbt_project_path` that resolves at parse time can therefore be missing on the worker, surfacing as a bare `FileNotFoundError` that is hard for support/CRE to triage from logs alone.

This PR enriches the two places the missing-project error is raised so the log is self-explanatory:

- `ProjectConfig.validate_project()` and `create_symlinks()` now append a hint — **on Airflow 3, for an absolute path** — telling the user to set `dbt_project_path` relative to their DAG file, e.g. `(Path(__file__).parent / "dbt/my_dbt_project").absolute().as_posix()`, linking the Astro getting-started docs.
- `create_symlinks()` now raises `CosmosValueError` instead of a bare `FileNotFoundError` when the project directory does not exist.

## Why absolute-path detection alone isn't enough

The officially recommended pattern (`(Path(__file__).parent / …).absolute()`) is itself an absolute path, so a proactive "you used an absolute path" warning would false-positive on the correct usage. The reliable signal is the actual "path not found" failure, which is what this PR enriches.

## Detecting Astronomer

When running on Astronomer, the hint is **definitive** ("you are running on Astronomer … the versioned bundle relocates the project"); otherwise it stays **conditional** so it remains accurate for non-versioned bundles (e.g. the default `dags-folder` bundle, which does not relocate).

Astronomer is detected via `settings.in_astro`, which keys on the **presence** of the `ASTRONOMER_ENVIRONMENT` variable rather than its value. The Astronomer Runtime sets it on every Astro image (`cloud` on hosted deployments, `local` for local dev) and it is present on workers, where the runtime failure occurs. It is an *undocumented* platform variable, so keying on presence — not on `"cloud"` — keeps the check working if Astronomer introduces new values, while a generic Airflow-3 fallback still fires when it is absent.

To avoid a circular import (`cosmos.config` is imported very early via the plugin/listeners, and a module-level import of `cosmos.dbt.project` pulls in a partially-initialised `cosmos.settings`), `validate_project()` imports the hint helper lazily.

## Testing

- New unit tests for the hint gating (Airflow 3 vs 2, absolute vs relative vs `None`) and the incisive-vs-conditional wording, plus `create_symlinks` raising `CosmosValueError` with the hint.
- Updated `test_validate_project_missing_fails` to be version-aware.
- Verified: AF2 (`py3.11-2.10-1.9`) full unit suite green; AF3 (`py3.11-3.0-1.9`) touched files green; mypy, pre-commit, and docs build all pass.
- Live-verified in cosmos-demo on Astro Runtime 3.2 (`ASTRONOMER_ENVIRONMENT=local`, so `in_astro=True`): a DAG with a hardcoded absolute `dbt_project_path` fails to parse with the incisive message, while a DAG deriving the path from `Path(__file__)` parses cleanly.

## Related issues

Closes astronomer/oss-integrations-private#418

Closes BOSS-382 — https://linear.app/astronomer/issue/BOSS-382/improved-cosmos-logging-for-common-failure-modes


tatiana:
@pankajkoti  on this one, please check on Astro deployments which envvar we set that would allow us to identify it is an Astro deployment and be more incisive about the path

pankajkoti:
> @pankajkoti  on this one, please check on Astro deployments which envvar we set that would allow us to identify it is an Astro deployment and be more incisive about the path

I thought about this but other thought I had is what if the Astro env var changes later and so thought to keep it generic. But happy to have a hard check if that's the suggestion @tatiana

tatiana (approved):
Thanks for implementing this, @pankajkoti ! It will be very helpful to guide users who may see this common error, particular for Astro customers using Airflow 3

PR #2930, opened 2026-07-28
Speed up integration tests: parse only the example DAG under test

tatiana:
## Summary

`test_example_dag` parametrized over dag_ids, which forced building the **full** DagBag — rendering every example dbt project (some via `dbt ls`) — at collection, even though each `pytest-split` job runs only ~1/3 of the DAGs. This parametrizes over **DAG files** and loads only the file under test, so each job parses only the DAGs it actually runs.

**Flag-controlled** (so we can switch back when Airflow parsing improves): `TEST_ISOLATE_EXAMPLE_DAGS`, default **on**. Set `TEST_ISOLATE_EXAMPLE_DAGS=false` to restore the old full-parse path.

## Verified locally — py3.10 / Airflow 3.0 / dbt 1.12, real Postgres
- **Collection: 0.29s with no DagBag parse** (isolate) vs a full render of *all* example DAGs (legacy).
- `basic_cosmos_dag` + `example_cosmos_sources` → **2 passed**, unchanged behaviour.
- Bonus robustness: a DAG that can't import (e.g. the S3-backed `manifest_example` without creds) now fails only *its own* test instead of breaking the whole collection.

## Notes
- Same test set runs: `-k` filters in `integration.sh` keep working (their names are substrings of the filenames); version-gated/IGNORED files are excluded from the parametrize.
- `.test_durations` **updated** in this PR: keys remapped dag_id → filename (files with >1 DAG summed) from the existing CI-measured times. Approximate — CI's next `--store-durations` will refine (e.g. add per-file parse overhead).
- Files defining >1 DAG now run their DAGs within one test (coverage unchanged).

Draft until we confirm the desired behaviour + CI timing.


tatiana:
Before vs after (Run-Integration-Tests main matrix, same filter, per-job total time):

```
┌──────────────────────────┬──────┬─────┬──────────┬──────┬───────────────┐
│                          │ jobs │ min │  median  │ max  │ total compute │
├──────────────────────────┼──────┼─────┼──────────┼──────┼───────────────┤
│ Before (main, no change) │ 78   │ 5.8 │ 14.3 min │ 20.6 │ 17.5 h        │
├──────────────────────────┼──────┼─────┼──────────┼──────┼───────────────┤
│ After (#2930)            │ 77   │ 5.2 │ 14.5 min │ 21.2 │ 17.2 h        │
└──────────────────────────┴──────┴─────┴──────────┴──────┴───────────────┘
```
The medians are identical within noise (14.3 → 14.5; the latter is even marginally higher). Total compute is flat. So as a speed change, this is not good enough — it produced no measurable per-job speedup.

PR #2933, opened 2026-07-28
Reduce `create_test_task_metadata` complexity

tatiana:
## Summary

Extract the WATCHER-aware test operator-class resolution out of `create_test_task_metadata` into a small `_calculate_test_operator_class` helper, to reduce the function's cyclomatic complexity.

## Why

`create_test_task_metadata` currently sits at complexity **exactly 10**, which is the ruff `C901` ceiling (`mccabe.max-complexity = 10`). It passes on `main` today, but it has **zero headroom**: any added branch tips it to 11 and fails CI. This already bit #2920, whose one-line `on_warning_callback` fix trips `C901` (11 > 10) purely because the function was at the limit.

Rather than `# noqa` the symptom, this gives the function room to breathe.

## Change

Pure, behavior-preserving extraction:
- New module-level `_WATCHER_TO_FALLBACK_EXECUTION_MODE ` map + `_calculate_test_operator_class(execution_mode, render_config)` helper holding the exact same AFTER_ALL / WATCHER* logic (the `in _WATCHER_TO_FALLBACK_EXECUTION_MODE ` membership check is the same set as the original tuple).
- `create_test_task_metadata` now calls the helper.

Result: `create_test_task_metadata` drops from 10 to ~7; ruff `C901`, `ruff check`, and `ruff format --check` all pass. No behavior change, so the existing `create_test_task_metadata` tests cover it.

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -192,6 +192,28 @@ def _override_profile_if_needed(task_kwargs: dict[str, Any], profile_kwargs_over
         task_kwargs["profile_config"] = modified_profile_config
 
 
+_WATCHER_TO_TEST_EXECUTION_MODE = {
+    ExecutionMode.WATCHER: ExecutionMode.LOCAL,
+    ExecutionMode.WATCHER_KUBERNETES: ExecutionMode.KUBERNETES,
+    ExecutionMode.WATCHER_GCP_GKE: ExecutionMode.GCP_GKE,
+}
+
+
+def _calculate_test_operator_class(execution_mode: ExecutionMode, render_config: RenderConfig | None) -> str:
+    """Resolve the operator class path for a dbt test task.
+
+    For an ``AFTER_ALL`` test running under a ``WATCHER*`` execution mode, the test itself runs in the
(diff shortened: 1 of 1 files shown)


cosmos/airflow/graph.py
```diff
@@ -192,6 +192,28 @@ def _override_profile_if_needed(task_kwargs: dict[str, Any], profile_kwargs_over
         task_kwargs["profile_config"] = modified_profile_config
 
 
+_WATCHER_TO_TEST_EXECUTION_MODE = {
```
tatiana: Need to change in the places that use this const:
```suggestion
_WATCHER_TO_FALLBACK_EXECUTION_MODE = {
```

tatiana: Done in fc912c8 — renamed the constant to `_WATCHER_TO_FALLBACK_EXECUTION_MODE` and updated both usages in the helper.

pankajkoti (approved):
LGTM. The PR description seems to be stale, we now seem to have `_WATCHER_TO_FALLBACK_EXECUTION_MODE` in the diff than the metioned `_WATCHER_TO_TEST_EXECUTION_MODE` in ther PR description. Would be nice to update the PR description before merging!

tatiana:
Thanks, @pankajkoti , addressed!

PR #2939, opened 2026-07-30
Raise minimum supported dbt-core version to 1.8

### LINKED ISSUE #2902: Raise the minimum supported dbt-core version
## Context

While working on #2773 (adding dbt-core 1.12 to the test matrix), we hit a real dependency conflict trying to make CI's pre-install step fail loudly (instead of silently substituting a fallback lockfile) whenever a specific `(airflow, dbt)` pair has no pinned lockfile.

Testing `airflow=3.0, dbt=1.5` (the combo `Run-Integration-Tests-DBT-1-5-4` exercises) directly against `uv`'s strict resolver:

```
And because apache-airflow-core==3.0.0 depends on jinja2>=3.1.5 and
apache-airflow==3.0.0 depends on apache-airflow-core==3.0.0, we can
conclude that apache-airflow==3.0.0 and dbt-duckdb>=1.5.0,<=1.5.2 are
incompatible.
And because you require dbt-duckdb>=1.5,<1.6 and apache-airflow==3.0, we
can conclude that your requirements are unsatisfiable.
```

dbt-core 1.5 (released 2023) and modern Airflow (3.0+) don't have a mutually-compatible `jinja2` under a strict resolver. This isn't a pinning bug — it's a genuine conflict in the package universe. It's presumably why `scripts/test/integration-dbt-1-5-4.sh` (and the other `*-setup.sh` re-pin scripts for old dbt minors) use plain `pip install` rather than `uv pip install`: pip's legacy resolver doesn't hard-block on full dependency-graph consistency the way `uv` does, so it can land in a state that happens to work even though `uv` calls it unsatisfiable.

## Proposal

Raise `astronomer-cosmos`'s minimum supported dbt-core version (currently 1.5, per `pyproject.toml`'s hatch test matrix and `docs/policy/compatibility-policy.rst`), dropping the oldest, most conflict-prone dbt minors. This would:

- Let `scripts/test/pre-install-airflow.sh` fail loudly on any missing `(airflow, dbt)` lockfile instead of silently falling back to a dbt-1.11 baseline for unpinned combos (see discussion on #2773).
- Remove the need for the permissive-`pip`-based re-pin scripts (`integration-dbt-1-5-4.sh`, parts of `integration-dbt-async.sh`) that exist specifically to work around old-dbt/new-Airflow incompatibilities.
- Reduce the size and maintenance burden of the CI test matrix.

## Open questions

- What's the actual floor we want (1.7? 1.8? 1.9?) — needs a look at real user adoption / support-policy conventions for how long a dbt minor is kept.
- Whether to do this as one bump or a deprecation cycle (warn for N releases, then drop).

cc @tatiana

pankajastro:
## Summary
dbt-core 1.5, 1.6, and 1.7 have reached End of Life per the dbt Labs support policy, and 1.5 specifically has an unresolvable jinja2 dependency conflict with modern Airflow (>=3.0) under a strict resolver. This drops them from the supported/tested range.

- `pyproject.toml`: hatch dbt test matrix now starts at 1.8.
- `.github/workflows/test.yml`: removed the dedicated dbt-1.5.4 job in favor of pinning it to 1.8 (the new floor); dropped 1.6/1.7 from the DBT-Async matrix.
- `scripts/test/integration-dbt-1-5-4.sh` → `integration-dbt-1-8.sh`, repinned to the 1.8 adapter series.
- `scripts/test/integration-dbt-async.sh`: dropped the 1.6-only pydantic/google-cloud-bigquery workarounds.
- `docs/policy/compatibility-policy.rst`: bumped the minimum/supported dbt Core versions in `Currently supported versions`, and dropped the separate EOL callout sentence for 1.5-1.7 so the section reads as what's supported rather than what was dropped.
- `execution-modes-local-conflicts.rst`: updated conflict table to match.
- `watcher-execution-mode.rst`: updated the minimum-dbt-version statement to match, and kept the "older versions are EOL, use `ExecutionMode.LOCAL` instead" note, updated to reference dbt-core 1.8 instead of 1.5.

Why Dbt-core 1.8?

dbt-core 1.8 and Airflow 2.9 were released around the same time.
- dbt-core 1.8.0: released 2024-05-09
- apache-airflow 2.9.0: released 2024-04-08

Success CI: https://github.com/astronomer/astronomer-cosmos/actions/runs/30535323883/job/90847155049?pr=2939. the failed CI is using old configuration.

Fixes #2902.

## Test plan
- [x] CI passes on the updated dbt test matrix (1.8+)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main, bump-min-dbt-core-1.8]
```
```suggestion
    branches: [main]
```

docs/guides/run_dbt/airflow-worker/watcher-execution-mode.rst
```diff
@@ -439,18 +439,19 @@ Known limitations
 Minimum supported dbt version
 +++++++++++++++++++++++++++++
 
-``ExecutionMode.WATCHER`` requires **dbt-core 1.5 or newer**. The watcher
-relies on ``--log-format json`` to consume structured per-model events from
-dbt, with the flag placed after the dbt subcommand (e.g.
-``dbt build --log-format json …``). dbt-core 1.5 introduced finer-grained
-log CLI parameters (see the
+``ExecutionMode.WATCHER`` requires **dbt-core 1.5 or newer**, which Cosmos's
```
pankajkoti: We still say here dbt core 1.5? What do you think about reconciling this with the newer releases, committing to 1.8+

pankajkoti (approved):
LGTM

cc: @tatiana @kaxil @phanikumv @yetudada if you have a reservation

docs/guides/run_dbt/airflow-worker/watcher-execution-mode.rst
pankajastro: Good catch — rather than reconcile the wording, I removed the "Minimum supported dbt version" subsection entirely. It was documenting that `ExecutionMode.WATCHER` needs dbt-core 1.5+, which was a real limitation back when Cosmos's own floor was lower than 1.5. Now that this PR raises Cosmos's global minimum to 1.8 (above 1.5), it's no longer a watcher-specific limitation — just the general compatibility policy repeated, so the section was redundant. Pushed in ec067c39.

PR #2940, opened 2026-07-30
fix: Honor `on_warning_callback` passed via `operator_args`

### LINKED ISSUE #2869: `on_warning_callback` silently ignored when passed via `operator_args` (should warn or honor it)

tatiana:
## Description

`on_warning_callback` is a first-class `DbtDag` / `DbtTaskGroup` argument, but `operator_args` is also forwarded to the underlying operators, so passing it there is a reasonable mental model. It was silently dropped: each of the three sites that wires the callback copies `task_args` (which already holds anything from `operator_args`) and then unconditionally overwrites the callback with the top-level value, which defaults to `None`.

This isolates the resolution into `_honor_on_warning_callback` and applies it at all three sites, so the top-level argument still wins when set, but an `operator_args` callback is preserved when it is not:

- `create_test_task_metadata`, for `dbt test` tasks
- `create_task_metadata`, `TestBehavior.BUILD` branch, for `dbt build` tasks
- `create_task_metadata`, `DbtResourceType.SOURCE` branch, for `dbt source freshness` tasks

All three were confirmed broken by rendering real DAGs and inspecting `on_warning_callback` on the instantiated operators, not by reading the code. The key is always set, so an operator is still handed an explicit `None` when neither form supplies a callback.

This supersedes #2918 and #2920, which each fixed the first site with an equivalent implementation. Both are credited as co-authors.

## Tests

- Per-site regression tests for all three sites. Each was verified to fail without the fix (reverting only the three call sites, keeping the helper, yields exactly one failure per site).
- A precedence test (the top-level argument wins when both are given) and a parametrised unit test for the helper, including the "neither set" case that pins the always-set-the-key invariant.
- An end-to-end `DbtDag` test over `AFTER_EACH` / `BUILD` / `AFTER_ALL`, asserting every rendered operator that supports the callback actually receives it.

`tests/airflow/test_graph.py` 159 passed; the new `tests/test_converter.py` test 3 passed; `tests/airflow/ tests/test_converter.py tests/dbt/test_pruning.py` 229 passed with 3 pre-existing failures that also fail on `main` (missing `airflow.providers.docker` locally). All 24 pre-commit hooks pass, including mypy. `sphinx-build -W` succeeds.

## Docs

Documents that both forms work and which one wins, in `testing-behavior.rst` and `managing-sources.rst`. #2869 asked for a docs note either way.

## Not covered here

Found while verifying, all pre-existing and out of scope for this issue, worth separate issues:

- The watcher test sensors (`DbtTestWatcherOperator` and its Kubernetes/GKE variants) have no `on_warning_callback` support at all, so even the top-level argument is swallowed.
- `DbtBuildKubernetesOperator` / `DbtBuildGcpGkeOperator` likewise swallow it, as they are not `DbtWarning*Operator` subclasses.
- Custom `render_config.node_converters` still receive the unresolved top-level value.

## Related Issue(s)

Closes #2869

## Breaking Change?

No. Behaviour only changes when `on_warning_callback` is present in `operator_args` and the top-level argument is unset, which was previously a silent no-op.

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -176,6 +176,22 @@ def forward_render_exclude_to_test(task_args: dict[str, Any], render_config: Ren
     task_args["exclude"] = _convert_list_to_str(merged)
 
 
+def _honor_on_warning_callback(task_args: dict[str, Any], on_warning_callback: Callable[..., Any] | None) -> None:
+    """
+    Resolve the effective ``on_warning_callback`` for a node command that supports it, in-place.
+
+    The top-level ``DbtDag`` / ``DbtTaskGroup`` / ``DbtToAirflowConverter`` argument takes precedence. When it is
+    unset (``None``), a callback the user passed through ``operator_args`` -- which reaches us already inside
+    ``task_args`` -- is preserved instead of being silently dropped.
+
+    The key is always set, so the operator is handed an explicit ``None`` when neither source supplies a callback.
+    See https://github.com/astronomer/astronomer-cosmos/issues/2869.
+    """
+    if on_warning_callback is None:
+        on_warning_callback = task_args.get("on_warning_callback")
+    task_args["on_warning_callback"] = on_warning_callback
+
+
 def _override_profile_if_needed(task_kwargs: dict[str, Any], profile_kwargs_override: dict[str, Any]) -> None:
     """
     Changes in-place the profile configuration if it needs to be overridden.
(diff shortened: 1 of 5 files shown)


pankajastro (approved):
Question: #2918 and #2920 are open PRs from first-time contributors targeting the same issue, and this PR credits both as co-authors. If this bug isn't urgent, would it make sense to give them a chance to  extend their PRs to the remaining sites before we merge this instead? Happy to proceed here if there's a reason it needs to land now.

tatiana:
Thanks @pankajastro, I agree - and I've always given the community time to solve things. In this particular case, they had one week to improve after I gave the feedback, and since we want to ship this in 1.15.1, I made the decision to iterate with AI for 30min to fix it

Issue #2857, opened 2026-06-30
[Bug] Cosmos re-parses full dbt manifest on every worker pod in Airflow 3.0 with KubernetesExecutor


juraj-kotisa:
### Astronomer Cosmos Version

1.14.0

### dbt Core or Fusion version

1.11.8

### Versions of dbt adapters

dbt-databricks==1.11.7

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

LOCAL

### InvocationMode

DBT_RUNNER

### airflow version

3.2.0

### Operating System

Linux (AKS — Azure Kubernetes Service, Ubuntu-based worker pods)

### Deployment

Official Apache Airflow Helm Chart

### Deployment details

- Helm chart: 1.18.0
- Executor: KubernetesExecutor
- Every task runs in a fresh pod (GITSYNC_ONE_TIME=true)
- DAG files are cloned via git-sync init container on every worker pod
- 36 Cosmos DbtTaskGroups in a single DAG
- dbt project: 3,238 nodes, 822 models
- Cosmos cache_dir configured at /opt/airflow/logs/cosmos-cache (shared PVC)
- Python 3.13, PostgreSQL 16.13

### What happened?

After upgrading from Airflow 2.11 to Airflow 3.0, every worker pod re-parses the entire 
dbt manifest for all 36 Cosmos task groups before executing any task — even a task that 
only belongs to one group and runs for 3 seconds.

Evidence from worker pod log:
- DAG file is loaded at task start: "Filling up the DagBag from ingestion_dag.py"
- All 36 task groups are parsed sequentially, each calling LoadMode.DBT_MANIFEST
- Total Cosmos parsing overhead per task: ~20 seconds
- dbt then performs a full project parse (3,238 nodes): ~43 seconds
- Actual dbt model execution: ~3 seconds

Total task duration: ~90 seconds. Actual useful work: 3 seconds.

In Airflow 2.11 this did not happen — workers received a serialized task instance 
and did not re-parse the full DAG file. After the upgrade to Airflow 3.0, the new 
Task Execution API requires workers to load the full DAG, triggering Cosmos to 
initialize all task groups on every single task execution.

We are aware that cache_dir caches the dbt connection profile. However it does NOT 
cache the manifest parsing itself across worker pods. Each pod still parses all 36 
groups from scratch.

Expected behaviour: Cosmos should parse only the task groups needed for the specific 
[...]

PR #2942, opened 2026-08-01
Skip generated dirs when hashing the dbt project folder

goingforstudying-ctrl:
## Description

I ran into #2857 while looking at DAG parse times on a network-backed dbt project folder, and the hash walk turned out to be the culprit pankajastro suspected. `_create_folder_version_hash` does an unpruned `os.walk` over the whole project and reads every file in full — including `target/`, `dbt_packages/`, `logs/` and `.git/`. With `enable_dag_versioning` on (the Airflow 3 default) that happens on every DAG parse, so on something like Azure File Share you're re-reading thousands of generated files per parse for no reason.

This prunes those directory names out of the walk. The list defaults to `.git, target, dbt_packages, logs` and is configurable through a new `[cosmos] project_hash_excluded_dirs` setting for projects that rename their target/packages paths in dbt_project.yml. Files are now also read in 1MB chunks instead of one shot, which matters if something big survives the pruning.

While in there I noticed the hash only covers file contents, so renaming a model file without touching its SQL produces the same hash — but dbt derives node names from file names, so the project has meaningfully changed and the partial-parse cache should be invalidated. I mixed each file's relative path into the hash to catch that. Happy to split that part out if you'd rather keep this PR to the pruning only.

Also added the parse-time log line showing the resolved cache dir that was mentioned in the issue, so an env var silently overriding the config block shows up in the logs.

Benchmark on a synthetic project (822 models, ~3.5k generated files under target/dbt_packages/logs): 1179ms -> 204ms per hash on local disk. The gap should be a lot wider on network storage where per-file reads dominate.

One heads-up: the hash values change with this (paths mixed in, generated dirs excluded), so on upgrade each project will look "modified" once and the affected caches refresh once. Seemed acceptable but worth calling out.

## Related Issue(s)

Refs #2857

## Breaking Change?

Not breaking, but the project content hash changes once on upgrade (see above), which triggers a one-time cache refresh and one new DAG version entry on Airflow 3.

## Checklist

Generated with an AI coding agent


cosmos/settings.py
```diff
@@ -15,13 +15,22 @@
...
 cache_dir: Path = Path(conf.get("cosmos", "cache_dir", fallback=DEFAULT_CACHE_DIR) or DEFAULT_CACHE_DIR)
 enable_cache: bool = conf.getboolean("cosmos", "enable_cache", fallback=True)
 enable_dag_versioning = conf.getboolean("cosmos", "enable_dag_versioning", fallback=True)
+# Directory names pruned from the dbt project content hash walk (cosmos/versioning.py). Defaults to
+# dbt/VCS-generated folders; setting this replaces the default entirely. See #2857.
+_dag_versioning_excluded_dirs_conf = conf.get("cosmos", "project_hash_excluded_dirs", fallback="")
+project_hash_excluded_dirs: frozenset[str] = (
+    frozenset(dirname.strip() for dirname in _dag_versioning_excluded_dirs_conf.split(",") if dirname.strip())
+    if _dag_versioning_excluded_dirs_conf.strip()
+    else DEFAULT_PROJECT_HASH_EXCLUDED_DIRS
```
pankajastro, line 32: With `LoadMode.DBT_MANIFEST`, `manifest.json` under `target/` can be the sole source of truth for the rendered DAG (see `dev/dags/cosmos_manifest_example.py`, which sets `dbt_project_path=.../jaffle_shop` and `manifest_path=.../jaffle_shop/target/manifest.json`). Since `target/` is pruned by default, a regenerated manifest with no other file changes won't change the project hash computed in `_add_dbt_project_hash_to_dag_docs` (`converter.py`), so Airflow 3 DAG-versioning won't detect it. Could this exempt `manifest_path` from pruning (hash it explicitly) when manifest mode is active?

cosmos/settings.py
```diff
@@ -15,13 +15,22 @@
...
 )
 
 # In MacOS users may want to set the envvar `TMPDIR` if they do not want the value of the temp directory to change
 DEFAULT_CACHE_DIR = Path(tempfile.gettempdir(), DEFAULT_COSMOS_CACHE_DIR_NAME)
 cache_dir: Path = Path(conf.get("cosmos", "cache_dir", fallback=DEFAULT_CACHE_DIR) or DEFAULT_CACHE_DIR)
 enable_cache: bool = conf.getboolean("cosmos", "enable_cache", fallback=True)
 enable_dag_versioning = conf.getboolean("cosmos", "enable_dag_versioning", fallback=True)
+# Directory names pruned from the dbt project content hash walk (cosmos/versioning.py). Defaults to
+# dbt/VCS-generated folders; setting this replaces the default entirely. See #2857.
+_dag_versioning_excluded_dirs_conf = conf.get("cosmos", "project_hash_excluded_dirs", fallback="")
```
pankajastro, line 28: Nit: this var backs `project_hash_excluded_dirs` (also used by the dbt-ls/selector caches, not just DAG versioning) — `_dag_versioning_excluded_dirs_conf` reads as scoped to just one of its consumers. Maybe `_project_hash_excluded_dirs_conf`?

cosmos/settings.py
```diff
@@ -15,13 +15,22 @@
...
+project_hash_excluded_dirs: frozenset[str] = (
```
pankajastro, line 29: This replaces the default list entirely rather than merging with it (documented, but worth double-checking the UX): a user who sets this to add one custom generated dir has to remember to re-list `.git, target, dbt_packages, logs` too, or those silently stop being pruned and start affecting the hash again. Consider making it additive, or logging the effective final list once at startup.

cosmos/versioning.py
```diff
@@ -2,38 +2,63 @@
...
+        excluded_dirs = settings.project_hash_excluded_dirs
+
     hasher = hashlib.md5()
     filepaths = []
+    pruned_dirs = 0
 
     for root_dir, dirs, files in os.walk(dir_path):
+        if excluded_dirs:
+            before = len(dirs)
+            dirs[:] = [dirname for dirname in dirs if dirname not in excluded_dirs]
```
pankajastro, line 46: Prunes by name at any depth, not just at the project root (confirmed intentional by the nested-dir test below). A user-authored directory that happens to be named `logs`/`target`/etc. anywhere in the tree would be silently excluded from the hash too. Probably fine given these are dbt-reserved names, but worth a one-line doc callout that this isn't root-scoped.

cosmos/versioning.py
```diff
@@ -2,38 +2,63 @@
...
         paths = [os.path.join(root_dir, filepath) for filepath in files]
         filepaths.extend(paths)
 
+    if pruned_dirs:
+        logger.debug("Pruned %s excluded directories while hashing the dbt project folder %s", pruned_dirs, dir_path)
+
     for filepath in sorted(filepaths):
+        # Include the path so that renaming a file also changes the hash; dbt derives node
+        # names from file names, so a content-preserving rename still changes the project
+        hasher.update(os.path.relpath(filepath, dir_path).encode())
```
pankajastro, line 57: Nit: no delimiter between the path bytes and the content bytes (or between one file's content and the next file's path) — e.g. path `models/a` + content `bc` hashes identically to path `models/ab` + content `c`. Low practical risk (needs a coincidental rename+edit), but a one-line fix (`hasher.update(b"\0")` as a separator) removes the ambiguity.

Separate nit: `os.path.relpath` uses the OS path separator, so identical project content will now hash differently on Windows vs. POSIX (previously only file content mattered). No Windows CI here so low severity, but `Path(filepath).relative_to(dir_path).as_posix()` avoids it for free.

pankajastro (commented):
Overall approach (pruning generated dirs from the hash walk) makes sense and is well benchmarked. A few things worth resolving before merge, plus some smaller nits inline.

pankajastro (approved):
Approving to merge — the outstanding review comments (manifest-mode DAG-versioning gap, settings naming/semantics, hash-construction nits) are all non-blocking. I'll address them in a follow-up PR crediting this one. Thanks for the contribution!

cosmos/settings.py
pankajastro, line 28: Also, having this configuration would lead to wrong result in dag versioning

PR #2944, opened 2026-08-04
Address #2942 review feedback: manifest-mode hashing, additive excluded dirs, hash construction

pankajastro:
## Description

Follow-up to #2942 (Skip generated dirs when hashing the dbt project folder), addressing the review feedback from https://github.com/astronomer/astronomer-cosmos/pull/2942#pullrequestreview-4843989274 that wasn't resolved before merging.

- **Manifest-mode DAG-versioning gap**: `_add_dbt_project_hash_to_dag_docs` now folds in a content checksum of the manifest file under `LoadMode.DBT_MANIFEST`, via a new standalone `_calculate_manifest_version_hash` (not `ObjectStoragePath.checksum()`, whose fsspec-default metadata — mtime/inode locally, generation/updated on GCS — changes on a byte-identical re-sync). Since `target/` (where the manifest conventionally lives) is pruned from the folder hash walk, a manifest regenerated with no local source changes would otherwise never bump the DAG version hash. The fold-in runs in its own `try`/`except`, so a manifest read failure falls back to the folder hash alone instead of dropping the hash suffix entirely for that parse.
- **`project_hash_excluded_dirs` is now additive**, not a full replacement, on top of the always-excluded `.git`/`target`/`dbt_packages`/`logs`. A user adding one custom generated dir can no longer accidentally stop excluding the defaults.
- **Warns (doesn't block) on suspicious values**: if the configured list includes a dbt project content directory (`models`, `macros`, `seeds`, `snapshots`, `analyses`, `tests`), Cosmos emits a `UserWarning`, since excluding one of those would silently stop DAG versioning / caches from reacting to real project changes.
- **Hash construction**: `_create_folder_version_hash` now separates each file's path from its content with a null byte (previously `path_bytes + content_bytes` had no delimiter, so e.g. path `a` + content `bc` hashed identically to path `ab` + content `c`), hashes the POSIX-form relative path so the digest no longer varies by OS path separator, and sorts files by that same POSIX path so the hash iteration order can't diverge across OSes either.

This changes the resulting hash values again (on top of #2942's own one-time change), so any deployments that already picked up #2942 will see one more DAG-version bump / cache refresh when this lands — same one-time-refresh tradeoff #2942 already called out, not a new concern.

## Design notes

- `_calculate_manifest_version_hash` is a separate function from `_create_folder_version_hash`, not merged into it. An earlier version of this PR did merge them (adding a `manifest_path` parameter to `_create_folder_version_hash`), but that was reverted at @pankajkoti's request — no definite reason required it, and it changed a function signature with no upside.
- The manifest hash is a plain content hash of the file, no more and no less. An earlier version stripped dbt-generated `metadata` fields (`generated_at`, `invocation_id`, ...) before hashing, on the theory that a background recompile could regenerate the manifest with unchanged content. That was dropped: in the intended workflow the manifest is checked into git alongside the DAG code, so any change to it reflects a real, intentional recompile.

## Review feedback addressed

- **Copilot**: OS-dependent hash iteration order, missing delimiter/falsy-checksum handling on the manifest fold-in (now moot — see Design notes), and a docs wording mismatch (`UserWarning` vs. "logs a warning"). Left the `stacklevel` suggestion on the settings warning as-is — it fires from module-level code at import/reload time with no single meaningful caller frame, same as the existing `watcher_dbt_execution_queue` deprecation warning a few lines below it.
- **@kaxil**: `ObjectStoragePath.checksum()` isn't content-derived on every fsspec backend, so a byte-identical manifest re-sync would still bump the DAG version — switched to hashing the manifest's actual content. Also flagged that the manifest fold-in shared a `try` with the folder hash, so a failure there dropped the whole hash suffix rather than just the manifest signal; now isolated in its own `try`/`except`. Docs for `enable_dag_versioning` now mention the manifest fold-in.
- **@pankajkoti**: confirmed `DbtGraph.load()` always resolves `LoadMode.AUTOMATIC` to a concrete mode before this code runs, so no `AUTOMATIC` branch is needed. Renamed `DBT_AUTHORED_CONTENT_DIRS` to `DBT_PROJECT_CONTENT_DIRS` (dbt doesn't author those dirs, users do) and updated the warning/docs wording to match. Pushed back on merging the manifest hash into `_create_folder_version_hash` (see Design notes) — reverted.

## Tests

- New: `test_dag_versioning_hash_folds_in_manifest_checksum_for_manifest_mode`, `test_dag_versioning_hash_survives_manifest_checksum_failure`, `test_calculate_manifest_version_hash_is_content_derived`, `test__create_folder_version_hash_no_path_content_boundary_ambiguity`, `test_project_hash_excluded_dirs_env_var_is_additive` (renamed/updated from the old "replaces default" test), `test_project_hash_excluded_dirs_warns_on_dbt_project_content_dir`.
- Updated expected hash digests in `tests/dbt/test_graph.py` for the new construction.
- `tests/test_versioning.py`, `tests/test_settings.py`, `tests/test_converter.py` all pass, verified under the Airflow 3.2 test matrix (where the manifest-mode tests execute rather than skip).
- `pre-commit` (incl. mypy) and `sphinx-build -W` both clean.

## Related

Follow-up to #2942. Credit to @goingforstudying-ctrl for the original PR.

### DIFF
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -38,6 +38,11 @@ DEFAULT_PROJECT_HASH_EXCLUDED_DIRS = frozenset(
     {".git", DBT_TARGET_DIR_NAME, DBT_DEFAULT_PACKAGES_FOLDER, DBT_LOG_DIR_NAME}
 )
 
+# Conventional dbt resource-content directory names. cosmos.settings warns (but does not block) if a
+# user adds any of these to project_hash_excluded_dirs, since excluding them would silently stop DAG
+# versioning / the partial-parse and dbt-ls caches from reacting to real project changes.
+DBT_AUTHORED_CONTENT_DIRS = frozenset({"models", "seeds", "snapshots", "analyses", "macros", "tests"})
+
 DEFAULT_OPENLINEAGE_NAMESPACE = "cosmos"
 OPENLINEAGE_PRODUCER = "https://github.com/astronomer/astronomer-cosmos/"
 
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -38,9 +38,7 @@ DEFAULT_PROJECT_HASH_EXCLUDED_DIRS = frozenset(
     {".git", DBT_TARGET_DIR_NAME, DBT_DEFAULT_PACKAGES_FOLDER, DBT_LOG_DIR_NAME}
 )
 
-# Conventional dbt resource-content directory names. cosmos.settings warns (but does not block) if a
-# user adds any of these to project_hash_excluded_dirs, since excluding them would silently stop DAG
-# versioning / the partial-parse and dbt-ls caches from reacting to real project changes.
+# Conventional dbt resource-content directory names; see cosmos.settings for how these are used.
 DBT_AUTHORED_CONTENT_DIRS = frozenset({"models", "seeds", "snapshots", "analyses", "macros", "tests"})
 
 DEFAULT_OPENLINEAGE_NAMESPACE = "cosmos"
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -38,8 +38,8 @@ DEFAULT_PROJECT_HASH_EXCLUDED_DIRS = frozenset(
     {".git", DBT_TARGET_DIR_NAME, DBT_DEFAULT_PACKAGES_FOLDER, DBT_LOG_DIR_NAME}
 )
 
-# Conventional dbt resource-content directory names; see cosmos.settings for how these are used.
-DBT_AUTHORED_CONTENT_DIRS = frozenset({"models", "seeds", "snapshots", "analyses", "macros", "tests"})
+# Conventional dbt project content directory names; see cosmos.settings for how these are used.
+DBT_PROJECT_CONTENT_DIRS = frozenset({"models", "seeds", "snapshots", "analyses", "macros", "tests"})
 
 DEFAULT_OPENLINEAGE_NAMESPACE = "cosmos"
 OPENLINEAGE_PRODUCER = "https://github.com/astronomer/astronomer-cosmos/"
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 import copy
 import hashlib
 import inspect
-import json
 import os
 import platform
 import time
@@ -62,23 +61,11 @@ from cosmos.versioning import _create_folder_version_hash
 
 logger = get_logger(__name__)
 
-# dbt stamps these into every manifest's `metadata` on every invocation, even when nothing else
-# in the project changed, so they must be dropped before hashing or the hash would too.
-_VOLATILE_MANIFEST_METADATA_KEYS = ("generated_at", "invocation_id", "invocation_started_at")
-
 
 def _calculate_manifest_version_hash(manifest_path: Any) -> str:
-    """Content checksum of a dbt manifest file, ignoring metadata fields that change on every
-    dbt invocation even when the project itself hasn't (``generated_at``, ``invocation_id``, ...).
-    """
+    """Content checksum of a dbt manifest file."""
     with manifest_path.open("rb") as fp:
-        manifest = json.load(fp)
-    metadata = manifest.get("metadata")
-    if isinstance(metadata, dict):
-        for key in _VOLATILE_MANIFEST_METADATA_KEYS:
-            metadata.pop(key, None)
-    canonical = json.dumps(manifest, sort_keys=True, separators=(",", ":")).encode()
-    return hashlib.md5(canonical).hexdigest()
+        return hashlib.md5(fp.read()).hexdigest()
 
 
 def migrate_to_new_interface(
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -57,9 +57,11 @@ from cosmos.exceptions import CosmosValueError
 from cosmos.listeners.task_instance_listener import _get_profile_config_attribute
 from cosmos.log import get_logger
 from cosmos.telemetry import _compress_telemetry_metadata, should_emit
-from cosmos.versioning import _HASH_READ_CHUNK_SIZE, _create_folder_version_hash
+from cosmos.versioning import _create_folder_version_hash
 
 logger = get_logger(__name__)
+
+_MANIFEST_HASH_READ_CHUNK_SIZE = 1024 * 1024
 
 
 def migrate_to_new_interface(
@@ -450,7 +452,7 @@ class DbtToAirflowConverter:
                 try:
                     manifest_hasher = hashlib.md5()
                     with manifest_path.open("rb") as fp:
-                        while chunk := fp.read(_HASH_READ_CHUNK_SIZE):
+                        while chunk := fp.read(_MANIFEST_HASH_READ_CHUNK_SIZE):
                             manifest_hasher.update(chunk)
                     dbt_project_hash = hashlib.md5(
                         f"{dbt_project_hash}\0{manifest_hasher.hexdigest()}".encode()
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -446,7 +446,8 @@ class DbtToAirflowConverter:
             if self.dbt_graph.load_method == LoadMode.DBT_MANIFEST and manifest_path is not None:
                 # target/ is pruned from the folder walk above, so fold in the manifest's own checksum.
                 manifest_checksum = manifest_path.checksum()
-                dbt_project_hash = hashlib.md5(f"{dbt_project_hash}{manifest_checksum}".encode()).hexdigest()
+                if manifest_checksum:
+                    dbt_project_hash = hashlib.md5(f"{dbt_project_hash}\0{manifest_checksum}".encode()).hexdigest()
 
             hash_suffix = f"\n\n**dbt project hash:** `{dbt_project_hash}`"
 
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import copy
 import hashlib
 import inspect
+import json
 import os
 import platform
 import time
@@ -61,20 +62,26 @@ from cosmos.versioning import _create_folder_version_hash
 
 logger = get_logger(__name__)
 
-_MANIFEST_HASH_READ_CHUNK_SIZE = 1024 * 1024
+# dbt stamps these into every manifest's `metadata` on every invocation, even when nothing else
+# in the project changed, so they must be dropped before hashing or the hash would too.
+_VOLATILE_MANIFEST_METADATA_KEYS = ("generated_at", "invocation_id", "invocation_started_at")
 
 
 def _calculate_manifest_version_hash(manifest_path: Any) -> str:
-    """Content checksum of a dbt manifest file, streamed in chunks.
+    """Content checksum of a dbt manifest file, ignoring metadata fields that change on every
+    dbt invocation even when the project itself hasn't (``generated_at``, ``invocation_id``, ...).
 
     A separate, named call site (mirroring ``_create_folder_version_hash``) rather than inlined
     in the caller, so it can be overridden independently of the folder hash.
     """
-    hasher = hashlib.md5()
     with manifest_path.open("rb") as fp:
-        while chunk := fp.read(_MANIFEST_HASH_READ_CHUNK_SIZE):
-            hasher.update(chunk)
-    return hasher.hexdigest()
+        manifest = json.load(fp)
+    metadata = manifest.get("metadata")
+    if isinstance(metadata, dict):
+        for key in _VOLATILE_MANIFEST_METADATA_KEYS:
+            metadata.pop(key, None)
+    canonical = json.dumps(manifest, sort_keys=True, separators=(",", ":")).encode()
+    return hashlib.md5(canonical).hexdigest()
 
 
 def migrate_to_new_interface(
--- a/cosmos/converter.py
+++ b/cosmos/converter.py
@@ -70,9 +70,6 @@ _VOLATILE_MANIFEST_METADATA_KEYS = ("generated_at", "invocation_id", "invocation
 def _calculate_manifest_version_hash(manifest_path: Any) -> str:
     """Content checksum of a dbt manifest file, ignoring metadata fields that change on every
     dbt invocation even when the project itself hasn't (``generated_at``, ``invocation_id``, ...).
-
-    A separate, named call site (mirroring ``_create_folder_version_hash``) rather than inlined
-    in the caller, so it can be overridden independently of the folder hash.
     """
     with manifest_path.open("rb") as fp:
         manifest = json.load(fp)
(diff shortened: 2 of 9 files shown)


cosmos/converter.py
```diff
@@ -440,6 +441,13 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
 
         try:
             dbt_project_hash = _create_folder_version_hash(self.dbt_graph.project_path)
+
+            manifest_path = self.dbt_graph.project.manifest_path
+            if self.dbt_graph.load_method == LoadMode.DBT_MANIFEST and manifest_path is not None:
+                # target/ is pruned from the folder walk above, so fold in the manifest's own checksum.
+                manifest_checksum = manifest_path.checksum()
```
kaxil: `ObjectStoragePath.checksum()` just delegates to `fs.checksum()`, and that isn't content-derived on every backend. For a local manifest fsspec falls back to `tokenize(fs.info(path))`, which includes `mtime` and `ino`, so re-syncing or recompiling a byte-identical manifest yields a different value (I checked: rewriting the same bytes, and an `os.replace`, both change it). gcsfs inherits the same default and its info carries `generation`/`updated`; only s3fs overrides it to use the ETag. So the same config behaves three different ways depending on where the manifest lives.

The comment at the top of `_create_folder_version_hash` deliberately rejects mtime ("does not work well for dag-only deployments where DAGs are constantly synced"), and since `doc_md` lands in the serialized DAG, an untouched-but-re-synced manifest would cut a new Airflow 3 DAG version on the next parse. Would streaming the manifest bytes (the way `cosmos.fs._calculate_file_checksum` does) be closer to the intent here?

cosmos/converter.py
```diff
@@ -440,6 +441,13 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
 
         try:
             dbt_project_hash = _create_folder_version_hash(self.dbt_graph.project_path)
+
+            manifest_path = self.dbt_graph.project.manifest_path
```
kaxil, line 451: This sits inside the same `try` as the folder hash, so a `checksum()` failure on a remote manifest takes the whole thing down and `doc_md` gets no hash at all for that parse. That costs more than the manifest signal: the hash line vanishes and comes back on the next parse, which is two spurious DAG versions. Wrapping just the fold in its own `try` would keep the folder hash landing.

docs/reference/configs/cosmos-conf.rst
```diff
@@ -47,13 +47,16 @@ This page lists all available `Apache Airflow® <https://airflow.apache.org/>`_
...
 
 `project_hash_excluded_dirs`_:
-    Comma-separated list of directory names that Cosmos skips when hashing the dbt project directory
-    (used by ``enable_dag_versioning`` and the partial-parse/dbt-ls cache keys). Generated folders such
-    as ``target/`` and ``dbt_packages/`` can be large, and reading them on every DAG parse is expensive
-    on network-backed storage. Setting this replaces the default list entirely. A matching directory
-    name is skipped wherever it appears in the project tree.
-
-    - Default: ``.git, target, dbt_packages, logs``
+    Comma-separated list of additional directory names that Cosmos skips when hashing the dbt project
```
kaxil, line 52: The `enable_dag_versioning` entry above still says Cosmos "computes a hash of the dbt project directory". Worth adding there that under `LoadMode.DBT_MANIFEST` the manifest's own checksum is folded in, since that's the user-visible half of this PR and it isn't documented anywhere.

kaxil (commented):
A few notes from reading through the manifest-hashing change. The `checksum()` semantics question below is the one I'd want settled before merge; the other two are minor.

cosmos/versioning.py
```diff
@@ -53,8 +53,10 @@ def _create_folder_version_hash(dir_path: Path, excluded_dirs: Collection[str] |
 
     for filepath in sorted(filepaths):
         # Include the path so that renaming a file also changes the hash; dbt derives node
-        # names from file names, so a content-preserving rename still changes the project
-        hasher.update(os.path.relpath(filepath, dir_path).encode())
+        # names from file names, so a content-preserving rename still changes the project.
+        # Null-byte separator avoids a path/content boundary ambiguity; as_posix() is OS-independent.
+        hasher.update(Path(filepath).relative_to(dir_path).as_posix().encode())
```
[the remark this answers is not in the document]
pankajastro: Good catch — fixed in f26db944. Sorted by the POSIX relative path (not the OS-native `filepath`) so the hash iteration order can't diverge across OSes either, on top of the already-OS-independent encoded bytes.

cosmos/converter.py
```diff
@@ -440,6 +441,13 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+            if self.dbt_graph.load_method == LoadMode.DBT_MANIFEST and manifest_path is not None:
+                # target/ is pruned from the folder walk above, so fold in the manifest's own checksum.
+                manifest_checksum = manifest_path.checksum()
+                dbt_project_hash = hashlib.md5(f"{dbt_project_hash}{manifest_checksum}".encode()).hexdigest()
+
```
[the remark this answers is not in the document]
pankajastro: Fixed in f26db944 — added a null-byte delimiter between the folder hash and manifest checksum (same fix already applied to the path/content boundary elsewhere in this PR), and skip the fold-in entirely when `checksum()` returns a falsy value instead of embedding `"None"` in the digest.

docs/reference/configs/cosmos-conf.rst
```diff
@@ -47,13 +47,16 @@ This page lists all available `Apache Airflow® <https://airflow.apache.org/>`_
...
+    directory (used by ``enable_dag_versioning`` and the partial-parse/dbt-ls cache keys). Generated
+    folders such as ``target/`` and ``dbt_packages/`` can be large, and reading them on every DAG parse
+    is expensive on network-backed storage. This setting is additive: ``.git``, ``target``,
+    ``dbt_packages`` and ``logs`` are always excluded, regardless of this value. A matching directory
+    name is skipped wherever it appears in the project tree. Cosmos logs a warning (but does not block)
+    if this includes a dbt-authored content directory, such as ``models`` or ``macros``, since excluding
+    one means real changes there won't be detected.
```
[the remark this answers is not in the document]
pankajastro: Fixed in f26db944 — docs now say Cosmos emits a `UserWarning` rather than "logs a warning," matching the actual `warnings.warn(...)` call.

cosmos/settings.py
```diff
@@ -23,14 +24,20 @@
...
+    dirname.strip() for dirname in _project_hash_excluded_dirs_conf.split(",") if dirname.strip()
 )
+_suspicious_project_hash_excluded_dirs = _project_hash_excluded_dirs_extra & DBT_AUTHORED_CONTENT_DIRS
+if _suspicious_project_hash_excluded_dirs:
+    warnings.warn(
+        f"`project_hash_excluded_dirs` includes {sorted(_suspicious_project_hash_excluded_dirs)}, which look like "
+        "dbt-authored content directories. Excluding them means real project changes there won't be detected by "
+        "DAG versioning or the partial-parse/dbt-ls caches.",
+        UserWarning,
+    )
```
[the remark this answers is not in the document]
pankajastro: Holding off on this one. This warning fires from module-level code at import/reload time (driven by the `AIRFLOW__COSMOS__PROJECT_HASH_EXCLUDED_DIRS` config, not a direct call from user code), so there isn't a single meaningful caller frame for `stacklevel` to point at — it would just attribute to whatever imports/reloads `cosmos.settings`. The existing `watcher_dbt_execution_queue` deprecation warning a few lines below (same file, same pattern) also has no `stacklevel` set, so adding it only here would be inconsistent. Happy to revisit if you see a concrete case where it'd help.

cosmos/converter.py
```diff
@@ -440,6 +441,14 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
 
         try:
             dbt_project_hash = _create_folder_version_hash(self.dbt_graph.project_path)
+
+            manifest_path = self.dbt_graph.project.manifest_path
+            if self.dbt_graph.load_method == LoadMode.DBT_MANIFEST and manifest_path is not None:
```
pankajkoti, line 452: Could you please check whether `LoadMode.AUTOMATIC` needs to be included here as an OR condition? It could also resolve to `LoadMode.DBT_MANIFEST` when `manifest_path` is provided, or does that already get resolved by the time the code reaches here? I am guessing it should be resolved by the time it reaches here, but good to double-check we're not missing it.

cosmos/converter.py
pankajastro: Great catch, and thanks for actually verifying the fsspec behavior across backends. Fixed in 62b6abd5 — now streams the manifest's own bytes (same chunked-read pattern as `_create_folder_version_hash`, reusing `_HASH_READ_CHUNK_SIZE`) and folds that content checksum in, instead of `manifest_path.checksum()`. That's content-derived regardless of backend, so a re-sync or recompile that doesn't change the bytes no longer bumps the DAG version.

cosmos/converter.py
pankajastro: Fixed in 62b6abd5 — the manifest read/hash is now wrapped in its own try/except (logs a warning on failure), so a remote manifest checksum failure no longer takes down the folder hash that already succeeded. Added `test_dag_versioning_hash_survives_manifest_checksum_failure` to cover it.

docs/reference/configs/cosmos-conf.rst
pankajastro: Fixed in 62b6abd5 — added a note under `enable_dag_versioning` that LoadMode.DBT_MANIFEST folds the manifest's own checksum into the hash, since target/ is excluded from the directory walk.

cosmos/converter.py
```diff
@@ -440,6 +441,23 @@ def _add_dbt_project_hash_to_dag_docs(self, dag: DAG | None) -> None:
...
+                # checksum -- not manifest_path.checksum(), whose fsspec-default metadata (mtime/inode
+                # for local files, generation/updated for GCS) changes on a byte-identical re-sync.
+                try:
+                    manifest_hasher = hashlib.md5()
+                    with manifest_path.open("rb") as fp:
+                        while chunk := fp.read(_HASH_READ_CHUNK_SIZE):
+                            manifest_hasher.update(chunk)
+                    dbt_project_hash = hashlib.md5(
+                        f"{dbt_project_hash}\0{manifest_hasher.hexdigest()}".encode()
+                    ).hexdigest()
```
pankajkoti: Looks like we're mixing 2 hashes here. I am wondering whether it makes more sense to wrap this logic to consider the manifest file into the `_create_folder_version_hash` function itself to include the manifest file while scanning the directory, so that computing the hash for the dbt project is scoped to that method. I guess that would have been the behaviour pre- #2942, right? That the manifest was getting considered in the `_create_folder_version_hash` method.  And `_create_folder_version_hash` has other callers too, like `cache.oy`, so it would stay consistent, and this change would likely be needed there too.

cosmos/constants.py
```diff
@@ -38,6 +38,9 @@
     {".git", DBT_TARGET_DIR_NAME, DBT_DEFAULT_PACKAGES_FOLDER, DBT_LOG_DIR_NAME}
 )
 
+# Conventional dbt resource-content directory names; see cosmos.settings for how these are used.
+DBT_AUTHORED_CONTENT_DIRS = frozenset({"models", "seeds", "snapshots", "analyses", "macros", "tests"})
```
pankajkoti: Can we name this appropriately, please? It reads as if dbt is authoring those folders, but it's not dbt that is authoring/generating those folders. Users would author in those folders.

Some suggestions: `DBT_PROJECT_CONTENT_DIRS`, `DBT_RESOURCE_CONTENT_DIRS`, `USER_AUTHORED_DBT_DIRS`

cosmos/converter.py
pankajastro: Good check — it's already resolved by the time we get here. `load()` never leaves `self.load_method` as `LoadMode.AUTOMATIC`: when `method == LoadMode.AUTOMATIC` and a manifest is available, it calls `self.load_from_dbt_manifest()` directly (cosmos/dbt/graph.py:803-805), which sets `self.load_method = LoadMode.DBT_MANIFEST` (line 1429) before `load()` returns. So by the time `_add_dbt_project_hash_to_dag_docs` runs, `load_method` always holds the concrete resolved mode, never `AUTOMATIC` itself — no OR-branch needed here.

cosmos/converter.py
pankajastro: Fair question — yes, pre-#2942 the manifest was picked up incidentally, because the walk covered the whole project tree including `target/` (checked the pre-#2942 version of `_create_folder_version_hash`: no excluded-dirs pruning existed at all). #2942 broke that by design (pruning `target/` for the network-storage-cost reason it describes), which is what created this gap.

I kept the fold-in in `converter.py` rather than threading manifest-awareness into `_create_folder_version_hash` itself because that function has two other callers in `cache.py` (`_calculate_dbt_ls_cache_current_version` and the selector cache), both of which only apply to `LoadMode.DBT_LS`-family loading — manifest mode never runs `dbt ls`, so there's no dbt-ls/selector cache to key for it. Adding an optional `manifest_path` param to the shared utility to serve a case two of its three callers can't hit felt like it'd complicate the general-purpose function for one caller's need. Happy to reconsider if you see a cleaner shape, though — I don't feel strongly attached to the current split.

cosmos/constants.py
pankajastro: Fair, fixed in 7545eb07 — renamed to `DBT_PROJECT_CONTENT_DIRS` and updated the warning message and docs wording from "dbt-authored" to "dbt project content" directories to match.

cosmos/converter.py
pankajastro: Follow-up: extracted the manifest-checksum step into its own named function, `_calculate_manifest_version_hash`, in c3b22bbd — same effect, but it's now a standalone call site (mirroring `_create_folder_version_hash`) rather than inlined in `_add_dbt_project_hash_to_dag_docs`. Keeps it independently overridable without touching `_create_folder_version_hash`'s signature.

cosmos/versioning.py
```diff
@@ -51,15 +69,43 @@ def _create_folder_version_hash(dir_path: Path, excluded_dirs: Collection[str] |
...
+
+def _manifest_content_checksum(manifest_path: Any) -> str:
+    """MD5 of a dbt manifest's content, ignoring metadata fields dbt stamps on every invocation
+    (``generated_at``, ``invocation_id``, ...) even when the project itself hasn't changed."""
+    with manifest_path.open("rb") as fp:
+        manifest = json.load(fp)
+    metadata = manifest.get("metadata")
+    if isinstance(metadata, dict):
+        for key in _VOLATILE_MANIFEST_METADATA_KEYS:
+            metadata.pop(key, None)
```
pankajkoti: Is this manoeuvring really needed? We didn't have to do this before, and I would prefer to avoid doing this unless there's a definite reason.

pankajastro: Agreed — reverted the merge in 4378a50f (`_create_folder_version_hash` back to its original signature), and also dropped the metadata-stripping from the manifest hash in a5453d3d since it was solving a problem that doesn't occur in practice (manifest.json is checked into git, so any diff is a real recompile).

pankajkoti (approved):
LGTM. Would be nice to test the desired expectations once.

PR #2950, opened 2026-08-07
WATCHER dataset emission silently disabled when profiles.yml uses dbt Jinja filters

### LINKED ISSUE #2948: [Bug] WATCHER silently emits NO datasets/assets when profiles.yml uses dbt's documented as_number/as_bool Jinja filters
### Astronomer Cosmos Version

`1.15.1`. Regression from `1.15.0`, introduced by astronomer/astronomer-cosmos#2879 (the fix for astronomer/astronomer-cosmos#2843).

dbt Core `1.11.11`

### Versions of dbt adapters

`dbt-redshift 1.10.1`

### LoadMode

DBT_LS_MANIFEST

### ExecutionMode

WATCHER

### InvocationMode

SUBPROCESS

### airflow version

3.2.1

### Operating System

Amazon Linux

*No response*

### Deployment

Amazon (AWS) MWAA

### Deployment details

*No response*

### What happened?

Our `profiles.yml` uses dbt's **documented profiles.yml filters** on numeric fields, e.g.

```yaml
threads: "{{ env_var('DBT_THREADS', 4) | as_number }}"
```

(`as_number`/`as_bool` are documented specifically for profiles.yml:<br>[https://docs.getdbt.com/reference/dbt-jinja-functions/as_number](<https://docs.getdbt.com/reference/dbt-jinja-functions/as_number>). dbt parses and runs<br>this profile fine.)

After upgrading 1.15.0 → 1.15.1, a WATCHER `DbtDag` run completes fully green but emits<br>**zero** Assets/dataset events — for every model, not just affected fields. Nothing in<br>any task log at INFO level; downstream asset-scheduled DAGs never trigger.

Cause: astronomer/astronomer-cosmos#2879 renders every profile string field through a bare `jinja2.Template` that<br>knows none of dbt's filters. `| as_number` raises<br>`TemplateAssertionError: No filter named 'as_number'`, which the broadened<br>`except TemplateError` in `get_dataset_namespace` swallows (DEBUG log only) and turns<br>into `return None` — and `None` disables dataset emission for the entire run. A filter<br>on a field the namespace resolvers never even read (`threads`, `connect_timeout`) kills<br>emission for every model.

Expected: the 1.15.0 behavior (datasets emitted), with astronomer/astronomer-cosmos#2879's rendering applied<br>per-field so an unrenderable irrelevant field can't abort namespace derivation — and a<br>WARNING, not a DEBUG line, if derivation genuinely fails.

### Relevant log output

```shell
```

### How to reproduce

Standalone, no Airflow run needed. `profiles.yml` (host is a plain literal; only<br>`threads` uses a dbt filter):

```yaml
repro:
  target: dev
  outputs:
    dev:
      type: postgres
      host: localhost
      port: 5432
      user: postgres
      password: postgres
      dbname: postgres
      schema: public
      threads: "{{ env_var('DBT_THREADS', 4) | as_number }}"
```

```python
from cosmos.dataset import get_dataset_namespace

pc = ProfileConfig(profile_name="repro", target_name="dev",
                   profiles_yml_filepath="/path/to/profiles.yml")
print(get_dataset_namespace(pc))
# 1.15.0 -> "postgres://localhost:5432"   (datasets emitted)
# 1.15.1 -> None                          (emission skipped for the whole run)
```

### Anything else :)?

Root cause: `_get_profile_dict` ([dataset.py:146](<http://dataset.py:146>), added by astronomer/astronomer-cosmos#2879) renders **every**<br>string field via `_resolve_env_var` (`cosmos/dbt/project.py:44-62` — bare<br>`jinja2.Template`, only `env_var` in context). The compile-time `TemplateAssertionError`<br>is caught by the broadened `except` ([dataset.py:170](<http://dataset.py:170>)), logged at DEBUG, and returned as<br>`None` ⇒ producer sets `_dataset_namespace = None` (watcher.py:472-474) ⇒ no outlet URIs<br>for any model. 1.15.0 emitted for this profile; 1.15.1 emits nothing, silently.

Proposed fix (**I'd like to submit a PR**):

1. Render per-field with fallback — on `TemplateError`, keep the raw value, log a<br>WARNING, continue. The namespace resolvers only read host/port-style keys, so<br>`threads`/`connect_timeout` failures become harmless.
2. Register dbt's profiles.yml filters (`as_number`, `as_bool`, `as_text`, `as_native`)<br>in `_resolve_env_var`'s environment — dbt's own text environment implements them as<br>identity functions, sufficient here since Cosmos renders to strings.
3. Raise the swallow in `get_dataset_namespace` from DEBUG to WARNING — "all asset<br>emission disabled" should be visible at default log levels.

Regression tests: literal host + `as_number` on `threads` ⇒ namespace resolves;<br>unrenderable field ⇒ raw value kept + WARNING.

Workaround for affected users: replace dbt-only filters with standard Jinja2 ones dbt<br>also accepts, e.g. `| as_number` → `| int` for integer fields.

### Contact Details

[ignaskr@gmail.com](<mailto:ignaskr@gmail.com>)

goingforstudying-ctrl:
Ran into the scenario from #2948 while poking at watcher-mode dataset emission, and it reproduces exactly as described there.

Any profiles.yml field using dbt's documented as_number/as_bool filters, e.g. `threads: "{{ env_var('DBT_THREADS', 4) | as_number }}"`, makes `get_dataset_namespace` return None. #2879 renders every string field through a bare `jinja2.Template` that doesn't know dbt's filters, the TemplateAssertionError gets swallowed at DEBUG level, and None disables dataset emission for the entire run. So a filter on `threads`, which the namespace resolvers never even read, silently kills every asset event. Downstream asset-scheduled DAGs just never fire.

What I changed:

`_resolve_env_var` now renders through a Jinja environment that registers dbt's as_number/as_bool/as_text/as_native filters as identity no-ops. That's what dbt's own text-mode environment does with them (`TEXT_FILTERS` in `dbt_common.clients.jinja`), so the semantics match dbt instead of inventing new ones.

`_get_profile_dict` renders fields independently now. If one field's template can't be rendered at all (unknown filter, undefined name), it keeps the raw value instead of aborting the whole profile.

The catch-all in `get_dataset_namespace` logs at WARNING instead of DEBUG, since returning None there means zero datasets for the run and nobody would notice otherwise.

One behavior change worth flagging: a field that genuinely can't render (e.g. `host: "db-{{ this.is_not_env_var }}.prod.internal"`) now degrades to its raw value in the namespace instead of disabling emission for the whole run. That restores the pre-1.15.1 behavior for that one field, and I updated the existing test to match. Seemed like the right trade-off given the alternative is silently dropping all asset events, but happy to revisit if you'd rather keep the None.

Verified with the profiles.yml from the issue: before, the namespace comes back None; after, `postgres://localhost:5432` as expected. TestGetDatasetNamespace (21 tests now) and the _resolve_env_var tests pass, black and ruff clean.

Credit to @hellowithchicken for the pinpoint root-cause analysis in the issue.

Fixes #2948

### DIFF
--- a/cosmos/dataset.py
+++ b/cosmos/dataset.py
@@ -156,12 +156,15 @@ def _render_profile_field(field: str, value: str, namespace_field: bool = False)
     Resolver-read fields fail instead so literal Jinja cannot enter an emitted namespace.
     """
     try:
-        return _resolve_profiles_yml_env_var(value)
+        rendered = _resolve_profiles_yml_env_var(value)
     except TemplateError as error:
         if namespace_field:
             raise TemplateError(f"Could not render Jinja in namespace field '{field}'") from error
         logger.debug("Could not render Jinja in profiles.yml field '%s'; using the raw value", field, exc_info=True)
         return value
+    if namespace_field and not rendered.strip():
+        raise TemplateError(f"Namespace field '{field}' rendered to an empty value")
+    return rendered
 
 
 def _get_profile_dict(profile_config: ProfileConfig) -> tuple[str, dict[str, Any]]:
--- a/cosmos/dataset.py
+++ b/cosmos/dataset.py
@@ -12,7 +12,7 @@ from packaging.version import Version
 
 from cosmos import settings
 from cosmos.constants import _DATASET_EMITTING_RESOURCE_TYPES, AIRFLOW_VERSION
-from cosmos.dbt.project import _resolve_env_var
+from cosmos.dbt.project import _resolve_profiles_yml_env_var
 from cosmos.log import get_logger
 
 if TYPE_CHECKING:
@@ -100,6 +100,24 @@ _ADAPTER_NAMESPACE_RESOLVERS: dict[str, Any] = {
     "duckdb": lambda p: f"duckdb://{p.get('path', '')}",
 }
 
+# Profile fields read by each resolver. An unrenderable value in one of these fields must abort
+# namespace derivation instead of allowing literal Jinja syntax into the emitted URI.
+_ADAPTER_NAMESPACE_FIELDS: dict[str, set[str]] = {
+    "postgres": {"host", "port"},
+    "redshift": {"host", "port"},
+    "bigquery": set(),
+    "snowflake": {"account"},
+    "databricks": {"host"},
+    "spark": {"host", "port", "method"},
+    "trino": {"host", "port"},
+    "clickhouse": {"host", "port"},
+    "sqlserver": {"server", "port"},
+    "dremio": {"software_host", "port"},
+    "athena": {"region_name"},
+    "glue": {"region", "account_id", "role_arn"},
+    "duckdb": {"path"},
+}
+
 
 def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_id: str) -> str:
     """
@@ -130,15 +148,18 @@ def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_i
     return "__".join(identifiers_list)
 
 
-def _render_profile_field(field: str, value: str) -> str:
-    """Render one profiles.yml field, keeping the raw value when its Jinja cannot be rendered.
+def _render_profile_field(field: str, value: str, namespace_field: bool = False) -> str:
+    """Render one profiles.yml field, keeping irrelevant raw values when necessary.
 
     Fields the namespace resolvers never read (e.g. ``threads``) may use Jinja that this
     renderer does not support; one such field must not abort derivation for the whole profile.
+    Resolver-read fields fail instead so literal Jinja cannot enter an emitted namespace.
     """
     try:
-        return _resolve_env_var(value)
-    except TemplateError:
+        return _resolve_profiles_yml_env_var(value)
+    except TemplateError as error:
+        if namespace_field:
+            raise TemplateError(f"Could not render Jinja in namespace field '{field}'") from error
         logger.debug("Could not render Jinja in profiles.yml field '%s'; using the raw value", field, exc_info=True)
         return value
 
@@ -160,12 +181,27 @@ def _get_profile_dict(profile_config: ProfileConfig) -> tuple[str, dict[str, Any
         with open(profile_config.profiles_yml_filepath) as f:
             profiles = yaml.safe_load(f)
         target = profiles[profile_config.profile_name]["outputs"][profile_config.target_name]
+        raw_adapter_type = target.get("type", "")
+        adapter_type = (
+            _render_profile_field("type", raw_adapter_type, namespace_field=True)
+            if isinstance(raw_adapter_type, str)
+            else raw_adapter_type
+        )
+        namespace_fields = _ADAPTER_NAMESPACE_FIELDS.get(adapter_type, set())
         # dbt renders env_var() Jinja in profiles.yml; replicate that since we read the file directly.
         # Render per field so one unrenderable field cannot abort the whole profile (#2948).
         target = {
-            key: _render_profile_field(key, value) if isinstance(value, str) else value for key, value in target.items()
+            key: (
+                (
+                    adapter_type
+                    if key == "type"
+                    else _render_profile_field(key, value, namespace_field=key in namespace_fields)
+                )
+                if isinstance(value, str)
+                else value
+            )
+            for key, value in target.items()
         }
-        adapter_type = target.get("type", "")
         return adapter_type, target
 
     return "", {}
@@ -188,9 +224,10 @@ def get_dataset_namespace(profile_config: ProfileConfig) -> str | None:
     """
     try:
         adapter_type, profile_dict = _get_profile_dict(profile_config)
-    except (AttributeError, KeyError, TypeError, OSError, yaml.YAMLError, TemplateError):
+    except (AttributeError, KeyError, TypeError, OSError, yaml.YAMLError, TemplateError) as error:
         logger.warning(
-            "Unable to extract profile info for dataset namespace derivation; dataset emission will be skipped.",
+            "Unable to extract profile info for dataset namespace derivation (%s); dataset emission will be skipped.",
+            error,
             exc_info=True,
         )
         return None
--- a/cosmos/dataset.py
+++ b/cosmos/dataset.py
@@ -130,6 +130,19 @@ def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_i
     return "__".join(identifiers_list)
 
 
+def _render_profile_field(field: str, value: str) -> str:
+    """Render one profiles.yml field, keeping the raw value when its Jinja cannot be rendered.
+
+    Fields the namespace resolvers never read (e.g. ``threads``) may use Jinja that this
+    renderer does not support; one such field must not abort derivation for the whole profile.
+    """
+    try:
+        return _resolve_env_var(value)
+    except TemplateError:
+        logger.debug("Could not render Jinja in profiles.yml field '%s'; using the raw value", field, exc_info=True)
+        return value
+
+
 def _get_profile_dict(profile_config: ProfileConfig) -> tuple[str, dict[str, Any]]:
     """
     Extract the adapter type and profile dict from a ProfileConfig.
@@ -148,7 +161,10 @@ def _get_profile_dict(profile_config: ProfileConfig) -> tuple[str, dict[str, Any
             profiles = yaml.safe_load(f)
         target = profiles[profile_config.profile_name]["outputs"][profile_config.target_name]
         # dbt renders env_var() Jinja in profiles.yml; replicate that since we read the file directly.
-        target = {key: _resolve_env_var(value) if isinstance(value, str) else value for key, value in target.items()}
+        # Render per field so one unrenderable field cannot abort the whole profile (#2948).
+        target = {
+            key: _render_profile_field(key, value) if isinstance(value, str) else value for key, value in target.items()
+        }
         adapter_type = target.get("type", "")
         return adapter_type, target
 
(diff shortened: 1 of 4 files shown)


hellowithchicken:
I was already working on my solution and had approved approach from the team. Is this normal behaviour?

pankajkoti:
> I was already working on my solution and had approved approach from the team. Is this normal behaviour?

Thanks to both of you, @hellowithchicken and @goingforstudying-ctrl, for caring about getting this fixed.

@hellowithchicken, your bug report is genuinely excellent. The root cause and proposed fix were spot on, and that write-up is what made this easy to act on. You are also right that @pankajastro approved your approach on the issue, so your question is fair.

Since Cosmos is a public repo, anyone can pick up an open issue and open a PR; we don't have much control over it. We have had cases before where two people sent PRs for the same bug, and we simply review both and go with the most complete and correct one. What matters to us is getting this fixed with a correct solution and passing CI.

Since you already had an approved approach and were mid-solution, we would really value your review on this PR, as you know the edge cases well. And if you would prefer to open your own PR with a better approach, that is completely welcome too. Either way, your contribution is highly valued and important to us.

Thanks again, both of you.

pankajkoti:
The failing linkcheck job was unrelated, we fixed it in #2952 and I have rebased this PR with a merge commit so we unblock the unrelated CI failure

hellowithchicken:
Thanks @pankajkoti, I appreciate your response.

Since this PR follows my proposal from #2948 step-by-step, I don't see value in opening a second PR that would just land in the same place. I'll take some time over the next few days to review #2950, but please don't wait on me. If it's ready to merge before then, go ahead.

cosmos/dataset.py
```diff
@@ -130,6 +130,19 @@ def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_i
     return "__".join(identifiers_list)
 
 
+def _render_profile_field(field: str, value: str) -> str:
```
pankajastro, line 133: _ADAPTER_NAMESPACE_RESOLVERS reads host/port/account, but this fallback keeps the raw unrendered string for any field that fails to render. If host fails to render, doesn't that bake literal `{{ }}` text into the namespace/URI instead of the previous safe `None`? Your new test (test_profiles_yml_filepath_unrenderable_field_keeps_raw_value) seems to assert that corrupted output as the intended behavior — is that intentional, or should resolver-read fields fall back the same way an all-fields failure does?

cosmos/dataset.py
```diff
@@ -130,6 +130,19 @@ def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_i
...
+    """Render one profiles.yml field, keeping the raw value when its Jinja cannot be rendered.
+
+    Fields the namespace resolvers never read (e.g. ``threads``) may use Jinja that this
+    renderer does not support; one such field must not abort derivation for the whole profile.
+    """
+    try:
+        return _resolve_env_var(value)
+    except TemplateError:
+        logger.debug("Could not render Jinja in profiles.yml field '%s'; using the raw value", field, exc_info=True)
```
pankajastro, line 144: Should this log at WARNING like the all-fields-failed branch above? Right now the case that can actually corrupt the namespace (partial fallback) logs quieter than the case that just skips emission.

cosmos/dbt/project.py
```diff
@@ -28,6 +28,12 @@
 
 logger = get_logger(__name__)
 
+# dbt renders profiles.yml values through a Jinja environment that registers its custom filters
+# (as_number, as_bool, as_text, as_native). In dbt's text-mode environment these filters are
+# identity functions; register the same no-ops so templates using them render here too.
+_JINJA_ENV = Environment()
+_JINJA_ENV.filters.update({name: lambda x: x for name in ("as_number", "as_bool", "as_text", "as_native")})
```
pankajastro: These filters are registered on the shared _JINJA_ENV, which get_dbt_packages_subpath() also uses for dbt_project.yml's packages-install-path. Could a filter name collision there now silently no-op instead of raising TemplateError? Should this use a separate Jinja env scoped to profiles.yml?

pankajastro:
@goingforstudying-ctrl just checking in — are you still planning to continue on this one, or do you have any questions on the review comments? Happy to help unblock.

goingforstudying-ctrl:
Yes — thanks for the follow-up. I have addressed all three review points in 5ebf56e:

- resolver-read fields now fall back to None instead of preserving unrendered braces;
- partial fallback now logs at WARNING;
- profiles.yml rendering now uses a dedicated Jinja environment, so dbt filters do not affect the shared environment.

Validation passed: all 23 TestGetDatasetNamespace tests, 6 focused rendering tests, and the changed-files pre-commit checks including mypy.

cosmos/dataset.py
```diff
@@ -130,6 +148,22 @@ def get_dataset_alias_name(dag: DAG | None, task_group: TaskGroup | None, task_i
...
 
+def _render_profile_field(field: str, value: str, namespace_field: bool = False) -> str:
+    """Render one profiles.yml field, keeping irrelevant raw values when necessary.
+
+    Fields the namespace resolvers never read (e.g. ``threads``) may use Jinja that this
+    renderer does not support; one such field must not abort derivation for the whole profile.
+    Resolver-read fields fail instead so literal Jinja cannot enter an emitted namespace.
+    """
+    try:
+        return _resolve_profiles_yml_env_var(value)
```
hellowithchicken, line 142: This can also return `""` silently instead of raising. `env_var`'s own default is `""`:

```python
def env_var(name: str, default: str = "") -> str:
    return os.getenv(name, default)
```

So a missing/renamed var (e.g. `REDSHIFT_HOST` unset, no default given) renders `host`
to `""` with no exception. For a namespace field that still produces a broken namespace like `postgres://:5432`, just without the obvious `{{ }}`.

Maybe we could add here for `namespace_field=True`:

```python
if namespace_field and not rendered:
    raise TemplateError(f"Namespace field '{field}' rendered to an empty value")
```

hellowithchicken: Might be worth adding a test for this too? something that checks a namespace field (one listed in _ADAPTER_NAMESPACE_FIELDS) can't silently end up empty after rendering, not just that it doesn't raise.

hellowithchicken (changes requested):

cosmos/dataset.py
goingforstudying-ctrl: Fixed in 7e9b6b10. Namespace fields now reject values that render to empty or whitespace-only text with TemplateError, while non-namespace fields keep the existing behavior. I also added a regression covering a missing REDSHIFT_HOST and confirmed the focused profiles.yml tests plus changed-files pre-commit checks pass.

pankajastro:
Thanks for addressing the feedback.

One issue with the new empty-value check in 8831033c: it aborts the whole profile when *any* field in `_ADAPTER_NAMESPACE_FIELDS` renders empty, but two resolvers treat an empty value as a legitimate signal rather than an error, so this disables emission for profiles that work on `main` today.

- `_resolve_glue_namespace` is written as `if "account_id" in profile and profile["account_id"]:` — the `and` is there specifically to fall through to `role_arn`. An empty `account_id` now raises before the resolver gets to use its own fallback. And when `account_id` *is* set, `role_arn` is never read at all, yet an empty `role_arn` still aborts.
- `_resolve_spark_namespace` reads `method` only to choose a default port, and only when `port` is absent. `method` never appears in the URI, so an empty `method` can't corrupt anything — but it disables emission.

Comparing `get_dataset_namespace()` on `main` vs this branch:

| profiles.yml | main | this branch |
|---|---|---|
| glue, `account_id: ""`, `role_arn` set | `arn:aws:glue:us-east-1:123456789012` | `None` |
| glue, `account_id` set, `role_arn: "{{ env_var('GLUE_ROLE_ARN', '') }}"` | `arn:aws:glue:us-east-1:123456789012` | `None` |
| spark, `port: 10000`, `method: "{{ env_var('SPARK_METHOD', '') }}"` | `spark://sparkhost:10000` | `None` |

This one passes on `main` and fails here:

```python
    def test_profiles_yml_filepath_glue_empty_role_arn_with_account_id(self, tmp_path):
        """role_arn is never read when account_id is set, so an empty role_arn is irrelevant
        to the namespace and must not disable dataset emission."""
        profiles_yml = tmp_path / "profiles.yml"
        profiles_yml.write_text("""
my_profile:
  outputs:
    dev:
      type: glue
      region: us-east-1
      account_id: "123456789012"
      role_arn: "{{ env_var('GLUE_ROLE_ARN', '') }}"
""")
        profile_config = MagicMock()
        profile_config.profile_mapping = None
        profile_config.profiles_yml_filepath = str(profiles_yml)
        profile_config.profile_name = "my_profile"
        profile_config.target_name = "dev"
        assert get_dataset_namespace(profile_config) == "arn:aws:glue:us-east-1:123456789012"
```

CI doesn't catch this because every existing glue/spark test goes through `profile_mapping`, which returns early from `_get_profile_dict` and never reaches `_render_profile_field` — nothing exercises those adapters via `profiles_yml_filepath`.

On direction: I'd suggest validating the final namespace string rather than individual fields — render everything with the raw-value fallback, then return `None` if the result contains `{{`/`{%` or has an empty component (extending the existing `endswith("://")` check). That's one source of truth instead of a field map that has to mirror each resolver's internal fallbacks, and it also catches `host:` (YAML null), which currently yields `postgres://None:5432` on both branches.

Also, minor: the reply above says the partial fallback now logs at WARNING, but `cosmos/dataset.py:163` is still `logger.debug`. That's arguably fine now that resolver-read fields raise — just flagging the mismatch.

goingforstudying-ctrl:
Good catch, thanks. Reworked the validation to check the final namespace string instead of individual fields. The per-field approach was aborting the whole profile when any resolver-read field rendered empty, but resolvers like glue and spark have their own internal fallback logic that treats empty values as legitimate signals.

Changes:
- Removed _ADAPTER_NAMESPACE_FIELDS and the per-field empty/unrenderable checks
- All fields now get the raw-value fallback in _render_profile_field
- Added final namespace validation in get_dataset_namespace: rejects unrendered Jinja ({{ or {%), empty host components (postgres://:5432), and scheme-only namespaces
- Added tests for glue empty role_arn with account_id, glue empty account_id falling through to role_arn, and spark empty method with explicit port

pankajastro (approved):
LGTM

PR #2952, opened 2026-08-07
ci(docs): exclude bot-blocked Stack Exchange links from linkcheck

pankajkoti:
## What

Add the Stack Exchange network (and Stack Overflow) to the `exclude` list in `lychee.toml`.

## Why

The "Build Sphinx documentation" → `linkcheck` job fails on a single pre-existing
external link — `https://devops.stackexchange.com/questions/676/...`, referenced in
`docs/guides/run_dbt/container/docker.rst`. Stack Exchange bot-blocks automated clients,
returning **403 even with a real browser User-Agent**, while the page resolves fine in a
real browser. This is a bot-block, not a broken link.

Because `docs-build.yml` triggers on any `cosmos/**` or `docs/**` change and lychee then
checks *all* docs, this breaks linkcheck on unrelated PRs (e.g. #2950, which only touches
Python). `main` itself would fail if the docs build were re-run.

## Fix

Exclude the whole `*.stackexchange.com` network plus `stackoverflow.com`, matching the
convention `lychee.toml` already documents for bot-blocked hosts (`docs.aws.amazon.com`,
`charts.bitnami.com`, `join.slack.com`). The `docker.rst` link is left in place — it's
useful and valid; per the file's own convention it's excluded from checking, not removed.

## Verification

Ran locally exactly as CI does:

    lychee --no-progress --config lychee.toml 'docs/**/*.rst'
    # 527 Total ✅ 447 OK 🚫 0 Errors 👻 80 Excluded

Confirmed the `devops.stackexchange.com` link now falls under Excluded rather than Errors.

### DIFF
--- a/lychee.toml
+++ b/lychee.toml
@@ -45,6 +45,6 @@ exclude = [
     'https://github\.com/your-org/',
     # Bot-blocked — Stack Exchange / Stack Overflow return 403 to automated
     # clients (even with a browser UA); pages resolve fine in a real browser.
-    'https://[a-z]+\.stackexchange\.com',
-    'https://stackoverflow\.com',
+    'https://[a-z0-9-]+\.stackexchange\.com',
+    'https://([a-z0-9-]+\.)?stackoverflow\.com',
 ]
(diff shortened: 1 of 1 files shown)


tatiana (approved):
Thanks for improving this, @pankajkoti !

pankajkoti:
The linkcheck job succeeded here: https://github.com/astronomer/astronomer-cosmos/actions/runs/31171271148/job/92843391496?pr=2952

Merging the PR to unblock other PR's CI runs. The remaining test runs are not related and hence not waiting on them to complete. 

Thanks for the quick review @tatiana

PR #2957, opened 2026-08-12
Use boto3 for Kubernetes dbt docs S3 uploads

### LINKED ISSUE #2556: Remove AWS CLI dependency from `DbtDocsS3KubernetesOperator`
The `DbtDocsS3KubernetesOperator` currently relies on the AWS CLI to upload dbt docs to S3. This introduces an unnecessary dependency and can complicate container setup and execution environments.

We should refactor the operator to use airflow hook. This would reduce external dependencies, and make the implementation more portable and easier to maintain.

Address: https://github.com/astronomer/astronomer-cosmos/pull/2058#discussion_r3083858170

jx2lee:
Closes #2556 cc. @pankajastro 

## Summary
This PR updates `DbtDocsS3KubernetesOperator` so Kubernetes dbt docs uploads happen inside the docs generator Pod with `boto3`, instead of requiring the AWS CLI.

Changes include:

- Replace the `aws s3 sync` upload command with a small in-Pod Python uploader using `boto3`.
- Resolve AWS credentials through `AwsBaseHook` and inject the resulting AWS environment variables into the Kubernetes Pod.
- Preserve the generated dbt command when wrapping it with `/bin/bash -c`.
- Update Kubernetes docs and test images to require `boto3` instead of `awscli`.
- Make the Kubernetes test setup a bit more idempotent for local runs.

## Motivation

For Kubernetes execution, `dbt docs generate` runs inside a separate Pod, so the generated `target/` directory is available in that Pod rather than on the Airflow worker. Uploading from inside the same Pod avoids requiring a shared volume between the worker and docs generator Pod.

Using `boto3` also removes the AWS CLI runtime dependency from Kubernetes images while still allowing Cosmos to resolve credentials through the Airflow AWS connection.

### DIFF
--- a/cosmos/operators/kubernetes.py
+++ b/cosmos/operators/kubernetes.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+import json
+import shlex
+import textwrap
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Sequence
 from typing import TYPE_CHECKING, Any
@@ -27,11 +30,6 @@ from cosmos.operators.base import (
     DbtSourceMixin,
     DbtTestMixin,
 )
-
-try:
-    from airflow.sdk.bases.hook import BaseHook
-except ImportError:  # Since Airflow 3.1, BaseHook is in the airflow.sdk.bases.hook module
-    from airflow.hooks.base import BaseHook
 
 
 class DbtKubernetesBaseOperator(AbstractDbtBase, KubernetesPodOperator):  # type: ignore[misc]
@@ -218,13 +216,21 @@ class DbtDocsCloudKubernetesOperator(DbtDocsKubernetesOperator, ABC):
     @abstractmethod
     def build_upload_shell_command(self, docs_target: str) -> str:
         """
-        Build the shell command that will upload the generated docs from
-        `docs_target` to cloud storage. Implemented by subclasses.
+        Build the shell command that uploads generated docs from `docs_target`
+        to cloud storage inside the Kubernetes Pod.
         """
 
     @abstractmethod
     def get_upload_env_vars(self) -> dict[str, str]:
         """Return env vars required by the upload command."""
+
+    @staticmethod
+    def _command_parts(command: Any) -> list[str]:
+        if not command:
+            return []
+        if isinstance(command, (list, tuple)):
+            return [str(part) for part in command]
+        return [str(command)]
 
     def build_and_run_cmd(
         self,
@@ -244,25 +250,20 @@ class DbtDocsCloudKubernetesOperator(DbtDocsKubernetesOperator, ABC):
         # the leading "dbt" is not dropped when folded into the bash -c string below (see PR #2488).
         cmds: Any = self.cmds  # type: ignore[has-type]
         arguments: Any = self.arguments  # type: ignore[has-type]
-        cmd_parts = list(cmds) if isinstance(cmds, (list, tuple)) else [cmds]
-        if isinstance(arguments, (list, tuple)):
-            cmd_parts.extend(arguments)
-        else:
-            cmd_parts.append(arguments)
-
-        dbt_cmd_str = " ".join([str(part) for part in cmd_parts])
+        cmd_parts = self._command_parts(cmds) + self._command_parts(arguments)
+        dbt_cmd_str = shlex.join(cmd_parts)
         docs_target = f"{self.project_dir}/target"
 
         upload_cmd = self.build_upload_shell_command(docs_target)
         shell_cmd = f"{dbt_cmd_str} && {upload_cmd}"
 
-        # Override container command and arguments
         self.cmds = ["/bin/bash", "-c"]
         self.arguments = [shell_cmd]
 
         self.log.info("Running command in Kubernetes Pod: %s", self.arguments)
         result = KubernetesPodOperator.execute(self, context)
         self.log.info(result)
+
         return result
 
     def inject_upload_env_vars(self, env_vars: dict[str, str]) -> None:
@@ -286,7 +287,7 @@ class DbtDocsCloudKubernetesOperator(DbtDocsKubernetesOperator, ABC):
 class DbtDocsS3KubernetesOperator(DbtDocsCloudKubernetesOperator):
     """
     Executes `dbt docs generate` inside a Kubernetes Pod and uploads the generated
-    documentation to S3 *also inside that Pod* using `aws s3 sync`.
+    documentation to S3 also inside that Pod using ``boto3``.
         - The Kubernetes Pod receives AWS credentials resolved from the supplied
           Airflow `connection_id`.
     """
@@ -306,45 +307,97 @@ class DbtDocsS3KubernetesOperator(DbtDocsCloudKubernetesOperator):
         self.folder_dir = folder_dir
 
     def build_upload_shell_command(self, docs_target: str) -> str:
-        if self.folder_dir:
-            s3_prefix = f"s3://{self.bucket_name}/{self.folder_dir}".rstrip("/")
-        else:
-            s3_prefix = f"s3://{self.bucket_name}"
+        folder_dir = self.folder_dir.rstrip("/") if self.folder_dir else ""
+        upload_script = textwrap.dedent(f"""
+            import json
+            import mimetypes
+            import os
+            from pathlib import Path
 
-        return f"aws s3 sync {docs_target} {s3_prefix}"
+            try:
+                import boto3
+            except ImportError as exc:
+                raise SystemExit("boto3 is required in the Kubernetes image to upload dbt docs to S3.") from exc
+
+            target_dir = Path({json.dumps(docs_target)})
+            bucket_name = {json.dumps(self.bucket_name)}
+            folder_dir = {json.dumps(folder_dir)}
+
+            client_kwargs = dict()
+            endpoint_url = os.environ.get("AWS_ENDPOINT_URL_S3")
+            if endpoint_url:
+                client_kwargs["endpoint_url"] = endpoint_url
+
+            client_config = json.loads(os.environ.get("COSMOS_AWS_CLIENT_CONFIG", "{{}}"))
+            if "verify" in client_config:
+                client_kwargs["verify"] = client_config["verify"]
+
+            config_kwargs = client_config.get("config_kwargs")
+            if config_kwargs:
+                from botocore.config import Config
+
+                client_kwargs["config"] = Config(**config_kwargs)
+
+            s3 = boto3.client("s3", **client_kwargs)
+            for file_path in target_dir.rglob("*"):
+                if not file_path.is_file():
+                    continue
+
+                relative_path = file_path.relative_to(target_dir).as_posix()
+                key = f"{{folder_dir}}/{{relative_path}}" if folder_dir else relative_path
+                content_type, _ = mimetypes.guess_type(str(file_path))
+                extra_args = {{"ContentType": content_type}} if content_type else None
+                print(f"Uploading {{file_path}} to s3://{{bucket_name}}/{{key}}")
+                if extra_args:
+                    s3.upload_file(str(file_path), bucket_name, key, ExtraArgs=extra_args)
+                else:
+                    s3.upload_file(str(file_path), bucket_name, key)
+            """).strip()
+        return f"$(command -v python3 || command -v python) - <<'PY'\n{upload_script}\nPY"
 
     def get_upload_env_vars(self) -> dict[str, str]:
         return self.aws_env_vars_from_connection(self.connection_id)
 
     def aws_env_vars_from_connection(self, connection_id: str) -> dict[str, str]:
-        conn = BaseHook.get_connection(connection_id)
-        conn_extra = conn.extra_dejson
+        try:
+            from airflow.providers.amazon.aws.hooks.base_aws import AwsBaseHook
+        except ImportError:
+            from cosmos.operators.lazy_load import MissingPackage
 
-        access_key = conn.login or conn_extra.get("aws_access_key_id")
-        secret_key = conn.password or conn_extra.get("aws_secret_access_key")
-        session_token = conn_extra.get("aws_session_token") or conn_extra.get("session_token")
+            AwsBaseHook = MissingPackage(
+                "airflow.providers.amazon.aws.hooks.base_aws.AwsBaseHook",
+                "amazon",
+            )
 
-        session_kwargs = conn_extra.get("session_kwargs", {})
-        config_kwargs = conn_extra.get("config_kwargs", {})
-        if not isinstance(session_kwargs, dict):
-            session_kwargs = {}
-        if not isinstance(config_kwargs, dict):
-            config_kwargs = {}
-        region_name = (
-            conn_extra.get("region_name")
-            or conn_extra.get("region")
-            or session_kwargs.get("region_name")
-            or config_kwargs.get("region_name")
-        )
+        hook = AwsBaseHook(aws_conn_id=connection_id, client_type="s3")
+        conn_config = hook.conn_config
+        conn_extra = conn_config.extra_config
+
+        config_kwargs = conn_extra.get("config_kwargs") or {}
+
+        region_name = hook.region_name or conn_extra.get("region") or config_kwargs.get("region_name")
+        endpoint_url = conn_config.get_service_endpoint_url("s3")
+        verify = hook.verify
 
         env_vars = {}
-        if access_key:
-            env_vars["AWS_ACCESS_KEY_ID"] = access_key
-        if secret_key:
-            env_vars["AWS_SECRET_ACCESS_KEY"] = secret_key
-        if session_token:
-            env_vars["AWS_SESSION_TOKEN"] = session_token
+        if conn_config.aws_access_key_id:
+            env_vars["AWS_ACCESS_KEY_ID"] = conn_config.aws_access_key_id
+        if conn_config.aws_secret_access_key:
+            env_vars["AWS_SECRET_ACCESS_KEY"] = conn_config.aws_secret_access_key
+        if conn_config.aws_session_token:
+            env_vars["AWS_SESSION_TOKEN"] = conn_config.aws_session_token
+        if endpoint_url:
+            env_vars["AWS_ENDPOINT_URL_S3"] = endpoint_url
         if region_name:
             env_vars["AWS_DEFAULT_REGION"] = region_name
+            env_vars["AWS_REGION"] = region_name
+
+        client_config = {}
+        if verify is not None:
+            client_config["verify"] = verify
+        if config_kwargs:
+            client_config["config_kwargs"] = config_kwargs
+        if client_config:
+            env_vars["COSMOS_AWS_CLIENT_CONFIG"] = json.dumps(client_config)
 
         return env_vars
--- a/docs/guides/dbt_docs/generating-docs.rst
+++ b/docs/guides/dbt_docs/generating-docs.rst
@@ -139,7 +139,7 @@ Requirements specific to Kubernetes:
 
 - The container image must include your dbt project files.
 - The container image or mounted files must include a ``profiles.yml`` file, because Kubernetes execution mode does not support :ref:`use-profile-mapping`.
-- The container image must have the AWS CLI available because Cosmos uploads the generated docs with ``aws s3 sync``.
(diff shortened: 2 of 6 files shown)


cosmos/operators/kubernetes.py
```diff
@@ -244,25 +250,20 @@ def build_and_run_cmd(
...
         arguments: Any = self.arguments  # type: ignore[has-type]
-        cmd_parts = list(cmds) if isinstance(cmds, (list, tuple)) else [cmds]
-        if isinstance(arguments, (list, tuple)):
-            cmd_parts.extend(arguments)
-        else:
-            cmd_parts.append(arguments)
-
-        dbt_cmd_str = " ".join([str(part) for part in cmd_parts])
+        cmd_parts = self._command_parts(cmds) + self._command_parts(arguments)
+        dbt_cmd_str = shlex.join(cmd_parts)
```
pankajastro, line 254: `shlex.join` genuinely fixes the common case here — a YAML-dumped `--vars` value containing spaces or newlines used to break the `&&` chain, so this is an improvement.

But it's also a silent behaviour change for DAGs that worked around the old naive `" ".join` by pre-embedding quotes themselves, e.g. `dbt_cmd_flags=["--vars", "'{key: value}'"]`. `shlex.quote` now escapes those quotes, so dbt receives them literally and fails to parse the vars. Only `DbtDocsCloudKubernetesOperator` goes through this path so the blast radius looks small — worth a changelog note so anyone with a pre-quoted workaround knows to drop it?

cosmos/operators/kubernetes.py
```diff
@@ -306,45 +307,54 @@ def __init__(
         self.folder_dir = folder_dir
 
     def build_upload_shell_command(self, docs_target: str) -> str:
```
pankajastro, line 309: `aws s3 sync <target> s3://…` previously published the **entire** `target/` directory; the new uploader publishes only the three `required_files`. Anyone consuming `run_results.json`, `graph_summary.json`, or other artifacts from that bucket would lose them silently, with no error.

Was that narrowing intentional? If so, could it get a changelog entry and a line in the docs — it's not mentioned in the PR body or the docs diff.

cosmos/operators/kubernetes.py
```diff
@@ -306,45 +307,54 @@ def __init__(
...
+            except ImportError as exc:
+                raise SystemExit("boto3 is required in the Kubernetes image to upload dbt docs to S3.") from exc
+
+            target_dir = Path({json.dumps(docs_target)})
+            bucket_name = {json.dumps(self.bucket_name)}
+            folder_dir = {json.dumps(folder_dir)}
+            required_files = {json.dumps(self.required_files)}
+
+            s3 = boto3.client("s3")
+            for filename in required_files:
```
pankajastro: Two questions about this loop:

1. `upload_file` hard-fails on a missing artifact where `aws s3 sync` would simply have skipped it. With `dbt_cmd_global_flags=["--no-write-json"]`, `check_static_flag` (lines 199-201) only removes `graph.gpickle`, so `manifest.json` stays in `required_files` and the pod dies with `FileNotFoundError` after the docs generate has already succeeded. Should a missing file warn-and-skip instead?
2. There's no all-or-nothing semantics — if the loop fails partway, the bucket is left with a fresh `index.html` pointing at a stale `catalog.json`, which renders as a confusing half-broken docs site. Would uploading `index.html` last make partial failures less surprising?

cosmos/operators/kubernetes.py
```diff
@@ -306,45 +307,54 @@ def __init__(
...
+                key = f"{{folder_dir}}/{{filename}}" if folder_dir else filename
+                content_type, _ = mimetypes.guess_type(str(file_path))
+                extra_args = {{"ContentType": content_type}} if content_type else None
+                print(f"Uploading {{file_path}} to s3://{{bucket_name}}/{{key}}")
+                if extra_args:
+                    s3.upload_file(str(file_path), bucket_name, key, ExtraArgs=extra_args)
+                else:
+                    s3.upload_file(str(file_path), bucket_name, key)
+            """).strip()
+        return f"python - <<'PY'\n{upload_script}\nPY"
```
pankajastro: This hard-codes `python`. A lot of dbt/base images ship only `python3` (Debian/Ubuntu without `python-is-python3`, RHEL/UBI), and the failure would surface as `bash: python: command not found` *after* the whole `dbt docs generate` has run — an expensive way to find out. Both test Dockerfiles here are `FROM python:3.12`, so CI can't catch it.

Could we use `python3`, or `$(command -v python3 || command -v python)` to cover both?

cosmos/operators/kubernetes.py
```diff
@@ -306,45 +307,54 @@ def __init__(
...
-            session_kwargs = {}
-        if not isinstance(config_kwargs, dict):
-            config_kwargs = {}
-        region_name = (
-            conn_extra.get("region_name")
-            or conn_extra.get("region")
-            or session_kwargs.get("region_name")
-            or config_kwargs.get("region_name")
-        )
+        from airflow.providers.amazon.aws.hooks.base_aws import AwsBaseHook
```
pankajastro: This adds a **worker-side** dependency on `apache-airflow-providers-amazon` — the implementation this replaces used only the generic `BaseHook` and needed no provider. The `kubernetes` extra in `pyproject.toml` pulls just `apache-airflow-providers-cncf-kubernetes`, and the docs changes in this PR only state that the *Pod image* needs `boto3`.

Wouldn't an existing user upgrading hit `ModuleNotFoundError: No module named 'airflow.providers.amazon'` at task runtime? CI passes because the pinned test lockfiles happen to include `apache-airflow-providers-amazon==9.21.0`, so it wouldn't show up there.

Should we document the worker-side requirement (and/or add it to the extra), or wrap the import with a `MissingPackage`-style message the way `cosmos/operators/lazy_load.py` does?

cosmos/operators/kubernetes.py
```diff
@@ -306,45 +307,54 @@ def __init__(
...
+
+        hook = AwsBaseHook(aws_conn_id=connection_id, client_type="s3")
```
pankajastro, line 372: The in-pod client is a bare `boto3.client("s3")`, so `endpoint_url`, `verify`, and `config_kwargs` from the connection are dropped. For a connection with `extra={"endpoint_url": "http://minio:9000"}` — MinIO, another S3-compatible store, or an S3 VPC endpoint — wouldn't this now upload to real AWS instead of the configured endpoint?

`DbtDocsS3LocalOperator` goes through `S3Hook` and honours `endpoint_url`, so K8s mode and local mode would behave differently for the same connection. Could we inject `AWS_ENDPOINT_URL_S3` (or thread `endpoint_url` into the generated script) to keep them consistent?

cosmos/operators/kubernetes.py
```diff
@@ -306,45 +307,54 @@ def __init__(
...
+        credentials = hook.get_credentials()
```
pankajastro: This is my main concern with the PR. `AwsBaseHook.get_credentials()` falls through to the boto3 default session when the Airflow connection defines no keys, which changes behaviour in two ways:

1. **Pod-side IRSA gets hijacked.** For someone on EKS who keeps a placeholder `aws` connection and relies on the *dbt pod's* service-account role, the old code injected nothing, so `aws s3 sync` used the pod identity. Now the *worker's* credentials are injected as `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN`, and those take precedence over `AWS_WEB_IDENTITY_TOKEN_FILE` inside the pod. Doesn't that make the upload run as the worker principal and fail with `AccessDenied`?
2. **Worker with no credentials at all** — `session.get_credentials()` returns `None`, so `credentials.access_key` on the next line raises `AttributeError: 'NoneType' object has no attribute 'get_frozen_credentials'` inside the hook, failing the task before the pod is even created. The previous implementation just returned `{}` here.

Would it work to only inject env vars when the connection actually defines credentials, and otherwise leave the pod's own identity chain untouched?

cosmos/operators/kubernetes.py
```diff
@@ -306,45 +307,54 @@ def __init__(
...
+        region_name = hook.region_name
```
pankajastro: The removed code also read `extra["region"]`, `extra["session_kwargs"]["region_name"]`, and `extra["config_kwargs"]["region_name"]`, but `AwsBaseHook.region_name` doesn't consult `config_kwargs`.

So for a connection whose only region source is `{"config_kwargs": {"region_name": "us-east-1"}}`, no region env var gets set and `boto3.client("s3")` in the pod raises `botocore.exceptions.NoRegionError` — again only after a full docs generate. Should we keep the `config_kwargs` fallback on top of `hook.region_name`?

cosmos/operators/kubernetes.py
```diff
@@ -306,45 +307,54 @@ def __init__(
...
+                print(f"Uploading {{file_path}} to s3://{{bucket_name}}/{{key}}")
+                if extra_args:
+                    s3.upload_file(str(file_path), bucket_name, key, ExtraArgs=extra_args)
+                else:
+                    s3.upload_file(str(file_path), bucket_name, key)
+            """).strip()
+        return f"python - <<'PY'\n{upload_script}\nPY"
 
     def get_upload_env_vars(self) -> dict[str, str]:
         return self.aws_env_vars_from_connection(self.connection_id)
```
pankajastro, line 359: Separate from the fallthrough question above: credentials are frozen here at the top of `build_and_run_cmd`, before the pod starts, but they're only consumed *after* `dbt docs generate` finishes.

For `assume_role` / web-identity connections `get_credentials()` returns short-lived credentials (15 min minimum, ~1 h typical), so any docs generate that outruns the TTL would fail with `ExpiredToken` at the very last step. The previous code exported the connection's long-lived static keys, which never expired.

Is that a caveat worth documenting, or should the pod resolve credentials itself (e.g. keep the role assumption in-pod) so it can't go stale?

docs/guides/dbt_docs/generating-docs.rst
```diff
@@ -138,8 +138,8 @@ This is important because the dbt ``target`` directory is created inside the Pod
 Requirements specific to Kubernetes:
 
 - The container image must include your dbt project files.
-- The container image or mounted files must include a ``profiles.yml`` file, because Kubernetes execution mode does not support :ref:`use-profile-mapping`.
-- The container image must have the AWS CLI available because Cosmos uploads the generated docs with ``aws s3 sync``.
+- The container image or mounted files must include a ``profiles.yml`` file, because Kubernetes execution mode does not support :doc:`../connect_database/use-profile-mapping`.
```
pankajastro: This changes `:ref:`use-profile-mapping`` back to `:doc:`../connect_database/use-profile-mapping``, which reverts commit 6dd07a74 ("Unify internal doc links to use `:ref:` role", #2873) that landed on main a few commits ago. The `_use-profile-mapping` label exists and every other reference in the repo now uses `:ref:`.

Looks like a stale-branch artifact rather than something intended by this PR — could you rebase and restore the `:ref:` form?

tests/operators/test_kubernetes.py
```diff
@@ -885,35 +885,42 @@ class CustomDbtDocsLocalOperator(DbtDocsKubernetesOperator):
...
-    )
+def test_dbt_docs_s3_kubernetes_operator_uses_connection_id(_mock_build_cmd, mock_kubernetes_execute):
     operator = DbtDocsS3KubernetesOperator(
         task_id="fake-task",
         project_dir="fake-dir",
         image="fake-image",
         profile_config=profile_config,
         connection_id="aws_s3_conn",
         bucket_name="fake-bucket",
+        folder_dir="docs",
```
pankajastro: This only exercises `folder_dir="docs"`. Could we `parametrize` the other new branches while we're here?

- `folder_dir=None` → `key = filename` (no prefix)
- `folder_dir="docs/"` → covers the new `rstrip("/")`
- `region_name=None` → neither `AWS_DEFAULT_REGION` nor `AWS_REGION` injected
- `credentials.token=None` → no `AWS_SESSION_TOKEN`

All four are cheap to add and they're the paths most likely to regress silently.

tests/operators/test_kubernetes.py
```diff
@@ -885,35 +885,42 @@ class CustomDbtDocsLocalOperator(DbtDocsKubernetesOperator):
...
-    # Guard the bash -c path: build_kube_args splits "dbt" into self.cmds, so the shell
-    # command must still start with the full "dbt docs generate" and not drop the executable
-    # (regression in the cloud-docs build_and_run_cmd, see PR #2488).
     assert operator.cmds == ["/bin/bash", "-c"]
-    assert operator.arguments[0].startswith("dbt docs generate")
-    assert " && aws s3 sync " in operator.arguments[0]
+    shell_cmd = operator.arguments[0]
+    assert shell_cmd.startswith("dbt docs generate --profile default --target dev --project-dir fake-dir")
+    assert " && python - <<'PY'" in shell_cmd
+    assert "import boto3" in shell_cmd
```
pankajastro: The uploader is now Python *source* assembled from a nested f-string with doubled braces plus `textwrap.dedent`, but the coverage here is substring matching only. Wouldn't a brace-escaping or indentation slip pass CI and then fail inside a Pod at the end of a real docs generate?

Could we extract the heredoc body and at minimum `ast.parse` it — or better, `exec` it against a stubbed `boto3` so the test asserts the actual bucket/key arguments and `ExtraArgs` rather than the text that produces them?

tests/operators/test_kubernetes.py
```diff
@@ -885,35 +885,42 @@ class CustomDbtDocsLocalOperator(DbtDocsKubernetesOperator):
...
+    assert env_vars["AWS_REGION"] == "us-east-1"
 
-    # Guard the bash -c path: build_kube_args splits "dbt" into self.cmds, so the shell
-    # command must still start with the full "dbt docs generate" and not drop the executable
-    # (regression in the cloud-docs build_and_run_cmd, see PR #2488).
     assert operator.cmds == ["/bin/bash", "-c"]
-    assert operator.arguments[0].startswith("dbt docs generate")
-    assert " && aws s3 sync " in operator.arguments[0]
+    shell_cmd = operator.arguments[0]
+    assert shell_cmd.startswith("dbt docs generate --profile default --target dev --project-dir fake-dir")
```
pankajastro: The comment explaining that these assertions guard the #2488 bug — `build_kube_args` splitting `dbt` into `self.cmds`, so the shell command must not drop the executable — was removed here, and `startswith("dbt docs generate")` became a hard-coded `--profile default --target dev --project-dir fake-dir`.

That makes the test both more brittle and less self-explanatory: any unrelated change to flag ordering now fails it with no indication of which invariant it was defending. Could we keep the explanatory comment and split it into two assertions — one for the `dbt docs generate` prefix (the actual regression guard) and one for the flags?

pankajastro (commented):
Thanks for moving this off the AWS CLI — the `shlex.join` fix and the in-pod uploader are both good directions, and I verified the heredoc mechanics render to valid Python and execute correctly (including the trailing `PY` delimiter and quoted newlines in `--vars`).

I have a set of questions inline, mostly around credential resolution. The one I'd like to resolve before merge is the `get_credentials()` fallthrough on `aws_env_vars_from_connection` — I think it changes behaviour for pod-side IRSA setups and for workers with no ambient credentials. The rest are smaller: dropped `endpoint_url`/`config_kwargs` handling, `python` vs `python3`, missing-file behaviour, an undeclared worker-side provider dependency, an unrelated docs revert, and a few test-coverage gaps.

cosmos/operators/kubernetes.py
jx2lee: I agree this behavior change is worth documenting, especially for users who may have manually quoted arguments as a workaround. I'll add a changelog entry clarifying the expected argument format.

cosmos/operators/kubernetes.py
jx2lee: Narrowing the upload to required_files wasn't intentional; the goal of this change is to replace the AWS CLI dependency rather than change what gets published. I'll update the boto3 uploader to recursively upload the files present under target/, preserving the previous upload scope. Fixed it to upload all target/ dir files.

cosmos/operators/kubernetes.py
jx2lee: The narrowing wasn't intentional.

I've updated the boto3 uploader to recursively walk the generated `target/` directory and upload every file it finds, preserving each file's path relative to `target/` under the configured S3 prefix. This keeps published artifact scope aligned with `aws s3 sync <target> s3://...` behavior, instead of limiting uploads to `required_files`.

required_files` is still kept for existing docs/static handling, but it's no longer used to decide which files get uploaded to S3 in Kubernetes mode.

cosmos/operators/kubernetes.py
jx2lee: Good point. Fixed it command python3 and python. 👍🏽

cosmos/operators/kubernetes.py
jx2lee: Thanks you for pointing out. I wrap import with `MissingPacakge`. 🤗

cosmos/operators/kubernetes.py
jx2lee: Yes. I injected AWS_ENDPOINT_URL_S3 and client kwargs to preserve options.

cosmos/operators/kubernetes.py
jx2lee: I agree this behavior change is worth documenting, especially for users who may have manually quoted arguments as a workaround. I'll add a changelog entry clarifying the expected argument format.

cosmos/operators/kubernetes.py
jx2lee: Narrowing the upload to required_files wasn't intentional; the goal of this change is to replace the AWS CLI dependency rather than change what gets published. I'll update the boto3 uploader to recursively upload the files present under target/, preserving the previous upload scope. Fixed it to upload all target/ dir files.

cosmos/operators/kubernetes.py
jx2lee: Thanks. Keep the `config_kwargs` on top of `hook.region_name`

tests/operators/test_kubernetes.py
jx2lee: Updated the S3 Kubernetes docs test to parametrize the folder prefix and optional AWS env branches. It now covers no folder prefix, trailing-slash normalization, missing region, and missing session token.

tests/operators/test_kubernetes.py
jx2lee: I split it into two assertions. Thanks!

tests/operators/test_kubernetes.py
jx2lee: I added a behavior-level test that extracts the generated heredoc body and executes it with a stubbed `boto3` client.  Now catches syntax/indentation issues in the generated Python source and also verifies the actual recursive S3 upload arguments, including bucket, key, and `ExtraArgs`. WDYT?

cosmos/operators/kubernetes.py
jx2lee: `get_credentials()` is too broad for this path because it can resolve the worker's boto3 credential chain and then inject those credentials into the dbt Pod.

I'll change this to read only the credentials parsed from the Airflow connection via `hook.conn_config` and skip AWS credential env vars when the connection does not define them. That keeps the Pod's own identity chain, including IRSA, untouched.

cosmos/operators/kubernetes.py
jx2lee: This is addressed by the same change that avoids calling `AwsBaseHook.get_credentials()` in the worker. The operator now only injects static credentials explicitly defined on the Airflow connection via `hook.conn_config`;otherwise it leaves the Pod's boto3 credential chain untouched.

That means assume-role/web-identity credentials are no longer pre-resolved before the Pod starts, so they cannot expire while `dbt docs generate` is still running.

jx2lee (commented):
Thanks again for the detailed review. I left replies inline and updated the implementation around the main credential concern.

Kubernetes S3 upload path no longer calls `AwsBaseHook.get_credentials()` on the worker. It only forwards credentials explicitly configured on the Airflow connection, and otherwise lets the Pod resolve credentials through its own boto3 chain.

I also restored the previous upload scope by recursively uploading all files under `target/`, preserved S3 endpoint/client config handling, and added test coverage for the generated upload script and env var injection.

I'd appreciate another look at whether this direction addresses the credential concerns.

pankajastro (approved):
LGTM

pankajastro:
PR: https://github.com/astronomer/astronomer-cosmos/pull/2965 should fix the linkcheck job 403 error

Issue #2960, opened 2026-08-13
[Bug] dbt ls cache key ignores requested --output-keys, so upgrades/config changes can silently serve stale cached node data


tatiana:
### Description

Cosmos's `dbt ls` result cache (`DbtGraph.should_use_dbt_ls_cache`/`save_dbt_ls_cache` in `cosmos/dbt/graph.py`, keyed via `cache._calculate_dbt_ls_cache_current_version` in `cosmos/cache.py`) decides whether to reuse a previously-cached `dbt ls --output json` result by hashing only:

- the dbt project folder's contents (`_create_folder_version_hash`), and
- `dbt_ls_cache_key_args`: `select`/`exclude`/`vars`/`selector`/`--no-partial-parse` (`dbt_ls_args`), plus env vars, plus a configurable allowlist of Airflow Variables (`RenderConfig.airflow_vars_to_purge_dbt_ls_cache`).

Nothing in that hash reflects: (a) which `--output-keys` Cosmos actually requested from `dbt ls`, (b) the Cosmos package version, or (c) `RenderConfig.source_rendering_behavior`, which independently controls whether `--output-keys` is passed at all (`run_dbt_ls`'s `specify_output_keys` logic in `cosmos/dbt/graph.py`).

This means two distinct scenarios silently return stale node data instead of re-running `dbt ls`:

1. **Config-driven (already possible today):** flipping `RenderConfig.source_rendering_behavior` between `NONE` and any other value changes whether `--output-keys` is passed to `dbt ls` at all (with vs without Cosmos's explicit key list), but that option is not part of `dbt_ls_cache_key_args`, so an existing cache entry keeps being served across the flip.
2. **Upgrade-driven:** if a future Cosmos release changes the `--output-keys` list itself (for example, #2959 proposes adding `database`/`schema`/`alias`/`relation_name`) or changes how `parse_dbt_ls_output` interprets the JSON, a user who upgrades Cosmos but doesn't also touch their dbt project files, `select`/`exclude`/`vars`, or the specific Airflow Variables in `airflow_vars_to_purge_dbt_ls_cache` keeps being served their pre-upgrade cached output - silently missing whatever the new Cosmos version expected to be there.

### Use case/motivation

[...]

Issue #2959, opened 2026-08-13
[Feature] Unify Asset/Dataset URI construction across Cosmos execution modes


tatiana:
### Description

Cosmos currently produces the Asset/Dataset URI for a dbt model through three different code paths, one per execution-mode family, and they don't all agree on how the URI is built:

- **WATCHER / WATCHER_KUBERNETES** - the producer calls `compute_model_outlet_uris(target/manifest.json, namespace)` (`cosmos/operators/_watcher/base.py`), which goes through `construct_dataset_uri`, and pushes the result via XCom; the consumer sensor emits `Asset(uri)` via `register_dataset` (`cosmos/operators/watcher.py`). Built from the manifest, no runtime OpenLineage involved.
- **LOCAL / VIRTUALENV** - `get_datasets()` builds the URI via `construct_dataset_uri(output.namespace, output.name)`, but only from `openlineage_events_completes` populated by parsing dbt's OpenLineage artifacts after the run (`cosmos/operators/local.py`), gated by `emit_datasets`. If `openlineage-integration-common` isn't installed, or artifact parsing raises one of several caught exceptions, emission silently drops to zero datasets with no warning.
- **AIRFLOW_ASYNC (BigQuery)** - `_register_event` hardcodes `f"bigquery://{gcp_project}/{dataset}/{table_name}"` (`cosmos/operators/_asynchronous/bigquery.py`), bypassing `construct_dataset_uri` entirely, so it uses a different URI scheme than every other mode.

Because URI construction isn't centralized, nothing guarantees that a URI computed from the manifest at parse time will match what a given execution mode actually emits at runtime for that same model.

### Use case/motivation

This surfaced while designing automatic cross-`DbtDag`/`DbtTaskGroup` dependency resolution (#1321): a consumer DAG needs to compute its external upstream's Asset URI at parse time and trust that it matches what the producer emits at runtime. That holds for WATCHER (both sides derive from the manifest), holds in practice but not by guarantee for LOCAL/VIRTUALENV (OpenLineage-dependent, can silently emit nothing), and never holds for AIRFLOW_ASYNC (divergent URI scheme).

Unifying emission onto one manifest/`construct_dataset_uri`-based path, mode by mode, would make dataset-based scheduling correctness independent of execution mode and unblock features like #1321 that need parse-time URIs to be trustworthy.

### Proposed approach (sketch, open for discussion)

[...]

Issue #1321, opened 2024-11-13
[Feature] Automate the dependency resolution between `DbtDag`s that represent sub-parts of a dbt project


tatiana:
### Description

Automatically create cross-DAG scheduling dependencies when the user splits a dbt project into multiple `DbtDag` instances.

### Use case

Cosmos 1.7 allows users to [map a dbt project into an Airflow DAG](https://astronomer.github.io/astronomer-cosmos/index.html#example-usage-jaffle-shop-project) or TaskGroup. It also allows users to easily convert a subset of dbt nodes into an Airflow DAG or Task Group using [selectors](https://astronomer.github.io/astronomer-cosmos/configuration/selecting-excluding.html). In both cases, it maps 1:1 dbt resources, such as models, into Airflow tasks and resolves their dependencies within the same Task Group or DAG.

However, the use case that Cosmos does not cover yet is to **resolve the dependency of a dbt project split into multiple Airflow DAGs**, affecting how they are triggered/scheduled. This feature aims to handle this use case by creating cross-DAG dependency on behalf of the user.

This feature assumes the DAGs are running within the same Airflow installation (deployment, in the case of Astro).

By resolving dependencies cross-DAGs automatically, it would also be great if we also allow users to easily:
* Assign different schedules per subset of a dbt project while still taking into account model dependencies;
* Assign different owners for subsets of a dbt project;
* Assign different sets of credentials to be used by subsets of a dbt project;

While delegating the responsibility to define the dbt models dependency between those DAGs to Cosmos.

### Current approach

Users can currently handle cross-DAG dependency manually with Cosmos 1.7. One approach is to leverage the Airflow Datasets generated by Cosmos and Airflow's [Data-Aware scheduling](https://astronomer.github.io/astronomer-cosmos/configuration/scheduling.html#data-aware-scheduling).

We've included below the use-case with Airflow Datasets using the [Jaffle shop](https://github.com/astronomer/cosmos-demo/tree/main/dbt/jaffle_shop):

![screenshot](https://astronomer.github.io/astronomer-cosmos/_images/jaffle_shop_dbt_graph.png)

Let's say a team is responsible for maintaining the upstream seeds (`raw_customers`, `raw_orders`, `raw_payments`), and another team is responsible for maintaining the downstream transformations (the other downstream nodes). Seeds are expected to be processed hourly, and once they are ready, the transformations are supposed to be triggered. The organisation wants to have separate DAGs per team for accountability.

[...]

PR #2961, opened 2026-08-13
Proposal for automating cross-DAG dbt dependency resolution

tatiana:
## Description

Cosmos maps a dbt project into one `DbtDag` or `DbtTaskGroup` and resolves dependencies within it. It does not resolve dependencies when a project is split across multiple `DbtDag`s or `DbtTaskGroup`s - for example, one team owns the seeds, another owns the downstream transforms, and the transforms should only run after the seeds are refreshed. Today, users hand-write `schedule=[Dataset(...)]` conditions or task sensors themselves, and that hand-wiring drifts whenever the project or its `select`/`exclude` scoping changes.

This PR adds a request-for-comments proposal for automating that dependency resolution.

This PR is proposal-only. It does not implement anything or change runtime behavior.

The proposed design:

- a parse-time structural resolver that figures out, for any selected subset of a dbt project, which of its parent nodes live outside that subset;
- `RenderConfig.auto_schedule` for separate `DbtDag`s, which turns those external parents into a Dataset/Asset condition on the DAG's schedule;
- a `DbtDependencyCoordinator` for `DbtTaskGroup`s split within one DAG, which wires direct task dependencies instead;
- explicit handling for the cases that would otherwise silently produce a wrong or missing dependency (ephemeral models, dbt sources, execution modes that can't emit an Asset at all, multi-owner or unowned nodes).

Two gaps surfaced while writing this proposal turned into their own tickets, which this proposal depends on:

- [astronomer-cosmos#2959](https://github.com/astronomer/astronomer-cosmos/issues/2959) - how Cosmos builds the Asset URI for a dbt model differs by execution mode today, so a URI computed at parse time isn't guaranteed to match what gets emitted at runtime.
- [astronomer-cosmos#2960](https://github.com/astronomer/astronomer-cosmos/issues/2960) - the `dbt ls` result cache doesn't account for changes to what Cosmos requests from `dbt ls`, so an upgrade can silently keep serving stale cached data.

The areas where feedback would be especially valuable:

1. Whether splitting the resolver into a structural piece (works under any dbt load mode) and a separate URI-enrichment piece (needs relation identity) is the right shape.
2. Whether a standalone `DbtDag` needs a `DbtDagGroup`-style construct (or a registry) to verify it isn't scheduling on an unowned or double-owned node, since it has no visibility into sibling DAGs today.
3. Whether the execution-mode limitations (which modes can emit an Asset at all, and the `WATCHER` family's single-producer architecture) are handled reasonably.

Full proposal: [Automate dependency resolution across a dbt project split into multiple DbtDags or DbtTaskGroups](plans/002-automate-cross-dag-dependency-resolution.md)

## Related Issue(s)

Relates to #1321. This proposal PR is intended to host the design discussion; it does not close the issue.

## Breaking Change?

No. This PR adds documentation only and does not change runtime or public API behavior.


pankajastro (commented):
Verified the codebase claims against 1.15.1 — line refs, the seven non-emitting modes, the WATCHER single-producer architecture, F2's init ordering, and the `schedule`-frozen/`timetable`-mutable finding (confirmed in both Airflow 3.0 and 3.2 sources) all hold. Seven things I think need changing before this is implementable, inline.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,676 @@
...
+  LOCAL emission is "silent-skip or hard-fail" depending on the error, never a guaranteed emit.
+- **F5 - the boundary resolver cannot verify that a matching producer exists or will emit, and this
+  splits into a structural half and a harder, mode/namespace half.** A node being outside a unit's own
+  selection only proves that unit doesn't own it; it proves nothing about whether another
+  `DbtDag`/`DbtTaskGroup` actually owns it, or whether exactly one unit claims it (vs. two overlapping
+  selections, or zero). This **structural** ownership question - does a rendered Airflow task exist for
+  this `unique_id`, and does that task do real work - is answerable today with no manifest and no
+  registry *when the peer units are all visible in one place*, which is exactly how
+  `DbtDependencyCoordinator` is constructed (`DbtDependencyCoordinator([tg_a, tg_b, ...])` - see the
+  coordinator section below). "Does real work" matters because `SeedRenderingBehavior.RENDER_ONLY`
```
pankajastro, line 172: `ephemeral_models_as_empty_operator` defaults to `True` (`cosmos/config.py:113`) and `graph.py:449` keeps the resulting `EmptyOperator` in `tasks_map`. Sources without freshness under `SourceRenderingBehavior.ALL` do the same (`graph.py:526-536`). Both pass this ownership check, so the coordinator accepts a producer that gates on nothing — and the ephemeral case is the default config. Placeholder set should be all three, not just `RENDER_ONLY`.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,676 @@
...
+def resolve_external_uris(
+    boundary: GraphBoundary, nodes: dict[str, DbtNode], namespace: str
+) -> dict[str, str]:
+    # external_parent_unique_id -> Asset URI, built from each DbtNode's relation identity (F1) -
+    # wherever that identity came from: the manifest today, or `dbt ls` once #2959/#2960 land.
+    # Takes the node map, not a manifest, so it isn't tied to LoadMode.DBT_MANIFEST.
+    ...
+```
+
+`resolve_graph_boundary` only needs `DbtNode.unique_id`/`depends_on`/`resource_type`/`config` - none of
```
pankajastro, line 339: `config` isn't guaranteed here. `parse_dbt_ls_output` does `config=node_dict.get("config") or {}` (`cosmos/dbt/graph.py:439`), so under `DBT_LS_FILE` with a user-supplied file, or the bare-`dbt ls` branch (`run_dbt_ls:832-857` — the comment there notes dbt Fusion returns only 5 keys), `has_ephemeral_materialization` is `False` for every node. That's indistinguishable from genuinely non-ephemeral, so the resolver emits a wrong boundary silently — the thing constraint 9 forbids.

Also: `LoadMode` has six values, not two. `AUTOMATIC` falls back to `load_via_custom_parser()` (no `config` at all) when `dbt ls` raises `FileNotFoundError`; `DBT_LS_FILE`/`DBT_LS_CACHE` aren't mentioned. Each needs an explicit supported/raise decision.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,676 @@
...
+Structural edges and Asset URIs are deliberately kept separate, because the two consumers need different
+things: `DbtDependencyCoordinator` (S3, below) only ever needs `owned`/`external_edges`/
+`external_sources` - it wires Airflow tasks directly and never touches a URI, so gating it on F1 would
+have rejected valid `DBT_LS`-based `DbtTaskGroup` splits for no reason. `auto_schedule` (S2, below) needs
+the structural boundary *and* `resolve_external_uris`'s output, since it schedules on Assets. Only the
+latter function needs relation identity (F1) - via the manifest today, or via `dbt ls` once
+[astronomer-cosmos#2959](https://github.com/astronomer/astronomer-cosmos/issues/2959)/[astronomer-cosmos#2960](https://github.com/astronomer/astronomer-cosmos/issues/2960)
+land (constraint 7); it should not be written against `LoadMode.DBT_MANIFEST` specifically.
+
+**Non-emitting-boundary rule (F3, mandatory, structural):** when an external parent cannot itself be
```
pankajastro, line 365: This is an Asset-emission rule, but S3 never touches Assets — it wires `task(e) >> task(c)` directly, which works precisely because ephemeral models are rendered as tasks to preserve the chain (`graph.py:449`). Making traversal mandatory makes S1 raise unsupported-topology under `DBT_LS` when the ephemeral ancestor isn't in the peer union, rejecting valid `DbtTaskGroup` splits — the outcome :356-360 says it wants to avoid.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,676 @@
...
+| `DbtDagGroup` / `DbtDagFamily` (ticket alternative b) | Not purely a follow-up: `auto_schedule` on a standalone `DbtDag` can't verify ownership across siblings without something like this (open question 5). Variants: split by dbt tags, per-model Cosmos config, or a Python grouper. |
+| Match on runtime-emitted Assets instead of manifest URIs | Rejected - the edge is only knowable after the run, but the consumer needs it at parse time. |
+| A producer registry stored in an Airflow Variable, used to *compute* the join key | Rejected as unnecessary for that - the URI is already a deterministic join key. A registry may still be needed to *verify* ownership across `DbtDag`s (open question 5); that's separate from this rejection. |
+
+## Changes by module (`astronomer-cosmos`)
+
+| Module | Change |
+|---|---|
+| `cosmos/dataset.py` (or a new `cosmos/dependencies.py`) | `resolve_graph_boundary` (structural, `owned`/`external_edges`/`internal_edges`/`external_sources`, no manifest needed) and a separate `resolve_external_uris` (needs relation identity, F1). Includes the F3 ephemeral traversal with the owned-re-entry check, the test-node exclusion rule, and fail-fast behavior. |
```
pankajastro, line 498: `compute_model_outlet_uris` silently `continue`s when `not all([database, schema, alias])` (`cosmos/dataset.py:317-321`). If `resolve_external_uris` resolves a URI for a node the producer skips, the schedule waits forever. Requiring the consumer to raise isn't enough — the producer side has to be made consistent, or #2959 has to cover it.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,676 @@
...
+| `cosmos/config.py` | New `RenderConfig.auto_schedule: DbtUpstreamUpdated \| None = None`; new `DbtUpstreamUpdated` enum (`AND`/`OR`) exported from `cosmos`. Reject `auto_schedule` on a `DbtTaskGroup`'s `RenderConfig` (it mutates a DAG's `schedule`; a `TaskGroup` isn't one). |
+| `cosmos/converter.py` (F2, constraint 8) | At the end of `DbtToAirflowConverter.__init__`, for `auto_schedule`: build the combined timetable, set `self.timetable`, call `attrs.validate(self)` to re-run `catchup`/`allowed_run_types`/`params` validation, and separately re-check `max_active_runs` vs `active_runs_limit`. No change needed to `cosmos/airflow/dag.py`. |
```
pankajastro, line 500: `set self.timetable` silently no-ops on `DbtTaskGroup`. `DbtToAirflowConverter` is a mixin shared with `DbtDag` (`cosmos/airflow/dag.py:14`); on a TaskGroup this just creates an unused attribute, no error. The rejection is load-bearing, but `RenderConfig` is a plain dataclass with no knowledge of its consumer — the guard has to live in the converter path.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,676 @@
...
+  silently never fires.
+- **Unsupported adapter** - `get_dataset_namespace` returns `None`, so there is no dataset to schedule
+  on; this must fail with a clear error rather than silently no-op.
+- **No Airflow version offers a native time-AND-dataset schedule.** Airflow exposes
+  `DatasetOrTimeSchedule` (Airflow 2) / `AssetOrTimeSchedule` (Airflow 3), but there is no AND
+  equivalent, including on Airflow `main` as of this writing. Short term: document the "emit a daily
+  dataset" workaround and let users control how Cosmos's dependency condition combines (`AND`/`OR`)
+  with their own datasets. This is not blocked solely on Airflow shipping a native schedule type -
+  Cosmos could emulate one (see the roadmap's follow-up slice below).
+- **Source freshness** is not handled by this proposal. **Cycles or ambiguous producers** (overlapping
```
pankajastro, line 541: Cycle/ambiguous-producer detection is implemented only in the coordinator row (:503), which uses its own peer list. A standalone `DbtDag` has zero peer visibility (open question 5), so two `DbtDag`s with mutually-external parents each gate on the other's Assets and neither ever runs — undetectable at parse time. This belongs in open question 5, not Edge cases.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,676 @@
...
+  selections claiming the same node) must be detected and raised, not silently resolved one way.
+- **F5, split by consumer.** For the **coordinator**, ownership (does exactly one peer own this node) is
+  checkable today from its own peer list - no gap. For **`auto_schedule`**, a standalone `DbtDag` can't
+  see its siblings at all, so it can't check ownership, `emit_datasets`, mode capability, or namespace
+  match without the peer-visibility mechanism open question 5 is about. Both must raise, never silently
+  omit the dependency, once the applicable check is in place.
+- **A source is legitimately ownerless - don't fail-closed on it.** Constraint 9 raises on an "ownerless"
```
pankajastro, line 548: Sources aren't always unwireable: under `SourceRenderingBehavior.ALL` / `WITH_TESTS_OR_FRESHNESS` they render as real `DbtSource*Operator` freshness tasks (`graph.py:500-536`). If `tg_a` owns the freshness task and `tg_b`'s models depend on those sources, routing them all to `external_sources` drops a real ordering edge. The exemption is right for `auto_schedule`, wrong for S3.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,630 @@
...
+  values, but these are pre-resolution and usually unset - not the resolved relation.
+  [astronomer-cosmos#2959](https://github.com/astronomer/astronomer-cosmos/issues/2959) is the ticket
+  covering this: extending the requested `--output-keys` and `DbtNode` to carry relation identity
+  independent of `LoadMode`.
+- **F2 - `schedule` is consumed before Cosmos parses the graph.** `DbtDag.__init__`
+  (`cosmos/airflow/dag.py`) calls `DAG.__init__` first, which normalizes `schedule`, and only then
+  calls `DbtToAirflowConverter.__init__`; the converter parses the dbt graph and builds the Airflow
+  graph later in its own `__init__` (`cosmos/converter.py`, `self.dbt_graph.load(...)` followed by
+  `build_airflow_graph(...)`). A converter-stage mutation of `schedule`/`timetable` is therefore too
+  late - confirmed still true against Cosmos 1.15.1.
```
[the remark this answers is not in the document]
tatiana: You are right that it was `schedule` reassignment that is too late, not `timetable` patching. Fixed in 8bddad1a: F2 now says only that a converter-stage mutation of `schedule` itself is too late, and defers `timetable` to constraint 8, which is where it is actually resolved.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,630 @@
...
+   never in the result. `DBT_MANIFEST` doesn't have this problem (the full project graph is always
+   loaded). So even after #2959/#2960 land, `auto_schedule` on a standalone `DbtDag` under `DBT_LS` may
+   still need to fail-fast on an external ephemeral ancestor it can't see - see open question 1.
+8. **Schedule injection (F2) - resolved: patch `timetable` inside the converter, don't move loading
+   before `DAG.__init__`.** Tested directly against Airflow 3.2.0rc1: `dag.schedule = X` after
+   construction raises `FrozenAttributeError` (schedule is frozen). `dag.timetable = X` after
+   construction works fine, and a serialize/deserialize round-trip confirms the patched timetable is
+   what the scheduler actually sees (`schedule` isn't even serialized). So `auto_schedule` can just set
+   `self.timetable` at the end of `DbtToAirflowConverter.__init__`, after computing the boundary and
+   URIs as usual - no need to load anything before `DAG.__init__`, and no risk of loading the graph
```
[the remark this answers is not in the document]
tatiana: Fixed in 4532fc02, at all three locations (this one, :413 and :469).

Worth recording the nuance, because it changes the fix: the converter is a mixin, not a standalone class that receives the DAG. `DbtDag(DAG, DbtToAirflowConverter)` (cosmos/airflow/dag.py:14) means `self` really is the DAG in that one case, which is why the wording survived this long. But `DbtTaskGroup(TaskGroup, DbtToAirflowConverter)` (cosmos/airflow/task_group.py:18) means `self` is a TaskGroup, and the converter can also be constructed directly with a `dag=` argument and no subclassing. So `self.timetable` is correct by accident in one of three cases.

The proposal now says to patch `dag.timetable`, reusing the DAG the converter already resolves for `build_airflow_graph` (`dag or (task_group and task_group.dag)`, cosmos/converter.py:408).

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,630 @@
...
+   Airflow validates several things against the timetable once, at construction, and none of it re-runs
+   automatically when `timetable` is patched afterward: `catchup`/`start_date`, `allowed_run_types`,
+   required-params-without-defaults, and `max_active_runs` vs `active_runs_limit`. Confirmed
+   empirically - constructing a DAG directly with an asset schedule and `catchup=True` but no
+   `start_date` raises; constructing with `schedule=None` (so the check doesn't fire) and then patching
+   `timetable` to the same asset condition does not, silently accepting an invalid combination.
+
+   Don't reimplement each check by hand - call `attrs.validate(self)` after patching. `catchup`,
+   `allowed_run_types`, and `params` are all `attrs` field validators that read `self.timetable`
+   dynamically, so `attrs.validate()` re-runs them against the patched value and correctly raises
```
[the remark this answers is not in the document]
tatiana: Fixed in 4532fc02: `attrs.validate(dag)`, not `attrs.validate(self)`, and the validators are described as reading the DAG's own `timetable`. Added the reason it matters beyond naming: on a `DbtTaskGroup` or a bare converter, `attrs.validate(self)` raises `NotAnAttrsClassError` rather than validating anything.

plans/002-automate-cross-dag-dependency-resolution.md
tatiana: Confirmed all three and fixed in 917a93da. F5 now carries a table of the placeholder cases with references:

- `SeedRenderingBehavior.RENDER_ONLY` (cosmos/airflow/graph.py:429-443)
- ephemeral models under `ephemeral_models_as_empty_operator`, default True (cosmos/config.py:113, cosmos/airflow/graph.py:449-475)
- sources with `has_freshness is False` under `SourceRenderingBehavior.ALL` (cosmos/airflow/graph.py:526-537)

Your point that the ephemeral case is the default is the one that made this a correctness problem rather than a completeness one, so the doc calls that out explicitly: a `RENDER_ONLY`-only screen is wrong in the most common configuration. The screen is now stated as "`EMPTY_OPERATOR_CLASS_PATH` in `tasks_map`", which covers all three.

I also added the two cases that render no task at all, since the check has to tell "placeholder" from "absent": `SeedRenderingBehavior.NONE` (graph.py:426-428), and a source dropped by `SourceRenderingBehavior.NONE`, by `WITH_TESTS_OR_FRESHNESS` with neither freshness nor tests (graph.py:504-509), or by `source_pruning` (graph.py:511-516).

One related consequence, in a3bbc771: an ephemeral placeholder is the one placeholder kind the coordinator may legitimately wire *through*, because its own upstream chain is intact inside its group. It is still an invalid terminal producer. The doc now states that so the two rules do not read as contradicting each other.

plans/002-automate-cross-dag-dependency-resolution.md
tatiana: Both confirmed and fixed in 917a93da, which replaces the two-mode framing with a six-row supported-or-raise matrix.

On `config`: `parse_dbt_ls_output` does `config=node_dict.get("config") or {}` (cosmos/dbt/graph.py:439), so absent config arrives as `{}`, not `None`, and no per-node check can tell it from a declared non-ephemeral. The doc now gates the ephemeral decision on the resolved `load_method` plus a positive check that the selection's model nodes carry a `materialized` key, and raises otherwise.

Two refinements while verifying:

- The bare `dbt ls` branch is narrower than "the bare branch": it fires only when `settings.pre_dbt_fusion` is set *and* `source_rendering_behavior` is `NONE` (cosmos/dbt/graph.py:831-833). dbt Fusion's default keys carry no `config` (the comment at :827-830), so it is a real hazard, just an opt-in one. Documented as its own path.
- `LoadMode.CUSTOM` is worse than "no config at all". It does build a `config` dict, from `config_selectors` (graph.py:1136), but those only carry materialized/schema/tags (parser/project.py:45) and `materialized:view` is *substituted* when the parser sees no materialization (parser/project.py:427-428). So an ephemeral model reports as a view: a false negative that looks like a confident positive. `CUSTOM` is therefore unsupported and raises.

`AUTOMATIC` is now decided from `DbtGraph.load_method` after `load()`, never the requested mode, precisely because of the `FileNotFoundError` fallback to `CUSTOM` you pointed at (graph.py:803-813). `DBT_LS_FILE` raises unless `config` is actually present; `DBT_LS_CACHE` follows `DBT_LS` once #2960 lands.

plans/002-automate-cross-dag-dependency-resolution.md
tatiana: Agreed, and this was the most consequential of the seven. Fixed in a3bbc771.

Traversal is now a `traverse_non_emitting` flag on `resolve_graph_boundary`, defaulting to `False`. S2 passes `True` because an ephemeral node emits no Asset. S3 leaves it off: as you say, ephemeral models are still rendered as tasks (graph.py:449-475), so `task(e) >> task(c)` is a real ordering edge - `e`'s own upstream `a` is wired to it by normal rendering inside its group, so `a >> e >> c` transitively gates `c` on `a`. `GraphBoundary` gained `ephemeral_external_parents` so the boundary reports them either way and each consumer decides, and there is a per-consumer table spelling out the four combinations.

This also removes the over-strict raise you identified: with the default flag, a `DBT_LS` peer union that does not contain the ephemeral ancestor no longer raises unsupported-topology, so the `DbtTaskGroup` split keeps working.

plans/002-automate-cross-dag-dependency-resolution.md
tatiana: Confirmed and fixed in 9969ee7d. You are right that a consumer-side raise cannot catch this - from the consumer's side nothing looks wrong.

The proposal now makes the producer-side fix a constraint 5 prerequisite for S2, either as part of #2959 or owned here: `compute_model_outlet_uris` must raise, or at minimum log per skipped node, instead of the bare `continue` at cosmos/dataset.py:317-322. I also added the wider version of the same hazard, cosmos/dataset.py:297-308, which returns `{}` at warning level for an unreadable or unparseable manifest, and the fact that the producer reads its run's `target/manifest.json` while the consumer resolves from the project manifest at parse time, so the two can disagree even when all three keys are present.

There is a corresponding test in the Testing section, scoped as a test of the producer change rather than of `resolve_external_uris`.

plans/002-automate-cross-dag-dependency-resolution.md
tatiana: Confirmed and fixed in 4532fc02. The guard now lives in the converter path, and the `cosmos/config.py` row says explicitly that the rejection cannot live in `RenderConfig` for the reason you give - it is a plain dataclass with no knowledge of its consumer.

One thing I found while verifying that makes the guard stronger than "silently no-ops": for a `DbtTaskGroup`, the DAG the converter resolves is the *enclosing user DAG* (`task_group.dag`, cosmos/converter.py:408). So an implementation that correctly avoids `self.timetable` and patches the resolved `dag` instead would silently rewrite the schedule of a DAG the user never asked Cosmos to own - worse than either a no-op or an error. The doc now requires a `CosmosValueError` when `auto_schedule` is set and `task_group` is not `None`, or when the converter does not exclusively own its `dag`, and the test asserts the enclosing DAG is left untouched.

plans/002-automate-cross-dag-dependency-resolution.md
tatiana: Agreed on both the gap and the placement. Fixed in 9969ee7d.

The Edge cases bullet no longer claims cycles and ambiguous producers are generally detectable; it says they are the coordinator's job where peers are visible in one parse, and open question 5 otherwise. Open question 5 now carries your mutual-deadlock case explicitly: two `DbtDag`s whose selections are each other's external parents each gate on the other's Assets and neither ever runs, with no parse-time error because neither parse can see the other DAG. That raises the bar on whatever mechanism answers the question - it has to make the peer set inspectable enough to reject that shape, not just confirm single ownership of one node at a time.

plans/002-automate-cross-dag-dependency-resolution.md
tatiana: Confirmed and fixed in a3bbc771, with a follow-up in 3322726a.

Proposed design now has a source-rendering table splitting the S2 and S3 answers per row: a source with freshness or tests under `ALL`/`WITH_TESTS_OR_FRESHNESS` renders a real `DbtSource*Operator` task (graph.py:500-537) and the coordinator wires `task(source) >> task(consumer)` like any other external parent; the freshness-less `ALL` placeholder and the no-task cases stay exempt. Your `tg_a` owns the freshness task / `tg_b` reads the source example is the test case.

Worth noting the freshness-less `ALL` placeholder is exempt for a different reason than the ephemeral placeholder is wireable: a source has no dbt upstreams, so that `EmptyOperator` succeeds immediately and gates nothing, whereas an ephemeral placeholder inherits a real upstream chain.

3322726a also scopes the F3 constraint text, which still said a source is "not a dependency to wait on" - true for Assets, wrong for the coordinator.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,909 @@
...
+
+**Status:** Draft - request for comments
+
+**Author:** [Tatiana Al-Chueyr](https://github.com/tatiana)
+
+**Created:** 2026-06
+
+**Discussion:** [astronomer-cosmos#1321](https://github.com/astronomer/astronomer-cosmos/issues/1321)
+
+**Tracking ticket:** BOSS-269 (internal Linear ticket, "Automate the dependency resolution between DbtDags that represent sub-parts of a dbt project")
```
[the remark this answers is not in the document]
tatiana: Removed in c8e40d63. #1321 is already listed as the public discussion link, so the internal identifier was redundant as well as unhelpful for external contributors.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,909 @@
...
+
+### Every dbt model already has a location-independent identity
+
+A model, seed, or snapshot maps to a stable Asset/Dataset URI built from its warehouse relation - the
+`postgres://0.0.0.0:5432/postgres.public.raw_customers` shape the ticket shows:
+
+- `construct_dataset_uri(namespace, "database.schema.alias")` (`cosmos/dataset.py:206`).
+- `compute_model_outlet_uris(manifest_path, namespace)` (`cosmos/dataset.py:279`) reads
+  `manifest.json`, filters to model/seed/snapshot, and returns
+  `{unique_id: [f"{namespace}/{database}.{schema}.{alias}"]}`.
```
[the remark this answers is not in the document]
tatiana: Fair, and checking this turned up a real inaccuracy rather than just a maintenance risk: the section was headed "verified against the last released version of Cosmos, 1.15.1", but the line numbers were taken from `main`, and the v1.15.1 tag and `main` had diverged - so several citations pointed at the wrong line in the version the header named.

Fixed in c8e40d63 by pinning rather than delisting. The header now states that every `path:line` is against `main` at 6dd07a74 (with a permalink), that each citation also names its symbol so it stays findable once lines move, and that the behaviour claims hold for released 1.15.1 even though the same code sits at different lines there. The "verified against 1.15.1" heading is gone, since the numbers were never from the tag.

Kept the line numbers rather than replacing them with symbols alone: a lot of these point at a specific branch or `continue` inside a long function, where the symbol name alone would not locate the claim. The last loose one ("around line 831") is now precise.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,912 @@
...
+runtime-emitted URI:
+
+Cosmos has 13 `ExecutionMode` values (`cosmos/constants.py`): `WATCHER`, `WATCHER_KUBERNETES`,
+`WATCHER_GCP_GKE`, `LOCAL`, `VIRTUALENV`, `AIRFLOW_ASYNC`, `DOCKER`, `KUBERNETES`, `AWS_EKS`, `AWS_ECS`,
+`AZURE_CONTAINER_INSTANCE`, `GCP_CLOUD_RUN_JOB`, `GCP_GKE`. Most of them cannot emit an Asset at all - a
+distinct, worse category than "conditional emission":
+
+| Mode | Emitted URI source | Matches the manifest URI? |
+|---|---|---|
+| WATCHER / WATCHER_KUBERNETES / WATCHER_GCP_GKE | Producer parser calls `compute_model_outlet_uris(target/manifest.json, namespace)` (`cosmos/operators/_watcher/base.py`, shared by the Kubernetes/GKE variants via `_k8s_common.py`) and pushes it via XCom; the consumer sensor emits `Asset(uri)` via `register_dataset` (`cosmos/operators/watcher.py`). No runtime OpenLineage events involved. All three share the single-producer architecture discussed in the `DbtTaskGroup` coordinator section below. | Yes - built straight from the manifest. |
```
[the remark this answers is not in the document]
tatiana: Correct on both counts - fixed in 0ebc40ac.

`cosmos/operators/watcher.py` holds the *producer*, which overrides `_handle_datasets` to a no-op precisely because emission is the consumer's job. The consumer emits in `DbtWatcherConsumerBaseOperator._emit_datasets` (`cosmos/operators/_watcher/base.py`), which builds `Asset(uri=...)` per URI and delegates to `cosmos.dataset.register_dataset_on_task` - explicitly not the local operator's `LocalDbtBaseOperator.register_dataset`, so it works for consumers that don't inherit from it. `register_dataset` unqualified was also wrong: that name only exists on the local operator.

Your suppressed comment on the `AssetAll(*uris)` example is right too and is fixed in the same commit: `resolve_external_uris` returns strings, the schedule expressions take `Asset` objects, so the compatibility matrix now wraps them first - the same construction `_emit_datasets` uses.

plans/002-automate-cross-dag-dependency-resolution.md
```diff
@@ -0,0 +1,937 @@
...
+| LOCAL / VIRTUALENV | After the run, `get_datasets()` builds URIs via `construct_dataset_uri(output.namespace, output.name)` from `openlineage_events_completes` (`cosmos/operators/local.py`), gated by `emit_datasets`. | In practice yes (same OpenLineage standard the ticket's Cosmos-1.7 venv example relies on), **but emission is conditional** - see F4. |
+| AIRFLOW_ASYNC (BigQuery) | `_register_event` hardcodes `f"bigquery://{gcp_project}/{dataset}/{table_name}"` (`cosmos/operators/_asynchronous/bigquery.py`). | No - a divergent scheme that bypasses `construct_dataset_uri` and won't match. |
```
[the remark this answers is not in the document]
tatiana: Correct - fixed in 26734611. `_register_event` (`cosmos/operators/_asynchronous/bigquery.py:259-270`) branches on `AIRFLOW_VERSION.major`, mirroring `construct_dataset_uri`'s delimiter switch without calling it, so the row now shows both forms.

Correcting it also let me sharpen the argument it was supporting, because the delimiter is not actually the reason these URIs fail to match. Three things diverge on *both* versions: the scheme is a hardcoded `bigquery://` rather than `get_dataset_namespace`'s output; the name is `unique_id.split(".")[-1]`, so a model with a custom `alias` gets its model name rather than the resolved relation; and the name is never `urllib.parse.quote`d. So AIRFLOW_ASYNC would not match even on the version whose delimiter convention it happens to follow - which is why #2959 routing it through `construct_dataset_uri` is the fix, not a per-version adjustment.

PR #2972, opened 2026-08-20
Vuln: Require authentication on the Airflow 3 Cosmos dbt docs routes

iancmoritz:
## Description

The Airflow 3 plugin mounts its FastAPI sub-app under `/cosmos` with `app.mount()`. Airflow 3 attaches auth per-router via `Depends(...)` rather than as global middleware, so a mounted sub-app inherits nothing — every dbt docs route (`dbt_docs`, `dbt_docs_index.html`, `manifest.json`, `catalog.json`) was reachable without credentials, exposing model SQL and schema metadata. The Airflow 2 plugin guards the equivalent views with `@has_access(MENU_ACCESS_PERMISSIONS)`; this restores parity.

```python
app = FastAPI(dependencies=cosmos_auth_dependencies())
# cosmos_auth_dependencies() -> [Depends(requires_access_view(AccessView.PLUGINS))]
# import failure -> [Depends(deny_all)]  # 403, fail closed
```

`AccessView.PLUGINS` maps to the viewer role in `SimpleAuthManager`, matching the AF2 `RESOURCE_WEBSITE` read permission, and `get_user` accepts the API-server JWT cookie, so the in-UI iframe keeps working.

Reproduced and verified against a local Airflow 3.3 cluster (Astro CLI runtime `3.3-4`, `SimpleAuthManager` with `all_admins=False`, one project configured via `[cosmos] dbt_docs_projects`):

| request | before | after |
| --- | --- | --- |
| unauthenticated `GET /cosmos/jaffle/manifest.json` | 200 + full manifest | 401 |
| unauthenticated `catalog.json` / `dbt_docs_index.html` / `dbt_docs` | 200 | 401 |
| `Authorization: Bearer <token>` or `_token` cookie | 200 | 200 |

## Related Issue(s)

Fixes the code scan finding "Airflow 3 FastAPI dbt-docs plugin routes have no authentication (regression from Airflow 2)" (`sfind-94bfb1c9f5af4cd48621524b82e3ce85`).

## Breaking Change?

Callers that scraped these endpoints anonymously must now authenticate. This is the intended behavior and matches Airflow 2.

## Checklist

Devin-Org: engineering

Link to Devin session: https://app.devin.ai/sessions/e7bbf6409f75432c8acf3264279353dc

---

#### Devin Review
| Status | Commit |
|---|---|
| ⚪ Not started | — |

> [Run Devin Review](https://staging.itsdev.in/review/COG-GTM/astronomer-cosmos/pull/1?analyze=true)

<a href="https://staging.itsdev.in/review/COG-GTM/astronomer-cosmos/pull/1" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-staging-dark.svg?v=1">
    Open in Devin Review (Staging)
  </picture>
</a>

### DIFF
--- a/cosmos/plugin/airflow3.py
+++ b/cosmos/plugin/airflow3.py
@@ -18,7 +18,7 @@ except ImportError:
     from airflow.configuration import conf
 from airflow.plugins_manager import AirflowPlugin
 from airflow.sdk import ObjectStoragePath
-from fastapi import FastAPI
+from fastapi import Depends, FastAPI, HTTPException, status
 from fastapi.responses import HTMLResponse, JSONResponse, Response
 from packaging.version import Version
 
@@ -130,9 +130,32 @@ def _load_projects_from_conf() -> dict[str, dict[str, str | None]]:
     return projects
 
 
+def cosmos_auth_dependencies() -> list[Any]:
+    """
+    Build the authentication/authorization dependencies applied to every Cosmos route.
+
+    Airflow 3 attaches authentication to its own routers, so a plugin sub-app mounted with
+    ``app.mount()`` does not inherit it. This is the AF3 equivalent of the ``@has_access`` guard used
+    by the Airflow 2 plugin: the caller must be authenticated and allowed to read plugin views.
+    If the Airflow security helpers cannot be imported, the routes fail closed.
+    """
+    try:
+        from airflow.api_fastapi.auth.managers.models.resource_details import AccessView
+        from airflow.api_fastapi.core_api.security import requires_access_view
+    except ImportError:
+        logger.warning("Airflow API auth helpers are unavailable: denying access to the Cosmos dbt docs routes.")
+
+        def deny_all() -> None:
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Forbidden")
+
+        return [Depends(deny_all)]
+
+    return [Depends(requires_access_view(AccessView.PLUGINS))]
+
+
 def create_cosmos_fastapi_app() -> FastAPI:  # noqa: C901
     ensure_airflow_version_supported()
-    app = FastAPI()
+    app = FastAPI(dependencies=cosmos_auth_dependencies())
 
     projects = _load_projects_from_conf()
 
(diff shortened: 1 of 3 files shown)


cosmos/plugin/airflow3.py
```diff
@@ -130,9 +130,32 @@ def _load_projects_from_conf() -> dict[str, dict[str, str | None]]:
...
+        from airflow.api_fastapi.core_api.security import requires_access_view
+    except ImportError:
+        logger.warning("Airflow API auth helpers are unavailable: denying access to the Cosmos dbt docs routes.")
+
+        def deny_all() -> None:
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Forbidden")
+
+        return [Depends(deny_all)]
+
+    return [Depends(requires_access_view(AccessView.PLUGINS))]
```
tatiana, line 153: Would it make sense to use `AccessView.WEBSITE` instead?

 My understanding is that, with the `FabAuthManager`, `AccessView.PLUGINS` maps to the Plugins resource, which standard `Viewer` and `User` roles cannot read. This means only users with roles like Op and Admin would be able to use the dbt Docs plugin in Airflow 3. This would prevent those Airflow 3 users from accessing dbt Docs - different from the Airflow 2 users.

To restrict access to Op/Admin feels like a breaking change from Cosmos has behaved so far in Airflow 3 - and in relation to the Airflow 2 behaviour.

cosmos/plugin/airflow3.py
```diff
@@ -130,9 +130,32 @@ def _load_projects_from_conf() -> dict[str, dict[str, str | None]]:
...
+
+
 def create_cosmos_fastapi_app() -> FastAPI:  # noqa: C901
     ensure_airflow_version_supported()
-    app = FastAPI()
+    app = FastAPI(dependencies=cosmos_auth_dependencies())
```
tatiana, line 158: Could we make this compatible with Airflow 3.1.0? It seems cookie support was added in 3.1.1.

In Airflow, 3.1.0,  `get_user dependency` only reads bearer headers - so we can't rely only on this.

tatiana (commented):
@iancmoritz thanks a lot for this vulnerability fix! It would be great if you could address the two inline comments before we approve and merge.

PR #2977, opened 2026-08-25
Keep ProjectConfig relative paths when the project root comes from RenderConfig

### LINKED ISSUE #2926: [Bug] `ProjectConfig` relative paths are ignored when the project root comes from `RenderConfig`
## Description

Cosmos supports configuring the dbt project root either through:

1. `ProjectConfig.dbt_project_path`; or
2. separate `RenderConfig.dbt_project_path` and `ExecutionConfig.dbt_project_path` values.

However, `ProjectConfig` only stores its configured model, seed, and snapshot relative paths when `ProjectConfig.dbt_project_path` is set.

When the project root comes from `RenderConfig`/`ExecutionConfig`, explicit relative paths are silently discarded. Under `LoadMode.CUSTOM`, `LegacyDbtProject` consequently falls back to the default `models`, `seeds`, and `snapshots` directories and may render an incomplete or empty DAG.

This is a pre-existing limitation of the singular `models_relative_path`/`seeds_relative_path`/`snapshots_relative_path` arguments, not a regression introduced by #2875. The new plural arguments added in #2875 inherit the same behaviour.

## Reproduction

Given a project containing:

```text
split_path_project/
└── custom/
    └── models/
        └── my_model.sql
```

the following configuration discards `custom/models`:

```python
project_config = ProjectConfig(
    manifest_path=manifest_path,
    project_name="split_path_project",
    models_relative_paths=["custom/models"],
)

render_config = RenderConfig(
    load_method=LoadMode.CUSTOM,
    dbt_project_path=project_dir,
)

    dbt_project_path=project_dir,
)
```

Observed:

```python
assert project_config.models_paths == []
```

When the custom parser runs, it searches the default `models/` directory instead of `custom/models/`, so `my_model` is not included in the rendered graph.

The deprecated singular argument has the same limitation:

```python
ProjectConfig(
    manifest_path=manifest_path,
    project_name="split_path_project",
    models_relative_path="custom/models",
)
```

## Expected behaviour

Relative model, seed, and snapshot paths should be retained independently of where the absolute project root is configured.

For DAG rendering, Cosmos should resolve them against the effective rendering project path, including when that path comes from `RenderConfig.dbt_project_path`.

## Possible approach

- Preserve the normalized relative paths even when `ProjectConfig.dbt_project_path` is absent.
- Resolve them against `RenderConfig.project_path` when loading through `LoadMode.CUSTOM`.
- Preserve the distinction between an omitted value and an explicitly empty list.

## Suggested tests

- Custom parser with the project root supplied through `RenderConfig` and `ExecutionConfig`.
- Nested relative paths such as `custom/models`.
- Multiple model, seed, and snapshot directories.
- An explicitly empty relative-path list, ensuring it does not fall back to the default directory.
- Backward compatibility for the deprecated singular arguments.

00yhj22-debug:
Closes #2926.

## What was happening

`ProjectConfig` only stored the configured `models` / `seeds` / `snapshots` relative paths when its own `dbt_project_path` was set:

```python
if dbt_project_path:
    self.dbt_project_path = Path(dbt_project_path)
    self.models_paths = [self.dbt_project_path / Path(p) for p in _as_path_list(models_relative_paths)]
    ...
```

When the project root was supplied through `RenderConfig` / `ExecutionConfig` instead, the relative paths were discarded with no warning. Under `LoadMode.CUSTOM`, `_relative_dirs()` then returned `None` and `LegacyDbtProject` fell back to the default `models` / `seeds` / `snapshots` folders, so a project laid out as `custom/models/` rendered an incomplete or empty DAG.

## The change

- `ProjectConfig` now normalizes and keeps `models_relative_paths`, `seeds_relative_paths` and `snapshots_relative_paths` regardless of where the project root is configured. Absolute `*_paths` are still only built when `dbt_project_path` is set, so nothing changes for existing configurations.
- `DbtGraph.load_via_custom_parser()` falls back to those relative paths when the absolute ones are unavailable, via a small `_configured_dirs()` helper.

The distinction between an omitted value and an explicitly empty list is preserved: an omitted argument still resolves to the dbt default (`models`), while `models_relative_paths=[]` stays empty.

## Reproduction

Using the configuration from the issue:

```python
ProjectConfig(
    manifest_path=manifest_path,
    project_name="split_path_project",
    models_relative_paths=["custom/models"],
)
```

| | `models_relative_paths` | directory searched by the custom parser |
|---|---|---|
| Before | discarded | `models/` |
| After | `[Path("custom/models")]` | `custom/models/` |

## Tests

Added to `tests/test_config.py`:

- `test_relative_paths_kept_without_project_path` — relative paths survive when `dbt_project_path` is unset, while `models_paths` stays empty.
- `test_relative_paths_default_when_not_configured` — an omitted argument falls back to `models`; an explicitly empty list stays empty.

## Note on local test runs

Roughly 70 tests in `tests/test_config.py` and `tests/dbt/test_graph.py` fail on my Windows machine both with and without this change (path separators and subprocess behaviour). I compared the failing sets before and after and they are identical apart from the two new tests flipping to passing, so this introduces no regressions there.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -53,6 +53,38 @@ def test_init_with_project_path_and_list_relative_paths():
     ]
 
 
+def test_relative_paths_kept_without_project_path():
+    """
+    The configured relative paths must survive even when ProjectConfig.dbt_project_path is not set,
+    because the project root may instead come from RenderConfig/ExecutionConfig. Previously they were
+    silently discarded and rendering fell back to the default `models`/`seeds`/`snapshots` folders.
+    """
+    project_config = ProjectConfig(
+        manifest_path=DBT_PROJECTS_ROOT_DIR / "manifest.json",
+        project_name="split_path_project",
+        models_relative_paths=["custom/models"],
+        seeds_relative_paths=["custom/seeds"],
+        snapshots_relative_paths=["custom/snapshots"],
+    )
+
+    assert project_config.dbt_project_path is None
+    assert project_config.models_paths == []
+    assert project_config.models_relative_paths == [Path("custom/models")]
+    assert project_config.seeds_relative_paths == [Path("custom/seeds")]
+    assert project_config.snapshots_relative_paths == [Path("custom/snapshots")]
+
+
+def test_relative_paths_default_when_not_configured():
+    """An omitted relative path falls back to the dbt default, an explicitly empty list stays empty."""
+    defaulted = ProjectConfig(manifest_path=DBT_PROJECTS_ROOT_DIR / "manifest.json", project_name="p")
+    assert defaulted.models_relative_paths == [Path("models")]
+
+    emptied = ProjectConfig(
+        manifest_path=DBT_PROJECTS_ROOT_DIR / "manifest.json", project_name="p", models_relative_paths=[]
+    )
+    assert emptied.models_relative_paths == []
+
+
 def test_models_paths_not_shared_across_instances():
     """
     models_paths/seeds_paths/snapshots_paths must not be a shared mutable default - mutating one
--- a/cosmos/config.py
+++ b/cosmos/config.py
@@ -266,6 +266,9 @@ class ProjectConfig:
     models_paths: list[Path]
     seeds_paths: list[Path]
     snapshots_paths: list[Path]
+    models_relative_paths: list[Path]
+    seeds_relative_paths: list[Path]
+    snapshots_relative_paths: list[Path]
     project_name: str
 
     def __init__(
@@ -301,6 +304,12 @@ class ProjectConfig:
             "snapshots_relative_paths",
         )
         # Assigned here, not as a class-level default, so instances don't share one mutable list.
+        # The relative paths are kept regardless of whether dbt_project_path is set on this config:
+        # the project root may instead be supplied through RenderConfig/ExecutionConfig, and rendering
+        # still needs to know which directories the user configured.
+        self.models_relative_paths = [Path(p) for p in _as_path_list(models_relative_paths)]
+        self.seeds_relative_paths = [Path(p) for p in _as_path_list(seeds_relative_paths)]
+        self.snapshots_relative_paths = [Path(p) for p in _as_path_list(snapshots_relative_paths)]
         self.models_paths = []
         self.seeds_paths = []
         self.snapshots_paths = []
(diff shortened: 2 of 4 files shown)


tests/test_config.py
```diff
@@ -53,6 +53,38 @@ def test_init_with_project_path_and_list_relative_paths():
     ]
 
 
+def test_relative_paths_kept_without_project_path():
```
pankajkoti, line 56: These two tests assert the `ProjectConfig` attributes, but the actual fix that closes #2926 is `_configured_dirs` and its use in `load_via_custom_parser`, and nothing exercises that path. They would still pass if `_configured_dirs` were reverted to `_relative_dirs`, which would quietly bring the bug back.

Could you add a functional test that reproduces the split-path CUSTOM scenario? `test_load_via_custom_parser_with_nested_model_dir` in `tests/dbt/test_graph.py` is a ready template. The only change needed is to supply the project root through RenderConfig/ExecutionConfig and leave `ProjectConfig.dbt_project_path` None (manifest_path + project_name + `models_relative_paths="custom/models"`), then assert the model under `custom/models` is crawled. That locks in the fallback this PR adds.

pankajkoti (approved):
Thanks for contributing the PR @00yhj22-debug , looks good to me, one comment inline that could also help increase the patch test coverage

tests/test_config.py
00yhj22-debug: Good catch — you're right that the two config assertions would survive a revert of `_configured_dirs`. Added `test_load_via_custom_parser_with_nested_model_dir_and_split_project_path`, built on the template you pointed at: `ProjectConfig` gets `manifest_path` + `project_name` + `models_relative_paths="custom/models"` and no `dbt_project_path`, the root comes from `RenderConfig`/`ExecutionConfig`, and it asserts the model under `custom/models` is crawled.

Verified it actually locks the fix in: with line 96 of `graph.py` temporarily returning `None` (i.e. `_configured_dirs` behaving like the old `_relative_dirs`), the new test fails with `assert 'my_model' in {}` while the two neighbouring tests still pass. Restored, it passes. Full `tests/dbt/test_graph.py` + `tests/test_config.py` failure lists are byte-identical before and after this commit on my machine — the remaining failures there are pre-existing and environment-related (Windows, no dbt binary).

PR #2978, opened 2026-08-25
Fix bare selector root resolution under LoadMode.DBT_MANIFEST

### LINKED ISSUE #2962: Manifest load mode: graph operator "+" only resolves exact node names, not folder/package/fqn roots
## Environment

| Field | Value |
|---|---|
| Astronomer Cosmos version | 1.15.1 |
| dbt Core version | 1.12.0 |
| dbt adapter versions | Not applicable (rendering from a committed `manifest.json`, no adapter invoked) |
| LoadMode | `DBT_MANIFEST` |
| ExecutionMode | `LOCAL` (rendering only, not relevant to the bug) |
| InvocationMode | Not applicable |
| Airflow version | 3.3.0 |
| Python | 3.11 |
| Operating System | macOS (Darwin) |

## Summary

In `LoadMode.DBT_MANIFEST`, a selector that combines the graph operator (`+`) with a bare **folder name** or **package name** (for example `my_folder+` or `my_package+`) resolves to an empty set. Cosmos resolves the graph-operator root by exact node name only, so folder and package tokens match nothing and the select or exclude silently has no effect. The same selectors work under `dbt ls`, which delegates resolution to dbt.

For `exclude` this is a correctness issue. The intended nodes are not removed and remain in the rendered DAG.

## Background: what #2357 addressed

PR #2357 ("Fix inclusion of package models, and selecting/excluding") added manifest-mode support for:

- Including installed-package models in the graph with correct file paths.
- `package:<name>` select and exclude.
- Bare identifiers resolving the dbt way (by package name, node name, or folder name).
- Select and exclude by folder name.

All of the above work. The gap below is the graph-operator (`+`) case layered on top of the folder and package tokens.

## What still needs addressing

Graph-operator root resolution should match dbt. The token attached to `+` should resolve through the same methods as a bare identifier (node name, folder name, package name, fqn, path), not by exact node name alone.

## How to reproduce

Project: [`fhir-dbt-analytics`](https://github.com/google/fhir-dbt-analytics) (public), loaded from its `target/manifest.json` with `LoadMode.DBT_MANIFEST`.

```python
from cosmos.config import ProjectConfig, RenderConfig, ExecutionConfig
from cosmos.constants import LoadMode
from cosmos.dbt.graph import DbtGraph

def n_models(selector):
    g = DbtGraph(
        project=ProjectConfig(dbt_project_path=DBT_DIR, manifest_path=DBT_DIR / "target" / "manifest.json"),
        render_config=RenderConfig(load_method=LoadMode.DBT_MANIFEST, select=[selector]),
        execution_config=ExecutionConfig(dbt_project_path=DBT_DIR),
    )
    g.load(method=LoadMode.DBT_MANIFEST)
    return sum(1 for x in g.filtered_nodes.values() if x.resource_type.value == "model")
```

Results:

| kind | selector | models selected | status |
|---|---|---|---|
| node name | `metric_definition` | 1 | PASS |
| node + descendants | `metric_definition+` | 7 | PASS |
| folder name | `post_processing` | 6 | PASS |
| **folder + descendants** | **`post_processing+`** | **0** | **FAIL** |
| path | `path:models/post_processing` | 6 | PASS |
| path + descendants | `path:models/post_processing+` | 11 | PASS |
| bare package | `fhir_dbt_utils` | 25 | PASS |
| **bare package + descendants** | **`fhir_dbt_utils+`** | **0** | **FAIL** |
| package selector | `package:fhir_dbt_utils` | 25 | PASS |
| package selector + descendants | `package:fhir_dbt_utils+` | 182 | PASS |

So `<node>+`, `path:...+`, and `package:...+` all work. Only `<folder>+` and `<package>+` (a bare non-node token plus the operator) return nothing.

## Root cause

In `cosmos/dbt/selector.py`, `GraphSelector` resolves its root through a `node_by_name[node.name]` lookup (exact node name). If the token is not a node name it logs `Selector <x> not found` and returns without selecting anything. Explicit method prefixes inside a graph selector (`path:`, `package:`) are handled, which is why `path:...+` and `package:...+` work, but bare folder, package, and fqn tokens are not. The non-graph bare-identifier path already resolves folder and package names, so the two paths are inconsistent.

## Impact

Any `<folder>+` or `<package>+` selector in manifest mode silently matches nothing, which diverges from `dbt ls`. On `exclude` the targeted nodes are not excluded.

## Workaround

Add an explicit method prefix so the graph selector can resolve the root. Use `package:<name>+` or `path:<folder>+` instead of the bare `<name>+`.

## Suggested fix

When the `GraphSelector` root token is not an exact node name, fall back to the same resolution the non-graph bare-identifier path already uses (folder name, package name, fqn), so `<folder>+` and `<package>+` behave as they do under dbt. Adding manifest-mode selector tests for these two shapes would lock in the parity.

## Are you willing to submit a PR?

Yes, I am willing to submit a PR.

pankajkoti:
## Description

Under `LoadMode.DBT_MANIFEST`, a selector combining the graph operator `+` with a bare folder or package name resolved to an empty set. `GraphSelector` resolved its root by exact node name only, so bare folder/package tokens matched nothing and the select or exclude silently had no effect. For `exclude` this was a correctness issue. The same selectors work under `dbt ls`.

## Changes

- Resolve a bare selector root with dbt's default fqn selector semantics (ported from dbt's `is_selected_node`): match the model name (leaf) or an ordered prefix of the fqn, over both the fqn and the package-less fqn. So the package, a top-level folder, a dotted folder path, and the model name match, a middle folder segment does not, and versioned models match by name and by the `name_version` task name.
- In `GraphSelector.filter_nodes`, resolve the bare graph-operator root by scanning all nodes (the fqn match, plus the dot-to-underscore patched task name), so folder/package/name roots all resolve and duplicate names across packages all contribute.
- `NodeSelector._is_bare_identifier_matching` delegates to the same resolver, keeping the graph and non-graph bare paths identical.
- Nodes without an fqn (e.g. `LoadMode.CUSTOM`) fall back to package name, node name, or a project-relative folder segment (directories only).
- Tests, a changelog entry, and a docs note.

This also corrects the pre-existing non-graph bare path, which keyed off the absolute file path and over-matched the model-paths root, filesystem ancestors, and middle folders.

## Behavior (parity with `dbt ls`, verified)

| selector | before | after (matches `dbt ls`) |
| -- | -- | -- |
| `<top-folder>+` / `<package>+` | 0 nodes | folder/package nodes + descendants |
| `<name>+` (leaf, any depth) | node only if exact name | the model + descendants |
| `<name>+` where a folder shares the name | node only | node + folder nodes + descendants |
| `<middle-folder>` (e.g. `core` in `marts/core`) | matched (over-match) | nothing |
| `models+` / `<ancestor-dir>+` | whole project | nothing |
| versioned `customers` / `customers_v1` | broken or no-op | the versioned model |
| `path:<folder>+`, `package:<name>+` | worked | unchanged |

Verified against a project with nested folders and a versioned model: cosmos `DBT_MANIFEST` matches `dbt ls` on every token tested, and the ticket's `fhir-dbt-analytics` counts are unchanged (`post_processing+` = 11, `fhir_dbt_utils+` = 182).

Closes #2962
Fixes BOSS-615


cosmos/dbt/selector.py
```diff
@@ -58,6 +58,22 @@ def _fqn_matches(node_fqn: str, fqn_selector_value: str) -> bool:
...
 
+def _node_matches_bare_identifier(node: DbtNode, identifier: str) -> bool:
+    """
+    True if a bare token matches the node the dbt way: as an element of the node's fqn
+    (package name, a folder under the model paths, or the node name).
+    Falls back to package name, node name, or project-relative folder segment when fqn is
+    absent (e.g. LoadMode.CUSTOM does not populate it).
+    """
+    if node.fqn:
+        return identifier in node.fqn
```
pankajastro, line 69: Nice fix overall, but I think this reintroduces the bug for versioned models. `_node_matches_bare_identifier` only checks `node.fqn` when it's set, and skips `node.name` — but dbt keeps a model's version as a separate trailing fqn element (e.g. `["proj", "customers", "v1"]`), so it's never joined into `customers_v1`.

In practice: `select_nodes(nodes, select=["customers_v1"])` returns `{}` on this branch, though it works fine on `main`. `exclude=["customers_v1"]` ends up being a silent no-op too — basically the same failure mode this PR is fixing, just triggered by a versioned model instead.

Could we check `node.name` alongside `node.fqn` here, and add a regression test for versioned models while we're at it?

pankajastro: Here's a repro test (fails on this branch, passes on `main`) — could drop this in next to `test_bare_root_resolves_via_fqn_not_absolute_path`:

```python
def test_bare_identifier_matches_versioned_model_name():
    """Bare select/exclude must still match a versioned model by its patched task name (BOSS-615 regression).

    dbt keeps a versioned model's version as a separate trailing fqn element
    (e.g. ["proj", "customers", "v1"]), never joined into "customers_v1". A bare
    identifier lookup that only checks `identifier in node.fqn` therefore never
    matches the model's actual (dot-to-underscore patched) task name, silently
    breaking both `select=["customers_v1"]` and `exclude=["customers_v1"]`.
    """
    versioned = DbtNode(
        unique_id="model.proj.customers.v1",
        resource_type=DbtResourceType.MODEL,
        depends_on=[],
        path_base=SAMPLE_PROJ_PATH,
        original_file_path=Path("models/customers.sql"),
        tags=[],
        config={},
        package_name="proj",
        fqn=["proj", "customers", "v1"],
    )
    other = _fqn_node("orders", "")
    nodes = {n.unique_id: n for n in [versioned, other]}

    assert versioned.name == "customers_v1"  # sanity check on the patched task name

    selected = select_nodes(project_dir=SAMPLE_PROJ_PATH, nodes=nodes, select=["customers_v1"])
    assert set(selected.keys()) == {versioned.unique_id}

    excluded = select_nodes(project_dir=SAMPLE_PROJ_PATH, nodes=nodes, exclude=["customers_v1"])
    assert set(excluded.keys()) == {other.unique_id}
```

pankajkoti: Confirmed and fixed. On the branch `select=['customers_v1']` returned nothing and the exclude was a no-op. I reworked the resolver to follow dbt's fqn selector directly (ported `is_selected_node`): it matches the leaf name or an ordered fqn prefix, over both the fqn and the package-less fqn. That matches a versioned model by its name and by the `name_version` form dbt joins on `_`, so `customers`, `customers.v1`, and `customers_v1` all resolve, while a lone `v1` does not. Added your repro test plus a fqn-prefix test. I checked the full matrix against `dbt ls` (nested folders, the versioned model, package name, model-paths root) and cosmos now matches it exactly. Thanks for the catch and the ready-made test.

cosmos/dbt/selector.py
```diff
@@ -58,6 +58,22 @@ def _fqn_matches(node_fqn: str, fqn_selector_value: str) -> bool:
...
+    return (
+        (node.package_name or "") == identifier
+        or node.name == identifier
+        or identifier in node.original_file_path.parts
+    )
```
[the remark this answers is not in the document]
pankajkoti: Fixed. The fqn-absent fallback now matches `original_file_path.parent.parts`, so it matches folder segments only and not the file name. This path runs only when the node has no fqn (e.g. LoadMode.CUSTOM).

cosmos/dbt/selector.py
```diff
@@ -324,16 +340,17 @@ def filter_nodes(self, nodes: dict[str, DbtNode]) -> set[str]:  # noqa: C901
...
+                node_id for node_id, node in nodes.items() if _node_matches_bare_identifier(node, self.node_name)
+            )
+            node_by_name = {node.name: node_id for node_id, node in nodes.items()}
             node_name_patched = self.node_name.replace(".", "_")
-
             if node_name_patched in node_by_name:
-                root_id = node_by_name[node_name_patched]
-                root_nodes.add(root_id)
-            else:
+                root_nodes.add(node_by_name[node_name_patched])
```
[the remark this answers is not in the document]
pankajkoti: Fixed. The bare graph-operator root is now resolved by scanning all nodes instead of a name-keyed dict, so duplicate node names across packages all contribute roots. Dotted and versioned roots resolve through the same dbt fqn matching.

cosmos/dbt/selector.py
```diff
@@ -58,6 +58,47 @@ def _fqn_matches(node_fqn: str, fqn_selector_value: str) -> bool:
...
+    (e.g. "marts.core") matches consecutive fqn segments from the start. Versioned models keep the
+    version as a separate trailing fqn element, so the model name is fqn[-2] and "name_version" also
+    matches (dbt joins the last two segments on "_").
+    """
+    selector_parts = selector.split(".")
+    if is_versioned:
+        if len(fqn) >= 2 and (fqn[-2] == selector or "_".join(fqn[-2:]) == "_".join(selector_parts[-2:])):
+            return True
+    elif fqn and fqn[-1] == selector:
+        return True
```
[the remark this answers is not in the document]
pankajkoti: This mirrors dbt on purpose. `_fqn_selector_matches` is a port of dbt's own `is_selected_node`, and that `name_version` early-return (comparing only the last two segments) is dbt's exact logic. Verified against `dbt ls`: `dbt ls -s bogus.customers.v1` and `marts.customers.v1` both select a versioned model even when the earlier segments do not match its fqn, so this is dbt's behavior, not an added over-match. cosmos matches `dbt ls` on all of these. Making it stricter would diverge from dbt, which is the parity this change is aiming for, so leaving it as-is.

cosmos/dbt/selector.py
```diff
@@ -733,12 +775,8 @@ def _is_package_matching(self, node: DbtNode) -> bool:
...
 
     def _is_bare_identifier_matching(self, node: DbtNode) -> bool:
-        """Bare identifiers match by package_name, node name, or path segment (e.g. folder name)."""
-        if (node.package_name or "") in self.config.bare_identifiers or node.name in self.config.bare_identifiers:
-            return True
-        # Match by path segment (folder name): e.g. "folder_a" matches nodes under .../folder_a/...
-        path_parts = node.file_path.parts
-        return any(bare in path_parts for bare in self.config.bare_identifiers)
+        """Bare identifiers match the dbt way, against the node's fqn (package, folder, or node name)."""
+        return any(_node_matches_bare_identifier(node, bare) for bare in self.config.bare_identifiers)
```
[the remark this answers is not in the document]
pankajkoti: Fixed. Updated the docstring to point at `_node_matches_bare_identifier` and note the package/name/folder fallback used when the node has no fqn.

cosmos/dbt/selector.py
```diff
@@ -324,16 +365,17 @@ def filter_nodes(self, nodes: dict[str, DbtNode]) -> set[str]:  # noqa: C901
...
-
-            if node_name_patched in node_by_name:
-                root_id = node_by_name[node_name_patched]
-                root_nodes.add(root_id)
-            else:
+            root_nodes.update(
+                node_id
+                for node_id, node in nodes.items()
+                if _node_matches_bare_identifier(node, self.node_name) or node.name == node_name_patched
+            )
```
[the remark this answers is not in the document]
pankajkoti: Good catch, fixed. The `node.name == node_name_patched` fallback was only needed for fqn-absent nodes (LoadMode.CUSTOM), where a dotted selector maps to the underscore task name. For fqn-present nodes it over-selected, e.g. `marts.core` also grabbed a model named `marts_core`. I moved that patched-name match into the fqn-absent branch of `_node_matches_bare_identifier` and dropped it from `filter_nodes`, so fqn nodes follow dbt's fqn semantics only. Verified against `dbt ls`: `marts.core` selects only the nested model, not `marts_core`, and added a regression test.

cosmos/dbt/selector.py
```diff
@@ -58,6 +58,47 @@ def _fqn_matches(node_fqn: str, fqn_selector_value: str) -> bool:
...
+    """
+    Port of dbt's ``is_selected_node`` (``dbt.graph.selector_methods``) without wildcard support:
+    a bare selector matches the leaf (model name) or an ordered prefix of the fqn. A dotted selector
+    (e.g. "marts.core") matches consecutive fqn segments from the start. Versioned models keep the
+    version as a separate trailing fqn element, so the model name is fqn[-2] and "name_version" also
+    matches (dbt joins the last two segments on "_").
+    """
+    selector_parts = selector.split(".")
+    if is_versioned:
+        if len(fqn) >= 2 and (fqn[-2] == selector or "_".join(fqn[-2:]) == "_".join(selector_parts[-2:])):
```
pankajastro, line 71: For a versioned model, this branch returns `True` as soon as the last two segments match (`fqn[-2:]` vs the selector's last two parts), without checking that earlier segments of a dotted selector also match the corresponding earlier `fqn` segments. Is that intentional? I reproduced it directly: for `fqn=['proj','marts','customers','v1']`, both `_fqn_selector_matches(fqn, 'staging.customers.v1', True)` and `_fqn_selector_matches(fqn, 'zzz_nonexistent_pkg.customers.v1', True)` return `True` — any garbage prefix is accepted as long as the trailing two tokens are `customers.v1`. Should the prefix check further down (the `flat_fqn`/`selector_parts` comparison) also run for the versioned case before returning `True`?

cosmos/dbt/selector.py
```diff
@@ -58,6 +58,47 @@ def _fqn_matches(node_fqn: str, fqn_selector_value: str) -> bool:
...
+    """
+    if node.fqn:
+        is_versioned = node.resource_type == DbtResourceType.MODEL and "." in node.resource_name
+        return _fqn_selector_matches(node.fqn, identifier, is_versioned) or _fqn_selector_matches(
+            node.fqn[1:], identifier, is_versioned
+        )
+    return (
+        (node.package_name or "") == identifier
+        or node.name == identifier.replace(".", "_")
+        or identifier in node.original_file_path.parent.parts
```
pankajastro, line 98: For nodes without an `fqn` (the `LoadMode.CUSTOM` fallback), `identifier in node.original_file_path.parent.parts` still lets a bare selector like `select=['models']` match the top-level models-dir folder segment. I traced this through `load_via_custom_parser` (`cosmos/dbt/graph.py`): `DbtNode.fqn` defaults to `None` and is never set there, and `original_file_path` is built from `model.path.relative_to(project_path)` where `model.path` comes from `models_dir.rglob(...)` with `models_dir = project_dir / "models"` — so `.parent.parts` includes `'models'`. This PR's own new test `test_bare_root_resolves_via_fqn_not_absolute_path` asserts the opposite for `DBT_MANIFEST`: "dbt never selects on the 'models' directory... it matches nothing (matching dbt)". Should this `LoadMode.CUSTOM` fallback also exclude the model-paths root segment for consistency, or is the inconsistency acceptable since `LoadMode.CUSTOM` is deprecated?

pankajastro (approved):

cosmos/dbt/selector.py
pankajkoti: Intentional, and it mirrors dbt exactly. `_fqn_selector_matches` is a port of dbt's `is_selected_node`, and that `name_version` early-return is dbt's own logic. I checked your examples against `dbt ls` with `fqn=['proj','marts','customers','v1']`: `dbt ls -s staging.customers.v1` and `dbt ls -s zzz_nonexistent_pkg.customers.v1` both select the model, same as cosmos. dbt accepts any prefix as long as the last two tokens are `name.version`. Adding the prefix check for the versioned case would also reject `customers.v1` for a nested model, which dbt matches through the same early-return, so it would diverge from dbt on a realistic selector. This PR targets dbt parity, so keeping it faithful is deliberate.

cosmos/dbt/selector.py
pankajkoti: This is limited to `LoadMode.CUSTOM`, where the node has no fqn, so the fallback matches folder segments from `original_file_path.parent.parts` and that still includes the `models` root. It is pre-existing (on main it matched through the absolute `file_path.parts`) and only affects the deprecated CUSTOM parser, which does not carry the model-paths config into this helper. The fqn path (DBT_MANIFEST and DBT_LS) matches dbt and excludes the model-paths root. Given CUSTOM is deprecated, keeping the fallback as-is is fine and out of scope for this PR.

PR #2983, opened 2026-08-27
Fix AttributeError masking the real dbt error on run-operation failure

### LINKED ISSUE #2982: [Bug] run-operation failure raises AttributeError and hides the real dbt error
# [Bug] `run-operation` failure raises `AttributeError` and hides the real dbt error

## What happens

When `DbtRunOperationLocalOperator` fails (dbt invocation mode), the task error is
`AttributeError: 'RunResultOutput' object has no attribute 'node'` instead of the
macro's actual error. The real cause never reaches the task log.

## Why

`extract_message_by_status()` reads `node_result.node.name` for every failed entry:

```python
# cosmos/dbt/runner.py
for node_result in result.result.results:
    if node_result.status in status_levels:
        node_names.append(str(node_result.node.name))   # <-- assumes RunResult
        node_results.append(str(node_result.message))
```

That assumption holds for `run`/`build`/`test`, whose entries are `RunResult`.
But `run-operation` produces `RunResultOutput`, which has **no `node` attribute** —
it carries `unique_id` instead. So the failure path itself raises, and the raised
`AttributeError` replaces the error the operator was trying to report.

On top of that, dbt always sets `message=None` for run-operation results
(`dbt/task/run_operation.py`), so even after guarding the `node` access there is
nothing useful in the result object. The macro's error text only exists in the
event stream (`RunningOperationCaughtError`).

## Reproduce

Any failing macro is enough — no warehouse needed.

```python
from dbt.cli.main import dbtRunner

result = dbtRunner().invoke(
    ["run-operation", "boom", "--project-dir", project, "--profiles-dir", project]
)
entry = result.result.results[0]
print(repr(entry))
# RunResultOutput(status='error', unique_id='macro.probe.boom', message=None)
print(hasattr(entry, "node"))   # False
```

Feeding that `result` to `AbstractDbtLocalBase.handle_exception_dbt_runner()` raises
`AttributeError` rather than surfacing the macro error.

Versions: astronomer-cosmos 1.15.0 and 1.15.1 (identical code in both), dbt-core 1.10.13,
Airflow 3.

## Suggested fix

1. Guard the node access in `extract_message_by_status()` — fall back to `unique_id`
   when `node` is absent, so the failure path never raises:

   ```python
   label = getattr(getattr(node_result, "node", None), "name", None) or getattr(
       node_result, "unique_id", "unknown"
   )
   ```

2. For `run-operation`, the message is not recoverable from the result object at all.
   Either collect `RunningOperationCaughtError` / `RunningOperationUncaughtError` from
   the dbt event stream inside the operator, or document that callers must pass
   `dbt_runner_callbacks` to obtain the macro error.

Happy to send a PR for (1) if the direction looks right.

barking-code:
## Description

`extract_message_by_status()` reads `node_result.node.name` for every failed entry.
That assumption holds for `run`, `build` and `test`, whose entries are `RunResult`,
but `run-operation` yields `RunResultOutput`, which has no `node` attribute at all —
it identifies the macro through `unique_id` instead.

Because the access happens *while reporting a failure*, the resulting
`AttributeError` replaces the error being reported, and the macro's real error never
reaches the task log. A failing `DbtRunOperationLocalOperator` currently surfaces:

```
AttributeError: 'RunResultOutput' object has no attribute 'node'
```

This falls back to `unique_id` when `node` is absent, so the failure path no longer
raises and the reported label stays meaningful.

Reproducible without a warehouse:

```python
result = dbtRunner().invoke(["run-operation", "boom", "--project-dir", p, "--profiles-dir", p])
entry = result.result.results[0]
# RunResultOutput(status='error', unique_id='macro.probe.boom', message=None)
hasattr(entry, "node")   # False
```

Note this only unblocks the *label*. dbt sets `message=None` for run-operation
results (`dbt/task/run_operation.py`), so the macro text itself still only exists in
the event stream (`RunningOperationCaughtError`). That part is discussed in the issue
and is not addressed here.

## Related Issue(s)

Fixes #2982

## Breaking Change?

No. Entries that carry a `node` keep reporting the node name — covered by
`test_extract_message_by_status_still_prefers_node_name`.

## Checklist

- [x] I have made corresponding changes to the documentation (not needed — internal helper)

Both new tests were verified against the change: reverting the fix makes
`test_extract_message_by_status_falls_back_to_unique_id_without_node` fail while the
rest keep passing.

### DIFF
--- a/cosmos/dbt/runner.py
+++ b/cosmos/dbt/runner.py
@@ -107,6 +107,12 @@ def run_command(
     return result
 
 
+def _node_label(node_result: Any) -> str:
+    """``run-operation`` yields ``RunResultOutput``, which has no ``node`` (#2982)."""
+    node = getattr(node_result, "node", None)
+    return str(getattr(node, "name", None) or getattr(node_result, "unique_id", None) or "unknown")
+
+
 def extract_message_by_status(
     result: dbtRunnerResult, status_levels: list[str] | None = None
 ) -> tuple[list[str], list[str]]:
@@ -128,7 +134,7 @@ def extract_message_by_status(
 
     for node_result in result.result.results:
         if node_result.status in status_levels:
-            node_names.append(str(node_result.node.name))
+            node_names.append(_node_label(node_result))
             node_results.append(str(node_result.message))
 
     return node_names, node_results
--- a/tests/dbt/test_runner.py
+++ b/tests/dbt/test_runner.py
@@ -15,6 +15,8 @@ from cosmos.operators.watcher import DbtProducerWatcherOperator
 
 sys.modules.pop("dbt.cli.main", None)
 
+from types import SimpleNamespace
+
 import pytest
 
 import cosmos.dbt.runner as dbt_runner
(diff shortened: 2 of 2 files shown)


pankajkoti (approved):
lgtm, thanks for contributing the PR @barking-code

PR #2986, opened 2026-08-31
Surface the macro error on run-operation failure with dbt < 1.12

barking-code:
## What

- Registers an internal dbt event callback for `run-operation` invocations.
- Fills `RunResultOutput.message` from the collected macro error when dbt left it unset.
- Follow-up to #2983, which covered part 1 of #2982. This covers part 2.

## Why

- dbt hardcoded `message=None` on `run-operation` results until 1.12.
- The macro error text was only emitted as an event: `RunningOperationCaughtError` / `RunningOperationUncaughtError`.
- With part 1 fixed, `extract_message_by_status()` completes but renders `macro.my_project.my_macro: None` on older dbt.
- dbt-labs/dbt-core#12730 populates `message` from dbt 1.12 onwards, so this change is a no-op there.
- It can be dropped once dbt < 1.12 is no longer supported.

## Verification

Real runs against a minimal duckdb project whose macro executes `select * from table_that_does_not_exist`:

| dbt-core | raised `CosmosDbtRunError` |
|---|---|
| 1.10.23 | `macro.repro.boom: Encountered an error while running operation: Runtime Error / Catalog Error: Table with name table_that_does_not_exist does not exist!` |
| 1.12.3 | `macro.repro.boom: Runtime Error / Catalog Error: Table with name table_that_does_not_exist does not exist!` |

- On dbt < 1.12 the text carries dbt's own event prefix; on 1.12+ the value dbt set is untouched.
- `tests/dbt/test_runner.py`: 18 passed.
- The two `test_handle_exception_if_needed_after_*` failures reproduce identically on `main`; they need a Postgres env.
- `black`, `ruff check` and `mypy --config-file ./pyproject.toml cosmos/dbt/runner.py`: no new findings.

## Notes

- Only `InvocationMode.DBT_RUNNER` is affected.
- `SUBPROCESS` runs dbt out of process, so no Python callback can attach — and none is needed there.
- `handle_exception_subprocess` already logs the full dbt output, which contains the macro error. This answers the question in #2982.
- The callback attaches only when the subcommand is `run-operation`, so other commands keep using the cached runner.
- Both lists are created per invocation, so no state is shared between runs.
- The callback never raises; dbt wraps a raising callback as `GenericExceptionOnRun`, as `operators/watcher.py` already notes.
- `target/run_results.json` is written by dbt before the result returns, so the file on disk still has `message: null` on dbt < 1.12.

Refs #2982

### DIFF
--- a/cosmos/dbt/runner.py
+++ b/cosmos/dbt/runner.py
@@ -148,7 +148,8 @@ def run_command(
     # to the dbt executable.
     cli_args = command[1:]
     macro_errors: list[str] = []
-    # Global flags precede the subcommand (see ``build_cmd``), so match anywhere.
+    # ``build_cmd`` puts flags on both sides of the subcommand — dbt global flags before it,
+    # ``add_global_flags()``/``dbt_cmd_flags`` after it — so a positional check is not safe here.
     if "run-operation" in cli_args:
         callbacks = [*(callbacks or []), _collect_macro_errors(macro_errors)]
     # ``exclude_dags_folder_from_sys_path`` must enter *before* ``change_working_directory`` so it
--- a/cosmos/dbt/runner.py
+++ b/cosmos/dbt/runner.py
@@ -21,6 +21,9 @@ else:  # pragma: no cover
 
 logger = get_logger(__name__)
 
+# dbt events carrying the macro error text of a failed ``run-operation``.
+MACRO_ERROR_EVENTS = ("RunningOperationCaughtError", "RunningOperationUncaughtError")
+
 if TYPE_CHECKING:  # pragma: no cover
     from dbt.cli.main import dbtRunner, dbtRunnerResult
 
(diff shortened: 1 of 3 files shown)


cosmos/dbt/runner.py
```diff
@@ -81,6 +84,33 @@ def _cleanup_dbt_adapters() -> None:
...
+    return collect
+
+
+def _fill_missing_messages(result: dbtRunnerResult, macro_errors: list[str]) -> None:
+    """dbt < 1.12 hardcodes ``message=None`` on run-operation results (dbt-labs/dbt-core#12730)."""
+    if not macro_errors:
+        return
+
+    message = "\n".join(macro_errors)
+    for node_result in getattr(result.result, "results", None) or []:
```
pankajastro: If a future dbt-core version renames/drops `.results` on the run-operation result object, this silently iterates zero items instead of erroring. Worth a debug log (or assert) here so a contract change surfaces instead of quietly un-fixing #2982-style errors?

cosmos/dbt/runner.py
```diff
@@ -81,6 +84,33 @@ def _cleanup_dbt_adapters() -> None:
     gc.collect()
 
 
+def _collect_macro_errors(collected: list[str]) -> Callable[[Any], None]:
+    def collect(event: Any) -> None:
+        # dbt wraps a raising callback as GenericExceptionOnRun, so never raise from here.
+        try:
+            info = getattr(event, "info", None)
```
pankajastro: `watcher.py`'s `_event_callback` already parses these events via `MessageToJson` + `store_dbt_resource_status_from_log` rather than raw `getattr` on `event.info`. Any reason not to reuse that path here too, so a future `EventMsg` field change only has to be handled once?

cosmos/dbt/runner.py
```diff
@@ -91,6 +121,10 @@ def run_command(
     # command that is used by `InvocationMode.SUBPROCESS`, and in that scenario the first command is necessarily the path
     # to the dbt executable.
     cli_args = command[1:]
+    macro_errors: list[str] = []
+    # Global flags precede the subcommand (see ``build_cmd``), so match anywhere.
```
pankajastro: `build_cmd` actually appends `add_global_flags()` output and `dbt_cmd_flags` *after* `base_cmd`, not just before — the comment reads as if flags only ever precede the subcommand. Mind rewording so a future refactor doesn't assume a positional check is safe here?

cosmos/dbt/runner.py
barking-code: @pankajastro Done in e93e0ae — extracted `dbt_event_to_json()`, and both this collector and the watcher callback go through it, so a protobuf field rename has one place to fix. `parse()` wants the JSON string, so the collector does `json.loads()` and reads two keys: ~21µs per event, and only for `run-operation`. It still swallows instead of raising, or dbt would replace the error with `GenericExceptionOnRun`.

cosmos/dbt/runner.py
barking-code: @pankajastro Agreed — 9949fdf logs at debug when macro errors were collected but `result.result.results` is missing or empty, then returns. Not an assert: raising would report a Cosmos-side shape problem instead of the run-operation failure this path exists to surface. Pinned by `test_fill_missing_messages_logs_when_there_is_nowhere_to_attach`.

cosmos/dbt/runner.py
barking-code: @pankajastro Right, the comment was half the story — 61d07ee. `build_cmd` puts global flags before `base_cmd` and `add_global_flags()`/`dbt_cmd_flags` after it, so the comment now says flags surround the subcommand, which is the real reason a positional check is unsafe. The test carrying the same claim was renamed to `..._when_flags_surround_run_operation` with flags on both sides of the argv.

PR #2991, opened 2026-09-08
Run the Cosmos test suite against Astronomer Runtime alpha images

pankajkoti:
## What and why

Adds a manually dispatched (`workflow_dispatch`) GitHub Actions workflow that runs cosmos's own test suites (from the checked-out ref, latest `main` by default) against a published **Astronomer Runtime alpha** image, so cosmos's compatibility with the alpha's exact Airflow build is validated before that alpha graduates to a stable release.

Run it with the alpha version:
```
gh workflow run test-runtime-alpha.yml -f runtime_version=3.3-2-alpha1
```
Inputs: `runtime_version` (required), `registry` (choice, default `astrocrpublic.azurecr.io/runtime-dev`), `python_versions` (JSON, default `["3.12"]`), `run_paid_suites` (boolean, default `false`, runs the paid Databricks + Snowflake suites).

## How it works

The Runtime image supplies Airflow. Each suite runs inside the image as a job container, installs cosmos + dbt adapters + pytest tooling on top of the image's pinned Airflow via `scripts/test/runtime-alpha-setup.sh` (shared by a small composite action), and reuses cosmos's existing `scripts/test/*.sh` run scripts. Because the image fixes the Airflow version, the python x airflow x dbt matrix collapses to python (from the image tag) x dbt. Providers come from the image's own constraints, so each suite runs against the exact provider versions the alpha ships rather than an overridden set.

## What is included

- `.github/workflows/test-runtime-alpha.yml`, a `setup-cosmos-runtime` composite action, `scripts/test/runtime-alpha-setup.sh`, and `scripts/test/requirements-test-tools.txt`.
- A few cosmos test-harness fixes needed to run on Airflow 3.1+ inside a container:
  - `serialize_dag_to_db` helper so perf and async tests serialize the DAG before `dag.test()` (Airflow 3.1+ requires it).
  - Host-agnostic Asset URI assertion in `test_local.py` (was hardcoding the runner's Postgres host).
  - The dbt-loom job symlinks the runner path its config hardcodes to the container workspace.

## Validation (alpha `3.3-2-alpha1`, py3.12)

Green against the alpha's own provider set: unit, integration (x3), telemetry, performance (x5), dbt-1.8, dbt-loom. The install seam, Postgres, secret injection, and the matrix are all exercised.

Deferred to **BOSS-662**: the dbt-async suite (its async BigQuery operator reassigns `__bases__` to the google provider's operator and breaks on the google `22.4.0` the alpha ships, disabled via the `RUN_DBT_ASYNC_SUITE` repo variable), the Kubernetes suite (pod-execution failures plus a watcher hang, disabled), the dbt-Fusion suite (same Airflow 3.1+ serialization pattern, then Snowflake/BigQuery validation), and one Databricks python-models test. The paid Databricks and Snowflake suites run only when `run_paid_suites` is set.

## Runtime image issues found

1. The alpha image ships `/pyproject.toml` with an invalid PEP440 `version`, which makes `uv pip install` abort inside the image. Worked around in the setup helper by neutralizing that line.
2. `airflow version` prints plugin warnings to stdout, breaking scripts that parse it for `db migrate` vs the AF3-removed `db init`. Worked around with `AIRFLOW__LOGGING__LOGGING_LEVEL=ERROR` scoped to setup.

## Review hardening

- `registry` is a `type: choice` allowlist so a dispatcher cannot redirect the secret-bearing jobs to an arbitrary image.
- Python 3.14 is excluded from the suites that force an older dbt with no 3.14 wheels (dbt-loom, dbt-1.8, async), mirroring test-unprivileged.yml.
- Removed the inert `dbt_versions` axis (the unit and integration suites never installed the matrix dbt).
- The paid suites sit behind the `run_paid_suites` boolean, a durable gate that stays off once the workflow is dispatch-only.
- A `Report-Suite-Coverage` job writes a ran-vs-skipped manifest to the run summary, so a green run states which suites were gated off.
- Version guards use release tuples (pre-release safe), and the pyproject version patch is asserted with tomllib.

Closes astronomer/oss-integrations-private#613
Closes BOSS-610

### DIFF
--- a/.github/workflows/test-runtime-alpha.yml
+++ b/.github/workflows/test-runtime-alpha.yml
@@ -73,7 +73,7 @@ jobs:
     container:
       # The Runtime image tag is a workflow input (an arbitrary alpha version), so it cannot be
       # digest-pinned. Suppress zizmor's unpinned-images finding for these container images.
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
+      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
       options: --user root
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -102,7 +102,7 @@ jobs:
     container:
       # The Runtime image tag is a workflow input (an arbitrary alpha version), so it cannot be
       # digest-pinned. Suppress zizmor's unpinned-images finding for these container images.
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
+      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
       options: --user root
     services:
       postgres:
@@ -173,7 +173,7 @@ jobs:
     container:
       # The Runtime image tag is a workflow input (an arbitrary alpha version), so it cannot be
       # digest-pinned. Suppress zizmor's unpinned-images finding for these container images.
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
+      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
       options: --user root
     env:
       # The telemetry test asserts Scarf analytics IS on, so override the workflow-global value.
@@ -205,7 +205,7 @@ jobs:
     container:
       # The Runtime image tag is a workflow input (an arbitrary alpha version), so it cannot be
       # digest-pinned. Suppress zizmor's unpinned-images finding for these container images.
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
+      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
       options: --user root
     services:
       postgres:
@@ -257,7 +257,7 @@ jobs:
     container:
       # The Runtime image tag is a workflow input (an arbitrary alpha version), so it cannot be
       # digest-pinned. Suppress zizmor's unpinned-images finding for these container images.
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
+      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
       options: --user root
     services:
       postgres:
@@ -310,7 +310,7 @@ jobs:
     container:
       # The Runtime image tag is a workflow input (an arbitrary alpha version), so it cannot be
       # digest-pinned. Suppress zizmor's unpinned-images finding for these container images.
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
+      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
       options: --user root
     services:
       postgres:
@@ -372,7 +372,7 @@ jobs:
     container:
       # The Runtime image tag is a workflow input (an arbitrary alpha version), so it cannot be
       # digest-pinned. Suppress zizmor's unpinned-images finding for these container images.
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
+      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
       options: --user root
     services:
       postgres:
@@ -431,7 +431,7 @@ jobs:
     container:
       # The Runtime image tag is a workflow input (an arbitrary alpha version), so it cannot be
       # digest-pinned. Suppress zizmor's unpinned-images finding for these container images.
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
+      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
       options: --user root
     services:
       postgres:
@@ -490,7 +490,7 @@ jobs:
     container:
       # The Runtime image tag is a workflow input (an arbitrary alpha version), so it cannot be
       # digest-pinned. Suppress zizmor's unpinned-images finding for these container images.
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
+      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }} # zizmor: ignore[unpinned-images]
       options: --user root
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -593,7 +593,7 @@ jobs:
         env:
           # Secrets/contexts pass through the environment (never interpolated into the command line)
           # and are forwarded into the container with `docker run -e NAME` (value-from-environment).
-          RUNTIME_IMAGE: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }}
+          RUNTIME_IMAGE: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }}
           AIRFLOW_CONN_AWS_S3_CONN: ${{ secrets.AIRFLOW_CONN_AWS_S3_CONN }}
           COSMOS_CONN_POSTGRES_PASSWORD: ${{ secrets.COSMOS_CONN_POSTGRES_PASSWORD }}
         run: |
--- a/.github/workflows/test-runtime-alpha.yml
+++ b/.github/workflows/test-runtime-alpha.yml
@@ -37,12 +37,6 @@ on:
         default: '["1.12"]'
         type: string
 
-  # TEMP (remove before merge): run in cosmos CI on pushes to this branch. workflow_dispatch only
-  # works once the workflow is on the default branch, so we use push + input fallbacks to validate.
-  push:
-    branches:
-      - pankajkoti/boss-610-integration-of-cosmos-tests-with-runtime-alpha-release
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.runtime_version }}
   cancel-in-progress: true
@@ -63,10 +57,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ${{ fromJSON(inputs.python_versions || '["3.12"]') }}
-        dbt-version: ${{ fromJSON(inputs.dbt_versions || '["1.12"]') }}
+        python-version: ${{ fromJSON(inputs.python_versions) }}
+        dbt-version: ${{ fromJSON(inputs.dbt_versions) }}
     container:
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }}
+      image: ${{ inputs.registry }}:${{ inputs.runtime_version }}-python-${{ matrix.python-version }}
       options: --user root
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -89,11 +83,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ${{ fromJSON(inputs.python_versions || '["3.12"]') }}
-        dbt-version: ${{ fromJSON(inputs.dbt_versions || '["1.12"]') }}
+        python-version: ${{ fromJSON(inputs.python_versions) }}
+        dbt-version: ${{ fromJSON(inputs.dbt_versions) }}
         split-group: [1, 2, 3]
     container:
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }}
+      image: ${{ inputs.registry }}:${{ inputs.runtime_version }}-python-${{ matrix.python-version }}
       options: --user root
     services:
       postgres:
@@ -159,10 +153,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ${{ fromJSON(inputs.python_versions || '["3.12"]') }}
+        python-version: ${{ fromJSON(inputs.python_versions) }}
         dbt-version: ["1.9"]
     container:
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }}
+      image: ${{ inputs.registry }}:${{ inputs.runtime_version }}-python-${{ matrix.python-version }}
       options: --user root
     env:
       # The telemetry test asserts Scarf analytics IS on, so override the workflow-global value.
@@ -188,11 +182,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ${{ fromJSON(inputs.python_versions || '["3.12"]') }}
+        python-version: ${{ fromJSON(inputs.python_versions) }}
         dbt-version: ["1.9"]
         num-models: [1, 10, 50, 100, 500]
     container:
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }}
+      image: ${{ inputs.registry }}:${{ inputs.runtime_version }}-python-${{ matrix.python-version }}
       options: --user root
     services:
       postgres:
@@ -239,10 +233,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ${{ fromJSON(inputs.python_versions || '["3.12"]') }}
+        python-version: ${{ fromJSON(inputs.python_versions) }}
         dbt-version: ["1.11"]
     container:
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }}
+      image: ${{ inputs.registry }}:${{ inputs.runtime_version }}-python-${{ matrix.python-version }}
       options: --user root
     services:
       postgres:
@@ -290,10 +284,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ${{ fromJSON(inputs.python_versions || '["3.12"]') }}
+        python-version: ${{ fromJSON(inputs.python_versions) }}
         dbt-version: ["1.8"]
     container:
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }}
+      image: ${{ inputs.registry }}:${{ inputs.runtime_version }}-python-${{ matrix.python-version }}
       options: --user root
     services:
       postgres:
@@ -350,10 +344,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ${{ fromJSON(inputs.python_versions || '["3.12"]') }}
+        python-version: ${{ fromJSON(inputs.python_versions) }}
         dbt-version: ["1.8", "1.9", "1.10", "1.11"]
     container:
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }}
+      image: ${{ inputs.registry }}:${{ inputs.runtime_version }}-python-${{ matrix.python-version }}
       options: --user root
     services:
       postgres:
@@ -399,7 +393,6 @@ jobs:
           POSTGRES_PORT: 5432
 
   Run-Integration-Tests-Expensive:
-    if: ${{ github.event_name == 'workflow_dispatch' }} # TEMP: skip real Databricks on the pre-PR push run
     runs-on: ubuntu-latest
     timeout-minutes: 60
     permissions:
@@ -407,10 +400,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ${{ fromJSON(inputs.python_versions || '["3.12"]') }}
+        python-version: ${{ fromJSON(inputs.python_versions) }}
         dbt-version: ["1.9"]
     container:
-      image: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-7-alpha2' }}-python-${{ matrix.python-version }}
+      image: ${{ inputs.registry }}:${{ inputs.runtime_version }}-python-${{ matrix.python-version }}
       options: --user root
     services:
       postgres:
@@ -456,7 +449,6 @@ jobs:
           AIRFLOW__COSMOS__REMOTE_TARGET_PATH_CONN_ID: aws_s3_conn
 
   Run-Integration-dbt-Fusion-Tests:
-    if: ${{ github.event_name == 'workflow_dispatch' }} # TEMP: skip real Snowflake on the pre-PR push run
     runs-on: ubuntu-latest
     timeout-minutes: 60
     permissions:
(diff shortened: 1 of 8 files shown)


.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,646 @@
...
+  # TEMP (remove before merge): run this workflow on the PR itself with a default image, so
+  # reviewers can see it execute. pull_request events do not populate workflow_dispatch inputs,
+  # so the jobs fall back to the defaults in the `|| '...'` expressions below. The paid Databricks
+  # and Snowflake suites are gated to workflow_dispatch only, so they do not run on the PR.
+  pull_request:
+    branches: [main]
+  # Also run on every push to the PR branch (in addition to pull_request), so each commit runs.
+  push:
+    branches:
+      - pankajkoti/boss-610-integration-of-cosmos-tests-with-runtime-alpha-release
```
pankajkoti, line 49: remove before merge

pankajkoti, line 49:
```suggestion
```

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,646 @@
...
+      - name: Run kubernetes tests on the Runtime image (Airflow from the alpha)
+        env:
+          # Secrets/contexts pass through the environment (never interpolated into the command line)
+          # and are forwarded into the container with `docker run -e NAME` (value-from-environment).
+          RUNTIME_IMAGE: ${{ inputs.registry || 'astrocrpublic.azurecr.io/runtime-dev' }}:${{ inputs.runtime_version || '3.3-2-alpha1' }}-python-${{ matrix.python-version }}
+          AIRFLOW_CONN_AWS_S3_CONN: ${{ secrets.AIRFLOW_CONN_AWS_S3_CONN }}
+          COSMOS_CONN_POSTGRES_PASSWORD: ${{ secrets.COSMOS_CONN_POSTGRES_PASSWORD }}
+        run: |
+          docker pull "$RUNTIME_IMAGE"
+          docker run --rm --network host --user root \
```
pankajastro, line 619: Nothing downstream of this `docker run` actually shells out to `docker`/`kind` (the KinD cluster and image loading already happened on the host at lines 589–598, and `integration-kubernetes.sh` only talks to the k8s API via kubeconfig) — so what is `-v /var/run/docker.sock:/var/run/docker.sock` here for? Since the image comes from a `workflow_dispatch` input (`registry`/`runtime_version`), mounting the socket gives any image pointed at by those inputs root-equivalent control of the runner. Can we drop the socket mount and make the kubeconfig mount read-only (`-v "$HOME/.kube":/root/.kube:ro`) before `RUN_KUBERNETES_SUITE` gets flipped on?

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,646 @@
...
+        description: 'JSON list of dbt minor versions for the unit + integration suites, e.g. ["1.12"]'
+        required: false
+        default: '["1.12"]'
+        type: string
+
+  # TEMP (remove before merge): run this workflow on the PR itself with a default image, so
+  # reviewers can see it execute. pull_request events do not populate workflow_dispatch inputs,
+  # so the jobs fall back to the defaults in the `|| '...'` expressions below. The paid Databricks
+  # and Snowflake suites are gated to workflow_dispatch only, so they do not run on the PR.
+  pull_request:
```
pankajastro, line 44: This block is commented `# TEMP (remove before merge)`, but it's still here — is it intentionally staying for one more round of review, or was removing it before merge missed? As-is, `pull_request: branches: [main]` will fire on every future PR to `main` once this merges, not just this branch.

scripts/test/runtime-alpha-setup.sh
```diff
@@ -0,0 +1,66 @@
...
+# Pin apache-airflow-providers-* to the versions cosmos tests against (from its per-Airflow
+# lockfile), so provider operators match what cosmos supports. The async BigQuery operator, for
+# example, reassigns __bases__ to the google provider's operator and breaks when the image resolves
+# a newer google provider than cosmos supports. Keyed off the image's Airflow minor. Providers are
+# dbt-agnostic, so the dbt-1.12 lockfile is used regardless of the suite's dbt version. Only
+# providers are constrained, so the image's Airflow and the per-suite dbt re-pins are left untouched.
+AF_MINOR="$(AIRFLOW__LOGGING__LOGGING_LEVEL=ERROR airflow version 2>/dev/null | grep -oE '^[0-9]+\.[0-9]+' | head -1 || true)"
+LOCKFILE="${REPO_ROOT}/requirements/requirements-airflow-${AF_MINOR}-dbt-1.12.txt"
+PROVIDER_CONSTRAINT_ARGS=()
+if [ -n "$AF_MINOR" ] && [ -f "$LOCKFILE" ]; then
```
pankajastro, line 44: When `AF_MINOR` has no matching lockfile, this falls through to "installing providers unpinned" with just a log line (line 49) — should that instead be a hard failure? Otherwise a future Airflow minor with no committed lockfile would silently run every suite unpinned with no CI signal beyond a log line in an otherwise green run.

scripts/test/runtime-alpha-setup.sh
```diff
@@ -0,0 +1,66 @@
...
+# that `uv pip install` stays pinned to the image's Airflow build. uv discovers this file by
+# walking up from the install directory. Some alpha images write an invalid PEP440
+# [project].version there (e.g. "3.3-7-alpha2" -- valid TOML, but not a valid version), which
+# makes uv abort parsing the file and therefore abort EVERY install run under it. We never use
+# that version, so neutralise just that line to a valid placeholder; the constraint-dependencies
+# below it are left untouched and still applied. Guarded on existence so older images are a no-op.
+RUNTIME_UV_PYPROJECT="${RUNTIME_UV_PYPROJECT:-/pyproject.toml}"
+if [ -f "$RUNTIME_UV_PYPROJECT" ]; then
+  echo "Neutralising [project].version in ${RUNTIME_UV_PYPROJECT} (Runtime image ships an invalid PEP440 version that breaks uv)."
+  sed -i -E '0,/^version = /{s/^version = .*/version = "0.0.0"/}' "$RUNTIME_UV_PYPROJECT"
```
pankajastro, line 30: This `sed` neutralizes the *first* `version = ` line anywhere in the file, not one scoped to `[project]` — is that guaranteed safe given what these Runtime images' `pyproject.toml` look like today, or should this be anchored to the `[project]` table specifically so a future image with an earlier `version =` under another table doesn't get silently mismatched?

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,646 @@
...
+    timeout-minutes: 60
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ${{ fromJSON(inputs.python_versions || '["3.12"]') }}
+        dbt-version: ${{ fromJSON(inputs.dbt_versions || '["1.12"]') }}
+        split-group: [1, 2, 3]
+    container:
```
pankajastro, line 108: The `container:`/checkout/`setup-cosmos-runtime` block (~19 lines) repeats verbatim across all 9 jobs (~170 duplicated lines total) — worth factoring into a reusable workflow or composite action in a follow-up so a future change (e.g. to the image-tag expression) only needs to land once? Non-blocking for this PR either way.

.github/workflows/test-runtime-alpha.yml
pankajkoti: Good catch, done. Dropped the `docker.sock` mount and the now-dead in-container `kind` install (nothing here shells out to docker or kind, the suite reaches the cluster over kubeconfig), and made the kubeconfig mount read-only. In f69e42c3.

.github/workflows/test-runtime-alpha.yml
pankajkoti: Intentional for now. The `pull_request` and `push` triggers are the temporary scaffolding that lets the suites run on this PR, and they get removed before merge, reverting to `workflow_dispatch`-only. Left in for one more round of review.

scripts/test/runtime-alpha-setup.sh
pankajkoti: This is moot now. We dropped the provider-lockfile pinning entirely in f69e42c3, so there is no unpinned fallback path anymore. Providers come from the image's own `/pyproject.toml` constraint-dependencies, so each suite runs against the exact provider versions the alpha ships. The one suite that genuinely breaks on the alpha's providers (dbt-async on google 22.4.0) is disabled and tracked in BOSS-662.

scripts/test/runtime-alpha-setup.sh
pankajkoti: Fair point, hardened in 948b0443. Scoped the substitution to the `[project]` table so an earlier `version =` under another table cannot be hit. For reference the image generates this file with `[project]` first and a single `version =` (astro-runtime `image/Dockerfile`), so the first-match version was correct in practice, but the scoped version is robust if that ever changes.

.github/workflows/test-runtime-alpha.yml
pankajkoti: Agreed, non-blocking. The container/checkout/setup block does repeat across the suites. A `workflow_call` reusable workflow is the right factoring, worth a follow-up once the suite set settles, since each suite's env still diverges. Noted for BOSS-662 scope.

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,649 @@
...
+    inputs:
+      runtime_version:
+        description: "Astronomer Runtime alpha version (image tag), e.g. 3.0-1-alpha1"
+        required: true
+        type: string
+      registry:
+        description: "Runtime image repository (without tag)"
+        required: false
+        default: "astrocrpublic.azurecr.io/runtime-dev"
+        type: string
```
kaxil: `registry` being free-form means the dispatcher picks which image runs here, and that image runs `--user root` in jobs whose env carries the AWS/GCP/Azure conns, `AIRFLOW_CONN_DATABRICKS_DEFAULT`, `DATABRICKS_CLUSTER_ID` and the six `SNOWFLAKE_*` secrets. Dispatching only needs `actions: write`, not `contents`, so that is a path to every repo secret with nothing in git history to show for it. Could this be `type: choice` with the two real values (`runtime-dev` plus the released repo)? The `zizmor: ignore[unpinned-images]` is fair for the tag, an alpha version genuinely cannot be digest-pinned, but the registry half does not have to stay open. The other route is an `environment:` on the secret-carrying jobs, the way `test.yml` scopes them through `Authorize`.

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,649 @@
...
+      python_versions:
+        description: 'JSON list of Python versions the alpha publishes, e.g. ["3.12","3.13"]'
```
kaxil, line 35: Runtime 3.3 publishes `-python-3.14` tags, and this one list feeds every suite, including the four that hardcode dbt 1.8/1.9/1.11. `test-unprivileged.yml` deliberately excludes `python-version: "3.14"` + `dbt-version: "1.11"` because dbt-core 1.11 has no 3.14 classifier, so dispatching with `["3.14"]` would turn the dbt-1.8, performance, telemetry and loom suites red for a reason that has nothing to do with the alpha. Worth carrying that exclude into those jobs, or saying here that they need a 3.12/3.13 tag.

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,649 @@
...
+        required: false
+        default: '["3.12"]'
+        type: string
+      dbt_versions:
+        description: 'JSON list of dbt minor versions for the unit + integration suites, e.g. ["1.12"]'
```
kaxil: I don't think this input reaches either of the two suites it names. `unit.sh` never reads `DBT_VERSION` and the unit job never sets it, and `integration-setup.sh` takes it as `$1` and computes `NEXT_MINOR_VERSION` but installs nothing, because the real `dbt-core~=$DBT_VERSION` pin lives in `pre-install-airflow.sh` (line 75) which this workflow deliberately skips. So dbt ends up as whatever `.[dbt-postgres,...]` resolves to, and `-f dbt_versions='["1.8","1.12"]'` yields 2 identical unit jobs and 6 identical integration jobs while reading as two dbt minors validated. Telemetry's hardcoded `["1.9"]` is inert the same way. Either pin dbt inside `runtime-alpha-setup.sh` (perf, loom, dbt-1.8 and async all re-pin afterwards, so they would be unaffected) or drop the axis from the suites that cannot honour it, and once it is wired `["1.11","1.12"]` would match test-unprivileged.yml. Separate from the provider-pinning change in f69e42c3, which was about providers rather than dbt.

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,649 @@
...
+          AIRFLOW_CONN_EXAMPLE_CONN: postgres://postgres:postgres@postgres:5432/postgres
+          POSTGRES_HOST: postgres
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+          POSTGRES_SCHEMA: public
+          POSTGRES_PORT: 5432
+
+  Run-Integration-Tests-Expensive:
+    if: ${{ github.event_name == 'workflow_dispatch' }} # TEMP: paid Databricks, dispatch-only (validated once, remaining failure tracked in BOSS-662)
```
kaxil: Once the TEMP `pull_request`/`push` triggers come out, `workflow_dispatch` is the only trigger left, so this condition is always true and the paid Databricks suite (plus the Snowflake one on line 499) runs on every dispatch. If the intent is "only when explicitly asked for", a `boolean` input survives that cleanup where the event check does not. The mirror image is `vars.RUN_DBT_ASYNC_SUITE` and `vars.RUN_KUBERNETES_SUITE`: both default to unset, so an alpha validation comes back green with 2 of 9 suites never executed and nothing in the run saying which. A `::warning::` or a line in the job summary when a suite is gated off would make that verdict honest.

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,649 @@
...
+            -e DATABRICKS_CLUSTER_ID=mock -e DATABRICKS_HOST=mock \
+            -e DATABRICKS_WAREHOUSE_ID=mock -e DATABRICKS_TOKEN=mock \
+            -e POSTGRES_HOST=localhost -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres \
+            -e POSTGRES_DB=postgres -e POSTGRES_SCHEMA=public -e POSTGRES_PORT=5432 \
+            "$RUNTIME_IMAGE" bash -c '
+              set -euxo pipefail
+              apt-get update && apt-get install -y --no-install-recommends build-essential libpq-dev curl
+              # kubectl for the in-container test setup. The suite reaches the cluster over the
+              # mounted kubeconfig, so no docker socket or kind binary is needed in here.
+              curl -fsSL -o /usr/local/bin/kubectl "https://dl.k8s.io/release/$(curl -fsSL https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
kaxil: kubectl comes from whatever `stable.txt` points at on the day, with no checksum, straight into `/usr/local/bin` in a root container. Everything else here is pinned (checkout by SHA, postgres by digest, kind-action by SHA), so this is the one floating binary, and it also means two runs of the same workflow can pick up different kubectl minors. Pinning the version and checking it against the published `.sha256` would keep it in line with the rest.

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,649 @@
...
+              chmod +x /usr/local/bin/kubectl
+              bash scripts/test/runtime-alpha-setup.sh "dbt-postgres,kubernetes"
+              bash scripts/test/integration-kubernetes.sh
```
kaxil, line 660: `integration-kubernetes.sh` force-installs `apache-airflow-providers-cncf-kubernetes==10.8.0` and `protobuf==6.33.2` whenever Airflow is 3.x, so this suite swaps out the alpha's own cncf-kubernetes for a fixed 10.8.0 and stops measuring the thing the workflow exists to measure. It is the only provider pin left in the run scripts (the rest live in `pre-install-airflow.sh`, which you skip). It has to stay conditional rather than go away, since test.yml's own k8s job calls the same script through the hatch `test-kubernetes` alias. BOSS-662 territory given the job is off, but worth capturing there.

.github/workflows/test-runtime-alpha.yml
pankajkoti: Good catch, fixed in 22d666a9. Made `registry` a `type: choice` constrained to the trusted registry, so a dispatcher can no longer point the secret-bearing jobs at an arbitrary image. Went with the choice rather than an environment gate to keep dispatch friction low, but if you'd prefer the `Authorize`/`environment` approval that test.yml uses for the residual arbitrary-tag risk, happy to add it.

.github/workflows/test-runtime-alpha.yml
pankajkoti: Fixed in 22d666a9. Added a `python-version: "3.14"` matrix exclude to the suites whose run scripts force an old dbt (dbt-loom on 1.11, dbt-1.8 on 1.8, and async on 1.8-1.11), mirroring the exclude and reasoning in test-unprivileged.yml. The unpinned suites resolve a 3.14-capable dbt, so they are left as-is.

.github/workflows/test-runtime-alpha.yml
pankajkoti: You're right, it was a phantom axis. Fixed in 22d666a9 by dropping it rather than faking it: removed the `dbt_versions` input and the `dbt-version` axis from unit, integration, telemetry, performance, and expensive (none of them installed the matrix dbt, since the real pin lives in pre-install-airflow.sh which this workflow skips). The suites that genuinely vary dbt (dbt-1.8, loom, async) keep their own matrices. This workflow's focus is the image's Airflow, and cosmos's main CI already covers the dbt matrix.

.github/workflows/test-runtime-alpha.yml
pankajkoti: Both fixed in 22d666a9. Replaced the `event_name == 'workflow_dispatch'` gate on the paid Databricks and Snowflake suites with a `run_paid_suites` boolean input (off by default), so they stay off after the TEMP triggers are removed instead of flipping to always-true. Added a `Report-Suite-Coverage` job that writes a ran-vs-skipped manifest to the run summary, so a green run states which suites were gated off rather than silently validating nothing.

.github/workflows/test-runtime-alpha.yml
pankajkoti: Confirmed and captured in BOSS-662 (item 7), alongside two more of the same kind the review surfaced (`pydantic==2.11.0` in integration-dbt-async.sh, `cadwyn!=5.4.0` in integration-dbtf-setup.sh). Since integration-kubernetes.sh is shared with test.yml's own k8s job and the suite is disabled, I left the script as-is rather than change a `==` pin cosmos CI depends on. The follow-up is to move these to a `>=` floor under the image constraints so the alpha's shipped versions stand.

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,728 @@
...
+        required: false
+        default: false
+        type: boolean
+
+  # TEMP (remove before merge): run this workflow on the PR itself with a default image, so
+  # reviewers can see it execute. pull_request events do not populate workflow_dispatch inputs,
+  # so the jobs fall back to the defaults in the `|| '...'` expressions below. The paid Databricks
+  # and Snowflake suites are gated behind the run_paid_suites input (off by default), so they do not
+  # run on the PR or on a default dispatch.
+  pull_request:
```
pankajastro: Is this `pull_request`/`push` trigger still meant to be removed before merge? There's no `Authorize`-style gate here (unlike `test.yml`), so as submitted this would run the secret-consuming, root-container jobs on every future PR to `main`.

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,728 @@
...
+    branches: [main]
+  # Also run on every push to the PR branch (in addition to pull_request), so each commit runs.
+  push:
+    branches:
+      - pankajkoti/boss-610-integration-of-cosmos-tests-with-runtime-alpha-release
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.runtime_version }}
```
pankajastro, line 46: Since `inputs.runtime_version` is only populated on `workflow_dispatch`, doesn't this group key differ between the `pull_request` and `push` events on the same commit (`refs/pull/<n>/merge` vs `refs/heads/<branch>`)? If so, `cancel-in-progress` can't dedupe them and every push runs the whole suite twice.

scripts/test/runtime-alpha-setup.sh
```diff
@@ -0,0 +1,52 @@
...
+# walking up from the install directory. Some alpha images write an invalid PEP440
+# [project].version there (e.g. "3.3-7-alpha2" -- valid TOML, but not a valid version), which
+# makes uv abort parsing the file and therefore abort EVERY install run under it. We never use
+# that version, so neutralise it to a valid placeholder, scoped to the [project] table so only that
+# key is touched and the constraint-dependencies below stay intact. Guarded on existence so older
+# images are a no-op.
+RUNTIME_UV_PYPROJECT="${RUNTIME_UV_PYPROJECT:-/pyproject.toml}"
+if [ -f "$RUNTIME_UV_PYPROJECT" ]; then
+  echo "Neutralising [project].version in ${RUNTIME_UV_PYPROJECT} (Runtime image ships an invalid PEP440 version that breaks uv)."
+  sed -i -E '/^\[project\]/,/^\[/{s/^version = .*/version = "0.0.0"/}' "$RUNTIME_UV_PYPROJECT"
```
pankajastro, line 31: This `sed` only matches a `version = ...` line formatted exactly this way. Should we assert the substitution actually happened (e.g. `grep` the file afterward and fail loudly), so a differently-formatted version line in a future image doesn't silently leave the invalid PEP440 version in place?

tests/operators/test_local.py
```diff
@@ -818,10 +819,12 @@ def test_run_operator_dataset_inlets_and_outlets_airflow_3_onwards(caplog):
...
     assert "Assigning outlets with DatasetAlias in Airflow 3" in caplog.text
-    assert (
-        "Outlets: [Asset(name='postgres://0.0.0.0:5432/postgres/public/stg_customers', uri='postgres://0.0.0.0:5432/postgres/public/stg_customers'"
-        in caplog.text
-    )
+    # The Asset URI host:port comes from the example_conn connection, which differs between setups
+    # (0.0.0.0 on a runner, the service hostname inside a container). Derive it so the assertion is
+    # host-agnostic.
+    conn = urlparse(os.environ.get("AIRFLOW_CONN_EXAMPLE_CONN", "postgres://user:pass@0.0.0.0:5432/postgres"))
+    expected_asset = f"postgres://{conn.hostname}:{conn.port or 5432}/postgres/public/stg_customers"
```
pankajastro: Host/port are derived from `AIRFLOW_CONN_EXAMPLE_CONN`, but `postgres` (the database name) is still hardcoded in `expected_asset`. Should this come from `conn.path` too, to fully match the "host-agnostic" goal?

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,728 @@
...
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/setup-cosmos-runtime # zizmor: ignore[self-repository]
+        with:
+          extras: ${{ env.COSMOS_TEST_EXTRAS }}
+      - name: Run performance tests (${{ matrix.num-models }} models)
+        run: |
+          # ERROR log level scoped to setup: keep the image's plugin warnings out of the
+          # `airflow version` output the setup script parses (db migrate vs init routing).
+          AIRFLOW__LOGGING__LOGGING_LEVEL=ERROR bash scripts/test/performance-setup.sh "$DBT_VERSION"
```
pankajastro, line 229: This (and `integration-dbt-loom.sh`, `integration-dbt-1-8.sh`, `integration-dbt-async.sh`) reinstall dbt adapters with plain `pip` instead of `uv pip install --system`. Since only `uv` reads the image's `[tool.uv].constraint-dependencies` pin, doesn't plain `pip` here risk an unconstrained resolve against the image's pinned Airflow build?

pankajastro (approved):

.github/workflows/test-runtime-alpha.yml
pankajkoti: Yes, these are TEMP scaffolding and come out before merge, reverting the workflow to workflow_dispatch-only (no PR trigger, so the secret-consuming jobs never run on a PR). On the missing Authorize gate, it is deliberate, and here is why it is safe.

- workflow_dispatch is reachable only by an actor with `actions: write`, that is repo write access. Fork and read-only contributors cannot trigger it at all, over the API or the UI.
- The `registry` input is a `type: choice` allowlist (only `astrocrpublic.azurecr.io/runtime-dev`), and GitHub enforces choice options on API dispatches too, not just the UI dropdown. So a dispatcher cannot point the secret-bearing root containers at an arbitrary image.
- test.yml's Authorize/environment gate exists for untrusted fork PRs (`pull_request_target` from an external repo, routed to the reviewer-protected `external` environment). Same-repo write-access actors get the ungated `internal` path there. A workflow_dispatch-only workflow is reachable only by that same write-access tier, so the fork threat the gate protects against cannot reach it.

We also plan to trigger this automatically from astro-runtime CI when an alpha publishes, so we specifically want it dispatchable without an approval gate. The registry allowlist is what keeps that automated, gate-free path from being an arbitrary-image hole.

.github/workflows/test-runtime-alpha.yml
pankajkoti: Good eye. That double-run is real, but it only happens because the TEMP pull_request and push triggers both fire on the same commit with different github.ref values. Once those come out (dispatch-only before merge), there is a single event and the group key dedupes normally. So it resolves with the descaffold rather than needing a group-key change.

scripts/test/runtime-alpha-setup.sh
pankajkoti: Added in 37ff7462 (file handle later tidied into a context manager in d81cb04b). After the sed, a tomllib check reads the [project].version regardless of how the line is formatted and fails loudly if it is still not valid PEP440, so a future image whose version line the sed does not match aborts setup with a clear message instead of letting uv reject the file later with a confusing one. Verified it exits non-zero on a mismatched version and is a no-op when there is no version line.

tests/operators/test_local.py
pankajkoti: Fixed in 37ff7462. The database name now comes from conn.path too (conn.path.lstrip('/') or 'postgres'), so host, port, and db are all derived from the connection and the assertion is fully host-agnostic.

.github/workflows/test-runtime-alpha.yml
pankajkoti: Real catch, and it is the same constraint-fidelity family as the provider overrides. Those dbt-adapter reinstalls in integration-dbt-loom.sh, integration-dbt-1-8.sh, and integration-dbt-async.sh use plain pip, which does not read the image's [tool.uv].constraint-dependencies, so they can resolve unconstrained against the image's pinned Airflow. The scripts are shared with test.yml's own suites, so a pip to uv pip switch needs validating on both paths. Tracking it in BOSS-662 to do that carefully rather than in this PR.

tests/utils.py
```diff
@@ -64,30 +64,37 @@ def check_dag_state(dag_run: DagRun | None, expected_dag_state: DagRunState = Da
...
 
-        try:
-            from airflow.dag_processing.dagbag import sync_bag_to_db
-        except ImportError:
-            from airflow.models.dagbag import sync_bag_to_db
+    Airflow 3.1+ requires a DAG to be serialized before ``dag.test()`` / ``create_dagrun()``,
+    which check for DagVersion and DagModel records. No-op on older Airflow.
+    """
+    if AIRFLOW_VERSION < version.Version("3.1"):
+        return
```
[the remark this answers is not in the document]
pankajkoti: Good catch, and thanks for flagging both spots. The serialize_dag_to_db guard is fixed, and I completed it in b906954d for new_test_dag too (lines 96 and 99, the ones your comment pointed at), which still used the Version("3.1") / Version("3.0") comparisons. All three version guards in the file now use release-tuple comparisons, matching make_dag_bag.

scripts/test/runtime-alpha-setup.sh
```diff
@@ -0,0 +1,71 @@
...
+  # Confirm the neutralisation took. tomllib reads the version regardless of how the line is
+  # formatted, so this catches a future image whose version line the sed did not match, which would
+  # otherwise leave an invalid PEP440 version for uv to reject later with a confusing error.
+  python - "$RUNTIME_UV_PYPROJECT" <<'PY'
+import sys, tomllib
+from packaging.version import InvalidVersion, Version
+
+path = sys.argv[1]
+version = tomllib.load(open(path, "rb")).get("project", {}).get("version")
```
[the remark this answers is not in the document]
pankajkoti: Done, the tomllib load is now wrapped in a `with` context manager.

scripts/test/runtime-alpha-setup.sh
```diff
@@ -0,0 +1,72 @@
...
+if version is not None:
+    try:
+        Version(version)
+    except InvalidVersion:
+        raise SystemExit(
+            f"ERROR: {path} still has an invalid PEP440 [project].version ({version!r}) after "
+            "neutralisation. The version-line format likely changed. Update the sed in "
+            "runtime-alpha-setup.sh."
+        )
+PY
```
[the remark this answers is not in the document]
pankajkoti: This one is a false positive. In sed, the end of a `/re1/,/re2/` range is matched on the lines after the start, so `/^\[/` matches the next table header (`[tool.uv]`), not the `[project]` line that opened the range. Verified against the real image layout and adversarial fixtures: the substitution rewrites `[project].version` and leaves other tables untouched. And the tomllib assertion right after would fail loudly if a differently-formatted line ever did slip past, so an invalid version cannot silently remain.

.github/workflows/test-runtime-alpha.yml
```diff
@@ -0,0 +1,728 @@
...
+  # so the jobs fall back to the defaults in the `|| '...'` expressions below. The paid Databricks
+  # and Snowflake suites are gated behind the run_paid_suites input (off by default), so they do not
+  # run on the PR or on a default dispatch.
+  pull_request:
+    branches: [main]
+  # Also run on every push to the PR branch (in addition to pull_request), so each commit runs.
+  push:
+    branches:
+      - pankajkoti/boss-610-integration-of-cosmos-tests-with-runtime-alpha-release
+
```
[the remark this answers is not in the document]
pankajkoti: Yes, these are TEMP scaffolding removed before merge, reverting the workflow to workflow_dispatch-only. Doing that as the final step once the fixes are green.

pankajkoti:
On Copilot's note about the checkout ref: this is intentional. The jobs check out the dispatched ref rather than pinning main, so the workflow validates whatever ref you dispatch. Latest main is the default (gh workflow run and the dispatch API default to the repository default branch when no ref is given, which is what the automated astro-runtime dispatch uses), and dispatching a branch lets us validate an in-progress cosmos fix against an alpha without merging it to main first. Updated the PR description to say "checked-out ref, latest main by default" so it matches the behavior.

PR #2996, opened 2026-09-09
Make`emit_datasets` templatable via `operator_args`

tatiana:
Change `emit_datasets` to be, at the operator level, an Airflow template field. This change does not affect `RenderConfig.emit_datasets`, which stays the parse-time default.

The motivating case is keeping backfills from triggering downstream asset-scheduled DAGs:
```
    operator_args={
        "emit_datasets": "{{ dag_run.run_type != 'backfill' }}"
    }
```
`operator_args` now takes precedence over `RenderConfig.emit_datasets`. Previously, `emit_datasets` passed in `operator_args` was discarded without any warnings.

On Airflow 2.10 and 2.11 the DatasetAlias is still registered while the task is built, because the runtime gate needs an alias to add events to; a run that renders to false leaves the alias registered and emits nothing to it.


pankajkoti (approved):

pankajastro:
`full_refresh` check changed from `is True` to `bool(...)`:

```python
full_refresh = 1
full_refresh is True   # False -> old: no --full-refresh
bool(full_refresh)     # True  -> new: --full-refresh appended
```

Since `--full-refresh` is destructive, worth double-checking this is intentional — could silently trigger it if `full_refresh` ever resolves to `1`/`0` instead of `True`/`False` (e.g. templated off `dag_run.conf`).

PR #3000, opened 2026-09-10
Add ExecutionMode.WATCHER_AWS_ECS

### LINKED ISSUE #2988: [Feature] ExecutionMode.WATCHER_AWS_ECS: run the watcher producer in an ECS task
### Description

Add `ExecutionMode.WATCHER_AWS_ECS`: the watcher counterpart to the existing `AWS_ECS` mode, so the producer runs `dbt build` in an ECS task while consumers poll its per-node status XCom.

Cosmos already ships two containerised watcher variants, `WATCHER_KUBERNETES` and `WATCHER_GCP_GKE`. There is no AWS equivalent, so on AWS the choice is between running the producer on the Airflow worker (`WATCHER`) or one container per node (`AWS_ECS`).

### Use case/motivation

On MWAA the worker is small and shared. An `mw1.medium` worker is 2 vCPU and 4 GB, and by default it runs up to 10 concurrent Airflow tasks. `WATCHER` runs dbt as a subprocess inside that worker, so producers belonging to different dbt projects can land on the same worker and compete for the same 4 GB.

What this looks like in practice: three dbt projects scheduled at the same time, each producer starting `dbt build` with the profile's `threads`. The producers survive the view phase and are SIGKILLed seconds after several snapshots start in parallel. Exit code -9, no traceback, nothing in the dbt logs, because the process is killed by the kernel rather than failing. Staggering the schedules makes the same projects pass, which points at contention rather than at any one project's size.

`AWS_ECS` removes the contention but replaces one `dbt build` with one container per node, which costs container start-up per node and loses the property that makes the watcher attractive.

`WATCHER_AWS_ECS` would keep the single build and move it off the worker, matching what Kubernetes users already have.

### Notes on feasibility

The transport that the Kubernetes variant gets from the pod manager callback appears to have an ECS equivalent already in the provider. `EcsRunTaskOperator` starts `AwsTaskLogFetcher`, a thread whose `run()` calls `logger.info()` once per CloudWatch log event, and the logger is passed in the constructor. A logger-like object that parses dbt's JSON events and pushes the per-node status XCom looks like a comparable hook, in the same spirit as `_StdoutFilter` in the local watcher, and it would need no change to the provider.

Two constraints worth raising before any implementation:

1. `EcsRunTaskOperator.execute()` defers before starting the log fetcher, so the producer would have to run with `deferrable=False` and hold a worker slot for the duration of the build. For the memory problem above that trade is acceptable, since the slot would hold a log-following thread rather than dbt itself, but it is a real difference from the Kubernetes variant.
2. The producer XCom backup and restore on retry hangs off `KubernetesPodOperatorCallback` in `_k8s_common`. `EcsRunTaskOperator` has no equivalent callback surface, so that part would need Airflow-level task callbacks instead.

Happy to be told this is out of scope or that a different shape is preferred before writing any code.

### Related issues

None found for a watcher variant on ECS.

fpiped:
## Description

Adds `ExecutionMode.WATCHER_AWS_ECS`: the watcher producer runs one `dbt build` in an ECS task and publishes per-node statuses read from the task's CloudWatch log stream; the consumer sensors are the existing ones. Same contract as `WATCHER_KUBERNETES`, on ECS. Motivation and the MWAA failure it addresses are in #2988.

Built on #2999, which moves the container-agnostic producer helpers out of `_k8s_common` so this module imports without the Kubernetes stack. Until #2999 merges this diff includes its commit; I'll rebase once it lands.

### How it works

`EcsRunTaskOperator.execute` (non-deferrable path, `awslogs_group` + `awslogs_stream_prefix` set) starts an `AwsTaskLogFetcher` thread through `_get_task_log_fetcher()`, waits for the task, then stops and joins the fetcher. The producer overrides three seams of that flow, all present since `apache-airflow-providers-amazon` 8.3.0 (`_ECS_WATCHER_MIN_AMAZON_PROVIDER_VERSION`, checked at import):

- `_get_task_log_fetcher()` returns `WatcherEcsLogFetcher`, a subclass that queues every CloudWatch event message instead of only logging it. Its `run()` also fixes a gap that matters here: the provider loop is `while not stopped: sleep; fetch`, so after the task stops the events written since the last poll — the last `NodeFinished` lines — are never read. The subclass keeps polling after `stop()` until dbt's `CommandCompleted` event is read or `drain_timeout` (default 60 s) elapses, and logs a warning if it gives up.
- `_wait_for_task_ended()` polls `describe_tasks` with the operator's `waiter_delay` / `waiter_max_attempts` and parses the queued lines between polls, so `store_dbt_resource_status_from_log` and every XCom push run on the operator's thread, as they do for the Kubernetes producer (the fetcher thread does I/O only).
- `_after_execution()` parses the drained tail before the provider's `_check_success_task()`.

Status latency for a consumer is bounded by `awslogs_fetch_interval` (default 5 s for this producer; the provider's 30 s is meant for displaying logs) plus the consumer's `poke_interval`.

`execute()` delegates to the shared `execute_watcher_producer`: no retries of the build, XCom backup restored and the attempt skipped on `try_number > 1`, backup flushed from the `on_retry` / `on_failure` callbacks.

### Rejected arguments

`awslogs_group` and `awslogs_stream_prefix` are required (no log stream, no statuses); `deferrable=True`, `reattach=True` and `wait_for_completion=False` raise `CosmosValueError` with the reason. Deferrable is not offered in this PR: the provider's deferred path (`TaskDoneTrigger`) yields once, at task completion, and does not re-enter the operator on an interval the way `KubernetesPodTrigger` does with `logging_interval`; a Cosmos trigger with that shape is a follow-up I intend to open after this one.

### Files

- `cosmos/operators/watcher_aws_ecs.py`: `WatcherEcsLogFetcher`, `DbtProducerWatcherAwsEcsOperator`, `DbtConsumerWatcherAwsEcsSensor` and the seed / snapshot / run / test / source / build operators, mirroring `watcher_kubernetes.py` (test sensor falls back to `dbt test --select <model>` in an ECS task; `DbtBuildWatcherAwsEcsOperator` raises).
- `cosmos/constants.py`: `ExecutionMode.WATCHER_AWS_ECS`, `_ECS_WATCHER_MIN_AMAZON_PROVIDER_VERSION`.
- `cosmos/airflow/graph.py`: the mode in `_WATCHER_TO_FALLBACK_EXECUTION_MODE` (→ `AWS_ECS` for `AFTER_ALL` tests); the producer-creation branch now keys off that map instead of a hard-coded tuple.
- `tests/operators/test_watcher_aws_ecs.py`: 27 tests, all mocked — operator class resolution and fallback map; producer defaults and rejected arguments; fetcher bound to the operator state; `collect()` queues messages and detects `CommandCompleted` (JSON only); drain after `stop()` reads the tail and gives up on timeout; queued lines parsed with the execution context and the producer state; `_wait_for_task_ended` parses between status polls and fails on a missing or never-stopping task; `_after_execution` order; retry skip and first-attempt flow; consumer first poke, retry fallback to an ECS `dbt run --select`, test sensor fallback; build operator raises.
- Docs: `docs/guides/run_dbt/container/watcher-aws-ecs-execution-mode.rst` (how it works, test behaviour, provider version, rejected arguments, `awslogs` driver in `blocking` mode, IAM), listed in the container index, the execution-modes list and the comparison table.

### Verification

Unit suite and the new tests green locally; docs build clean. Cosmos CI has no ECS cluster, so an end-to-end run against a real cluster is pending; I'll report the result in this PR before asking for review.

## Related Issue(s)

closes #2988

## Breaking Change?

No. New execution mode; existing modes untouched apart from the refactor in #2999.

## Checklist

### DIFF
--- a/cosmos/airflow/graph.py
+++ b/cosmos/airflow/graph.py
@@ -212,6 +212,7 @@ _WATCHER_TO_FALLBACK_EXECUTION_MODE = {
     ExecutionMode.WATCHER: ExecutionMode.LOCAL,
     ExecutionMode.WATCHER_KUBERNETES: ExecutionMode.KUBERNETES,
     ExecutionMode.WATCHER_GCP_GKE: ExecutionMode.GCP_GKE,
+    ExecutionMode.WATCHER_AWS_ECS: ExecutionMode.AWS_ECS,
 }
 
 
@@ -1243,7 +1244,7 @@ def build_airflow_graph(
     if execution_mode == ExecutionMode.AIRFLOW_ASYNC:
         # This property is only relevant for the setup task, not the other tasks:
         virtualenv_dir = task_args.pop("virtualenv_dir", None)
-    elif execution_mode in (ExecutionMode.WATCHER, ExecutionMode.WATCHER_KUBERNETES, ExecutionMode.WATCHER_GCP_GKE):
+    elif execution_mode in _WATCHER_TO_FALLBACK_EXECUTION_MODE:
         setup_operator_args = getattr(execution_config, "setup_operator_args", None) or {}
         # We are intentionally creating the producer task ahead of the consumer tasks
         # Airflow priority weight is not being respected in multiple versions of the library, including 3.1
--- a/cosmos/constants.py
+++ b/cosmos/constants.py
@@ -125,6 +125,7 @@ class ExecutionMode(Enum):
     GCP_GKE = "gcp_gke"
     WATCHER_GCP_GKE = "watcher_gcp_gke"
     WATCHER_KUBERNETES = "watcher_kubernetes"
+    WATCHER_AWS_ECS = "watcher_aws_ecs"
 
 
 class InvocationMode(Enum):
@@ -236,5 +237,8 @@ TELEMETRY_TIMEOUT = 1.0
 _AIRFLOW3_MAJOR_VERSION = 3
 
 _K8s_WATCHER_MIN_K8S_PROVIDER_VERSION = Version("10.8.0")
+# First release with `AwsTaskLogFetcher._get_log_events(skip_token)` and the `_wait_for_task_ended` /
+# `_after_execution` seams of `EcsRunTaskOperator` that the ECS watcher producer overrides.
+_ECS_WATCHER_MIN_AMAZON_PROVIDER_VERSION = Version("8.3.0")
 _DBT_STARTUP_EVENTS_XCOM_KEY = "dbt_startup_events"
 _PRODUCER_CMD_FLAGS_XCOM_KEY = "producer_cmd_flags"
--- a/cosmos/operators/_k8s_common.py
+++ b/cosmos/operators/_k8s_common.py
@@ -15,7 +15,6 @@ from os import PathLike
 from typing import TYPE_CHECKING, Any, Protocol
 
 import kubernetes.client as k8s
-from airflow.exceptions import AirflowException
 from airflow.providers.cncf.kubernetes.backcompat.backwards_compat_converters import convert_env_vars
 from airflow.providers.cncf.kubernetes.callbacks import client_type
 from airflow.providers.cncf.kubernetes.operators.pod import KubernetesPodOperator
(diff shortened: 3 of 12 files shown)


fpiped:
Ran this end to end on ECS Fargate (Airflow 3.1.8 standalone, `apache-airflow-providers-amazon` 9.36.0, `dbt-oss` 2.0.5 in the container, a two-model duckdb project with three tests).

- Producer: `dbt build --log-format json` in one Fargate task; the fetcher found the CloudWatch stream ~15 s after `RunTask`, read the five `NodeFinished` events and dbt's `CommandCompleted` about 7 s after dbt wrote them, and pushed the statuses. Task exit code 0.
- Consumers: all four pulled their status from the producer's XCom keys (one from its deferred trigger, the rest on the first poke); no fallback runs. DAG run 60 s wall time.

Two things the run taught, now in the docs: `awslogs_stream_prefix` has to include the container name (the provider builds `<prefix>/<task id>`, the driver writes `<prefix option>/<container>/<task id>`), and the image's working directory must be the dbt project because container modes pass no `--project-dir`.

Not exercised here: the producer retry path and the consumer fallback when a status is missing; both remain unit-tested only. Marking the PR ready for review.

